PR #5, opened 2016-09-29
define INSTALL_LIB_PATH for cmd-record.c

thepaul:
and not uftrace.c, because that symbol is only used in cmd-record.c. without this change, it looks like passing --libdir to configure has no effect.


namhyung:
@thepaul Thanks for your contribution!  It's indeed a bug and your fix looks good!

PR #7, opened 2016-10-06
Build fix when objdir is set

ganadist:
Hello namhyung.

I tried to make uftrace  package for personal use. but I found there are some build errors when set objdir option.
Please look patches and apply them.

### DIFF
--- a/configure
+++ b/configure
@@ -65,6 +65,7 @@ LD=${LD:-${CROSS_COMPILE}ld}
 
 srcdir=$(readlink -f $(dirname $0))
 objdir=${objdir:-${PWD}}
+[ ${objdir:0:1} != '/' ] && objdir=${PWD}/${objdir}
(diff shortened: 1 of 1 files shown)


namhyung:
@ganadist Thanks for your contribution.

How do you use the objdir then?  I thought following usecase which don't need objdir.

```
$ mkdir build
$ cd build
$ ../uftrace/configure
$ make
```

Yes, it still needs to have install target though..

namhyung:
It seems you now removed bash dependency from the configure, right?  And I still don't know how to use this patch.  Could you please share your build settings?

namhyung:
@ganadist I'm closing this PR since I think it's not necessary now and it doesn't apply to the current code.  Oh, I already cherry-picked the last commit which adds 'install' rule in the generated Makefile.  If you have any concern, please let me know.

PR #9, opened 2016-10-06
review/time filter v2

honggyukim:
This PR fixes a replay time bug for time filtered functions with 'trace' trigger enabled.


namhyung:
Thank you for the fix.  However this code is not yet released and the bug was fixed already in my local branch so I think it's better not to apply this PR, sorry.

honggyukim:
Thanks for implementing time filter support in replay command. This feature was really necessary for me.

PR #10, opened 2016-10-07
minor documentation spelling and grammar

thepaul:
Hopefully this isn't offensive. I noticed some spelling errors in a man page (or, the Debian "lintian" tool caught them for me), and in the course of correcting them, I found a few other things that could be clarified, and then I ended up going through all the man pages to try and make sure my changes were in keeping with the original intended meanings and that they were consistent between pages.

If you don't care about this sort of thing, do feel free to close the PR without merging. I can suppress the lintian warnings on my end if I need to.

Aside: I love these docs. Very few advanced tools like this have good documentation. Thanks for doing that.


namhyung:
Wow, thanks a lot!! In fact I highly appreciate this kind of documentation patch.  As a non-native English speaker I always worry about the quality of the doc.  I only have a few of minor nitpicks and will add inline comments.  Again, thanks for your work!

doc/uftrace-report.md
```diff
@@ -23,45 +23,45 @@ OPTIONS
...
-:   Show average, min, max of each functions self time.
+:   Show average, min, max of each function's self time.
 
 \--diff=*DATA*
-:   Report difference between the input trace data and the given DATA.
+:   Report differences between the input trace data and the given DATA.
 
 \--sort-column=*IDX*
-:   When --diff option is used, 3 columns will be shown for each total time, self time and call count.  This is option is to select the index of column to be used as a sort key.  The index 0 is for original data given by --file option, and index 1 is for data given by --diff option, and index 2 is for (percentage of) difference between the two data.
+:   When `--diff` is used, 3 columns will be shown: total time, self time and call count.  This option selects the index of the column to be used as a sort key.  Index 0 is for original data given by the `--file` option, index 1 is for data given by the `--diff` option, and index 2 is for (percentage) differences between the two data.
```
namhyung: I know it's not your problem, but could you please change `--file` option to `--data` option?  It was changed long ago but the doc update was missed.

doc/uftrace-report.md
```diff
@@ -23,45 +23,45 @@ OPTIONS
...
 
 -k, \--kernel
-:   Trace kernel functions as well as user functions.  Only kernel functions inside user functions will be shown.
+:   Show kernel functions as well as user functions.  Only kernel functions called inside user functions will be shown.
 
 --kernel-full
-:   Show all kernel functions called outside of user functions.  Implies \--kernel option.
+:   Show all kernel functions, including those called outside of user functions.
```
namhyung: Do you think it's ok to remove the "Implies `--kernel`.`" part?

doc/uftrace-report.md
thepaul: Sure thing.

doc/uftrace-report.md
thepaul: I did that just because the `--kernel` description specifically says "Only kernel functions called inside user functions will be shown". If `--kernel-full` implies `--kernel`, then we would expect both "Show all kernel functions, including those called outside of user functions" and "Only kernel functions called inside user functions will be shown", which is confusing.

namhyung: Ok then.

PR #12, opened 2016-10-07
don't try to spawn pagers during test

thepaul:
uftrace correctly avoids spawning a pager when stdout is not a tty, and
runtest.py correctly runs uftrace connected to pipes, but (many? most?)
individual tests spawn uftrace themselves without connecting it to
any pipes.

thus, if you try to run runtests.py yourself in an interactive shell
with a tty, uftrace may try to spawn many many pagers in the course of
the test. and depending on your pager availability and configuration,
*each pager* might require interaction with the user before it will
quit.

you can, of course, use `runtests.py | cat` to avoid the whole pager
mess entirely, but this also has the effect of removing the coloring
from runtests.py output, which is extremely useful.

instead, this change adds `--no-pager` to every uftrace invocation under
tests.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -21,7 +21,7 @@ class TestBase:
     TEST_SUCCESS_FIXED = -8
 
     objdir = os.environ['objdir'] or '..'
-    ftrace = objdir + '/uftrace -L' + objdir
+    ftrace = objdir + '/uftrace --no-pager -L' + objdir
 
     default_cflags = ['-fno-inline', '-fno-builtin', '-fno-omit-frame-pointer']
 


namhyung:
Does it require interaction from user for each test?  The runtest.py will execute the uftrace by sp.Popen with stdout=sp.PIPE so it should not spawn pager.  In addition the default pager is "less" with "-F" option which terminates itself if the result is fit on the screen.

When I ran the runtest.py manually I saw a different error instead.

```
$ ./runtest.py 
  File "./runtest.py", line 7, in <module>
    class TestBase:
  File "./runtest.py", line 23, in TestBase
    objdir = os.environ['objdir'] or '..'
  File "/usr/lib/python3.5/os.py", line 725, in __getitem__
    raise KeyError(key) from None
KeyError: 'objdir'
```
I'd go to fix it.  But after the fix I cannot see any problem.

Tip: I usually run the testcases with "make test".  I can give options or test names via "TESTARG" variable.  The -d option shows the diff, -p option only selects "-pg" build and -O option selects compiler optimazation level.  So when some test failed, I'd run it again with following command:

```
$ make test TESTARG='-d 100'

or

$ make test TESTARG='-dpO2 time'
```

thepaul:
> Does it require interaction from user for each test? The runtest.py will execute the uftrace by
> sp.Popen with stdout=sp.PIPE so it should not spawn pager. In addition the default pager is
> "less" with "-F" option which terminates itself if the result is fit on the screen.

It shouldn't require interaction, certainly, no, but it does if you already have `$LESS` set. Try this in a terminal:

    PAGER=less LESS=  make test TESTARG='-d 23'

I thought the problem was when `$PAGER` was unset, but it turns out that the problem manifests instead when `$LESS` is set (because on line 36 of utils/pager.c, the overwrite parameter to `setenv` is 0, so if `$LESS` is already in the environment, it is not changed to `FSRX`).

As for `sp.Popen(..., stdout=sp.PIPE)`, that is true, when uftrace is being invoked that way there should be no problem with pagers. But it's not _always_ invoked that way, as I mentioned above. See, for example, `tests/t023_replay_filter.py`. It runs uftrace without putting `stdout` through a pipe. Several other tests do the same.

> When I ran the runtest.py manually I saw a different error instead.

Right, yes, you have to set `$objdir` in order to run `runtest.py` directly. I think that's fine, and I am fine with running it under `make test`. I just wanted to narrow the problem description to `runtest.py` specifically.

> Tip: I usually run the testcases with "make test". I can give options or test names via "TESTARG"
> variable. The -d option shows the diff, -p option only selects "-pg" build and -O option selects compiler
> optimazation level. So when some test failed, I'd run it again with following command:

Yep. I did see that, and it's a very nice construct. Thanks for that.

namhyung:
> As for sp.Popen(..., stdout=sp.PIPE), that is true, when uftrace is being invoked that way there should be no problem with pagers. But it's not always invoked that way, as I mentioned above. See, for example, tests/t023_replay_filter.py. It runs uftrace without putting stdout through a pipe. Several other tests do the same.

Ah, you're right.  They use `sp.call()` to run `uftrace record` for the test.

PR #14, opened 2016-10-08
doc: Update time filter option in other man pages

honggyukim:
Since time filter is supported in replay, report, dump, and graph
commands now, their man pages have to be updated as well.


honggyukim:
@namhyung Thanks for the comment. I have updated and sent a new pull request in check/man-update-v2. Please have a look at it.

PR #15, opened 2016-10-09
Check/man update v2

honggyukim:
Since time filter is supported in replay, report, dump, and graph
commands now, their man pages have to be updated as well.


namhyung:
Thanks both of you!

PR #16, opened 2016-10-11
use "LINKFLAGS" for `ld` invocations

thepaul:
instead of "LDFLAGS", as that is supposed to be used for invocations of
gcc that are intended to call ld underneath, and not ld directly.
before this change, if a user overrode LDFLAGS for the main Makefile,
then it would also be overridden for these Makefiles as well, and would
often breaks (try, for example, `make LDFLAGS=-Wl,-z,now`).

### DIFF
--- a/arch/arm/Makefile
+++ b/arch/arm/Makefile
@@ -1,5 +1,5 @@
 ASFLAGS = -fPIC
-LDFLAGS = -r
+LINKFLAGS = -r
 
 sdir = $(srcdir)/arch/arm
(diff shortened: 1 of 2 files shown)


thepaul:
No. I don't know of any standard name for a partial-linkage command like this. Any other name that doesn't collide should be fine, if you'd prefer something else.

namhyung:
I just wanted to know whether it's a standard name. I'm ok with the name of LINKFLAGS.

PR #17, opened 2016-10-11
don't build libtraceevent modules with -fPIC

thepaul:
(a) because this project uses libtraceevent as a static lib to be linked
    directly into an executable, where -fPIC doesn't make sense. from
    what I've heard, this is how everyone uses libtraceevent.

(b) because there is no way for a higher-level Makefile to disable this
    use of -fPIC; adding -fPIE to the options at the end would override
    the -fPIC, but there is no way with the Makefile as previously
    written to insert options after the -fPIC.


namhyung:
Yes, afaik other projects including `trace-cmd` (which is the original source of the libtraceevent), `perf` and `powertop` build it as a static library.  Although there's a plan to publish the libtraceevent as a separate shared library, we should use static one before finishing that work (I expect it'll take some time).

PR #18, opened 2016-10-11
honor CPPFLAGS

thepaul:
since that is a pretty standard way of handing off preprocessor flags
to build systems.


namhyung:
LGTM :)

PR #19, opened 2016-10-11
allow overriding CFLAGS/LDFLAGS per piece

thepaul:
so that a user's build instruction can specify some flags to be used only when building uftrace, some to be used only when building libtraceevent, and some to be used only when building the libs (currently, only libmcount*).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,13 +69,14 @@ COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
 
 #
 # Note that the plain CFLAGS and LDFLAGS can be changed
-# by config/Makefile later but LIB_*FLAGS are not.
+# by config/Makefile later but *_*FLAGS can not.
 #
-UFTRACE_CFLAGS = $(COMMON_CFLAGS) $(CFLAGS_$@)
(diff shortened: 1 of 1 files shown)


namhyung:
@thepaul At this point, I need to tell you that I want you to read the [DCO](http://elinux.org/Developer_Certificate_Of_Origin) and add Signed-off-by line in your commit message (`git` has -s option for this).  It was from the Linux kernel project, but I think it's good to use it here as well to clarify possible license issues later.  I'll add it to the documentation.

Also I think it's better to combine your PRs into a single PR since they are all related to build and I don't want to add many merge commits if possible.  It's also a good idea to have a subject word in the commit header line like "build: allow overriding ...".

thepaul:
Sure, a DCO signoff is no problem. Some projects only want authorized committers to add signed-off-by, so I didn't want to presume.

Sorry if there are too many PRs- I tend to try and separate them into as many logical pieces as possible in case some parts are not wanted. If you don't want too many merge commits, Github now provides the option of accepting PRs by cherry-pick/rebase if you want.

But sure, I'll combine these.

namhyung:
Didn't know that github supports that kind of flexibility.  But I still prefer seeing a single PR containing all the related commits with proper changelogs.

In the kernel, the S-o-b is a must and not related to any authority.  I wrote it down on the doc/contributing.md.

PR #20, opened 2016-10-13
Tweaks to build flags

thepaul:
Taken together, this will allow the Debian packaging to add the recommended hardening flags to the various pieces of the final result.


namhyung:
Looks good.  Thanks for your work.

PR #26, opened 2016-10-19
mcount: Disable 'recover' trigger in cygprof_entry

honggyukim:
'recover' trigger assumes that return addresses in the current call
stack are replaced to mcount_exit by mcount_entry in binaries built with
'-pg' option.  So this patch simply disables 'recover' trigger in
cygprof_entry generated by '-finstrument-functions' option.

We may have to consider some cases when -pg and -finstrument-functions
are mixed in a single binary in the future.


honggyukim:
Thanks!

PR #29, opened 2016-10-21
kernel: Honor --kernel-only for replay and graph

honggyukim:
The --kernel-only option was used only to dump kernel functions only.
It can be useful for replay and graph commands as well especially in
some systems that do not exactly match between kernel and user
timestamps.

In that case, we can only see kernel functions by ignoring user funtions
so that timestamps of user functions cannot confuse kernel function
record.

### DIFF
--- a/cmd-graph.c
+++ b/cmd-graph.c
@@ -395,6 +395,11 @@ static int build_graph(struct opts *opts, struct ftrace_file_handle *handle,
 		struct sym *sym = NULL;
 		char *name;
 
+		if (opts->kernel_only && !task->is_kernel) {
+			/* skip user functions if --kernel-only is set */
+			continue;
+		}
+
 		if (opts->kernel_skip_out) {
 			/* skip kernel functions outside user functions */
 			if (!task->user_stack_count &&
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -572,6 +572,11 @@ int command_replay(int argc, char *argv[], struct opts *opts)
 	while (read_rstack(&handle, &task) == 0 && !ftrace_done) {
 		uint64_t curr_time = task->rstack->time;
 
+		if (opts->kernel_only && !task->is_kernel) {
+			/* skip user functions if --kernel-only is set */
+			continue;
+		}
+
 		/*
 		 * data sanity check: timestamp should be ordered.
 		 * But print_graph_rstack() may change task->rstack
--- a/utils/fstack.c
+++ b/utils/fstack.c
@@ -1310,11 +1310,13 @@ retry:
 user:
 		utask->rstack = &utask->ustack;
 		task = utask;
+		task->is_kernel = false;
 	}
 	else {
 kernel:
 		ktask->rstack = &ktask->kstack;
 		task = ktask;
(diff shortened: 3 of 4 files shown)


honggyukim:
@namhyung This is for avoiding timestamp issues in some ARM target boards so that we can only see kernel function records. I just made a pull request for using --kernel-only option for replay and graph commands.

However, I need to also handle backtrace information in graph commands. The backtrace still contains user function record even if I put --kernel-only option. So I would be grateful if you can take a look at it. Thanks!

namhyung:
Or you can use the existing `is_kernel_address()` helper.

namhyung:
You may want to add support for report as well.  In fact I'd like to be in sync between them as much as possible.  Also please update the documentation. :)

namhyung:
@honggyukim For graph backtrace, you can use this patch

```
diff --git a/cmd-graph.c b/cmd-graph.c
index 237798e..efce17f 100644
--- a/cmd-graph.c
+++ b/cmd-graph.c
@@ -44,6 +44,7 @@ struct task_graph {
 
 struct uftrace_graph {
        char *func;
+       bool kernel_only;
        struct ftrace_session *sess;
        struct uftrace_graph *next;
        struct graph_backtrace *bt_curr;
@@ -69,9 +70,17 @@ static int create_graph(struct ftrace_session *sess, void *func)
        return 0;
 }
 
-static void setup_graph_list(char *func)
+static void setup_graph_list(struct opts *opts, char *func)
 {
+       struct uftrace_graph *graph;
+
        walk_sessions(create_graph, func);
+
+       graph = graph_list;
+       while (graph) {
+               graph->kernel_only = opts->kernel_only;
+               graph = graph->next;
+       }
 }
 
 static struct uftrace_graph * get_graph(struct ftrace_task_handle *task)
@@ -128,15 +137,21 @@ out:
 static int save_backtrace_addr(struct task_graph *tg)
 {
        int i;
+       int skip = 0;
        int len = tg->task->stack_count;
        unsigned long addrs[len];
        struct graph_backtrace *bt;
 
+       if (tg->graph->kernel_only) {
+               skip = tg->task->user_stack_count;
+               len -= skip;
+       }
+
        if (len == 0)
                return 0;
 
        for (i = len - 1; i >= 0; i--)
-               addrs[i] = tg->task->func_stack[i].addr;
+               addrs[i] = tg->task->func_stack[i + skip].addr;
 
        list_for_each_entry(bt, &tg->graph->bt_list, list) {
                if (len == bt->len &&
@@ -387,7 +402,7 @@ static int build_graph(struct opts *opts, struct ftrace_file_handle *handle,
        struct uftrace_graph *graph;
        uint64_t prev_time = 0;
 
-       setup_graph_list(func);
+       setup_graph_list(opts, func);
 
        while (!read_rstack(handle, &task) && !ftrace_done) {
                struct ftrace_ret_stack *frs = task->rstack;
```

honggyukim:
@namhyung I saw your backtrace patch now so I will update the new pull request again.

honggyukim:
@namhyung The backtrace patch works fine so I made a new pull request #30.
I will close this PR. Thanks for your help!

PR #34, opened 2016-11-20
record: Fix to handle shared object executable

honggyukim:
Since some shared object files are executable, check_binary has to allow
it can be traced without --force option.

### DIFF
--- a/cmd-record.c
+++ b/cmd-record.c
@@ -1281,7 +1281,7 @@ static void check_binary(struct opts *opts)
 	if (read(fd, &e_type, sizeof(e_type)) < 0)
 		pr_err("Cannot read '%s'", opts->exename);
 
-	if (e_type != ET_EXEC && !opts->force)
+	if (e_type != ET_EXEC && e_type != ET_DYN && !opts->force)
 		pr_err_ns(OBJTYPE_MSG, opts->exename);
 
 	if (read(fd, &e_machine, sizeof(e_machine)) < 0)


honggyukim:
I found a problem when running d8 binary that is compiled by ninja build.  After building it, d8 binary is shown as below:

```
$ file d8
d8: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.26, BuildID[sha1]=a1a4f3a51a054f72331c9d667c2bc35c9eedd8fa, not stripped

$ readelf -h d8
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x15220
  Start of program headers:          64 (bytes into file)
  Start of section headers:          1067176 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         9
  Size of section headers:           64 (bytes)
  Number of section headers:         42
  Section header string table index: 41

$ ldd d8
        linux-vdso.so.1 =>  (0x00007fffa439c000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f907d9ba000)
        libv8.so => /home/honggyu/work/v8/git/v8/out.gn/x64.debug.pg/././libv8.so (0x00007f907beaa000)
        libv8_libbase.so => /home/honggyu/work/v8/git/v8/out.gn/x64.debug.pg/././libv8_libbase.so (0x00007f907be72000)
        libv8_libplatform.so => /home/honggyu/work/v8/git/v8/out.gn/x64.debug.pg/././libv8_libplatform.so (0x00007f907be22000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f907ba9a000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f907b78a000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f907b3ba000)
        /lib64/ld-linux-x86-64.so.2 (0x000055c582e03000)
        libicui18n.so => /home/honggyu/work/v8/git/v8/out.gn/x64.debug.pg/././libicui18n.so (0x00007f907b0d2000)
        libicuuc.so => /home/honggyu/work/v8/git/v8/out.gn/x64.debug.pg/././libicuuc.so (0x00007f907aee2000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f907acda000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f907aac2000) 
```

But it can be executed as if it is just a normal execuable file.
```
$ ./d8 -e 'print("Hello")'
Hello
```

namhyung:
Thanks for the patch.  I know many programs were built as a dynamic library for some reason.  But handling the binaries usually required some more changes in the uftrace IIRC.  I'll take a look at what they need more.

honggyukim:
I do not know the exact reasons why some binaries are created as shared object, but this patch simply allows those binaries to be traced as of now. We can examine the other cases later on.

honggyukim:
Please check this again :)

namhyung:
Hmm.. those binaries (built as a shared lib) usually enable some subtle build options like `LD_BIND_NOW`.  But yes, it's not related to this change and can be addressed in eparate patches.

PR #38, opened 2016-12-15
tests: run record sub-processes with --no-pager

bkircher:
Prevent opening a pager during `make test` runs.

This patch allows me to add a `%check` section to a RPM spec file that I am preparing for inclusion in Fedora. Without it, rpmbuild, or a script it runs, will halt with a pager (usually `/usr/bin/less` or something) open during test execution and thus making the build fail.

### DIFF
--- a/tests/t023_replay_filter.py
+++ b/tests/t023_replay_filter.py
@@ -22,7 +22,7 @@ class TestCase(TestBase):
 """)
 
     def pre(self):
-        record_cmd = '%s record -d %s %s' % (TestBase.ftrace, TDIR, 't-allocfree')
+        record_cmd = '%s --no-pager record -d %s %s' % (TestBase.ftrace, TDIR, 't-allocfree')
         sp.call(record_cmd.split())
         return TestBase.TEST_SUCCESS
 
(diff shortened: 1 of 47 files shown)


namhyung:
First, thanks for sending a PR and your packaging work. :)

This looks strange though, because @thepaul reported same issue before and he added `--no-pager` to `TestBase.ftrace` in the commit ce8d4663 so the test cases should run with pager disabled.

bkircher:
> This looks strange though, because @thepaul reported same issue before and he added --no-pager to TestBase.ftrace in the commit ce8d466 so the test cases should run with pager disabled.

I see. Well, the pager only opens when in this rpmbuild environment. When I manually invoke `make test` everything is fine. I'll investigate further until I really understand whats happening here.

bkircher:
Hi! Holidays 🎄☃️🕎🌠✌️️❤️ are over so here we go:

> This looks strange though, because @thepaul reported same issue before and he added --no-pager to TestBase.ftrace in the commit ce8d466 so the test cases should run with pager disabled.

`git tag --contains ce8d466` reveals that @thepaul 's commit is not in 0.6 release which is what I am packaging (and testing) right now.

`%check` section works on current master so my PR is not needed. I'll leave the `%check` section out of the spec file until there is a 0.7 release.

This PR can be closed.

namhyung:
Thank your for comfirmation.  I plan to release a next version next week but it's not gonna be 0.7 :)

PR #47, opened 2016-12-30
doc: Add missing explanation for -r/--time-range

honggyukim:
Since the notation for elapsed time is not written before, this adds the
missing explanation and updates the description a bit more including
live command man page.


honggyukim:
Previously, "<start>~<stop>" was shown "~" only so I also added escape marks.

PR #50, opened 2017-01-01
build: Fix a permission problem in configure

honggyukim:
The current 'make clean' does not clean up the timestamp files that are
generated during 'configure' executed to prevent the permission problems
such as the following example.

  $ sudo ./configure

  $ ./configure --prefix=$HOME/usr
  touch: cannot touch 'check-tstamp': Permission denied

  $ sudo make clean
      ...

  $ ./configure --prefix=$HOME/usr
  touch: cannot touch 'check-tstamp': Permission denied

### DIFF
--- a/check-deps/Makefile
+++ b/check-deps/Makefile
@@ -21,6 +21,7 @@ $(CHECK_LIST): %: __%.c
 
 check-tstamp: PHONY
 	@touch $@
+	@if [ `id -u` -eq 0 ]; then chmod 666 $@; fi
 
 check-clean:
 	@$(RM) $(CHECK_LIST) check-tstamp *.o
--- a/configure
+++ b/configure
@@ -99,6 +99,9 @@ echo override LDFLAGS = $LDFLAGS >> $output
 echo                             >> $output
 echo override srcdir := $srcdir  >> $output
 echo override objdir := $objdir  >> $output
+if [ $(id -u) -eq 0 ]; then
+    chmod 666 $output
+fi
 
 if [ "$srcdir" != "$objdir" ]; then
     cat > $objdir/Makefile <<EOF
@@ -129,4 +132,7 @@ test: all
 
 .PHONY: all clean prepare test install
 EOF
+    if [ $(id -u) -eq 0 ]; then
+        chmod 666 $objdir/Makefile
+    fi
 fi


namhyung:
I'd like to keep the configuration after `make clean`.  Users might get confused if their settings changed when rebuild.  It seems that the problem occurs only when first configure was called using sudo and then subsequent configure was from a normal user.  IOW the other way around has no problem.

I'd suggest replacing it with a new file - simply `rm -f` and `touch` would work - in this case.  Or maybe you can add write permission to the check-tstamp file so that normal users can update the timestamp.

honggyukim:
Thanks for the comment. If you agree with the below change, I will create another PR.
```
index 681d358..231cd8a 100755
--- a/configure
+++ b/configure
@@ -84,6 +84,10 @@ if [ ! -e ${srcdir}/check-deps/have_libelf ]; then
     exit 1
 fi

+touch $output
+if [ $(id -u) -eq 0 ]; then
+  chmod 666 $output
+fi
 echo "# this file is generated automatically" > $output
 echo override prefix := $prefix  >> $output
 echo override bindir := $bindir  >> $output
```
I tested that this works fine for me.

namhyung:
But AFAIK the output is `.config` not `check-tstamp`, how does it fixes the problem?

honggyukim:
Oops, I've changed my patch with recent updates in the master. But it contains all the commits in between. Do you have any ideas to properly upload the updated patch?

honggyukim:
Or should I create another PR instead?

namhyung:
It seems you need to rebase your branch onto the current master.

honggyukim:
@namhyung Thanks, there was a tiny mistake so I've just updated it again.

honggyukim:
The permission error comes from not only check-tstamp but also .config file.  I didn't put the error message of .config but it shows the following error message as well.
```
$ ./configure
./configure: line 92: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 93: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 94: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 95: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 96: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 97: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 98: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 99: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 100: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 101: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 102: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 103: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 104: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 105: /home/honggyu/work/uftrace/.config: Permission denied
./configure: line 106: /home/honggyu/work/uftrace/.config: Permission denied
```

honggyukim:
Thanks for your comment. I've updated as you mentioned so please have a look at it again.

Issue #40, opened 2016-12-21
Assertion error: Assertion `sess' failed when reading multithreaded data


tahini:
I don't know if it's me, a bug or just lack of documentation, but I get this message when trying to 'uftrace report' on data from a multithreaded application (compiled with -lpthread):

`uftrace: /home/me/sources/uftrace/utils/session.c:198: add_session_ref: Assertion `sess' failed.`

This is my task.txt file. It appears to have no way to identify the session for tid 9461. Shouldn't the SESS argument be the pid instead of the tid?

```
[me@me multithread]$ cat uftrace.data/task.txt 
SESS timestamp=2525026.536462738 tid=9460 sid=c979e932a604f4f8 exename="/home/me/sources/test/multithread/multithread"
TASK timestamp=2525026.536514704 tid=9460 pid=9457
TASK timestamp=2525026.536579834 tid=9461 pid=9457
```

I can work on a fix if need be.

namhyung:
@tahini hi, thanks for using uftrace and reporting the bug.

Looking at the task file, the task 9460 seems not a main thread, right?  I'm not sure why task 9457 (the main thread, I guess) was not recorded and the session came from task 9460.  If your program is not too big, could you please share it to investigate more?

tahini:
oh the program is not big at all! It's a toy program I did just to test and benchmark uftrace for students that may work with it next semester. 

See here for the code: https://git.dorsal.polymtl.ca/~gbastien/?p=testProjects;a=blob;f=multithread/main.cpp;h=5f1bc834a866b486136bf2b0ddcffd1f6339030c;hb=c12d510fc0261bbaf66280d6a2a00fed73e107b0

The sample I ran had 2 threads created (n=2), probably 9460 and 9461. The main thread did not leave the main function. Maybe that's it?

namhyung:
Thanks for sharing.  Unfornately it works well on my system (I also changed to use 2 threads) like below.  How did you compile your program?  Is it reproducible?

```
$ g++ -o multithread -pg multithread_main.cpp -pthread

$ uftrace --report multithread
return 887459712#
# uftrace report
#
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================================
   14.032  s    9.151  s           2  effect
    7.118  s    6.477 us           1  main
    7.118  s    7.118  s           2  pthread_join
    4.881  s    4.881  s   100000000  bar
  185.632 us    0.712 us           1  iter
[...]

PR #52, opened 2017-01-09
session: Save sessions per pid instead of tid

tahini:
Prevents issue #40 from happening in the future, where a task does not have
a session because the session is identified by tid instead of pid.

Signed-off-by: Geneviève Bastien <gbastien@versatic.net>

### DIFF
--- a/utils/data-file.c
+++ b/utils/data-file.c
@@ -104,6 +104,7 @@ int read_task_txt_file(char *dirname, bool needs_session, bool sym_rel_addr)
 	struct ftrace_msg_task task;
 	struct ftrace_msg_sess sess;
 	char *exename, *pos;
+	char dummy[2];
 
 	xasprintf(&fname, "%s/%s", dirname, "task.txt");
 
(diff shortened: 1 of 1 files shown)


tahini:
Sorry about that, I'm not very familiar with c, but indeed, if string pattern specifiers are possible, it will be much clearer. I'll work on that and the tests.
How do you prefer me to do the changes? Additional commits on this branch or a new pull request?

namhyung:
Thanks for doing this.  I prefer another pull request with "v2" in the subject.

Also you can get rid of the dummy array by adding "*" to the pattern. :)

namhyung:
Please check the current master.  I simplified t010 testcase for your convenience. :)

tahini:
Thanks for the tips. 

Unfortunately, with the changes you did for 0.6.1, I cannot seem to be able to generate uftrace data where there is no task for the pid. Maybe it is fixed for good? Or I just don't have enough imagination to figure something... Anyway, I'll open a new pull-request for the pid on the session line, with a test with previous data with tid.

PR #57, opened 2017-01-20
Dump: Adding nanoseconds precision for "ts" field in Trace Event Format output

abenbachir:
chrome://tracing support nanoseconds.

### DIFF
--- a/cmd-dump.c
+++ b/cmd-dump.c
@@ -533,8 +533,8 @@ static void print_chrome_task_rstack(struct uftrace_dump_ops *ops,
 
 	if (frs->type == FTRACE_ENTRY) {
 		ph = 'B';
-		pr_out("{\"ts\":%"PRIu64",\"ph\":\"%c\",\"pid\":%d,\"name\":\"%s\"",
-			frs->time / 1000, ph, task->tid, name);
+		pr_out("{\"ts\":%"PRIu64".%03d,\"ph\":\"%c\",\"pid\":%d,\"name\":\"%s\"",
+			frs->time / 1000, frs->time % 1000, ph, task->tid, name);
 		if (frs->more) {
 			str_mode |= HAS_MORE;
 			get_argspec_string(task, spec_buf, sizeof(spec_buf), str_mode);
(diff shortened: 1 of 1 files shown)


abenbachir:
I wanted to use defaultTimeUnit:"ns" configuration discussed at this thread https://github.com/google/trace-viewer/issues/624. but i think adding decimal point precision is more simple.

abenbachir:
There is an bug happing in chrome://tracing with the decimal precision as in the picture 
trace-with-ns-decimal-precision

When I removed the decimal precision and used only nanseconds "frs->time" with "defaultTimeUnit":"ns"  the call stack aligned very will see picture
trace-with-defaulttimeunit-ns

This is the both trace file related to both previous pictures [trace-decimal-precision.json](https://gist.github.com/abenbachir/da7543d6f796348b37b09e74308d92bf), [trace-with-defaultTimeUnit.json](https://gist.github.com/abenbachir/dc9d94e638c1885eb82d0b9ab52cfd59)


@namhyung I think we should use the attribute "defaultTimeUnit":"ns" and only use nanoseconds what do you think ?.

namhyung:
@honggyukim Could you please take a look?

abenbachir:
@namhyung it looks like @honggyukim is not responding

namhyung:
@abenbachir sorry for late response.

I was off due to the one of the biggest holiday in Korea - luna new year.  I bet @honggyukim is same.  I'll try to check it by myself today.  Thanks!

honggyukim:
@abenbachir Sorry for answering late. I will check it and answer back to you today.

honggyukim:
@abenbachir Please apply the patch below, then you will see the problem is gone.
```
diff --git a/cmd-dump.c b/cmd-dump.c
index 572805d..0496366 100644
--- a/cmd-dump.c
+++ b/cmd-dump.c
@@ -598,7 +598,7 @@ static void print_chrome_footer(struct uftrace_dump_ops *ops,
        ctime_r(&statbuf.st_mtime, buf);
        buf[strlen(buf) - 1] = '\0';
 
-       pr_out("\n], \"metadata\": {\n");
+       pr_out("\n], \"displayTimeUnit\": \"ns\", \"metadata\": {\n");
        if (handle->hdr.info_mask & (1UL << CMDLINE))
                pr_out("\"command_line\":\"%s\",\n", handle->info.cmdline);
        pr_out("\"recorded_time\":\"%s\"\n", buf);
```
I found that the original name was "defaultTimeUnit":"ns" but it has been renamed to "displayTimeUnit":"ns". So please test it and let me know if it works for you.

honggyukim:
@abenbachir You can find the formal document for trace event format from the link below:

https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit

namhyung:
@honggyukim thanks for checking!

@abenbachir please also add your sign-off in the commit message after reading doc/contributing.md, thanks!

abenbachir:
Thanks @honggyukim for the help it works now.
@namhyung I pushed my changes.

namhyung:
@abenbachir thanks for your persistence!

PR #58, opened 2017-01-27
Minor refactoring about color setting

Taeung:
Hi, Namhyung :smile:

I send minor commits about color setting

  - Add new enum color_setting
  - Make setup_color function be more compact

Thanks,
Taeung


Taeung:
Thank you!!

PR #59, opened 2017-02-03
ftrace: Change warning msg color from red to yellow

Taeung:
Currently not only __pr_err but also __pr_warn
use red color. So I suggest changing to yellow color.
And I think it is better to add a prefix "WARN:" such as pr_err


Taeung:
I understood! :smile: 
I resend a changed commit.

Thanks,
Taeung

PR #60, opened 2017-02-03
record: Check trace functions regardless of '--force' option

Taeung:
Util now only if force option isn't used,
we check trace functions. But I think
it is better to check them whether using force option or not.

Because the user should know that their binary file was
compiled without -pg or -finstrument-functions flag
and therefore the output could be a just series of library calls.


Taeung:
I have second thoughts about this waring.
I think you are right ! :smile: 

Thanks,
Taeung

namhyung:
Yes please.

PR #68, opened 2017-02-15
uftrace: Remove "+" prefix in --time-range option

honggyukim:
-r/--time-range is used to specify time range either timestamp or
elapsed time.  The given data is parsed as timestamp by default or
parsed as elapsed time when it has "+" prefix and <time_unit> postfix,
for example '+100ms'.

But it'd be better to remove "+" prefix to identify whether it is
timestamp or elapsed time.  It can automatically be identified only with
<time_unit> postfix.

So this patch removes "+" prefix for time range parsing and also updates
relevant man pages.


Taeung:
IMHO, I agree this PR
Because "+" letter can cause confusion from newbie's viewpoint.
I thought "+100ms" means "more 100ms" before.
So I thought the user can just use `+100ms` without "~" when using `time-range` option.
(But now I know the correct usage for `time-range` option)
If we remove "+", it is better to use `time-range` without some confusion.

uftrace.c
```diff
@@ -309,6 +309,15 @@ static uint64_t parse_time(char *arg)
 	return val;
 }
 
+static bool is_time_unit(const char *str)
+{
+	/* <time_unit>  :=  "ns" | "us" | "ms" | "s" */
+	if (str[strlen(str) - 1] == 's')
+		return true;
```
namhyung: Note that time unit can be "usec" and so on.  I think it's better to use `isalpha()`.  Also I think the name should be "has_time_unit"..

namhyung:
I fixed it up and pushed, thanks!

Issue #66, opened 2017-02-14
Display time according to --time-range option


Taeung:
Good morning, Namhyung :smile: 

I found a problem. Is it a problem ? or I did wrong usage for it ? :worried: 
I tested on current master branch (58639c)
```
$ uftrace ./a.out
# DURATION    TID     FUNCTION
   0.834 us [16683] | __monstartup();
   0.419 us [16683] | __cxa_atexit();
            [16683] | main() {
   5.181 us [16683] |   fgets();
            [16683] |   get_values_from() {
   0.576 us [16683] |     strdup();

... (It hasn't any problems)
```
But if using `-r` or `--time-range` with `+` prefix, it isn't work with a warning message as below.
```
$ uftrace -r +1us ./a.out
invalid time range: +1us (ignoring...)
# DURATION    TID     FUNCTION
   0.635 us [16628] | __monstartup();
   0.404 us [16628] | __cxa_atexit();
            [16628] | main() {
   4.102 us [16628] |   fgets();
            [16628] |   get_values_from() {
   0.561 us [16628] |     strdup();
...
```
Should we fix it ? and
Add test scripts for `time-range` such as `t053_filter_time.py` ? :smile:

namhyung:
@Taeung Hi, the `--time-range` option requires 2 time values separate by `~`.  You can omit one of them but not the `~`.  So you should use `+1us~` (if you want to see functions after 1us from the beginning) or `~+1us` (to see functions before 1us from the beginning) instead.

Taeung:
Ah, I understood it.

After `uftrace record`
```
$ uftrace record ./a.out
```
In accardance  with the **elapsed time**, we can see **a part** of function graph as below.
```
$ uftrace replay -f elapsed -r +20us~
#  ELAPSED    FUNCTION
  20.034 us |     } /* strdup */
  20.206 us |     strsep();
  20.432 us |     atoi();
  20.721 us |     atoi();
  20.954 us |   } /* get_values_from */
  21.060 us |   fgets();
  21.302 us |   get_values_from() {
...
```
But I think **duration** of default view can cause confusion from the point of view of newbies.
For example, If the user just do `uftrace replay -r +20us~` after recording,
```
$ uftrace replay -r +20us~
# DURATION    TID     FUNCTION
            [22335] |     } /* strdup */
   0.091 us [22335] |     strsep();
   0.150 us [22335] |     atoi();
   0.122 us [22335] |     atoi();
[...]

PR #72, opened 2017-03-02
Display timestamp or elapsed time by default when using time-range

Taeung:
Hi Namhyung :smile: 

This is small PR mentioned at #66 
Additionally I used `enum replay_field_id` instead of just integer values
for **index of field_table array** in order to better readability.

### DIFF
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -17,7 +17,7 @@ static int column_index;
 static int prev_tid = -1;
 
 enum replay_field_id {
-	REPLAY_F_NONE           = 0,
+	REPLAY_F_NONE           = -1,
 	REPLAY_F_DURATION,
 	REPLAY_F_TID,
 	REPLAY_F_ADDR,
@@ -172,6 +172,7 @@ static void print_header(void)
 	pr_out(" FUNCTION\n");
 }
 
+/* index of this table should be matched to replay_field_id */
 struct replay_field *field_table[] = {
 	&field_duration,
 	&field_tid,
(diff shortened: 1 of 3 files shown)


cmd-replay.c
```diff
@@ -239,6 +239,13 @@ static void setup_field(struct opts *opts)
 		s++;
 	}
 
+	if (opts->range.start > 0) {
+		if (opts->range.start_elapsed)
+			add_field(field_table[5]);
+		else
+			add_field(field_table[3]);
+	}
```
namhyung: The range can be set "stop" part only.  Also I'd like to add the time field only if user didn't set replay fields explicitly.  IOW, if user gave `-f` option, it should be honored as is IMHO.

cmd-replay.c
```diff
@@ -234,16 +234,16 @@ static void setup_field(struct opts *opts)
 
 	if (*str == '+') {
 		/* prepend default fields */
-		add_field(field_table[0]);  /* duration */
-		add_field(field_table[1]);  /* tid */
+		add_field(field_table[REPLAY_F_DURATION]);
+		add_field(field_table[REPLAY_F_TID]);
 		s++;
```
namhyung: OK, then I think it's better to add a coment on the field_table[] that it should match to replay_field_id.

Taeung: I got it.
will send v2 with the commit!

Taeung:
@namhyung  I sent v2 for this PR, too. I'd appreciate some feedback. 😄

namhyung:
It seems github not sending alarm on pushing a new code.  It'd be nice you could ping me like this. :)

Taeung:
@namhyung I also didn't know that github not sending alarm when PR is changed.. Okey I got it! 😄

cmd-replay.c
namhyung: It seems like you missed my comment on this: you need to check opts->range.stop also.  And I think you need to change default value of opts->fields (to `NULL`?) so that you can determine whether the field string is given by user or not.  If user gives field option, then it should ignore this logic and show the user-given fields only.

Taeung: @namhyung Ah.. I missed the above comment you wrote 20 days ago...
I understood what you said. I missed the case the user want to see output on particular fields they assign.
Will fix it as v3 !

Taeung:
@namhyung Could you check this PR v3 ? I think this PR can properly handle fields when user give particular fields with `-f` option.

Issue #56, opened 2017-01-18
can not find symbol in mcount support (on ARM)


iamwonseok:
I ran the ./runtest.py on machine. Most of test cases are passed but I got the error on two test cases as below:

root@kesl:~/uftrace/tests# ./runtest.py 076
Test case             pg             finstrument-fu
--------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6efd631
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f2b65f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f1565f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f4665f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f7a65d
076 lib_replay_F    : NG NG NG NG NG NG NG NG NG NG

root@kesl:~/uftrace/tests# ./runtest.py 077
Test case             pg             finstrument-fu
--------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6ef6631
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f2165f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f2665f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6fba65f
uftrace: /root/uftrace/arch/arm/mcount-support.c:263:mcount_arch_parent_location
 ERROR: cannot find symbol for b6f5b65d
077 lib_replay_T    : NG NG NG NG NG NG NG NG NG NG

I am not sure whether the error is my configuration problem or a bug.
Could you let me know what the problem is?

namhyung:
[...]

PR #73, opened 2017-03-07
symbol: Fix a symbol-finding bug on ARM

iamwonseok:
Hi @namhyung 

This is PR for #56.
I ran the record and replay command on both of arm32 and x86_64 using same data file from arm32.

**get uftrace.data on arm32**

| VMSPLIT   |      ARM      |  x86_64 |
|:----------:|:-----:|:-----:|
| 1G |  OK | OK |
| 2G |  OK | OK |
| 3G | OK | OK |
| 3GOPT | OK | OK |

**get uftrace.data on x86_64**

| VMSPLIT   |      ARM      |  x86_64 |
|:----------:|:-----:|:-----:|
| FIXED | ERR | OK |

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -910,6 +910,7 @@ static void mcount_startup(void)
 
 	mcount_exename = read_exename();
 	record_proc_maps(dirname, session_name(), &symtabs);
+	set_kernel_base(dirname, session_name());
 	load_symtabs(&symtabs, NULL, mcount_exename);
 
 #ifndef DISABLE_MCOUNT_FILTER
--- a/utils/data-file.c
+++ b/utils/data-file.c
@@ -11,6 +11,7 @@
 #include "uftrace.h"
 #include "utils/utils.h"
 #include "utils/fstack.h"
+#include "utils/symbol.h"
 #include "libmcount/mcount.h"
 
 
@@ -137,6 +138,8 @@ int read_task_txt_file(char *dirname, bool needs_session, bool sym_rel_addr)
 
 			sscanf(line + 5, "timestamp=%lu.%lu %*[^i]id=%d sid=%s",
 			       &sec, &nsec, &sess.task.pid, (char *)&sess.sid);
+
+			set_kernel_base(dirname, (char *)&sess.sid);
 
 			// Get the execname
 			pos = strstr(line, "exename=");
--- a/utils/symbol.c
+++ b/utils/symbol.c
@@ -15,6 +15,7 @@
 #include <gelf.h>
 #include <unistd.h>
 #include <assert.h>
+#include <limits.h>
 
 /* This should be defined before #include "utils.h" */
 #define PR_FMT     "symbol"
@@ -1276,13 +1277,6 @@ size_t count_dynsym(struct symtabs *symtabs)
 	return dsymtab->nr_sym;
 }
 
-unsigned long get_real_address(unsigned long addr)
-{
-	if (is_kernel_address(addr))
-		return addr | (-1UL << KADDR_SHIFT);
-	return addr;
-}
-
 struct sym * find_symtabs(struct symtabs *symtabs, unsigned long addr)
 {
 	struct symtab *stab = &symtabs->symtab;
@@ -1429,4 +1423,55 @@ void print_symtabs(struct symtabs *symtabs)
 		       i, sym->addr, name, sym->size);
 		symbol_putname(sym, name);
 	}
+}
+
+static unsigned long kernel_base_addr;
+
+static unsigned long get_kernel_base(char *str)
+{
+	unsigned long addr = strtoul(str, NULL, 16);
+
+	if (addr < 0x40000000UL) {
+		return 0x40000000UL;
+	} else if (addr < 0x80000000UL) {
+		return 0x80000000UL;
+	} else if (addr < 0xB0000000UL) {
+		return 0xB0000000UL;
+	} else if (addr < 0xC0000000UL) {
+		return 0xC0000000UL;
+	} else {
+		return 0x800000000000ULL;
+	}
+}
+
+void set_kernel_base(char *dirname, const char *session_id)
+{
+	FILE *fp;
+	char buf[4096];
+	char line[200];
+
+	snprintf(buf, sizeof(buf), "%s/sid-%s.map", dirname, session_id);
+
+	fp = fopen(buf, "r");
+	if (fp == NULL)
+		pr_err("open %s file failed", buf);
+
+	while (fgets(line, sizeof(line), fp) != NULL) {
+		if (strstr(line, "[stack") != NULL) {
+			kernel_base_addr = get_kernel_base(line);
+		}
+	}
+	fclose(fp);
+}
+
+bool is_kernel_address(unsigned long addr)
+{
+	return addr >= kernel_base_addr;
+}
+
+unsigned long get_real_address(unsigned long addr)
+{
+	if (is_kernel_address(addr) && kernel_base_addr > UINT_MAX)
+		return addr | (-1ULL << KADDR_SHIFT);
+	return addr;
 }
--- a/utils/symbol.h
+++ b/utils/symbol.h
@@ -74,11 +74,9 @@ struct symtabs {
 # define KADDR_SHIFT  31
 #endif
 
-static inline bool is_kernel_address(unsigned long addr)
-{
-	return !!(addr & (1UL << KADDR_SHIFT));
-}
+bool is_kernel_address(unsigned long addr);
 unsigned long get_real_address(unsigned long addr);
+void set_kernel_base(char *dirname, const char *session_id);
 
 struct sym * find_symtabs(struct symtabs *symtabs, unsigned long addr);
 struct sym * find_symname(struct symtab *symtab, const char *name);


libmcount/mcount.c
```diff
@@ -909,7 +909,9 @@ static void mcount_startup(void)
 		symtabs.flags &= ~SYMTAB_FL_SKIP_DYNAMIC;
 
 	mcount_exename = read_exename();
-	record_proc_maps(dirname, session_name(), &symtabs);
+	const char *session_id = session_name();
+	record_proc_maps(dirname, session_id, &symtabs);
+	set_kaddr(dirname, session_id);
```
namhyung: I don't like putting declarations in the middle.  Actually `session_name()` is intended to be called directly - IOW it's ok to call the function multiple times.  So you can just add:

```
set_kaddr(dirname, session_name());
```

utils/data-file.c
```diff
@@ -132,12 +133,14 @@ int read_task_txt_file(char *dirname, bool needs_session, bool sym_rel_addr)
 			create_task(&task, true, needs_session);
 		}
 		else if (!strncmp(line, "SESS", 4)) {
-			if (!needs_session)
-				continue;
-
```
namhyung: I think you should keep this check as is.  If `needs_session` is false, no need to set kaddr IMHO.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
 		symbol_putname(sym, name);
 	}
 }
+
+static unsigned long kaddr;
```
namhyung: I'd like to have a more descriptive name for this.  How about `kernel_base_addr`?

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+
+static unsigned long get_kaddr(char *str)
+{
+	char *pch;
+	unsigned long addr;
+
+	pch = strtok (str,"-");
+	addr = strtoul(pch, NULL, 16);
```
namhyung: I don't think the `strtok()` is needed, just calling `strtoul(str, NULL, 16)` should do the job IMHO.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+		return 0xB0000000UL;
+	} else if (addr < 0xC0000000UL) {
+		return 0xC0000000UL;
+#if __SIZEOF_LONG__ == 8
+	} else {
+		return 0x800000000000ULL;
+	}
+#else
+    }
+#endif
```
namhyung: Please do not use the `#if` directives for a different block boundary.  In fact, this function is meaninful only for 32-bit (ARM) systems so that you can omit the 64-bit part entirely IMHO.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+	FILE *fp;
+	char buf[4096];
+	char line[200];
+
+	snprintf(buf, sizeof(buf), "%s/sid-%s.map", dirname, session_id);
+
+	fp = fopen(buf, "r");
+
+	if (fp == NULL)
+		exit(1);
```
namhyung: I suggest to remove a blank line between `fopen()` and the if statement.  Also please you `pr_err()` instead of the plain `exit()` so that it can show the error message on failure.

In addition, you can share a single buffer and get rid of `line`.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+
+	while (fgets(line, sizeof(line), fp) != NULL) {
+		if (strstr(line, "[stack") != NULL) {
+			kaddr = get_kaddr(line);
+			break;
```
namhyung: I think we need to continue checking all mappings even if it found a "[stack" line.  A multi-threaded process might have multiple stacks for each thread.  So it'd be better to use the last stack address.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+		}
+	}
+	pclose(fp);
```
namhyung: Did you mean `fclose()`?

utils/symbol.c
namhyung: Also getter/setter functions should be renamed.  Like `get_kernel_base()` and `set_kernel_base()`.

utils/symbol.c
```diff
@@ -1430,3 +1423,71 @@ void print_symtabs(struct symtabs *symtabs)
...
+unsigned long get_real_address(unsigned long addr)
+{
+	if (is_kernel_address(addr)) {
+#if __SIZEOF_LONG__ == 8
+		return (kaddr == 0x800000000000ULL) ? 
+                    (addr | (-1UL << KADDR_SHIFT)) : 
+                    (addr | kaddr);
+#else
+		return addr | kaddr;
+#endif
```
namhyung: This function is meaningless for 32-bit systems.  We save function addresses into 48-bit to save disk space.  So it should be extended to 64-bit on those systems to be check the address correctly.  However 32-bit systems don't have such issues so it's ok to return the given address as is.  The only thing needed to 32-bit systems is a correct `is_kernel_address()` function.

iamwonseok:
Hi @namhyung 
Thank you for your advise.

I modified the code by your comment and sent the PR.

utils/symbol.c
```diff
@@ -1430,3 +1423,61 @@ void print_symtabs(struct symtabs *symtabs)
...
+	FILE *fp;
+	char buf[4096];
+	char line[200];
+
+	snprintf(buf, sizeof(buf), "%s/sid-%s.map", dirname, session_id);
+
+	fp = fopen(buf, "r");
+	if (fp == NULL) {
+		if (errno == EEXIST)
+			return;
```
namhyung: I think EEXIST cannot happen for `fopen(xxx, "r")`.

utils/symbol.c
```diff
@@ -1430,3 +1423,61 @@ void print_symtabs(struct symtabs *symtabs)
...
+		pr_err("cannot open %s file", buf);
+	}
+
+	while (fgets(line, sizeof(line), fp) != NULL) {
+		if (strstr(line, "[stack") != NULL) {
+			kernel_base_addr = get_kernel_base(line);
+			break;
```
namhyung: Again, please remove `break` here.

utils/symbol.c
```diff
@@ -1430,3 +1423,61 @@ void print_symtabs(struct symtabs *symtabs)
...
+
+unsigned long get_real_address(unsigned long addr)
+{
+	if (is_kernel_address(addr)) {
+		return (kernel_base_addr == 0x800000000000ULL) ?
+                    (addr | (-1UL << KADDR_SHIFT)) :
+                    (addr | kernel_base_addr);
+	}
+	return addr;
+}
```
namhyung: I think blindly OR-ing the kernel base on VMSPLIT_3GOPT might have problem.  As I said before, 32-bit system can skip it.  What about this?

```
unsigned long get_real_address(unsigned long addr)
{
#if __SIZEOF_LONG__ == 8
        if (is_kernel_address(addr))
                return addr | (-1UL << KADDR_SHIFT);
#endif
        return addr;
}
```

iamwonseok: @namhyung, your suggestion looks better, but I am still thinking about it.

As you mentioned, replay can be run on different machine. IMHO, It means that uftrace can replay any uftrace.data where it records from; e.g., the uftrace.data is recorded on arm32 and it can replay on x86_64.

Applying your suggestion, `uftrace replay --kernel` command can replay the uftrace.data from 64-bit machine only. I am attaching a tracing result on 64-bit machine with 32-bit uftrace.data:

```
uftrace replay --kernel
   4.000 us [ 2456] | __cxa_atexit();
   6.300 us [ 2456] | <4008002c>();
   4.000 us [ 2456] | main();
            [ 2456] | <4008002c>() {
            [ 2456] |   printf() {
   3.900 us [ 2456] |   } /* <4008002c> */
  13.400 us [ 2456] |   <40112108>();
            [ 2456] |   <4011df44>() {
  16.800 us [ 2456] |     <4000a3f4>();
  61.200 us [ 2456] |   } /* <4011df44> */
  34.500 us [ 2456] |   <400f5f2c>();
  33.800 us [ 2456] |   <4001efac>();
  81.900 us [ 2456] |   <4010e1a4>();
            [ 2456] |   <4008002c>() {
   6.400 us [ 2456] |   } /* printf */
 295.900 us [ 2456] | } /* <4008002c> */
   6.200 us [ 2456] | <4008002c>();

task: 2456
[0] <4002cfa8>
```

The reason why I wrote code as below:
```
unsigned long get_real_address(unsigned long addr)
{
       if (is_kernel_address(addr)) {
               return (kernel_base_addr == 0x800000000000ULL) ?
                    (addr | (-1UL << KADDR_SHIFT)) :
                    (addr | kernel_base_addr);
        return addr;
}
```

iamwonseok:
Hi, @namhyung 

Thank you for your feedback.
By the way, I have a question about  `get_real_address()`. Could you please look at my reply?

utils/symbol.c
namhyung: Oh, you're right.  But it reminds me of using 'unsigned long' for address is not safe for accessing 64-bit data on a 32-bit system. :-/

Anyway, I'd like to avoid the OR for 32-bit data.  What about this?

```
#include <limits.h>
...
unsigned long get_real_address(unsigned long addr)
{
        /* sign-extend to full 64-bit address */
        if (is_kernel_address(addr) && kernel_base_addr > UINT_MAX)
                    return addr | (-1ULL << KADDR_SHIFT);
        return addr;
}
```

iamwonseok: Great!

namhyung:
LGTM. Thanks for your endurance! :)

iamwonseok:
Without your help, I can not achieve it.
Thank you! :D

PR #74, opened 2017-03-16
build unittests with -no-pie

thepaul:
Newer releases of Debian and Ubuntu have begun making GCC build with
PIE by default. This is good for security, but breaks special uses
like the lookup of test cases through libelf in "unittests". To be more
specific, without this change, the unittests executable will quickly
segfault on Debian 9 (stretch) or Ubuntu 16.10 (yakkety) or newer
releases.

Add the -no-pie option to the GCC command line when building unittests.

This might break on systems where GCC does not yet understand that
option. I'm not sure yet how best to detect that situation.


namhyung:
Thanks for the report! But I think it'd be better fix the problem rather than using `-no-pie`.  I'll investigate it.

namhyung:
I just pushed the fix into `check/pie-fix` branch.  Could you please verify?

thepaul:
That seems to work! And it does look like a better solution.

namhyung:
Merged. Thanks for the comfirmation!

PR #75, opened 2017-03-16
suppress unittest colored output with #define

thepaul:
I need to be able to disable the ansi color sequences during debian builds. I was addressing this problem previously by piping the test output through a sed script, but that started causing some minor issues because of the buffering (like error messages from failed test runs being printed in the wrong place, because the stderr stream was not being buffered the same way).

You might have a better idea of what the Right Way is to opt out of colored output, but if nothing else, this should do!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ endif
 RM = rm -f
 INSTALL = install
 
-export ARCH CC AR LD RM srcdir objdir
+export ARCH CC AR LD RM srcdir objdir CFLAGS_unittest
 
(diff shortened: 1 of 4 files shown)


namhyung:
I'd rather add an option to suppress the color output.

thepaul:
That would be perfect.

namhyung:
Push the code to `master`.  Please check it. :)

thepaul:
Looks good to me! This will make the packaging for the next release quite a bit simpler.

PR #76, opened 2017-03-20
filter: Support a float time value for trigger

Taeung:
The time tigger cannot handle a float value
so fix it using existing parse_time().

@namhyung 
I moved parse_time() function into utils/utils.c
But if you don't agree the movement,
I'll resend v2 fixing it.

Before:
```
  $ uftrace record ./hello
  $ uftrace replay -T main@time=1.79us
     1.013 us [19810] | __monstartup();
     0.649 us [19810] | __cxa_atexit();
              [19810] | main() {
    11.739 us [19810] |   puts();
              [19810] |   foo() {
     1.785 us [19810] |     puts();
     2.357 us [19810] |   } /* foo */
    15.040 us [19810] | } /* main */
```
After:
```
  $ uftrace replay -T main@time=1.79us
     1.013 us [19810] | __monstartup();
     0.649 us [19810] | __cxa_atexit();
              [19810] | main() {
    11.739 us [19810] |   puts();
     2.357 us [19810] |   foo();
    15.040 us [19810] | } /* main */
```


Taeung:
Okey, I resend v2 as you said tomorrow !

Taeung:
@namhyung could you check this PR?
I sent v2 according to your comment. :smile:

namhyung:
Oh, I missed that.  Will check!

tests/t135_trigger_time2.py
```diff
@@ -0,0 +1,48 @@
...
+
+        record_cmd = '%s record -F main -d %s %s' % (TestBase.ftrace, TDIR, 't-' + self.name)
+        sp.call(record_cmd.split())
+
+        # find timestamp of function 'malloc'
+        replay_cmd = '%s replay -d %s -F malloc' % (TestBase.ftrace, TDIR)
+        p = sp.Popen(replay_cmd, shell=True, stdout=sp.PIPE, stderr=sp.PIPE)
+        r = p.communicate()[0].decode()
+        START, unit = r.split('\n')[1].split()[0:2] # skip header
+        START = str(float(START) + 0.001) # for time filtering
```
namhyung: I don't think converting float into str is a good idea here.  Also 'START' is not a good name, what about THRESHOLD or something?

tests/t135_trigger_time2.py
```diff
@@ -0,0 +1,48 @@
...
+        START += unit
+        p.wait()
+
+        return TestBase.TEST_SUCCESS
+
+    def runcmd(self):
+        return '%s replay -T main@time=%s -d %s' % (TestBase.ftrace, START, TDIR)
```
namhyung: Here the time filter value might be longer than 3 digits.  For my (failed) test case, it was `-T main@time=1.2799999999999998us`  so I think it's better to print it as float with a precision limit:

```
diff --git a/tests/t135_trigger_time2.py b/tests/t135_trigger_time2.py
index f16230d..268f844 100644
--- a/tests/t135_trigger_time2.py
+++ b/tests/t135_trigger_time2.py
@@ -4,7 +4,8 @@ from runtest import TestBase
 import subprocess as sp
 
 TDIR='xxx'
-START=0
+TIME=0
+UNIT=''
 
 class TestCase(TestBase):
     def __init__(self):
@@ -24,7 +25,7 @@ class TestCase(TestBase):
 """, sort='simple')
 
     def pre(self):
-        global START
+        global TIME, UNIT
 
         record_cmd = '%s record -F main -d %s %s' % (TestBase.ftrace, TDIR, 't-' + self.name)
         sp.call(record_cmd.split())
@@ -33,15 +34,14 @@ class TestCase(TestBase):
         replay_cmd = '%s replay -d %s -F malloc' % (TestBase.ftrace, TDIR)
         p = sp.Popen(replay_cmd, shell=True, stdout=sp.PIPE, stderr=sp.PIPE)
         r = p.communicate()[0].decode()
-        START, unit = r.split('\n')[1].split()[0:2] # skip header
-        START = str(float(START) + 0.001) # for time filtering
-        START += unit
+        TIME, UNIT = r.split('\n')[1].split()[0:2] # skip header
+        TIME = float(TIME) + 0.001 # for time filtering
         p.wait()
 
         return TestBase.TEST_SUCCESS
 
     def runcmd(self):
-        return '%s replay -T main@time=%s -d %s' % (TestBase.ftrace, START, TDIR)
+        return '%s replay -T main@time=%.3f%s -d %s' % (TestBase.ftrace, TIME, UNIT, TDIR)
 
     def post(self, ret):
         sp.call(['rm', '-rf', TDIR])
```

tests/t135_trigger_time2.py
Taeung: Okey, will change the name.

tests/t135_trigger_time2.py
Taeung: @namhyung  Sorry for my insufficient code,
and your changes definitely looks better, will send v2 with it !

Taeung:
@namhyung I sent v2 according to your changes including 'signed off you'. 😄

Taeung:
@namhyung  Oops, I missed changing the commit ID in the change log (https://github.com/Taeung/uftrace/commit/c9eabc6021f295632a70b85d9cf04396196fd936), so fixed it now !

Issue #89, opened 2017-04-21
Implicit fallthrough warning on GCC 7.x


Taeung:
@namhyung 

I found warning messages about `-Wimplicit-fallthrough` on `gcc-7.0.1`,
on current master branch.

The reason is like below.
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
It looks like gcc 7 has added a default warning about fall-through.

But there isn't the warning option in gcc 6:
https://gcc.gnu.org/onlinedocs/gcc-6.3.0/gcc/Warning-Options.html

Is it needed to handle this warning message on `gcc-7.x`?
If you want to handle it, what about this ?

Currently the warning messages occur when building `ufrace` by `gcc-7.0.1` like below.
```
$ gcc -dumpversion
7.0.1

$  git status | head -1
On branch master

$ sudo make clean && sudo make install
...
/home/taeung/git/uftrace/utils/utils.c: In function ‘parse_min’:
/home/taeung/git/uftrace/utils/utils.c:288:12: warning: this statement may fall through [-Wimplicit-fallthrough=]
    decimal *= 10;
    ~~~~~~~~^~~~~
/home/taeung/git/uftrace/utils/utils.c:289:3: note: here
   case 3:
   ^~~~
  CC       arch/x86_64/cpuinfo.o
  CC       arch/x86_64/regs.o
  LINK     arch/x86_64/uftrace.o
  FLAGS:   * new build flags or cross compiler
  CC FPIC  libtraceevent/event-parse.o
/home/taeung/git/uftrace/libtraceevent/event-parse.c: In function ‘process_arg_token’:
/home/taeung/git/uftrace/libtraceevent/event-parse.c:2914:6: warning: this statement may fall through [-Wimplicit-fallthrough=]
   if (strcmp(token, "(") == 0) {
      ^
/home/taeung/git/uftrace/libtraceevent/event-parse.c:2919:2: note: here
  case EVENT_OP:
  ^~~~
  CC FPIC  libtraceevent/event-plugin.o
  CC FPIC  libtraceevent/trace-seq.o
  CC FPIC  libtraceevent/parse-filter.o
/home/taeung/git/uftrace/libtraceevent/parse-filter.c: In function ‘pevent_filter_clear_trivial’:
/home/taeung/git/uftrace/libtraceevent/parse-filter.c:1625:7: warning: this statement may fall through [-Wimplicit-fallthrough=]
    if (filter_type->filter->boolean.value)
       ^
[...]

PR #90, opened 2017-04-23
Add the fallthrough comments to avoid the warning of gcc 7

Taeung:
gcc 7 has added a default warning about fallthrough unlike gcc 6:

  https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

So add the fallthrough comments into the switch statements
that are cause of the below warning. This commit handles #89
```
  /home/taeung/git/uftrace/utils/utils.c: In function ‘parse_min’:
  /home/taeung/git/uftrace/utils/utils.c:288:12: warning: this statement may fall through [-Wimplicit-fallthrough=]
      decimal *= 10;
      ~~~~~~~~^~~~~
  /home/taeung/git/uftrace/utils/utils.c:289:3: note: here
     case 3:
     ^~~~
    CC       arch/x86_64/cpuinfo.o
    CC       arch/x86_64/regs.o
    LINK     arch/x86_64/uftrace.o
    FLAGS:   * new build flags or cross compiler
    CC FPIC  libtraceevent/event-parse.o
  /home/taeung/git/uftrace/libtraceevent/event-parse.c: In function ‘process_arg_token’:
  /home/taeung/git/uftrace/libtraceevent/event-parse.c:2914:6: warning: this statement may fall through [-Wimplicit-fallthrough=]
     if (strcmp(token, "(") == 0) {
        ^
  /home/taeung/git/uftrace/libtraceevent/event-parse.c:2919:2: note: here
    case EVENT_OP:
    ^~~~
    CC FPIC  libtraceevent/event-plugin.o
    CC FPIC  libtraceevent/trace-seq.o
    CC FPIC  libtraceevent/parse-filter.o
  /home/taeung/git/uftrace/libtraceevent/parse-filter.c: In function ‘pevent_filter_clear_trivial’:
  /home/taeung/git/uftrace/libtraceevent/parse-filter.c:1625:7: warning: this statement may fall through [-Wimplicit-fallthrough=]
      if (filter_type->filter->boolean.value)
         ^
  /home/taeung/git/uftrace/libtraceevent/parse-filter.c:1627:3: note: here
     case FILTER_TRIVIAL_TRUE:
     ^~~~
```

### DIFF
--- a/libtraceevent/event-parse.c
+++ b/libtraceevent/event-parse.c
@@ -2916,6 +2916,7 @@ process_arg_token(struct event_format *event, struct print_arg *arg,
 			type = process_paren(event, arg, &token);
 			break;
 		}
+		/* fall through */
 	case EVENT_OP:
 		/* handle single ops */
 		arg->type = PRINT_OP;
--- a/libtraceevent/parse-filter.c
+++ b/libtraceevent/parse-filter.c
@@ -1624,9 +1624,11 @@ int pevent_filter_clear_trivial(struct event_filter *filter,
 		case FILTER_TRIVIAL_FALSE:
 			if (filter_type->filter->boolean.value)
 				continue;
+			break;
 		case FILTER_TRIVIAL_TRUE:
 			if (!filter_type->filter->boolean.value)
 				continue;
+			break;
 		default:
 			break;
 		}
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -304,6 +304,7 @@ static uint64_t parse_min(uint64_t min, uint64_t decimal, int decimal_places)
 			break;
 		case 2:
 			decimal *= 10;
+			/* fall through */
 		case 3:
 			decimal *= 10;
 			nsec += decimal * NSEC_PER_MSEC;


libtraceevent/event-parse.c
```diff
@@ -2905,6 +2905,7 @@ process_arg_token(struct event_format *event, struct print_arg *arg,
 		break;
 
 	case EVENT_DQUOTE:
+		/* fall through */
 	case EVENT_SQUOTE:
```
namhyung: Hmm.. I think this is normal for a switch case.  Does GCC7 shows warning for this too?

Taeung: Sorry, you are right. the above fallthrough comment isn't needed here..
I won't add it.

Taeung:
I got it. I sent v2 now. 😄

libtraceevent/parse-filter.c
```diff
@@ -1624,9 +1624,11 @@ int pevent_filter_clear_trivial(struct event_filter *filter,
 		case FILTER_TRIVIAL_FALSE:
 			if (filter_type->filter->boolean.value)
 				continue;
+			/* fall through */
 		case FILTER_TRIVIAL_TRUE:
 			if (!filter_type->filter->boolean.value)
 				continue;
+			/* fall through */
```
Taeung: Actually gcc 7 didn't show the warning message but I think it is better to add it equally.
What do you think about it ?

namhyung: I also think that it'd be better to add.

namhyung: Wait, I'm not sure what the code does.  Shouldn't it be "break" instead of fall-through?

Taeung: Right. I think so, too.. Use `break` instead of two the fall-through comment ?
```
diff --git a/libtraceevent/parse-filter.c b/libtraceevent/parse-filter.c
index 185774a..feb68e1 100644
--- a/libtraceevent/parse-filter.c
+++ b/libtraceevent/parse-filter.c
@@ -1624,11 +1624,11 @@ int pevent_filter_clear_trivial(struct event_filter *filter,
                 case FILTER_TRIVIAL_FALSE:
                         if (filter_type->filter->boolean.value)
                                 continue;
-                        /* fall through */
+                        break;
                 case FILTER_TRIVIAL_TRUE:
                         if (!filter_type->filter->boolean.value)
                                 continue;
-                        /* fall through */
+                        break;
                 default:
                         break;
                 }
```

Taeung: @namhyung  I think `trivial filter` means it's type is `FILTER_ARG_BOOLEAN`,
and this `pevent_filter_clear_trivial()` function to just remove the only `trivial filters` in all filters.
So I think it is fine to use the fall-through comment and `break;` like above.

But I don't know where `pevent_filter_clear_trivial()` is called in uftrace source code and linux kernel source code..
So weird.

namhyung: It's a library function so any user (external program) can use it.  Maybe trace-cmd?

But I'm not sure the fall-through is right for the FALSE case since it'll be hit always.

Taeung: I understood it, thank you.
I thought same thing about the case `FILTER_TRIVIAL_FALSE:`.
Just use `continue;` like below ? as another PR 
```
diff --git a/libtraceevent/parse-filter.c b/libtraceevent/parse-filter.c
index 21c92fb..60ad0a8 100644
--- a/libtraceevent/parse-filter.c
+++ b/libtraceevent/parse-filter.c
@@ -1622,11 +1622,11 @@ int pevent_filter_clear_trivial(struct event_filter *filter,
                         continue;
                 switch (type) {
                 case FILTER_TRIVIAL_FALSE:
-                        if (filter_type->filter->boolean.value)
-                                continue;
+                        continue;
                 case FILTER_TRIVIAL_TRUE:
                         if (!filter_type->filter->boolean.value)
                                 continue;
+                        break;
                 default:
                         break;
                 }

```

namhyung: This is what current code is dong but I think it's wrong.. :)  You could send a patch to upstream so that Steven Rostedt can confirm.

namhyung: I think you'd better sending the "break" version.

Taeung: @namhyung I sent [the patch](https://lkml.org/lkml/2017/4/26/393) and then Steven Rostedt told me he take this patch
but his remote repository isn't updated yet.

Apply the break version into this PR ? or wait for his upstream ?

namhyung: Please push the "break" version.  I don't apply the upstream commit directly and there may be some updates in the upstream.  Also I prefer adding "break" to the TRUE case too.

Taeung: I understood !
will send changed PR as next version. 😄

Taeung: @namhyung Do you mean that push "break" version as another PR , not this PR ?

namhyung: I mean this PR, thanks!

Taeung:
@namhyung I sent v3 now !

 * Add a commit for "break" version
 * a bit modified commit message
 * rebased on current `master`

Taeung:
@namhyung , I rebased onto the recent `master` branch, again!

PR #93, opened 2017-05-01
Fix dump test scripts

Taeung:
Hi @namhyung ,

I fixed several dump test scripts.
 * Correctly check architecture (64 bit or 32 bit) in fixup()
 * Use fixup() when the target program don't contain `build-id`

I found the problem when using GCC 7 (no default `build-id` option) as below.
(GCC 5 and 6 haven't this problem)
```
$ gcc -dumpversion
7.0.1

$ ./runtest.py dump
097 dump_basic          : NG NG NG NG NG NG NG NG NG NG
098 dump_tid            : NG NG NG NG NG NG NG NG NG NG
099 dump_filter         : NG NG NG NG NG NG NG NG NG NG
100 dump_depth          : NG NG NG NG NG NG NG NG NG NG
...
```
Detailedly
```
$ ./runtest.py -d dump
t097_dump_basic: diff result of -pg -O0
--- expect	2017-05-01 23:25:59.504314659 +0900
+++ result	2017-05-01 23:25:59.504314659 +0900
@@ -4,5 +4,5 @@
-uftrace file header: class         = 1 (32 bit)
+uftrace file header: class         = 2 (64 bit)
 uftrace file header: features      = 0x63 (PLTHOOK | TASK_SESSION | SYM_REL_ADDR | MAX_STACK)
-uftrace file header: info          = 0x3ff
+uftrace file header: info          = 0x3fd
 [entry] depth: 0 main
...
```
The reason is that expected output `info mask` (i.e. `0x3ff` 1111111111) is different from
the result (i.e. `0x3fd` 1111111101), because the test program (i.e. `t-abc`) don't contain `build-id`.

So I fix it and the first patch is to avoid side effect on the above situation.
If you don't want to allow that the target program don't contain `build-id`, I'll close this PR.

### DIFF
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -46,4 +46,8 @@ reading 5231.dat
         return ret
 
     def fixup(self, cflags, result):
-        return result.replace("2 (64 bit)", "1 (32 bit)")
+        import platform
+
+        if platform.architecture()[0] == '32bit':
+            result = result.replace("2 (64 bit)", "1 (32 bit)")
+        return result
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -50,4 +50,7 @@ reading 5231.dat
 
         if platform.architecture()[0] == '32bit':
             result = result.replace("2 (64 bit)", "1 (32 bit)")
+        p = sp.Popen(['file', 't-' + self.name], stdout=sp.PIPE)
+        if 'BuildID' not in p.communicate()[0].decode():
+            result = result.replace("0x3ff", "0x3fd")
         return result
--- a/tests/t098_dump_tid.py
+++ b/tests/t098_dump_tid.py
@@ -58,4 +58,8 @@ reading 5188.dat
         return ret
 
     def fixup(self, cflags, result):
-        return result.replace("2 (64 bit)", "1 (32 bit)")
+        import platform
+
+        if platform.architecture()[0] == '32bit':
+            result = result.replace("2 (64 bit)", "1 (32 bit)")
+        return result
--- a/tests/t098_dump_tid.py
+++ b/tests/t098_dump_tid.py
@@ -62,4 +62,7 @@ reading 5188.dat
 
         if platform.architecture()[0] == '32bit':
             result = result.replace("2 (64 bit)", "1 (32 bit)")
+        p = sp.Popen(['file', 't-' + self.name], stdout=sp.PIPE)
+        if 'BuildID' not in p.communicate()[0].decode():
+            result = result.replace("0x3ff", "0x3fd")
         return result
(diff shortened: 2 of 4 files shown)


Taeung:
I got it. I have three option like below to check build-id of target program.

 * Option 1: As you said, we can use `file` command like below.
```
diff --git a/tests/t097_dump_basic.py b/tests/t097_dump_basic.py
index 45d0800..ece7b53 100644
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -50,4 +50,7 @@ reading 5231.dat
 
         if platform.architecture()[0] == '32bit':
             result = result.replace("2 (64 bit)", "1 (32 bit)")
+        p = sp.Popen(['file','t-' + self.name], stdout=sp.PIPE)
+        if 'BuildID' not in p.communicate()[0].decode():
+            result = result.replace("0x3ff", "0x3fd")
         return result

```
 * Option 2: We can use [pwnlib.elf.elf.ELF.buildid](https://docs.pwntools.com/en/stable/elf/elf.html#pwnlib.elf.elf.ELF.buildid) after installing pwntools like below.

```
$ sudo pip install pwntools
```
```
diff --git a/tests/t097_dump_basic.py b/tests/t097_dump_basic.py
index 45d0800..8ec6f21 100644
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -47,7 +47,11 @@ reading 5231.dat
 
     def fixup(self, cflags, result):
         import platform
+        import pwnlib.elf.elf as pwnelf
 
         if platform.architecture()[0] == '32bit':
             result = result.replace("2 (64 bit)", "1 (32 bit)")
+        target_elf = pwnelf.ELF('t-' + self.name)
+        if not target_elf.buildid:
+            result = result.replace("0x3ff", "0x3fd")
```

 * Option 3: We can check build-id by `readelf -p .note.gnu.build-id`
```
diff --git a/tests/t097_dump_basic.py b/tests/t097_dump_basic.py
index 45d0800..ef7e6be 100644
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -50,4 +50,7 @@ reading 5231.dat
 
         if platform.architecture()[0] == '32bit':
             result = result.replace("2 (64 bit)", "1 (32 bit)")
+        p = sp.Popen(['readelf', '-p', '.note.gnu.build-id', 't-' + self.name], stderr=sp.PIPE)
+        if 'not exist' in p.communicate()[1].decode():
+            result = result.replace("0x3ff", "0x3fd")
         return result
```
What do you think about the above changes ?

namhyung:
I think the option 1 is simplest and portable.

Taeung:
@namhyung I sent v2 now 😄 based on your comment

Issue #100, opened 2017-05-18
PLT hook segfault from <__fentry__@plt> on GCC7


Taeung:
This problem is similar to #92 and #91 .
But this only `-pg -mfentry` case is like below.

 - Target program
```
$ cat  hello.c
#include <stdio.h>

void main()
{
	printf("hello\n");
}
```
 - Failed only `-pg -mfentry` case (with uftrace built by `GCC7`)
(Sure, `normal tracing` and `dynamic tracing` are fine)
```
$ gcc -o hello -pg -mfentry hello.c

$ uftrace ./hello
child terminated by signal: 11: Segmentation fault
# DURATION    TID     FUNCTION
   1.741 us [14660] | __monstartup();
   1.519 us [14660] | __cxa_atexit();

```
 - With `--no-libcall`, there isn't segfault. 
```
uftrace --no-libcall ./hello
hello
# DURATION    TID     FUNCTION
   2.212 us [15019] | main();
```

This problem is related to the 16-byte stack alignment required from GCC7 because of  `movaps`.
```
$ gdb ./hello core
...
Core was generated by `./hello'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa62271d612 in plthook_entry (ret_addr=0x7fff69e186f0, child_idx=4, 
    module_id=140351523635560, regs=0x7fff69e186b0)
    at /home/taeung/git/uftrace/libmcount/plthook.c:429
429		struct ftrace_trigger tr = {
(gdb) disas
Dump of assembler code for function plthook_entry:
   0x00007fa62271d5f0 <+0>:	push   %r15
   0x00007fa62271d5f2 <+2>:	push   %r14
   0x00007fa62271d5f4 <+4>:	push   %r13
   0x00007fa62271d5f6 <+6>:	push   %r12
   0x00007fa62271d5f8 <+8>:	push   %rbp
   0x00007fa62271d5f9 <+9>:	push   %rbx
   0x00007fa62271d5fa <+10>:	pxor   %xmm0,%xmm0
   0x00007fa62271d5fe <+14>:	sub    $0x58,%rsp
   0x00007fa62271d602 <+18>:	lea    0x217318(%rip),%rax        # 0x7fa622934921 <mcount_setup_done>
   0x00007fa62271d609 <+25>:	movq   $0x0,0x40(%rsp)
=> 0x00007fa62271d612 <+34>:	movaps %xmm0,0x20(%rsp)
[...]

PR #101, opened 2017-05-18
mcount: Fix PLT hook segfault about 16-byte alignment on GCC7

Taeung:
I'm not sure you agree this PR, but I'd appreciate some feedback on this. 😄 

If uftrace is built by GCC7, plthook_entry() has `movaps` insn.
So the stack pointer must be aligned on a 16-byte boundary.
This was handled by the commit 7b80ea8c2 ("mcount: Fix segfault on GCC7")

However, the commit didn't handle the below `-pg -mfentry` case:
```
  $ gcc -o hello -pg -mfentry hello.c

  $ uftrace ./hello
     1.741 us [14660] | __monstartup();
     1.519 us [14660] | __cxa_atexit();
```
So fix it, this situation was mentioned on #100.

### DIFF
--- a/arch/x86_64/plthook.S
+++ b/arch/x86_64/plthook.S
@@ -19,6 +19,11 @@ plt_hooker:
 	movq %r9, 8(%rsp)
 	.cfi_offset r9, -64
 
+	/* check stack alignment before calling plthook_entry() */
+	lea (%rsp), %rdi
+	callq is_aligned
+	cmpq $0, %rax
+
 	/* child idx */
 	movq 64(%rsp), %rsi
 	/* address of parent ip */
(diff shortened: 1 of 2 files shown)


namhyung:
As I said in #100, it should be done differently.  Also using a magic value (72) can be a problem later.

Taeung:
I a bit changed this PR,
But I'm not sure that this is good solution of #100 ..

Can we check whether `plt_hooker()` is called from `<__fentry__@plt>` or not ?
Hum.. Or can we use the new trampoline for `__fentry__@plt` ?

namhyung:
Maybe we can check the child ip is `__fentry__` or something, but I don't want to add it to the fast path.

Taeung:
The fast path you said means at the very beginning of `plt_hooker()` ?

And I thought we can know only child idx(i.e. got offset) in plthook.S
because _dl_runtime_resolve() isn't called before plt_hooker().. no ?

What about checking `__fentry__` in `.rela.plt` section when reading target binary ?

PR #102, opened 2017-05-19
mcount: Only call mcount_startup() one time

Taeung:
I found this situation with GDB.

Currently mcount_init() is first called before __monstartup(),
so check mcount_setup_done in __monstartup(), not mcount_init(),
to not call mcount_startup() twice.

Fixes: cb38d982 ("mcount: Staticize startup and cleanup routines")

Or,
check the `mcount_setup_done` variable on both ?

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -1236,12 +1236,10 @@ __visible_default void * dlopen(const char *filename, int flags)
  */
 void __visible_default __monstartup(unsigned long low, unsigned long high)
 {
-	mcount_startup();
 }
 
 void __visible_default _mcleanup(void)
 {
-	mcount_cleanup();
 }
 
 void __visible_default mcount_restore(void)
@@ -1271,8 +1269,7 @@ void __visible_default __cyg_profile_func_exit(void *child, void *parent)
 static void __attribute__((constructor))
 mcount_init(void)
 {
-	if (!mcount_setup_done)
-		mcount_startup();
+	mcount_startup();
 }
 
 static void __attribute__((destructor))


Taeung:
Okey, I modified this PR based on your comment rebasing on current master.

And I wonder why do you let two functions (i.e. mcount_init() and __monstartup()) call mcount_startup().

namhyung:
I don't think it fixes the commit you mentioned since it's not called twice, right?

The `__monstartup()` is called automatically from the binary when it's compiled with `-pg`.  That's how it initialized itself, but I changed to use the constructor mechanism to have the function called before any other library routines.

Taeung:
Currently `mcount_startup()` is called twice like below.

 1. by `mcount_init()`
 2. by `__monstartup()`

```
$ gcc -o hello -pg hello.c

$ gdb uftrace
...
(gdb) set follow-fork-mode child
(gdb) b plt_hooker
Function "plt_hooker" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (plt_hooker) pending.
(gdb) b mcount_startup
Function "mcount_startup" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 2 (mcount_startup) pending.
(gdb) r --no-pager ./hello
...
Thread 2.1 "hello" hit Breakpoint 2, mcount_startup ()
    at /home/taeung/git/uftrace/libmcount/mcount.c:930
930	{
(gdb) bt
#0  mcount_startup () at /home/taeung/git/uftrace/libmcount/mcount.c:930
#1  0x00007ffff7de74ea in call_init (l=<optimized out>, argc=argc@entry=1, 
    argv=argv@entry=0x7fffffffde28, env=env@entry=0x7fffffffde38) at dl-init.c:72
#2  0x00007ffff7de75fb in call_init (env=0x7fffffffde38, argv=0x7fffffffde28, argc=1, 
    l=<optimized out>) at dl-init.c:30
#3  _dl_init (main_map=0x7ffff7ffe168, argc=1, argv=0x7fffffffde28, env=0x7fffffffde38)
    at dl-init.c:120
#4  0x00007ffff7dd7cfa in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
#5  0x0000000000000001 in ?? ()
#6  0x00007fffffffe1da in ?? ()
#7  0x0000000000000000 in ?? ()
(gdb) c
Continuing.

Thread 2.1 "hello" hit Breakpoint 2, mcount_startup ()
    at /home/taeung/git/uftrace/libmcount/mcount.c:930
930	{
(gdb) bt
#0  mcount_startup () at /home/taeung/git/uftrace/libmcount/mcount.c:930
#1  0x00007ffff7bca595 in plt_hooker () from /usr/local/lib/libmcount-fast.so
#2  0x0000000000000000 in ?? ()

```
And I also thought that we can just call `mcount_startup()` one time by `mcount_init()`, not using `__monstartup()`
because a target program can be built without `-pg` or `-finstrument-functions`.
Is it wrong way ?

namhyung:
Okay, please make the `__monstartup()` empty then.

Taeung:
I changed this PR as you said but what do you think about `mcount_cleanup()`
that also is called twice (i.e. by `_mcleanup()` and `mcount_fini()` )?

namhyung:
Hmm. right. please fix it too, thanks!

Taeung:
I got it. I changed this PR, again.
If you want to a bit modify this commit log message, I'll do it. 😄

PR #104, opened 2017-05-25
uftrace: Simple refactoring setup_field() and setup_sort()

Taeung:
Hi @namhyung ,
It is simple refactoring PR about checking invalid field name and sort key.

No check `if (i == ARRAY_SIZE())` each time,
and use `next:` goto statement if valid name or key.

Thanks,
Taeung

### DIFF
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -262,18 +262,16 @@ static void setup_field(struct opts *opts)
 				continue;
 
 			add_field(field);
-			break;
+			goto next;
 		}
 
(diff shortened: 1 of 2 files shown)


namhyung:
Hi @Taeung,

I think it's just a matter of taste, but I'd like to keep the error handling code in a if block.  It makes me feel that it's not in a normal code flow.

Taeung:
Okey, I intended to reduce a if statement but understood your comment !!
will close this PR.

PR #105, opened 2017-06-10
std::string support for argument/retval display

honggyukim:
Since std::string is widely used in C++ programs, so it'd be useful to support it when displaying arguments and return values.
Currently, it only supports std::string in libstdc++ library.

### DIFF
--- a/cmd-dump.c
+++ b/cmd-dump.c
@@ -304,8 +304,9 @@ static void pr_args(struct fstack_arguments *args)
 		if (spec->idx == RETVAL_IDX)
 			continue;
 
-		if (spec->fmt == ARG_FMT_STR) {
-			char buf[64];
+		if (spec->fmt == ARG_FMT_STR ||
+		    spec->fmt == ARG_FMT_STD_STRING) {
+			char buf[64] = {0};
 			const int null_str = -1;
 
 			size = *(unsigned short *)ptr;
@@ -314,7 +315,10 @@ static void pr_args(struct fstack_arguments *args)
 			if (!memcmp(buf, &null_str, 4))
 				strcpy(buf, "NULL");
 
-			pr_out("  args[%d] str: %s\n", i , buf);
+			if (spec->fmt == ARG_FMT_STD_STRING)
+				pr_out("  args[%d] std::string: %s\n", i , buf);
+			else
+				pr_out("  args[%d] str: %s\n", i , buf);
 			size += 2;
 		}
 		else {
@@ -344,8 +348,9 @@ static void pr_retval(struct fstack_arguments *args)
 		if (spec->idx != RETVAL_IDX)
 			continue;
 
-		if (spec->fmt == ARG_FMT_STR) {
-			char buf[64];
+		if (spec->fmt == ARG_FMT_STR ||
+		    spec->fmt == ARG_FMT_STD_STRING) {
+			char buf[64] = {0};
 			const int null_str = -1;
 
 			size = *(unsigned short *)ptr;
@@ -354,7 +359,10 @@ static void pr_retval(struct fstack_arguments *args)
 			if (!memcmp(buf, &null_str, 4))
 				strcpy(buf, "NULL");
 
-			pr_out("  retval[%d] str: %s\n", i , buf);
+			if (spec->fmt == ARG_FMT_STD_STRING)
+				pr_out("  retval[%d] std::string: %s\n", i , buf);
+			else
+				pr_out("  retval[%d] str: %s\n", i , buf);
 			size += 2;
 		}
 		else {
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -486,7 +486,8 @@ void get_argspec_string(struct ftrace_task_handle *task,
 			break;
 		}
 
-		if (spec->fmt == ARG_FMT_STR) {
+		if (spec->fmt == ARG_FMT_STR ||
+		    spec->fmt == ARG_FMT_STD_STRING) {
 			unsigned short slen;
 			unsigned short newline = 0;
 			char last_ch;
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -252,14 +252,14 @@ The uftrace tool supports recording function arguments and/or return values usin
     <int_spec>    :=  "arg" N [ "/" <format> [ <size> ] ] [ "%" ( <reg> | <stack> ) ]
     <float_spec>  :=  "fparg" N [ "/" ( <size> | "80" ) ] [ "%" ( <reg> | <stack> ) ]
     <ret_spec>    :=  "retval" [ "/" <format> [ <size> ] ]
-    <format>      :=  "i" | "u" | "x" | "s" | "c" | "f"
+    <format>      :=  "i" | "u" | "x" | "s" | "c" | "f" | "S"
     <size>        :=  "8" | "16" | "32" | "64"
     <reg>         :=  <arch-specific register name>  # "rdi", "xmm0", "r0", ...
     <stack>       :=  "stack" [ "+" ] <offset>
 
 The `-A`/`--argument` option takes argN where N is an index of the arguments.  The index starts from 1 and corresponds to the argument passing order of the calling convention on the system.  Note that the indexes of arguments are separately counted for integer (or pointer) and floating-point type, and they can interfere depending on the calling convention.  The argN is for integer arguments and fpargN is for floating-point arguments.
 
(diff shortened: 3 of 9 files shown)


honggyukim:
Hi @namhyung,

This is a draft version of std::string support. I've wrote this based on the internal implementation of libstdc++, but didn't find the format document about the layout of std::string implementation. I know it may have some problems in other environments, so I would like to find that this layout is "correct" in a formal document. But anyhow, it'd be useful if it works in most of cases.

I will revise the code and commit messages based on your feedback. So please have a look at it.

Thanks,
Honggyu

honggyukim:
I've tested it in the below environments:
1. g++ with libstdc++
2. clang++ with libstdc++
3. clang++ with libc++

As mentioned this feature only supports libstdc++ so 1 and 2 works fine, but 3 doesn't work.

namhyung:
@honggyukim thanks for the interesting PR.  Do you have a test code?  I took a quick look at the gcc/libstdc++ code, and it looked like it might work only for short strings (less than 16).

I think we need to use `_M_dataplus._M_p` instead, how about this?

```
diff --git a/libmcount/record.c b/libmcount/record.c
index aa29e52..7c13c53 100644
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -253,9 +253,9 @@ static unsigned save_to_argbuf(void *argbuf, struct list_head *args_spec,
                        if (spec->fmt == ARG_FMT_STD_STRING) {
                                long *base = ctx->val.p;
                                long *_M_string_length = base + 1;
-                               char *_M_local_buf = (char*)(base + 2);
+                               char *_M_data = (char*)(*base);
                                len = *_M_string_length;
-                               str = _M_local_buf;
+                               str = _M_data;
                        }
 
                        if (str) {
```

honggyukim:
@namhyung Thanks for your feedback.

> I think we need to use _M_dataplus._M_p instead, how about this?
You're absolutely correct, so I've updated the commit. Please have a look at it again.

namhyung:
What if it's used with libc++, crash?  Also it'd be better the test code contains a longer string..

honggyukim:
> What if it's used with libc++, crash?

I cannot guarantee but it just shows empty string in the test program.

> Also it'd be better the test code contains a longer string..

Okay, I've already tested it but will add a longer string as well.

honggyukim:
I've updated the test program with a longer string as follows:
```
            [33070] | main() {
   2.640 us [33070] |   std_string_arg("Hello");
   0.213 us [33070] |   std_string_arg("World!");
   0.156 us [33070] |   std_string_arg("std::string support is done!");
   1.679 us [33070] |   std_string_ret::cxx11() = "Hello";
   0.193 us [33070] |   std_string_ret::cxx11() = "World!";
   0.233 us [33070] |   std_string_ret::cxx11() = "std::string support is done!";
   8.841 us [33070] | } /* main */
```

>> What if it's used with libc++, crash?

> I cannot guarantee but it just shows empty string in the test program.

I've tested it once again, but it is crashed when libc++ is used so I think it'd be better to detect if libc++ is used or not when 'S' format is used. Do we have such infrastructure in the code base?

namhyung:
You can refer to `check_libpthread` in utils/symbol.c

honggyukim:
@namhyung Thanks, I would like to check the code as follows:
```
diff --git a/cmd-record.c b/cmd-record.c
index 1e207f5..ac997fc 100644
--- a/cmd-record.c
+++ b/cmd-record.c
@@ -85,7 +85,7 @@ static void setup_child_environ(struct opts *opts, int pfd)
 {
        char buf[4096];
        char *old_preload, *old_libpath;
-       bool must_use_multi_thread = check_libpthread(opts->exename);
+       bool must_use_multi_thread = check_so_dependency(opts->exename, "libpthread.so.0");

        if (opts->lib_path) {
                strcpy(buf, opts->lib_path);
diff --git a/utils/filter.c b/utils/filter.c
index 94b2a0f..978be6d 100644
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -322,6 +322,10 @@ static int parse_spec(char *str, struct ftrace_arg_spec *arg, char *suffix)
                size = sizeof(double);
                break;
        case 'S':
+               if (check_so_dependency(opts->exename, "libc++.so")) {
+                       pr_use("libc++ is not supported for std::string\n");
+                       return -1;
+               }
                fmt = ARG_FMT_STD_STRING;
                break;
        default:
diff --git a/utils/symbol.c b/utils/symbol.c
index 6229f82..45b2962 100644
--- a/utils/symbol.c
+++ b/utils/symbol.c
@@ -68,7 +68,7 @@ static int namefind(const void *a, const void *b)
        return strcmp(name, sym->name);
 }

-bool check_libpthread(const char *filename)
+bool check_so_dependency(const char *filename, const char *soname)
 {
        int fd;
        Elf *elf;
@@ -119,8 +119,9 @@ bool check_libpthread(const char *filename)
                        goto elf_error;

                if (dyn.d_tag == DT_NEEDED) {
-                       char *soname = elf_strptr(elf, shstr_idx, dyn.d_un.d_ptr);
-                       if (!strcmp(soname, "libpthread.so.0")) {
+                       char *so = elf_strptr(elf, shstr_idx, dyn.d_un.d_ptr);
+                       size_t len = strlen(soname);
+                       if (!strncmp(so, soname, len)) {
                                ret = true;
                                break;
                        }
```
But in `utils/filter.c`, there's no way to get `opts->exename` here and it requires a lot of propagation from the parent functions. Is there a easier way to get the `exename`?

namhyung:
Hmm.. this is a different case and you'd better using `dl_iterate_phdr` and checking `info->dlpi_name` instead.

honggyukim:
@namhyung Thanks, I've just added a new patch for checking libc++.so in 2bb62bb.
Please take a look at it!

namhyung:
It would check the library everytime it sees the 'S' format.  You could save the result and use it.

Also I'd like to see if it works with the old ABI (in GCC 4.x) as well.  Please see https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

honggyukim:
I've tested it once again with g++ 4.8 and g++ 5.4 and it works fine. Please see the result:
```
$ g++-4.8 -pg -O3 s-std-string.cpp -o t-std-string-4.8

$ g++-5 -pg -O3 s-std-string.cpp -o t-std-string-5.4

$ uftrace -A ^std_string_arg@arg1/S -R ^std_string_ret@retval/S -F main -F ^std_string_ -D 1 t-std-string-4.8
            [83957] | main() {
   2.519 us [83957] |   std_string_arg("Hello");
   0.366 us [83957] |   std_string_arg("World!");
   0.326 us [83957] |   std_string_arg("std::string support is done!");
   1.110 us [83957] |   std_string_ret() = "Hello";
   0.247 us [83957] |   std_string_ret() = "World!";
   0.200 us [83957] |   std_string_ret() = "std::string support is done!";
   8.974 us [83957] | } /* main */

$ uftrace -A ^std_string_arg@arg1/S -R ^std_string_ret@retval/S -F main -F ^std_string_ -D 1 t-std-string-5.4
            [83976] | main() {
   1.668 us [83976] |   std_string_arg("Hello");
   0.343 us [83976] |   std_string_arg("World!");
   0.295 us [83976] |   std_string_arg("std::string support is done!");
   0.416 us [83976] |   std_string_ret::cxx11() = "Hello";
   0.175 us [83976] |   std_string_ret::cxx11() = "World!";
   0.502 us [83976] |   std_string_ret::cxx11() = "std::string support is done!";
   7.253 us [83976] | } /* main */
```
Both of them shows the output correctly but some of function name is changed. `std_string_ret` has become `std_string_ret::cxx11` when g++ 5.4 is used. I should change the expected test result based on the g++ version. Is it possible to provide multiple expected outputs in test infra?

namhyung:
Thanks for the confirmation.  Test cases can have fixup() function to modify output of known failures.

honggyukim:
@namhyung I've updated it as you suggested. Thanks!

honggyukim:
@namhyung Please check  it again!

namhyung:
The first commit which changes the dump command should be moved after the record change.  Otherwise, looks good to me.

honggyukim:
@namhyung Right, I've just changed the order of the commits. Thanks for your review!

honggyukim:
Here is GDB Pretty printers for libc++ so the structure can be considered later for libc++ support.
https://github.com/koutheir/libcxx-pretty-printers/blob/master/src/libcxx/v1/printers.py
```
class StdStringPrinter:
    "Print a std::basic_string of some kind"

    def __init__(self, typename, val):
        self.val = val
        self.typename = typename

    def to_string(self):
        # Make sure &string works, too.
        type = self.val.type
        if type.code == gdb.TYPE_CODE_REF:
            type = type.target()

        ss = self.val['__r_']['__first_']['__s']
        __short_mask = 0x1
        if (ss['__size_'] & __short_mask) == 0:
            len = (ss['__size_'] >> 1)
            ptr = ss['__data_']
        else:
            sl = self.val['__r_']['__first_']['__l']
            len = sl['__size_']
            ptr = sl['__data_']

        return ptr.string(length=len)

    def display_hint(self):
        return 'string'
```

PR #107, opened 2017-06-19
Add a method to trigger some command line when trace data received (during uftrace recv)

jb-seo:
Hi, @namhyung.

I hope to add a feature to execute command when uftrace data received.
With this, user can see the output of their own script to analysis something immediately the trace ends.

Originally, I hope to create some remote interface that start/end the trace and do something after the trace.
It may be a first step to reach that.

### DIFF
--- a/cmd-recv.c
+++ b/cmd-recv.c
@@ -13,6 +13,7 @@
 #include <linux/limits.h>
 #include <sys/stat.h>
 #include <errno.h>
+#include <sys/wait.h>
 
 #include "uftrace.h"
 #include "utils/utils.h"
@@ -61,6 +62,7 @@ static int signal_fd(struct opts *opts)
 	sigemptyset(&mask);
 	sigaddset(&mask, SIGINT);
 	sigaddset(&mask, SIGTERM);
+	sigaddset(&mask, SIGCHLD);
 
 	if (sigprocmask(SIG_BLOCK, &mask, NULL) < 0)
 		pr_err("signal block failed");
@@ -447,6 +449,20 @@ static void recv_trace_end(int sock, int efd)
 	close(sock);
 }
 
+static void execute_run_cmd(char **argv) {
+	if (!argv)
+		return;
+
+	int pid = fork();
+	if (pid < 0)
+		pr_err("cannot start child process");
+
+	if (pid == 0) {
+		execvp(argv[0], argv);
+		pr_err("Failed to execute '%s'", argv[0]);
+	}
+}
+
 static void epoll_add(int efd, int fd, unsigned event)
 {
 	struct epoll_event ev = {
@@ -479,7 +495,7 @@ static void handle_server_sock(struct epoll_event *ev, int efd)
 	pr_dbg("new connection added from %s\n", hbuf);
 }
 
-static void handle_client_sock(struct epoll_event *ev, int efd)
+static void handle_client_sock(struct epoll_event *ev, int efd, struct opts *opts)
 {
 	int sock = ev->data.fd;
 	struct uftrace_msg msg;
@@ -524,6 +540,7 @@ static void handle_client_sock(struct epoll_event *ev, int efd)
 	case UFTRACE_MSG_SEND_END:
 		pr_dbg2("receive UFTRACE_MSG_SEND_END\n");
 		recv_trace_end(sock, efd);
+		execute_run_cmd(opts->run_cmd);
 		break;
 	default:
 		pr_dbg("unknown message: %d\n", msg.type);
@@ -533,6 +550,7 @@ static void handle_client_sock(struct epoll_event *ev, int efd)
 
 int command_recv(int argc, char *argv[], struct opts *opts)
 {
+	struct signalfd_siginfo si;
 	int sock;
 	int sigfd;
 	int efd;
@@ -566,12 +584,17 @@ int command_recv(int argc, char *argv[], struct opts *opts)
 			pr_err("epoll wait failed");
 
 		for (i = 0; i < len; i++) {
-			if (ev[i].data.fd == sigfd)
-				uftrace_done = true;
+			if (ev[i].data.fd == sigfd) {
+				int nr = read(sigfd, &si, sizeof si);
+				if (nr > 0 && si.ssi_signo == SIGCHLD)
+					waitpid(-1, NULL, WNOHANG);
+				else
+					uftrace_done = true;
+			}
 			else if (ev[i].data.fd == sock)
 				handle_server_sock(&ev[i], efd);
 			else
-				handle_client_sock(&ev[i], efd);
+				handle_client_sock(&ev[i], efd, opts);
 		}
 	}
 
--- a/uftrace.c
+++ b/uftrace.c
@@ -83,6 +83,7 @@ enum options {
 	OPT_kernel_full,
 	OPT_kernel_only,
 	OPT_list_event,
+	OPT_run_cmd,
 };
 
 static struct argp_option ftrace_options[] = {
@@ -144,6 +145,7 @@ static struct argp_option ftrace_options[] = {
 	{ "patch", 'P', "FUNC", 0, "Apply dynamic patching for FUNCs" },
 	{ "event", 'E', "EVENT", 0, "Enable EVENT to save more information" },
 	{ "list-event", OPT_list_event, 0, 0, "List avaiable events" },
+	{ "run-cmd", OPT_run_cmd, "CMDLINE", 0, "Command line that want to execute after tracing data received" },
 	{ 0 }
 };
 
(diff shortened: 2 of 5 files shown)


namhyung:
@jb-seo Thanks for contributing to uftrace!  Just out of curiousity, what kind of work do you want to do with this?

The system(3) blocks the calling thread until the given task completed, so I want to avoid using it.  Note that a single uftrace recv process can handle multiple sessions at once.  You'd be better using one of the exec(3) family and handle SIGCHLD instead.

namhyung:
Also I'm very much interested in having the remote start/stop interface.  If you have an idea, please feel free to share and discuss.  Thanks!

jb-seo:
I've updated to use exec rather than system. Thanks for advise.

cmd-recv.c
```diff
@@ -447,6 +447,20 @@ static void recv_trace_end(int sock, int efd)
...
+	int pid = fork();
+	if (pid < 0)
+		pr_err("cannot start child process");
+
+	if (pid == 0) {
+		execvp(argv[0], argv);
+		pr_err("Failed to execute '%s'", argv[0]);
+	}
+}
+
```
namhyung: Then you need to wait for the child process like below..

```
diff --git a/cmd-recv.c b/cmd-recv.c
index fb01794..314f163 100644
--- a/cmd-recv.c
+++ b/cmd-recv.c
@@ -13,6 +13,7 @@
 #include <linux/limits.h>
 #include <sys/stat.h>
 #include <errno.h>
+#include <sys/wait.h>

 #include "uftrace.h"
 #include "utils/utils.h"
@@ -531,12 +532,19 @@ static void handle_client_sock(struct epoll_event *ev, int efd)
        }
 }

+static void wait_child(int sig)
+{
+       waitpid(-1, NULL, WNOHANG);
+}
+
 int command_recv(int argc, char *argv[], struct opts *opts)
 {
        int sock;
        int sigfd;
        int efd;

+       signal(SIGCHLD, wait_child);
+
        if (strcmp(opts->dirname, UFTRACE_DIR_NAME)) {
                char *dirname = "current";

```

uftrace.c
```diff
@@ -142,6 +143,7 @@ static struct argp_option ftrace_options[] = {
 	{ "time-range", 'r', "TIME~TIME", 0, "Show output within the TIME(timestamp or elapsed time) range only" },
 	{ "patch", 'P', "FUNC", 0, "Apply dynamic patching for FUNCs" },
 	{ "event", 'E', "EVENT", 0, "Enable EVENT to save more information" },
+	{ "recv-cmd", OPT_recv_cmd, "CMDLINE", 0, "Command line that want to execute after tracing data received" },
```
namhyung: What about changing name to "run-cmd" instead?  This is uftrace recv command, so repeating "recv" here seems unnatural.  Also help message can omit "that want" part IMHO.

uftrace.c
```diff
@@ -712,6 +722,7 @@ int main(int argc, char *argv[])
 		.color		= COLOR_AUTO,  /* default to 'auto' (turn on if terminal) */
 		.column_offset	= 8,
 		.comment	= true,
+		.recv_cmd	= NULL,
```
namhyung: In general, omitted fields will be initialized to zero so no need to do it explicitly.

utils/utils.c
```diff
@@ -392,3 +392,81 @@ char * strjoin(char *left, char *right, char *delim)
 	strcpy(new + len - rlen - 1, right);
 	return new;
 }
+
+#define QUOTE '\''
+#define DQUOTE '"'
+
+static int setargs(char *args, char **argv)
```
namhyung: This function is little bit complex so it'd be nice if you could add some comment here.

utils/utils.c
```diff
@@ -392,3 +392,81 @@ char * strjoin(char *left, char *right, char *delim)
...
+		++args;
+
+	while (*args) {
+		/* consider quotes and update argv */
+		if (*args == QUOTE) {
+			if (argv)
+				argv[count] = args + 1;
+			while (*args != QUOTE)
+				++args;
+		} else if (*args == DQUOTE) {
```
namhyung: The coding style saids that the "else" clause should start at the next line of the closing "}".

```
   if () {
      ...
   }
   else if () {
      ...
   }
```

utils/utils.c
```diff
@@ -392,3 +392,81 @@ char * strjoin(char *left, char *right, char *delim)
...
+			if (argv)
+				argv[count] = args + 1;
+			while (*args != DQUOTE)
+				++args;
+		} else if (argv) {
```
namhyung: Ditto.

utils/utils.c
```diff
@@ -392,3 +392,81 @@ char * strjoin(char *left, char *right, char *delim)
...
+
+#undef QUOTE
+#undef DQUOTE
+
+char **parse_cmdline(char *cmd, int *argc)
+{
+	char **argv = NULL;
+	int argn = 0;
+
+	if (cmd && *cmd
```
namhyung: I prefer returning ASAP if "cmd" is not avaiable.  It'd make the code simpler IMHO.

```
  if (cmd == NULL || *cmd == '\0')
      return NULL;
```

And please use a temp variable for allocated "cmd" string to reduce confusions.

utils/utils.c
```diff
@@ -392,3 +392,81 @@ char * strjoin(char *left, char *right, char *delim)
...
+
+	return argv;
+}
+
+void free_parsed_args(char **argv)
+{
+	if (argv) {
+		free(argv[-1]);
+		free(argv - 1);
+	}
```
namhyung: This doesn't look straight-forward.  You could either pre-decrement the "argv" with a comment or make `parse_cmdline()` return "&argv[0]" and use "argv[1...]" in `execute_recv_cmd()`.

 Also I think it'd be better making the function name consistent - like `free_parsed_cmdline()`?

cmd-recv.c
jb-seo: done

uftrace.c
jb-seo: done

uftrace.c
jb-seo: Done. But if so .fields also be removed.

utils/utils.c
jb-seo: done

utils/utils.c
jb-seo: done

utils/utils.c
jb-seo: done

utils/utils.c
jb-seo: done.

Also, fixed what I implemented wrong way. (I tested this on separated one but way of test was wrong before)

utils/utils.c
jb-seo: done

cmd-recv.c
jb-seo: It's not work with likes "signal(SIGCHLD, wait_child)" because it's crashed on epoll_wait.
So, I added SIGCHLD to be handled on epoll_wait and check signal type.

namhyung: ok!

uftrace.c
namhyung: Right, you can do it if you want. :)

utils/utils.c
namhyung: Oh, I meant a header commit which is right above the function signature and describes arguments and return value.  Especially for `argv`...  But I think it should be for `parse_cmdline` and `free_parsed_cmdline()`.

utils/utils.c
```diff
@@ -392,3 +392,95 @@ char * strjoin(char *left, char *right, char *delim)
...
+		count++;
+	}
+
+	return count;
+}
+
+#undef QUOTE
+#undef DQUOTE
+
+char **parse_cmdline(char *cmd, int *argc)
```
namhyung: Maybe something like this:

```
/**
 * parse_cmdline - parse given string to be executed via execvp(3)
 * @cmd:  full command line
 * @argc: pointer to number of arguments
 *
 * This function parses @cmd and split it into an array of string
 * to be executed by exec(3) like argv[] in main().  The @argc
 * will be set to number of argument parsed if it's non-NULL.
 * The resulting array contains a copy of input string (@cmd) in
 * the first element which other elements point to.  It returns a
 * pointer to the second element so that it can be used directly
 * in other functions like exec(3).
 *
 * The returned array should be freed by free_parsed_cmdline().
 */
char **parse_cmdline(char *cmd, int *argc)

jb-seo: added to the utils.h

namhyung: Sorry for bothering you.. but I think it'd be better placing it in the .c file.  Also please add a similar comment for free_parsed_cmdline() too.

jb-seo: done

jb-seo:
It's so much pended by my personal things.
Finally, I updated it!

PR #117, opened 2017-07-17
kernel: Fix the problem about enabled "tracing_on" even after kernel tracing was done

Taeung:
Hi @namhyung ,

I found this problem below case:

```
$ cat /sys/kernel/debug/tracing/tracing_on
0

$ sudo uftrace -k t-openclose 
   2.141 us [23687] | __monstartup();
   2.029 us [23687] | __cxa_atexit();
            [23687] | main() {
            [23687] |   open() {
  25.416 us [23687] |     sys_open();
...

$ cat /sys/kernel/debug/tracing/tracing_on
1
```
If it ordinarily works, "tracing_on" was disabled after kernel tracing, Right ?
So I fix it. Additionally I added a debug message for stop_kernel_tracing()

I'd appreciate some feedback on my patches. 😄

### DIFF
--- a/cmd-record.c
+++ b/cmd-record.c
@@ -1579,8 +1579,10 @@ int command_record(int argc, char *argv[], struct opts *opts)
 	}
 
 	stop_all_writers();
-	if (opts->kernel)
-		stop_kernel_tracing(&kernel);
+	if (opts->kernel) {
+		if (stop_kernel_tracing(&kernel) == 0)
+			pr_dbg("kernel tracing stopped\n");
+	}
 
 	if (fill_file_header(opts, status, &usage) < 0)
 		pr_err("cannot generate data file");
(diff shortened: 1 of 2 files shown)


Taeung:
@namhyung ,
Currently uftrace tool reset several configs by `reset_tracing_file()` as you know.
What do you think about restoring the previous configuration of ftrace  ?

namhyung:
It's not a problem and, in fact, it's the kernel's default setting.

Maybe it'd be nice to save all of the original settings and restore after uftrace.  But I'm not sure it's worth doing since if anyone wants to use kernel tracing, [s]he needs to tweak the settings anyway.  Simply reset to the default seems ok to me.

Taeung:
I got it  ! 😄 
I'll close this PR.

Taeung:
I got it ! 😄 
I'll close this PR

PR #128, opened 2017-08-08
mcount: Fix the size in proc statm event

honggyukim:
Since /proc/[pid]/statm prints the number of pages for each field, it'd
be better to keep the memory size in KB.

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -59,6 +59,8 @@ static enum filter_mode mcount_filter_mode = FILTER_MODE_NONE;
 static struct rb_root mcount_triggers = RB_ROOT;
 #endif /* DISABLE_MCOUNT_FILTER */
 
+int page_size_in_kb;
+
 uint64_t mcount_gettime(void)
 {
 	struct timespec ts;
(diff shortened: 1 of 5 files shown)


namhyung:
Oops I missed the units, thanks.  Btw, I think we can keep a static pagesize_kb variable instead to reduce the function call and division.

honggyukim:
Okay. I will keep the record logic, then just modify replay and dump side. Would it be okay?

namhyung:
IIRC replay and dump assume the numbers are in KB, so changing record side should be enough.

honggyukim:
I mean that I have to multiply the recorded data by 4 during `replay` and `dump`.

namhyung:
Hmm.. you can multiply the size during `record` and use it as is during `replay` and `dump` no?

honggyukim:
Sorry, I was busy on something so I misunderstood your point. I've just updated the commit.

namhyung:
LGTM

honggyukim:
Thanks! But by the way, what is `IIRC`? :)

namhyung:
Oh, it's just an acronym for "If I Recall Correctly". :)

honggyukim:
Ah.. Thanks! I still have some problems understanding jargons.

PR #131, opened 2017-08-09
Fix calc of message sizes in send_trace_metadata

thepaul:
It sometimes happens that (htonl(a) + htonl(b)) == htonl(a + b), but
sometimes (usually?) those are quite different.

This fixes t142_recv_multi.py so that it doesn't leave hung daemon
uftrace instances lying around when the memory maps file is of certain
sizes that happen to exercise this htonl bug.


namhyung:
LGTM.  But could you please add your "Signed-off-by" in the commit log? :)

thepaul:
Right! Right. Yes.

namhyung:
Thanks!

PR #132, opened 2017-08-10
Test trivial fix

sjp38:
Hi Namhyung,

I attended to your uftrace seminar today and got impressed so much.  During the time, I found two trivial nitpicks by coincidence.  So, here's the patches.

Thanks,
SeongJae Park


namhyung:
Thanks for your kind word.  The patches look good to me.

PR #138, opened 2017-08-28
test: Avoid -finstrument-functions bug in gcc

honggyukim:
Since older gcc such as gcc-4.xx and gcc-5.xx have a bug in
-finstrument-functions, some of tests are failed.

The problem is that gcc sometimes put '__cyg_profile_func_enter' and
'__cyg_profile_func_exit' for some library calls, which is not expected.

For example, the below example shows unexpected additional open call.

```
  $ gcc -finstrument-functions -O2 s-openclose.c
  $ uftrace a.out
              [18948] | main() {
              [18948] |   open() {
     7.493 us [18948] |     open();
    19.570 us [18948] |   } /* open */
     4.143 us [18948] |   close();
    25.447 us [18948] | } /* main */
```

Unlike this, clang doesn't have such problem for the same example.

```
  $ clang -finstrument-functions -O2 s-openclose.c
  $ uftrace a.out
              [18980] | main() {
    13.103 us [18980] |   open();
     4.433 us [18980] |   close();
    31.157 us [18980] | } /* main */
```

Since this bug has been fixed in recent gcc, so we better avoid in our
test framework to reduce confusion in old systems.

### DIFF
--- a/tests/s-alloca.c
+++ b/tests/s-alloca.c
@@ -5,7 +5,7 @@
 int foo(int c)
 {
 	char *ptr = alloca(c);
-	snprintf(ptr, c, "%s", "hello world\n");
+	strncpy(ptr, "hello world\n", c);
 	return c;
 }
 
--- a/tests/s-openclose.c
+++ b/tests/s-openclose.c
@@ -1,9 +1,7 @@
 #include <stdio.h>
-#include <unistd.h>
-#include <fcntl.h>
 
 int main(void)
 {
-	close(open("/dev/null", O_RDONLY));
+	fclose(fopen("/dev/null", "r"));
 	return 0;
 }
--- a/tests/t016_alloca.py
+++ b/tests/t016_alloca.py
@@ -9,14 +9,14 @@ class TestCase(TestBase):
   75.736 us [ 6681] | __cxa_atexit();
             [ 6681] | main() {
             [ 6681] |   foo() {
-   2.153 us [ 6681] |     snprintf();
+   2.153 us [ 6681] |     strncpy();
    3.073 us [ 6681] |   } /* foo */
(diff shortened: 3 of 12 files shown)


tests/s-alloca.c
```diff
@@ -5,7 +5,7 @@
 int foo(int c)
 {
 	char *ptr = alloca(c);
-	snprintf(ptr, c, "%s", "hello world\n");
+	strncpy(ptr, "hello world\n", c);
```
honggyukim: This example has the same issue so I changed it to use another call.
```
main() {
   foo() {
     snprintf() {
       snprintf();
     } /* snprintf */
   } /* foo */
   bar() {
     foo() {
       snprintf() {
         snprintf();
       } /* snprintf */
     } /* foo */
     foo() {
       snprintf() {
         snprintf();
       } /* snprintf */
     } /* foo */
   } /* bar */
 } /* main */
main() {
   foo() {
     snprintf();
   } /* foo */
   bar() {
     foo() {
       snprintf();
     } /* foo */
     foo() {
       snprintf();
     } /* foo */
   } /* bar */
 } /* main */
016 alloca              : OK OK OK OK OK OK NG NG NG NG
```

honggyukim:
Here is the test result with this patch on top of https://github.com/namhyung/uftrace/compare/check/pie-fix.
```
runtime test stats
====================
total  1650  Tests executed (success: 82.36%)
  OK:  1329  Test succeeded
  OK:    30  Test succeeded (with some fixup)
  NG:     7  Different test result
  SG:     4  Abnormal exit by signal
  TM:     0  Test ran too long
  BI:    40  Build failed
  SK:   240  Skipped
```
I can see many of NGs have become OKs.

PR #139, opened 2017-08-29
info: Add a info item 'record_date' to uftrace_info_bits

Taeung:
Currently we get a record date from the info file's mtime.
But it isn't reliable because it can be changed anytime.
So implement a new info item 'record_date' that will be
saved at the time of recording.

Note:
 - Rename `recorded date` to `record date`
 - Save a `record date` when tracing was stopped.
   Or, record the `record date` immediately after `start_tracing()` ?
 - Save a `record date` as just a date string, not a timestamp value (i.e. `time_t`)

I'd appreciate some feedback on this. 😄

### DIFF
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -29,6 +29,7 @@ struct fill_handler_arg {
 	int exit_status;
 	struct opts *opts;
 	struct rusage *rusage;
+	char *elapsed_time;
 };
 
 static char *copy_info_str(char *src)
@@ -687,7 +688,9 @@ static int fill_record_date(void *arg)
 
 	time(&current_time);
 
-	return dprintf(fha->fd, "record_date:%s", ctime(&current_time));
+	dprintf(fha->fd, "record_date:%s", ctime(&current_time));
+	dprintf(fha->fd, "elapsed_time:%s\n", fha->elapsed_time);
+	return 0;
 }
 
 static int read_record_date(void *arg)
@@ -703,6 +706,15 @@ static int read_record_date(void *arg)
 		return -1;
 
 	info->record_date = copy_info_str(&buf[12]);
+
+	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+		return -1;
+
+	if (strncmp(buf, "elapsed_time:", 13))
+		return -1;
+
+	info->elapsed_time = copy_info_str(&buf[13]);
+
 	return 0;
 }
 
@@ -712,7 +724,7 @@ struct ftrace_info_handler {
 };
 
 void fill_ftrace_info(uint64_t *info_mask, int fd, struct opts *opts, int status,
-		      struct rusage *rusage)
+		      struct rusage *rusage, char *elapsed_time)
 {
 	size_t i;
 	off_t offset;
@@ -721,6 +733,7 @@ void fill_ftrace_info(uint64_t *info_mask, int fd, struct opts *opts, int status
 		.opts = opts,
 		.exit_status = status,
 		.rusage = rusage,
+		.elapsed_time = elapsed_time,
 	};
 	struct ftrace_info_handler fill_handlers[] = {
 		{ EXE_NAME,	fill_exe_name },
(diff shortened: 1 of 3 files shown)


Taeung:
Or, what do you think about saving a record date when `mcount()` is called for the first time ?

namhyung:
Looks ok to me.  I don't think it matters when it saves the timestamp - so just keep the current code.

But I'd like to add elapsed time as well.  Since we already saves ts1 and ts2 in `writer_data`, it can simply be done by passing the delta to `fill_file_header()`.

Taeung:
Ok ! @namhyung , I resent this patchkit in accordance with your comment. 😄

cmd-info.c
```diff
@@ -680,13 +681,52 @@ static int read_arg_spec(void *arg)
...
+	info->record_date = copy_info_str(&buf[12]);
+
+	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+		return -1;
+
+	if (strncmp(buf, "elapsed_time:", 13))
+		return -1;
+
+	sprintf(buf, "%s\n# %-20s: %s", info->record_date, "elapsed time",
+		copy_info_str(&buf[13]));
```
Taeung: @namhyung , hum.. I thought it'd be better to has two items in uftrace_info (i.e. record_date and elapsed_time).
What do you think about it ?

cmd-record.c
```diff
@@ -1266,7 +1267,13 @@ static void print_child_time(struct timespec *ts1, struct timespec *ts2)
 		sec--;
 	}
 
-	pr_out("elapsed time: %"PRIu64".%09"PRIu64" sec\n", sec, nsec);
+	sprintf(buf, "%"PRIu64".%09"PRIu64" sec", sec, nsec);
+	return xstrdup(buf);
+}
```
namhyung: It's unnatural to keep it as a string but I think it's ok for this case since we need to do the same job twice.  And then please replace snprint + xstdup into xasprint.

cmd-info.c
namhyung: Yes, I think it's good to have uftrace_info->elapsed_time member.  What I was asking is not to add an item to `enum uftrace_info_bits`.

cmd-info.c
```diff
@@ -799,15 +827,10 @@ int command_info(int argc, char *argv[], struct opts *opts)
...
-	snprintf(buf, sizeof(buf), "%s/info", opts->dirname);
-
-	if (stat(buf, &statbuf) < 0)
-		return -1;
-
 	pr_out("# system information\n");
 	pr_out("# ==================\n");
 	pr_out(fmt, "program version", argp_program_version);
-	pr_out("# %-20s: %s", "recorded on", ctime(&statbuf.st_mtime));
+	pr_out("# %-20s: %s\n", "recorded on", handle.info.record_date);
```
namhyung: It should be

```
if (handle.hdr.info_mask & (1UL << RECORD_DATE))
    pr_out(fmt, "recorded on", handle.info.record_date);
else
    pr_out("# %-20s: %s", "recorded on", ctime(&statbuf.st_mtime));
```

cmd-info.c
Taeung: Ok !

cmd-record.c
Taeung: I got it.

cmd-info.c
Taeung: Understood !

Taeung:
@namhyung , I resent v3 ! I'd appreciate it, if you check this again. 😃

namhyung:
Almost done, I just want to print "elapsed time" info in process information section - maybe between "exit status" and "cpu time".

cmd-info.c
```diff
@@ -703,6 +706,15 @@ static int read_record_date(void *arg)
...
 
 	info->record_date = copy_info_str(&buf[12]);
+
+	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+		return -1;
+
+	if (strncmp(buf, "elapsed_time:", 13))
+		return -1;
+
+	info->elapsed_time = copy_info_str(&buf[13]);
```
namhyung: You need to free these two data after use.

Taeung: Got it !

Taeung:
@namhyung I did the two things you said ! Please take a look v4 😃

namhyung:
LGTM

PR #140, opened 2017-08-30
uftrace: Disable pager when using --nop option

honggyukim:
--nop option is for performance testing so doesn't require to use pager
so it's better to disable it.

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -862,6 +862,8 @@ int main(int argc, char *argv[])
 
 	if (opts.mode == UFTRACE_MODE_RECORD || opts.mode == UFTRACE_MODE_RECV)
 		opts.use_pager = false;
+	if (opts.nop)
+		opts.use_pager = false;
 
 	if (opts.use_pager)
 		start_pager();


namhyung:
OK, but I think `--nop` is used usually with record command so it's already disabled.

honggyukim:
If you see the commit 73950bda0f0d4b9c09c31c802b0bfaefb2d4a470, it just skips replay and do not disable pager.

namhyung:
Ah, ok.  So you use it from live command.  Technically it's undocumented but I think it's good to support it anyway.

But I think it'd be better checking it in the main() function right before calling `start_pager()`.

honggyukim:
What about disabling pager for `uftrace info` command? It's trivial but sometimes annoying.

honggyukim:
I've just updated it as you said. I didn't handle uftrace info.

uftrace.c
```diff
@@ -860,7 +860,9 @@ int main(int argc, char *argv[])
 	setup_color(opts.color);
 	setup_signal();
 
-	if (opts.mode == UFTRACE_MODE_RECORD || opts.mode == UFTRACE_MODE_RECV)
+	if (opts.mode == UFTRACE_MODE_RECORD
+	    || opts.mode == UFTRACE_MODE_RECV
+	    || opts.nop)
```
namhyung: It's just a coding style issue - I prefer putting the operator at the end of line.  And I think it'd be better splitting nop check into another line like below:

```
if (opts.mode = ... RECORD || ... RECV)
    opts.use_pager = false;
if (opts.nop)
    opts.use_pager = false;

honggyukim:
@namhyung Thanks for pointing out coding style issue! I followed your guide and updated it again.

PR #146, opened 2017-09-06
python: Fix to import matplotlib and numpy

honggyukim:
Since sys.path is not properly set before, there was some import errors.
This patch first gets the correct sys.path by running python
interpreter, then generate .py_sys_path.def file so that it can be
included by utils/script-python.c.

This also adds RTLD_GLOBAL flag to dlopen to complete the symbol
resolution of subsequently loaded shared objects.

Now, it's ready to use matplotlib and numpy in uftrace python script.

### DIFF
--- a/utils/script-python.c
+++ b/utils/script-python.c
@@ -218,7 +218,7 @@ int script_init_for_python(char *py_pathname)
 	script_uftrace_exit = python_uftrace_exit;
 	script_uftrace_end = python_uftrace_end;
 
-	python_handle = dlopen(libpython, RTLD_LAZY);
+	python_handle = dlopen(libpython, RTLD_LAZY | RTLD_GLOBAL);
 	if (!python_handle) {
 		pr_warn("%s cannot be loaded!\n", libpython);
 		return -1;
--- a/utils/script-python.c
+++ b/utils/script-python.c
@@ -90,17 +90,29 @@ static char *remove_py_suffix(char *py_name)
 	return py_name;
 }
 
-/* Import python module that is given by -p option. */
-static int import_python_module(char *py_pathname)
+static int set_python_path(char *py_pathname)
 {
 	char py_sysdir[PATH_MAX];
+	char *old_sysdir = getenv("PYTHONPATH");
+	char *new_sysdir = NULL;
+
 	if (absolute_dirname(py_pathname, py_sysdir) == NULL)
 		return -1;
 
-	/* Set path to import a python module. */
-	__PySys_SetPath(py_sysdir);
-	pr_dbg("PySys_SetPath(\"%s\") is done!\n", py_sysdir);
+	if (old_sysdir)
+		xasprintf(&new_sysdir, "%s:%s", old_sysdir, py_sysdir);
+	else
+		new_sysdir = xstrdup(py_sysdir);
 
+	setenv("PYTHONPATH", new_sysdir, 1);
+	free(new_sysdir);
+
+	return 0;
+}
+
+/* Import python module that is given by -p option. */
+static int import_python_module(char *py_pathname)
+{
 	char *py_basename = basename(py_pathname);
 	remove_py_suffix(py_basename);
 
@@ -250,6 +262,8 @@ int script_init_for_python(char *py_pathname)
 	INIT_PY_API_FUNC(PyDict_SetItem);
 	INIT_PY_API_FUNC(PyDict_SetItemString);
 	INIT_PY_API_FUNC(PyDict_GetItem);
+
+	set_python_path(py_pathname);
 
 	__Py_Initialize();
 


honggyukim:
Hi @namhyung ,

I've added a new script to generate sys.path but I'm not sure if there's a better way to get it.
Anyway, it allows uftrace to import matplotlib and numpy now, so please take a look at it.
If you know a better way, please let me know. Thanks!

namhyung:
You'd be better append the script path to `sys.path` instead of replacing it.  But I think changing the `PYTHONPATH` env variable is easier.

http://ask.xmodulo.com/change-syspath-pythonpath-python.html

honggyukim:
@namhyung Thanks! What do you think about the way I got sys.path from interpreter using `gen-py-sys-path.sh`? I would like to find a better way but couldn't find it yet.
We can set `PATHONPATH` after finding the list of python paths but is empty if I read it from a shell,

namhyung:
Hmm.. my approach is the opposite way.  Initially `sys.path` should have a list of system path to load modules.  So we should not change it and add (not replace, if any) the module's path to the `PYTHONPATH`.  The env would be searched in addition to the `sys.path`.

https://docs.python.org/2/using/cmdline.html?#envvar-PYTHONPATH

honggyukim:
Hm.. As far as I know, it's not possible to import uftrace script module without resetting `sys.path` with `PySys_SetPath`. I commented out `PySys_SetPath` and gave the full path of `test.py` but got the below error.
```
$ uftrace -S /home/honggyu/work/uftrace/git/uftrace/scripts/test.py --force --no-pager /usr/bin/uptime
AttributeError: 'module' object has no attribute 'uftrace_entry'
AttributeError: 'module' object has no attribute 'uftrace_exit'
 00:17:22 up 9 days, 13:08, 63 users,  load average: 1.44, 1.13, 1.05
   9.296 us [136237] | setlocale();
   2.213 us [136237] | bindtextdomain();
    ...
```

namhyung:
It seems `Py_Initialize()` checks the env and sets up the path.  So you need to set the env before calling it.  Below patch worked for me. ([set-python-path-before-init.patch.txt](https://github.com/namhyung/uftrace/files/1282993/set-python-path-before-init.patch.txt))

```
diff --git a/utils/script-python.c b/utils/script-python.c
index 3d0c67c..21bb631 100644
--- a/utils/script-python.c
+++ b/utils/script-python.c
@@ -94,17 +94,29 @@ static char *remove_py_suffix(char *py_name)
        return py_name;
 }
 
-/* Import python module that is given by -p option. */
-static int import_python_module(char *py_pathname)
+static int set_python_path(char *py_pathname)
 {
        char py_sysdir[PATH_MAX];
+       char *old_sysdir = getenv("PYTHONPATH");
+       char *new_sysdir = NULL;
+
        if (absolute_dirname(py_pathname, py_sysdir) == NULL)
                return -1;
 
-       /* Set path to import a python module. */
-       __PySys_SetPath(py_sysdir);
-       pr_dbg("PySys_SetPath(\"%s\") is done!\n", py_sysdir);
+       if (old_sysdir)
+               xasprintf(&new_sysdir, "%s:%s", old_sysdir, py_sysdir);
+       else
+               new_sysdir = xstrdup(py_sysdir);
+
+       setenv("PYTHONPATH", new_sysdir, 1);
+       free(new_sysdir);
 
+       return 0;
+}
+
+/* Import python module that is given by -S option. */
+static int import_python_module(char *py_pathname)
+{
        char *py_basename = basename(py_pathname);
        remove_py_suffix(py_basename);
 
@@ -491,6 +503,8 @@ int script_init_for_python(char *py_pathname)
        INIT_PY_API_FUNC(PyDict_SetItemString);
        INIT_PY_API_FUNC(PyDict_GetItem);
 
+       set_python_path(py_pathname);
+
        __Py_Initialize();
 
        /* Import python module that is passed by -p option. */
```

namhyung:
> I commented out PySys_SetPath and gave the full path of test.py but got the below error.
>
> $ uftrace -S /home/honggyu/work/uftrace/git/uftrace/scripts/test.py --force --no-pager /usr/bin/uptime
> AttributeError: 'module' object has no attribute 'uftrace_entry'
> AttributeError: 'module' object has no attribute 'uftrace_exit'

It looks like succeeded to load your script.  The error message shows that your script lacks the functions required by uftrace.  Hmm.. maybe no need to set the env at all?

honggyukim:
Ah.. I didn't have to set `sys.path` using `PySys_SetPath`. You're change smoothly works fine. I think I better close this PR so please merge your change above. Thanks!

namhyung:
No need to use `RTLD_GLOBAL` then?

honggyukim:
@namhyung Ah.. `RTLD_GLOBAL` is needed.

namhyung:
If you want to add my patch in your PR, please use description below:

script: Set python module load path properly

Currently it replace Python's sys.path to the dirname of
the given script.  But it makes to fail to load other modules
in the script.  We can fix it by using PYTHONPATH env
instead of replacing the sys.path.

PR #147, opened 2017-09-07
doc: Add a item 'elapsed time'

Taeung:
Hi @namhyung 

I send this patch because of a bit renewed info output with `elapsed time`.
But I didn't want to change all info output due to changed one thing,
so I just inserted one line with a elapsed time value from trace data recorded
with `t-abc` of tests/s-abc.c in my case.

What do you think about that ?


namhyung:
LGTM.  But it'd be nice to change the value since cpu time already have 0.003 sec of user time.  Maybe you can change the elapsed time also starts with 3 instead of 2 to prevent possible confusion later.

Taeung:
Understood, I resent this patch, is it ok ?

PR #149, opened 2017-09-07
record: Add a comment for 'addr' of uftrace_record

Taeung:
struct uftrace_record has a member 'addr'.
It can mean child ip or uftrace_event_id,
so add this comment.

### DIFF
--- a/uftrace.h
+++ b/uftrace.h
@@ -423,7 +423,7 @@ struct uftrace_record {
 	uint64_t more:   1;
 	uint64_t magic:  3;
 	uint64_t depth:  10;
-	uint64_t addr:   48;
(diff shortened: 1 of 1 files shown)


uftrace.h
```diff
@@ -423,6 +423,8 @@ struct uftrace_record {
 	uint64_t more:   1;
 	uint64_t magic:  3;
 	uint64_t depth:  10;
+
+	/* child ip or uftrace_event_id */
 	uint64_t addr:   48;
```
namhyung: I prefer having the comment on the (right) side unless it's long.

Taeung: Ok, I changed it !

PR #150, opened 2017-09-07
script: Add a new --record option

honggyukim:
It'd be better to have an option to record a new trace before running a
given script.  This patch adds --record option to do it in script
command.

### DIFF
--- a/cmd-script.c
+++ b/cmd-script.c
@@ -140,6 +140,19 @@ int command_script(int argc, char *argv[], struct opts *opts)
 		return -1;
 	}
 
+	if (opts->record) {
+		/* parse in-script record option - "uftrace_options" */
+		parse_script_opt(opts);
+
+		char *script_file = opts->script_file;
+		opts->script_file = NULL;
+
+		pr_dbg("start recording before running a script\n");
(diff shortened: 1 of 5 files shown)


namhyung:
Hmm.. thinking again.  I think this is needed since running script requires specific data from record side.  So allowing record before running script can make sure it sees proper data.  I'll add a way to describe recording time options in the script.

cmd-script.c
```diff
@@ -120,6 +120,16 @@ int command_script(int argc, char *argv[], struct opts *opts)
 		return -1;
 	}
 
+	if (opts->record) {
+		char *script_file = opts->script_file;
+		opts->script_file = NULL;
```
namhyung: To apply in-script record option, you need to call `parse_script_opt()` before making it NULL.

honggyukim:
> To apply in-script record option, you need to call parse_script_opt() before making it NULL.

@namhyung Thanks for your review. I've updated it to understand in-script uftrace option. Another patch relocates `check_binary()` routine after script option processing so that it can understand `--force` option.

honggyukim:
@namhyung Thanks!

PR #152, opened 2017-09-08
script: Add malloc-free.py for memory tracing

honggyukim:
This adds 'malloc-free.py' script to trace malloc/free and additionally
calloc/realloc.  It shows the below output.

  $ uftrace script -S malloc-free.py

     ELAPSED TIME   FUNCTION               ADDRESS           ALLOCATED BYTES
    =========================================================================
             1940 : malloc(80)           = 0x55fd007f4380    80
             8543 : free(0x55fd007f4380)                     0 (-80)
            44840 : malloc(240)          = 0x55fd007f43e0    240
            46546 : free(0x55fd007f43e0)                     0 (-240)
            59560 : malloc(248)          = 0x55fd007f4380    248

                    INVALID ADDRESS FREE
            60886 : free(0x55fd007f4388)                     248 (-xxx)
    =========================================================================

    * 248 bytes are not free-ed in 1 objects

      NON-FREE ADDRESS          SIZE (bytes)
    =========================================================================
        0x55fd007f4380          248
    =========================================================================

It also show objects that are not free-ed at the end of the script.


honggyukim:
Hm.. I found some strange case in gcc. It may have something else that can be `free`ed. I only traced `malloc`, `calloc`, and `realloc`. I tested again but it doesn't call `memalign`. I'm don't know what's missing yet.

honggyukim:
It seems that uftrace misses some of `malloc` related allocation functions that are invoked during startup time. This script cannot show invalid free or correct memory allocation status without it.

honggyukim:
I found that they are so many library calls that use `malloc` internally. Here is the list that I found so far:
```
#
#   void *malloc(size_t size);
#   void free(void *ptr);
#   void *calloc(size_t nmemb, size_t size);
#   void *realloc(void *ptr, size_t size);
#
#   char *strdup(const char *s);
#   char *strndup(const char *s, size_t n);
#   char *strdupa(const char *s);
#   char *strndupa(const char *s, size_t n);
#
#   int asprintf(char **strp, const char *fmt, ...);
#   int vasprintf(char **strp, const char *fmt, va_list ap);
#
#   int posix_memalign(void **memptr, size_t alignment, size_t size);
#   void *aligned_alloc(size_t alignment, size_t size);
#   void *valloc(size_t size);
# 
#   void *memalign(size_t alignment, size_t size);
#   void *pvalloc(size_t size);
#
#   char *realpath(const char *path, char *resolved_path);
#
```

honggyukim:
And are there any libc functions that internally call `free()`?

namhyung:
I don't have a full list of allocation functions, but it seems enough to handle those functions only.  But functions end with 'a' will use stack (using `alloca`) AFAIK.  And some functions return memory by double-pointer so you need to deference it somehow.

scripts/malloc-free.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+
+# uftrace-option: --force --nest-libcall -A malloc@arg1 -R malloc@retval -A free@arg1 -A calloc@arg1,arg2 -R calloc@retval -A realloc@arg1,arg2 -R realloc@retval -A memalign@arg1,arg2 -R memalign@retval -A malloc@libstdc++,arg1 -R malloc@libstdc++,retval -A free@libstdc++,arg1 -A calloc@libstdc++,arg1,arg2 -R calloc@libstdc++,retval -A realloc@libstdc++,arg1,arg2 -R realloc@libstdc++,retval -A memalign@libstdc++,arg1,arg2 -R memalign@libstdc++,retval
```
namhyung: I think it can be simply `--nest-libcall --auto-args` as you set `UFTRACE_FUNCS` below.

PR #155, opened 2017-09-10
uftrace: Make --nest-libcall imply --force option

honggyukim:
Since --nest-libcall can be used for binaries that are not recompiled
with -pg or -finstrument-functions, it can often be used with --force
option.  So it'd be better to make --nest-libcall implies --force
option to make it simpler for general users.

Before:
  $ uftrace --nest-libcall --force /usr/bin/gcc hello.c

After:
  $ uftrace --nest-libcall /usr/bin/gcc hello.c

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -682,6 +682,8 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
 		break;
 
 	case OPT_nest_libcall:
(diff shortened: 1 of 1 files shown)


honggyukim:
I think `--force` option is not user-friendly so better to avoid its usage. Please take a look at it.

honggyukim:
Merged into another PR #157.

PR #156, opened 2017-09-11
info: Allow to read data info inside uftrace.data

honggyukim:
This patch makes uftrace to read data files inside data directory.
It can be useful when uftrace.data can be renamed to something else.
It may be easiser to use '-d' option to specifiy the directory name.

Before:
  $ uftrace record -d uftrace.data.test t-abc
  $ uftrace info -d uftrace.data.test
  < ... shows info for uftrace.data.test ... >

After:
  $ uftrace record -d uftrace.data.test t-abc
  $ cd uftrace.data.test
  $ uftrace info
  < ... shows info for uftrace.data.test ... >


honggyukim:
Merged into another PR #157.

PR #157, opened 2017-09-11
Add various misc commits

honggyukim:
Change log:
- add missing script command in uftrace man page
- make --nest-libcall imply --force
- allow to read data file inside uftrace.data like directory
- fix 131 lib_dlopen test
- fix some tests by properly removing old directory
- add a check routine for --record execution in script to remove coverity warning

### DIFF
--- a/utils/data-file.c
+++ b/utils/data-file.c
@@ -357,51 +357,44 @@ static void check_data_order(struct ftrace_file_handle *handle)
 		pr_dbg("bitfield order is different!\n");
 }
 
-#define RECORD_MSG  "Was '%s' compiled with -pg or\n"		\
-"\t-finstrument-functions flag and ran with uftrace record?\n"
-
 int open_data_file(struct opts *opts, struct ftrace_file_handle *handle)
 {
 	int ret = -1;
 	FILE *fp;
 	char buf[PATH_MAX];
-	bool again = false;
 
 	snprintf(buf, sizeof(buf), "%s/info", opts->dirname);
 
-retry:
 	fp = fopen(buf, "rb");
-	if (fp == NULL) {
-		if (again) {
-			/* restore original file name for error reporting */
-			snprintf(buf, sizeof(buf), "%s/info", opts->dirname);
+	if (fp != NULL)
+		goto ok;
+
+	/* if default dirname is failed */
+	if (!strcmp(opts->dirname, UFTRACE_DIR_NAME)) {
+		/* try again inside the current directory */
+		fp = fopen("./info", "rb");
+		if (fp != NULL) {
+			opts->dirname = "./";
+			goto ok;
 		}
 
-		if (errno == ENOENT) {
-			if (!again && !strcmp(opts->dirname, UFTRACE_DIR_NAME)) {
-				/* retry with old default dirname */
-				snprintf(buf, sizeof(buf), "%s/info",
-					UFTRACE_DIR_OLD_NAME);
-
-				again = true;
-				goto retry;
-			}
-
-			pr_dbg("cannot find %s file!\n", buf);
-
-			if (opts->exename)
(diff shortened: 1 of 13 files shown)


utils/data-file.c
```diff
@@ -365,43 +365,39 @@ int open_data_file(struct opts *opts, struct ftrace_file_handle *handle)
...
 
-	if (again) {
-		/* found data in old dirname, rename it */
-		opts->dirname = UFTRACE_DIR_OLD_NAME;
+		/* restore original file name for error reporting */
+		snprintf(buf, sizeof(buf), "%s/info", opts->dirname);
 	}
 
+	/* data file loading is failed */
+	pr_err("cannot open %s file", buf);
```
Taeung: Hi @honggyukim ,

I tested this patch.
And I found a little shortcoming in the use case `uftrace info --symbols COMMAND` like below:

```
$ uftrace info --symbols hello
uftrace: /home/taeung/git/uftrace/utils/data-file.c:397:open_data_file
 ERROR: cannot open uftrace.data/info file: No such file or directory
```
But I think we can expect the below output even though there isn't a trace data (e.g. uftrace.data )
```
$ uftrace info --symbols hello
Normal symbols
==============
[ 0] 0x400580: _start (size: 42)
[ 1] 0x4005b0: __gmon_start__ (size: 65)
[ 2] 0x4006c6: bar (size: 39)
[ 3] 0x4006ed: foo (size: 49)
[ 4] 0x40071e: main (size: 32)
[ 5] 0x400740: __libc_csu_init (size: 101)
[ 6] 0x4007b0: __libc_csu_fini (size: 2)
[ 7] 0x4007c0: atexit (size: 41)


Dynamic symbols
===============
[ 0] 0x400520: puts (size: 16)
[ 1] 0x400530: printf (size: 16)
[ 2] 0x400540: __libc_start_main (size: 16)
[ 3] 0x400550: __monstartup (size: 16)
[ 4] 0x400560: mcount (size: 16)
[ 5] 0x400570: __cxa_atexit (size: 16)
```

Taeung: Additionally I think we can check a return value of `open_data_file()` like e999a76
So I'm not sure using `pr_err("cannot open %s file", buf);` on the 397 line is right.

honggyukim:
@Taeung Thanks for pointing out the problem. I've just updated it.

utils/data-file.c
```diff
@@ -357,51 +357,44 @@ static void check_data_order(struct ftrace_file_handle *handle)
 		pr_dbg("bitfield order is different!\n");
 }
 
-#define RECORD_MSG  "Was '%s' compiled with -pg or\n"		\
-"\t-finstrument-functions flag and ran with uftrace record?\n"
-
```
namhyung: Why did you remove this message?

honggyukim: The previous usage of RECORD_MSG was as follows:
```
$ uftrace info
WARN: cannot open data: uftrace.data: No such file or directory

$ uftrace info a.out 
WARN: Was 'a.out' compiled with -pg or
        -finstrument-functions flag and ran with uftrace record?
WARN: cannot open data: uftrace.data: No such file or directory

$ uftrace a.out 
Hello World!
   1.053 us [118894] | __monstartup();
   0.853 us [118894] | __cxa_atexit();
            [118894] | main() {
   6.146 us [118894] |   puts();
   3.757 us [118894] |   fflush();
  23.340 us [118894] |   malloc();
   0.826 us [118894] |   free();
  35.827 us [118894] | } /* main */
```
In this case, I thought the message is not informative or correct so removed it.

honggyukim:
I've just rebased and added one more commit to remove coverity warning. Please take a look at it.

utils/data-file.c
namhyung: Fair enough. :)

PR #158, opened 2017-09-12
automatic argument/retval display for well-known functions

honggyukim:
Hi @namhyung and @Taeung 

I'm writing some code for [item 3](https://github.com/namhyung/uftrace/wiki/TODO#3-automatic-argument--retval-display-for-known-libc-functions) and I wrote a subset of working prototype.

I would like to ask you about the interface and internals.
- a new option `--auto-args` is added
- `utils/autoargs.h` describes the list of arguments / retvals for well known library functions
- the arguments and retvals are passed as shown in the field of `uftrace info`.

I would like to hear your comments.

Thanks,
Honggyu


honggyukim:
Here is a example usage and output:
```
$ uftrace --auto-args a.out 
Hello World!
   1.514 us [85955] | __monstartup();
   1.090 us [85955] | __cxa_atexit();
            [85955] | main() {
   7.343 us [85955] |   puts("Hello World!");
   8.333 us [85955] |   fflush();
  22.924 us [85955] |   malloc(0x38d7ea4c68000) = 0;
   1.123 us [85955] |   free(0);
  42.811 us [85955] | } /* main */
```

honggyukim:
I would like to find a better way to specify arguments and retvals for a long list of functions.

libmcount/mcount.h
```diff
@@ -114,7 +114,7 @@ struct mcount_shmem {
 };
 
 /* first 4 byte saves the actual size of the argbuf */
-#define ARGBUF_SIZE  1024
+#define ARGBUF_SIZE  16384
```
namhyung: I don't think this is needed since it's not for passing argpsec.  It is used to save actual contents of arguments and return value.

honggyukim:
> Thanks for doing this. But I think we don't need to pass the auto arg list through the env.

Hmm.. I don't clearly understand your word. I didn't pass auto arg list through env because it may exceed the limit of environment variable length.

> Both of uftrace and libmcount can have the list, so it can just check the name of function and use it for argspec.

Sorry but I don't get your point here again. Can you please explain more in details?

libmcount/mcount.h
honggyukim: I see. I just limited the maximum length to 64 in 29ce50c. Thanks!

honggyukim:
>> Thanks for doing this. But I think we don't need to pass the auto arg list through the env.

> Hmm.. I don't clearly understand your word. I didn't pass auto arg list through env because it may exceed the limit of environment variable length. I just keep using argspec: form to save it in info file.

Clearly saying that I passed auto arg list directly to `libmcount.so` by including `argspec.h` during record time and `uftrace` also includes the same `argspec.h` to write `argspec` to `info` file. So it doesn't pass auto arg list but just use the same info for both.

namhyung:
> Clearly saying that I passed auto arg list directly to libmcount.so by including argspec.h during record time and uftrace also includes the same argspec.h to write argspec to info file. So it doesn't pass auto arg list but just use the same info for both.

Oh, sorry about the misunderstanding.  I can see the code not passing but adding the arg string directly.

cmd-record.c
```diff
@@ -352,9 +354,46 @@ static int fill_file_header(struct opts *opts, int status, struct rusage *rusage
 	if (write(fd, &hdr, sizeof(hdr)) != (int)sizeof(hdr))
 		pr_err("writing header info failed");
 
+	char *orig_args = opts->args;
+	char *orig_retval = opts->retval;
```
namhyung: I prefer declaring variables before any statement like in (old) C syntax.

cmd-record.c
```diff
@@ -352,9 +354,46 @@ static int fill_file_header(struct opts *opts, int status, struct rusage *rusage
...
+	if (opts->auto_args) {
+		/* add auto-args info to write into info file */
+		size_t alloc_size;
+
+		if (orig_args) {
+			alloc_size = strlen(auto_args_list) + strlen(orig_args) + 2;
+			opts->args = (char*)malloc(alloc_size);
```
namhyung: Please use xmalloc (and its friends) and you don't need to cast it in C.

cmd-record.c
```diff
@@ -352,9 +354,46 @@ static int fill_file_header(struct opts *opts, int status, struct rusage *rusage
...
+			sprintf(opts->args, "%s;%s", auto_args_list, orig_args);
+		}
+		else {
+			alloc_size = strlen(auto_args_list) + 1;
+			opts->args = (char*)malloc(alloc_size);
+			strncpy(opts->args, auto_args_list, alloc_size);
+		}
+		pr_dbg("opts->args = %s\n", opts->args);
```
namhyung: For the debug message, I think a level 1 message (`pr_dbg`) should be used to describe high level actions.  I don't want to too many messages when using `-v` alone.

In this case, it may be like "auto arg is used" or "extending args using builtin auto-arg list".  Messages like above can use level 2 or 3 according to the verbosity.

libmcount/mcount.c
```diff
@@ -1201,13 +1203,47 @@ static void mcount_startup(void)
...
+		pr_dbg("retval_str = %s\n", retval_str);
+	}
+	else {
+		size_t alloc_size;
+
+		char *arg_env = getenv("UFTRACE_ARGUMENT");
+		if (arg_env) {
+			alloc_size = strlen(auto_args_list) + strlen(arg_env) + 2;
+			argument_str = (char*)malloc(alloc_size);
+			sprintf(argument_str, "%s;%s", auto_args_list, arg_env);
```
namhyung: you can use `xasprintf()` instead.

libmcount/mcount.c
```diff
@@ -1201,13 +1203,47 @@ static void mcount_startup(void)
...
+		}
+		else {
+			alloc_size = strlen(auto_args_list) + 1;
+			argument_str = (char*)malloc(alloc_size);
+			strncpy(argument_str, auto_args_list, alloc_size);
```
namhyung: And `xstrdup()` for this.

utils/autoargs.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * Automatic arguments / retvals display for known library functions.
+ * It's useful especially for libc functions as of now and can be extended to
+ * many other well-known library functions.
+ *
+ * TODO: need to write a script to generate this form from prototypes
+ */
+static char *auto_args_list = "\
+malloc@arg1;free@arg1;\
```
namhyung: C concatenates adjacent string literals into one.  Also you could also make it a const array rather than a pointer variable.  So it could be:

```
static const char auto_arg_list[] =
  "AAA"
  "BBB"
  "CCC"
;
```

honggyukim:
@namhyung I appreciate your comments in details. I applied your suggestion and updated it as a working-in-progress status. I still need to think how to generate auto args/retvals list.
I just think that I can first support the list in `ltrace` as specified in the link below:
https://github.com/dkogan/ltrace/blob/master/etc/libc.so.conf

cmd-record.c
honggyukim: done!

cmd-record.c
honggyukim: done!

cmd-record.c
honggyukim: Thanks. I changed it to `pr_dbg2`.

libmcount/mcount.c
honggyukim: done!

libmcount/mcount.c
honggyukim: done!

utils/autoargs.h
honggyukim: I changed it as a single string and separated into many pieces of string literals.

honggyukim:
Besides that, I would like to remove library specifier when giving args/retval before writing `autoargs.h`.
i.e. removing `libstdc++` in `malloc@libstdc++,args1`

libmcount/record.c
```diff
@@ -267,6 +267,7 @@ static unsigned save_to_argbuf(void *argbuf, struct list_head *args_spec,
 			if (str) {
 				unsigned i;
 				char *dst = ptr + 2;
+				const unsigned str_limit = 48;
```
namhyung: For this kind of magic number, I prefer using macro constant with capital letters.  Also 48 seems to too small - I don't know what's the good size but 100 looks like reasonable for me.

namhyung: I also see a crash with a long string argument during `uftrace dump`.

namhyung:
> Besides that, I would like to remove library specifier when giving args/retval before writing autoargs.h.
i.e. removing libstdc++ in malloc@libstdc++,args1

Yep, I'm working on that now..

libmcount/record.c
honggyukim: > For this kind of magic number, I prefer using macro constant with capital letters.

Okay. I will make it as capital letter.

> Also 48 seems to too small - I don't know what's the good size but 100 looks like reasonable for me.
> I also see a crash with a long string argument during uftrace dump.

I made it as 64 before but crashed when I run `uftrace dump`. I'm not sure about it but /usr/bin/gcc does a lot of string comparison that exceeds my expectation.

honggyukim:
> @namhyung I appreciate your comments in details. I applied your suggestion and updated it as a working-in-progress status. I still need to think how to generate auto args/retvals list with a script.
I just think that I can first support the function list in ltrace as specified in the link below:
https://github.com/dkogan/ltrace/blob/master/etc/libc.so.conf

I would like to add better example headers in `latrace` by Jiri Olsa but found that it's GPL v3.
https://github.com/jkkm/latrace/tree/master/etc/latrace.d/headers

honggyukim:
>> Besides that, I would like to remove library specifier when giving args/retval before writing autoargs.h.
i.e. removing libstdc++ in malloc@libstdc++,args1

> Yep, I'm working on that now..

I appreciate that!

honggyukim:
FYI, here is some snippet of `gcc` trace result:
```
$ uftrace --nest-libcall --auto-args /usr/bin/gcc hello.c
    ...
   0.250 us [40905] | strcmp("cpp-output", "c");
   0.237 us [40905] | strcmp("c-header", "c");
   0.220 us [40905] | strcmp("c", "c");
   0.567 us [40905] | strncmp("mtune=generic", "E|M|MM:%(trad_capable_cpp) %(cpp_options) %(cp...", 1);
   0.480 us [40905] | strncmp("march=x86-64", "E|M|MM:%(trad_capable_cpp) %(cpp_options) %(cp...", 1);
   0.494 us [40905] | strncmp("mtune=generic", "M|MM:%(trad_capable_cpp) %(cpp_options) %(cpp_...", 1);
   0.476 us [40905] | strncmp("march=x86-64", "M|MM:%(trad_capable_cpp) %(cpp_options) %(cpp_...", 1);
   0.563 us [40905] | strncmp("mtune=generic", "MM:%(trad_capable_cpp) %(cpp_options) %(cpp_de...", 2);
   0.447 us [40905] | strncmp("march=x86-64", "MM:%(trad_capable_cpp) %(cpp_options) %(cpp_de...", 2);
   0.500 us [40905] | strncmp("mtune=generic", "E:%{!M:%{!MM:          %{traditional:%eGNU C n...", 1);
   0.463 us [40905] | strncmp("march=x86-64", "E:%{!M:%{!MM:          %{traditional:%eGNU C n...", 1);
   0.517 us [40905] | malloc(481) = 0xdeb7a0;
   0.433 us [40905] | memcpy(0xdeb7a0, 0x477384, 480);
   0.480 us [40905] | strncmp("mtune=generic", "M:%{!MM:          %{traditional:%eGNU C no lon...", 1);
   0.443 us [40905] | strncmp("march=x86-64", "M:%{!MM:          %{traditional:%eGNU C no lon...", 1);
   0.303 us [40905] | malloc(475) = 0xdeb990;
   0.243 us [40905] | memcpy(0xdeb990, 0xdeb7a5, 474);
    ...
```
In such cases, it may consume a lot of record data so I couldn't increase the limit more than 48.

honggyukim:
> This is a incomplete arbitrary list of standard library functions. I'm not sure how to maintain this list - it'd be great if some conversion script can be provided as well. Also it should be considered that how this can be useful in a long term.

I've added a script that generates `autoargs.h` from pre-defined `protytypes.h` header file in de478f8.
Due to some problems when I give a very long function lists, uftrace doesn record the trace correctly so commented out some of functions.
Please take a look at it!

honggyukim:
Here is the example usage:
```
$ uftrace --auto-args tests/t-autoargs hello 
hello 
            [11503] | main() {
   2.044 us [11503] |   strlen("autoargs test") = 13; 
   1.583 us [11503] |   calloc(1, 14) = 0x1b28a80;
   1.193 us [11503] |   free(0x1b28a80);
   1.344 us [11503] |   strcmp("hello", "hello") = 0;
   4.056 us [11503] |   puts("hello") = 6;
  18.513 us [11503] | } /* main */
```

honggyukim:
@namhyung It seems that this work is almost done but it always generates `autoargs.h` whenever I run `make`. If you know how to fix it, please let me know. Thanks!

utils/gen-autoargs.py
```diff
@@ -44,7 +44,7 @@
 type_specifier.extend(["std::string"])
 
 # The contents of libnames will be made to genearte library name specifier for arg spec
-libnames = ["", "libstdc++"]
+libnames = ["", "libstdc++", "libc++", "libbfd", "libLLVM"]
```
honggyukim: I know this is a bit weird to add such libraries but this commit is added for reference. If library specifier is removed, then we don't need this so we can also make the args/retval list shorter.

Makefile
```diff
@@ -193,6 +194,9 @@ $(filter-out $(objdir)/uftrace.o,$(UFTRACE_OBJS)): $(objdir)/%.o: $(srcdir)/%.c
 $(objdir)/version.h: PHONY
 	@$(srcdir)/misc/version.sh $@ $(VERSION_GIT)
 
+$(objdir)/autoargs.h: PHONY
+	@$(objdir)/utils/gen-autoargs.py $(objdir)/utils/prototypes.h
```
namhyung: I think we don't need to run `gen-autoargs.py` always.  If we provide autoargs.h, it will run only after prototype.h was changed.  Also I'd like put the script and prototype header in the "misc" directory rather than "utils".

honggyukim: Hi, I've just updated as you said. I removed the commit for `Makefile` and added `autoargs.h` instead. Also I moved those files to `misc`. Thanks.

namhyung:
Unlike `version.sh` it doesn't need to be called everytime and deleted on `make clean`.  As it depends on prototype.h and the script itself so the following patch should work IMHO.  But I think it'd be better if it could take an argument to specify name of output file.

```
diff --git a/Makefile b/Makefile
index a8e65d8..455371d 100644
--- a/Makefile
+++ b/Makefile
@@ -194,8 +194,8 @@ $(filter-out $(objdir)/uftrace.o,$(UFTRACE_OBJS)): $(objdir)/%.o: $(srcdir)/%.c
 $(objdir)/version.h: PHONY
        @$(srcdir)/misc/version.sh $@ $(VERSION_GIT)
 
-$(objdir)/autoargs.h: PHONY
-       @$(objdir)/utils/gen-autoargs.py $(objdir)/utils/prototypes.h
+$(objdir)/autoargs.h: $(srcdir)/utils/gen-autoargs.py $(srcdir)/utils/prototypes.h
+       $(QUIET_GEN)$(srcdir)/utils/gen-autoargs.py $(srcdir)/utils/prototypes.h
 
 $(objdir)/uftrace: $(UFTRACE_OBJS) $(UFTRACE_ARCH_OBJS) $(objdir)/libtraceevent/libtraceevent.a
        $(QUIET_LINK)$(CC) $(UFTRACE_CFLAGS) -o $@ $(UFTRACE_OBJS) $(UFTRACE_ARCH_OBJS) $(UFTRACE_LDFLAGS)
@@ -242,8 +242,7 @@ clean:
        $(Q)$(RM) $(objdir)/*.o $(objdir)/*.op $(objdir)/*.so $(objdir)/*.a
        $(Q)$(RM) $(objdir)/utils/*.o $(objdir)/utils/*.op $(objdir)/libmcount/*.op
        $(Q)$(RM) $(objdir)/gmon.out $(srcdir)/scripts/*.pyc $(TARGETS)
-       $(Q)$(RM) $(objdir)/uftrace-*.tar.gz
-       $(Q)$(RM) $(objdir)/version.h $(objdir)/autoargs.h
+       $(Q)$(RM) $(objdir)/uftrace-*.tar.gz $(objdir)/version.h
        @$(MAKE) -sC $(srcdir)/arch/$(ARCH) clean
        @$(MAKE) -sC $(srcdir)/tests ARCH=$(ARCH) clean
        @$(MAKE) -sC $(srcdir)/doc clean
diff --git a/utils/gen-autoargs.py b/utils/gen-autoargs.py
index b15451a..41e171b 100755
--- a/utils/gen-autoargs.py
+++ b/utils/gen-autoargs.py
@@ -270,5 +270,4 @@ def main(argv):
 
 
 if __name__ == "__main__":
-  print("  GEN      " + argspec_file)
   sys.exit(main(sys.argv))
```

honggyukim:
@namhyung Thanks but how can I go back to the previous patch series in the same branch. I changed to directly add `autoargs.h` the force updated the same branch.

namhyung:
`git help reflog`?

honggyukim:
Thanks! I've just updated as you said, especially modified `gen-autoargs.py` to accept more options. Please take a look at it.

namhyung:
Merged: aeeb2b3

PR #160, opened 2017-09-14
[WIP] universal trigger for both entry and exit

honggyukim:
This PR is to implement a prototype of read trigger for both entry and exit.
It's a draft version as of now so I will keep updating if this direction gets agreement.

### DIFF
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -393,7 +393,9 @@ void save_trigger_read(struct mcount_thread_data *mtdp,
 			event = &mtdp->event[mtdp->nr_events++];
 
 			event->id    = EVENT_ID_PROC_STATM;
-			event->time  = rstack->start_time;
+			event->time  = is_return_trigger ?
+					rstack->end_time :
+					rstack->start_time;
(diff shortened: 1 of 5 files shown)


libmcount/record.c
```diff
@@ -507,6 +511,7 @@ static int record_ret_stack(struct mcount_thread_data *mtdp,
 	if (type == UFTRACE_EXIT)
 		timestamp = mrstack->end_time;
 
+	/* TODO: need to explain in comments to show the behavior */
```
honggyukim: I thought it's almost done but I'm not sure how to write the record properly here.
@namhyung Could you please explain about the below code?

honggyukim:
I've studied event recording but I'm wondering why event[] is in `mtdp`, not in `rstack`. I found some problem handling time-filter for event recording. It seems that event doesn't go well with time-filter.

honggyukim:
Since 0c9c82ec0096fab1a4f1c13e6d722c8dbddcd162 has changed read trigger to work on both entry and exit. I will close this PR.

Issue #161, opened 2017-09-14
Time filter doesn't work for read trigger event


honggyukim:
Please see the below example:
```
$ uftrace -T malloc@read=proc/statm -t 100s a.out 
Hello World!
# DURATION    TID     FUNCTION
            [113780] | main() {
            [113780] |   malloc() {
            [113780] |     /* read:proc/statm (size=27368KB, rss=3172KB, shared=2904KB) */
  32.800 us [113780] |   } /* malloc */
            [113780] |   malloc() {
            [113780] |     /* read:proc/statm (size=27568KB, rss=3172KB, shared=2904KB) */
  12.630 us [113780] |   } /* malloc */
  85.644 us [113780] | } /* main */
```
It shows that current time filter doesn't work for read trigger event.

honggyukim:
I've sent a PR #163. Please take a look at it!

namhyung:
Yep, as I said in the PR, events ignore time filter for now.  Maybe we can handle the filter at replay time.

namhyung:
I wrote some patch to distinguish sync and async events in order to apply time filter.  Please take a look at [review/event-filter-v1](https://github.com/namhyung/uftrace/compare/review/event-filter-v1).

honggyukim:
I didn't fully read the internal changes but my test is working fine. Thanks!

honggyukim:
Hmm.. It seems that time filter for read trigger doesn't work in replay time.
```
$ uftrace record -T .*@read=proc/statm t-abc

$ uftrace replay -t 7us
# DURATION    TID     FUNCTION
            [13600] | __monstartup() {
            [13600] |   /* read:proc/statm (size=28736KB, rss=4612KB, shared=3012KB) */
  11.788 us [13600] | } /* __monstartup */
            [13600] | __cxa_atexit() {
            [13600] |   /* read:proc/statm (size=28736KB, rss=4612KB, shared=3012KB) */
   2.916 us [13600] | } /* __cxa_atexit */
            [13600] | main() {
            [13600] |   /* read:proc/statm (size=28736KB, rss=4612KB, shared=3012KB) */
            [13600] |   a() {
            [13600] |     /* read:proc/statm (size=28736KB, rss=4612KB, shared=3012KB) */
            [13600] |     b() {
            [13600] |       /* read:proc/statm (size=28736KB, rss=4612KB, shared=3012KB) */
            [13600] |       c() {
[...]

PR #163, opened 2017-09-15
event: Fix to enable timer-filter for events

honggyukim:
Currently, time-filter doesn't work for events so this patch fixes the
problem.

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -530,14 +530,6 @@ void mcount_entry_filter_record(struct mcount_thread_data *mtdp,
 				save_argument(mtdp, rstack, tr->pargs, regs);
 			if (tr->flags & TRIGGER_FL_READ)
 				save_trigger_read(mtdp, rstack, tr->read);
-
-			if (mtdp->nr_events) {
-				/*
-				 * Flush rstacks if event was recorded as it only has
-				 * limited space for the events.
-				 */
-				record_trace_data(mtdp, rstack, NULL);
-			}
 		}
 
 		/* script hooking for function entry */
@@ -634,6 +626,9 @@ void mcount_exit_filter_record(struct mcount_thread_data *mtdp,
 
 			if (record_trace_data(mtdp, rstack, retval) < 0)
 				pr_err("error during record");
+		} else {
+			/* invalidate events if filtered-out by time filter */
+			mtdp->nr_events = 0;
 		}
 
 		/* script hooking for function exit */
--- a/tests/s-time-to-sleep.c
+++ b/tests/s-time-to-sleep.c
@@ -1,0 +1,31 @@
+#include <stdlib.h>
+#include <unistd.h>
+
+void *mem_alloc(void)
+{
+	return malloc(0);
+}
+
+void mem_free(void *ptr)
+{
+	free(ptr);
+}
+
+void bar(int msec)
+{
+	usleep(msec);
+}
+
+void foo(void)
+{
+	void *p = mem_alloc();
+	bar(2000);
+	mem_free(p);
+}
+
+int main(void)
+{
+	foo();
+	bar(1000);
+	return 0;
+}
--- a/tests/t172_event_time_filter.py
+++ b/tests/t172_event_time_filter.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+
+import re
+from runtest import TestBase
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'time-to-sleep', result="""
+# DURATION    TID     FUNCTION
+            [48790] | main() {
+            [48790] |   foo() {
+            [48790] |     bar() {
+            [48790] |       /* read:proc/statm (size=27368KB, rss=3192KB, shared=2928KB) */
+   2.093 ms [48790] |       usleep();
+   2.124 ms [48790] |     } /* bar */
+   2.130 ms [48790] |   } /* foo */
+   3.257 ms [48790] | } /* main */
+""")
+
+    def runcmd(self):
+        uftrace = TestBase.ftrace
+        args    = '-F main -t 2ms -T bar@read=proc/statm'
+        prog    = 't-' + self.name
+        return '%s %s %s' % (uftrace, args, prog)
+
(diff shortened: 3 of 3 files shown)


Taeung:
Hi @honggyukim 
Long time no see..
Could you show Before/After output for this patch ?
Is the below test wrong way for this patch ?
```
$ uftrace record -E uftrace:event sdt

$ uftrace replay
   0.619 us [16849] | __monstartup();
   0.686 us [16849] | __cxa_atexit();
            [16849] | main() {
            [16849] |   foo() {
            [16849] |     /* uftrace:event */
   6.144 us [16849] |   } /* foo */
   7.551 us [16849] | } /* main */


$ uftrace replay -t 7us
            [16849] | main() {
            [16849] |   foo() {
            [16849] |     /* uftrace:event */
   6.144 us [16849] |   } /* foo */
   7.551 us [16849] | } /* main */

```

Taeung:
Oops, Sorry.. I missed #161
It means the read trigger event, right ?

honggyukim:
Hi @Taeung,

It's been a long time to talk to you. Hope you got better now. :)

> Oops, Sorry.. I missed #161
It means the read trigger event, right ?

Yes, right!

honggyukim:
I've pushed a test for this. You can see the difference as below:
```
$ uftrace -F main -t 1ms -T bar@read=proc/statm -A usleep@arg1 t-time-to-sleep 
            [55601] | main() {
            [55601] |   foo() {
            [55601] |     bar() {
            [55601] |       /* read:proc/statm (size=27368KB, rss=3132KB, shared=2864KB) */
   2.100 ms [55601] |       usleep(2000);
   2.132 ms [55601] |     } /* bar */
   2.141 ms [55601] |   } /* foo */
            [55601] |   bar() {
            [55601] |     /* read:proc/statm (size=27368KB, rss=3132KB, shared=2864KB) */
   1.096 ms [55601] |     usleep(1000);
   1.131 ms [55601] |   } /* bar */
   3.274 ms [55601] | } /* main */
```
There are two `usleep`s, one for 2ms and the other is 1ms. In the above example, time-filter is given 1ms, so both `bar()` shows `proc/statm` info. But if I put 2ms for the same program:
```
$ uftrace -F main -t 2ms -T bar@read=proc/statm -A usleep@arg1 t-time-to-sleep 
            [55656] | main() {
            [55656] |   foo() {
            [55656] |     bar() {
            [55656] |       /* read:proc/statm (size=27368KB, rss=3276KB, shared=3008KB) */
   2.066 ms [55656] |       usleep(2000);
   2.096 ms [55656] |     } /* bar */
   2.106 ms [55656] |   } /* foo */
   3.237 ms [55656] | } /* main */
```
I doesn't show the second `bar`s `proc/statm` info because it's under the time-threshold.

honggyukim:
Hmm.. I found some problem on this change.
```
$ uftrace -F main -t 2ms -T foo@read=proc/statm t-time-to-sleep 
            [63348] | main() {
            [63348] |   foo() {
            [63348] |     bar() {
   2.100 ms [63348] |       usleep();
   2.102 ms [63348] |     } /* bar */
   2.137 ms [63348] |   } /* foo */
   3.236 ms [63348] | } /* main */
```
It doesn't show `foo`'s `proc/statm` info here. I will take a look at it again.

Taeung:
Ok, got it

Wow new test script for this patch is really nice  !!
I think you are so great !  I'll follow you to understood latest your work

honggyukim:
I just copied Namhyung's code and modified a bit. Thanks but you have to say it to him! :)

honggyukim:
Please give me some comments on the recent work. I couldn't finish the job but would like to discuss the direction as well.

libmcount/mcount.c
```diff
@@ -530,14 +530,6 @@ void mcount_entry_filter_record(struct mcount_thread_data *mtdp,
...
 			if (tr->flags & TRIGGER_FL_READ)
 				save_trigger_read(mtdp, rstack, tr->read);
-
-			if (mtdp->nr_events) {
-				/*
-				 * Flush rstacks if event was recorded as it only has
-				 * limited space for the events.
-				 */
-				record_trace_data(mtdp, rstack, NULL);
-			}
```
Taeung: For the reason removing the above lines,
what about additionally appending comments ?  like below:

> The commit 9321fc5 ("mcount: Implement reading memory stat on trigger") flush rstacks
> if event was recorded at the entry but let it can work at the exit for time-filter

I'm not sure this comment is best but I thought we can add `The commit 9321fc5 ("mcount: Implement reading memory stat on trigger") ` to commit log message.

libmcount/mcount.c
```diff
@@ -634,6 +626,9 @@ void mcount_exit_filter_record(struct mcount_thread_data *mtdp,
 
 			if (record_trace_data(mtdp, rstack, retval) < 0)
 				pr_err("error during record");
+		} else {
+			/* invalidate events if filtered-out by time filter */
+			mtdp->nr_events = 0;
```
namhyung: The reason I do this is that SDT events are recorded in a signal handler context.  So it's possible to have multiple events before a function exit and it'll lose some since it can have up to 4 events at once.  So I wanted to flush events as soon as possible regardless of time-filter.

So you need to check whether a event belong to that function actually.  If may be or may be not (it belongs to a parent function).  Also some events (e.g. linux:schedule and kernel events) are not handled this way so we need to filter out them in replay time.

libmcount/mcount.c
honggyukim: @Taeung I think it's better for me to have such comment. Thanks!

libmcount/mcount.c
honggyukim: Hmm.. I think it's a bit confusing to think read trigger as other events. Why don't we separate read trigger records from other events and put it into `mcount_ret_stack` instead of keeping in `mcount_thread_data` so that it can better support time-filter?

namhyung: Adding another buffer for each rstack might not be a good option due to memory usage.  I'm ok if you could make it to reuse argbuf (when used with -A option).

namhyung:
Merged: 78201a3

PR #169, opened 2017-09-24
replay: Add 's' postfix to std::string

honggyukim:
std::string can be directly represented with operator""s in C++14.
So it'd be better to display std::string in C++14 style.


namhyung:
Looks OK.

PR #171, opened 2017-09-25
auto-args with filter module

honggyukim:
This PR includes review/filter-module branch of namhyung then modified check/auto-args based on it.


honggyukim:
I just found that filter-module branch and check/auto-args branch have conflicts so I resolved and created another PR. And I also removed the library name add logic in `gen-autoargs.py` You can just get some reference from it either directly merge this if you need.

honggyukim:
This is not necessary any more so close it.

PR #178, opened 2017-10-13
kernel: Skip functions that seem to be rarely used

Taeung:
Hi @namhyung , I send this PR.

Because we discussed the problem about the kernel functions
that seem needless (i.e. 'syscall_trace_enter_phase1', 'syscall_slow_exit_work',
'smp_irq_work_interrupt')

So, if users don't particularly want to see this functions,
it seems to be better to skip them.

If running simple hello example,
you can see the below kernel functions.

Before:
```
  $ sudo uftrace -k hello
     0.730 us [24138] | __monstartup();
     0.773 us [24138] | __cxa_atexit();
              [24138] | main() {
              [24138] |   printf() {
     0.569 us [24138] |     syscall_trace_enter_phase1();
     1.700 us [24138] |     sys_newfstat();
     0.611 us [24138] |     syscall_slow_exit_work();
     4.927 us [24138] |     __do_page_fault();
     0.734 us [24138] |     smp_irq_work_interrupt();
    14.950 us [24138] |   } /* printf */
    ...
```

After:
```
  $ sudo ufrace -k hello

   0.533 us [24055] | __monstartup();
   0.515 us [24055] | __cxa_atexit();
            [24055] | main() {
            [24055] |   printf() {
   2.197 us [24055] |     sys_newfstat();
   3.389 us [24055] |     __do_page_fault();
   9.987 us [24055] |   } /* printf */
   ...
```

### DIFF
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -467,6 +467,52 @@ out:
 	return -EINVAL;
 }
 
+static void skip_kernel_functions(struct uftrace_kernel_writer *kernel)
+{
+	unsigned int i;
+	struct kfilter *kfilter;
+	const char *skip_funcs[] = {
+		"syscall_trace_enter_phase1",
+		"smp_irq_work_interrupt",
+		"syscall_slow_exit_work",
+		/*
+		 * Some (old) kernel and architecture doesn't support VDSO
+		 * so there will be many sys_clock_gettime() in the output
+		 * due to internal call in libmcount.  It'd be better
+		 * ignoring them not to confuse users.  I think it does NOT
+		 * affect to the output when VDSO is enabled.
+		 */
+		"sys_clock_gettime",
+	};
+
+	for (i = 0; i < ARRAY_SIZE(skip_funcs); i++) {
+		bool add = true;
+		const char *name = skip_funcs[i];
+		struct kfilter *pos;
+
+		/* Don't skip it if user particularly want to see them*/
+		list_for_each_entry(pos, &kernel->filters, list) {
+			if (strcmp(pos->name, name)) {
+				add = false;
+				break;
+			}
+		}
+
+		list_for_each_entry(pos, &kernel->patches, list) {
+			if (strcmp(pos->name, name)) {
+				add = false;
+				break;
+			}
+		}
+
+		if (add) {
+			kfilter = xmalloc(sizeof(*kfilter) + strlen(name) + 1);
+			strcpy(kfilter->name, name);
+			list_add(&kfilter->list, &kernel->notrace);
+		}
+	}
+}
+
 /**
  * setup_kernel_tracing - prepare to record kernel ftrace data (binary)
  * @kernel : kernel ftrace handle
@@ -500,19 +546,8 @@ int setup_kernel_tracing(struct uftrace_kernel_writer *kernel, struct opts *opts
 	/* mark kernel tracing is enabled (for event tracing) */
 	opts->kernel = true;
 
-	if (opts->kernel_skip_out) {
-		/*
-		 * Some (old) kernel and architecture doesn't support VDSO
-		 * so there will be many sys_clock_gettime() in the output
-		 * due to internal call in libmcount.  It'd be better
-		 * ignoring them not to confuse users.  I think it does NOT
-		 * affect to the output when VDSO is enabled.
-		 *
-		 * If an user wants to see them, give --kernel-full option.
-		 */
-		build_kernel_filter(kernel, "!sys_clock_gettime@kernel",
-				    &kernel->filters, &kernel->notrace);
-	}
+	if (opts->kernel_skip_out)
+		skip_kernel_functions(kernel);
 
 	ret = __setup_kernel_tracing(kernel);
 	if (ret < 0)


Taeung:
>  kernel functions to be hided:
> syscall_trace_enter_phase1(), syscall_slow_exit_work(), smp_irq_work_interrupt()

Additionally I thought what is better among `kernel->notrace` and `kernel->nopatch`,
to skip the above functions..
(This patch use `kernel->notrace`, to hide the kernel functions)

What do you think about it ?

I also tested a patch using `kernel->nopatch`

But I found another functions can't be hided like below,
i.e do_audit_syscall_entry(), __audit_syscall_exit(), __audit_syscall_entry(), ...

I think this functions also seem to be rarely needed.
We need to hide all kernel functions that seem needless ?

```
$ sudo uftrace -k  hello
   0.635 us [29472] | __monstartup();
   0.776 us [29472] | __cxa_atexit();
            [29472] | main() {
            [29472] |   printf() {
   0.268 us [29472] |     do_audit_syscall_entry();
   1.082 us [29472] |     sys_newfstat();
   0.347 us [29472] |     __audit_syscall_exit();
   3.304 us [29472] |     __do_page_fault();
   0.212 us [29472] |     irq_enter();
   0.273 us [29472] |     __wake_up();
   0.216 us [29472] |     irq_exit();
  11.395 us [29472] |   } /* printf */
```

namhyung:
There's a similar code to filter out sys_clock_gettime() in `setup_kernel_tracing`.  But I'm not sure setting a filter to those internal functions which might be changed depending on versions.  Also you made it unconditional, but what if some user really want to see those functions?

utils/kernel.c
```diff
@@ -467,6 +465,46 @@ static int __setup_kernel_tracing(struct uftrace_kernel_writer *kernel)
...
+	};
+
+	for (i = 0; i < ARRAY_SIZE(skip_funcs); i++) {
+		bool found = false;
+		const char *name = skip_funcs[i];
+		struct kfilter *pos, *tmp;
+
+		/* Don't skip it if user particularly want to see them*/
+		list_for_each_entry_safe(pos, tmp, &kernel->filters, list) {
+			if (strcmp(pos->name, name))
```
Taeung: > Also you made it unconditional, but what if some user really want to see those functions?

@namhyung ,
This patch can handle the case you said.
Because `kernel->filters` and `kernel->patches` users gave are checked in the above for loop
before skip the particular kernel functions that seem to be rarely used.

Taeung:
> I'd like to leave the argument for the (possible) future change. It's not used now but maybe needed later..
OK, got it !

utils/kernel.c
namhyung: Ah, missed that part, sorry.

But I still want to see all functions when `--kernel-full` option is used.  And could you please change `sys_clock_gettime()` to be part of this change as well?

utils/kernel.c
```diff
@@ -465,6 +465,46 @@ static int __setup_kernel_tracing(struct uftrace_kernel_writer *kernel)
...
+				goto found;
+		}
+
+		list_for_each_entry_safe(pos, tmp, &kernel->patches, list) {
```
namhyung: You don't need to use the "_safe" version unless you want to delete an entry.  Same goes to the loop above.

utils/kernel.c
```diff
@@ -465,6 +465,46 @@ static int __setup_kernel_tracing(struct uftrace_kernel_writer *kernel)
...
+				found = true;
+				break;
+			}
+		}
+
+		if (found) {
+			found = false;
+found:
+			continue;
+		}
```
namhyung: This "found" logic looks not pretty.  What about this?

```
bool add = true;

list_for_each_entry(&kernel->filters)
  if (strcmp() == 0)
    add = false;
    break;

list_for_each_entry(*kernel->patches)
  if (strcmp() == 0)
    add = false;
    break;

if (add)
  list_add(&kernel->notrace);
```

Although it always traverses the 'filters' list and 'patches' list, I think it's fine as they're (usually) small lists.

Taeung:
Happy New Year ! @namhyung 😄 

First of all, sorry for too late update..
I modified this PR based on your comments.
 - When `--kernel-full` option is used, show all kernel functions
 -  When `--kernel-full` option isn't used, skip several functions including `sys_clock_gettime()` in new `skip_kernel_functions()`

I'd appreciate it, if you check this PR.

namhyung:
I need to think about it again.  This is only an interim solution and doesn't cover all cases as you mentioned before.  The function name can be changed anytime and might be different depending on arch.  Also a different Kconfig might add new noises.

For me, syscall tracing is off by default so only concern is the `smp_irq_work_interrupt`.  Actually it's because of the wake ups (for polling ftrace trace-pipe) AFAIK.  The right solution should be changing kernel not to wake up uftrace so often but it will take time.  Currently it's too often and messes up the execution flow so it needs to be filtered out in the meantime.

Taeung:
Hum.. I understood it !
What do you think about trying to solve this problem in kernel parts as you said ?
Or, first of all , you need to intentionally filter some kernel functions ?

namhyung:
I'm thinking of a patch that wakes up waiters of trace-pipe less frequently (like when a page is full).

Taeung:
OK, hum.. it seems that personally I need to more study the domain knowledge..
If you work the patch, could you include me in Cc: list ?

namhyung:
Sure thing, will do.

honggyukim:
@namhyung 
> I changed my mind since changing kernel is hard and it takes a long time. For a short term solution, I'd like to add those functions to the no-trace filter list.

I also think so.

namhyung:
Merged: 05177661

PR #182, opened 2017-10-20
info: Use dynamically allocated buffer for cmd-info.c

honggyukim:
Claudiu Balogh reported that uftrace cannot read data that has more than
1000 threads.  It's because the current implementation stores the
task list in a fixed size buffer (4096).

So this patch makes the buffer to be allocated dynamically to make it
work for any size of task list.

### DIFF
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -506,11 +506,12 @@ static int read_taskinfo(void *arg)
 {
 	struct ftrace_file_handle *handle = arg;
 	struct uftrace_info *info = &handle->info;
-	char buf[4096];
 	int i, lines;
 	int ret = -1;
+	char *buf = NULL;
+	size_t len = 0;
 
-	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+	if (getline(&buf, &len, handle->fp) < 0)
 		goto out;
 
 	if (strncmp(buf, "taskinfo:", 9))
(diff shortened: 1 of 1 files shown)


cmd-info.c
```diff
@@ -506,11 +506,12 @@ static int read_taskinfo(void *arg)
 {
 	struct ftrace_file_handle *handle = arg;
 	struct uftrace_info *info = &handle->info;
-	char buf[4096];
 	int i, lines;
 	int ret = -1;
 
-	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+	char* buf = NULL;
+	size_t len = 0;
```
namhyung: A minor coding-style nit-picking.

Please put declarations together at the beginning of a block.  I prefer keeping a blank line between declarations and statements.  Probably it's an old-style, but I used to it and it helps me understanding code a bit.  Thanks.

namhyung:
`read_arg_spec()` might have similar problem as it deals with a user-given string.

honggyukim:
@namhyung I've done your request and applied to same change to read_arg_spec as well.

@claudiubalogh It's good to hear that you don't have a problem with this fix. Thanks for your report!

namhyung:
@honggyukim LGTM

@claudiubalogh Thanks for testing!

PR #200, opened 2017-10-25
live: Add a new line after record before replay

honggyukim:
Some programs do not provide a new line when the program finishes.
In this case, uftrace replay header is concatenated to the original
program output, which makes the output ugly as follows.

  $ uftrace hello
  Hello World!# DURATION    TID     FUNCTION
     0.243 us [32159] | __monstartup();
     0.090 us [32159] | __cxa_atexit();
              [32159] | main() {
     1.529 us [32159] |   printf();
     1.617 us [32159] |   fflush();
     3.445 us [32159] | } /* main */

This adds a new line between record and replay in live command.

### DIFF
--- a/cmd-live.c
+++ b/cmd-live.c
@@ -163,6 +163,7 @@ int command_live(int argc, char *argv[], struct opts *opts)
 	if (!can_skip_replay(opts, ret)) {
 		int ret2;
 
+		pr_out("\n");
 		reset_live_opts(opts);
 
 		pr_dbg("live-record finished.. \n");


honggyukim:
I've done it and test works fine. Thanks!

honggyukim:
I thought about this again and concluded that it's better not to change the default output.
I just found that `printf` to `puts` transformation is only done by gcc and clang doesn't as follows:
```
$ gcc -pg hello.c
$ uftrace -F main a.out
Hello World!
            [30801] | main() {
   1.066 us [30801] |   puts();
   2.088 us [30801] |   fflush();
   3.898 us [30801] | } /* main */

$ clang -pg hello.c
$ uftrace -F main a.out
Hello World!
            [30881] | main() {
   2.424 us [30881] |   printf();
   3.586 us [30881] |   fflush();
   7.540 us [30881] | } /* main */
```
Thanks!

PR #201, opened 2017-10-26
graph: Make graph output header simpler

honggyukim:
Since the header of graph output was very long, it'd be better to make
it simpler so that it can be shown as a compact form.

Before:
  \#
  \# function graph for 'main' (session: a4016e70192f5080)
  \#

  backtrace
  \=====================================
   backtrace #0: hit 1, time  13.013 us
     [0] main (0x400cc9)

  calling functions
  \=====================================
    13.013 us : (1) main
      ...

After:
  \# Function Call Graph for 'main' (session: a4016e70192f5080)

  \=============== BACKTRACE ===============
   backtrace #0: hit 1, time  13.013 us
     [0] main (0x400cc9)

  \========== FUNCTION CALL GRAPH ==========
    13.013 us : (1) main

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,17 +164,12 @@ The `graph` command shows function call graph of given function.  In the above
 example, function graph of function 'main' looks like below:
 
     $ uftrace graph  main
-    #
-    # function graph for 'main' (session: 8823ea321c31e531)
-    #
-    
-    backtrace
-    ================================
+    # Function Call Graph for 'main' (session: 073f1e84aa8b09d3)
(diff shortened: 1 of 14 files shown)


honggyukim:
@namhyung I've updated all the related files. Please take a look at it. Thanks!

namhyung:
I'm pulling this but I don't think we need to update version in README everytime.

honggyukim:
I was thinking that the version was too much behind the current one so updated it. Thanks!

PR #208, opened 2017-10-29
record: warn if FTRACE not enabled in kernel

bansan85:
Signed-off-by: Vincent LE GARREC <legarrec.vincent@gmail.com>


namhyung:
Maybe it could be a different error but I think most likely it is.

bansan85:
Thx, I made this pull request because it's not said in the doc that this config is needed for "-k". And it took me some time to found out.

PR #215, opened 2017-11-01
doc: Fix uftrace live man page

honggyukim:
The man page of uftrace live is not still updated yet.  This fixes the
document just like uftrace record man page.


namhyung:
LGTM

Issue #148, opened 2017-09-07
compilation error for arm (softfp)


honggyukim:
@namhyung I've got the following compilation error when I try to build it for arm arch.
```
  CC FPIC  libmcount/dynamic.op
  LINK     arch/arm/uftrace.o
arm-linux-gcc: error: unrecognized command line option '-mno-sse2'; did you mean '-fno-dse'?
make: *** [/home/honggyu/src/uftrace/git/arm/uftrace/libmcount/dynamic.op] Error 1
make: *** Waiting for unfinished jobs....
  CC FPIC  libtraceevent/event-parse.o
```

honggyukim:
Hm.. I reset the base to before making such error and got another error. fb70735
```
  CC FPIC  libmcount/plthook-fast.op
  CC FPIC  libmcount/mcount-single.op
/tmp/ccvbOWZe.s: Assembler messages:
/tmp/ccvbOWZe.s:1760: Error: selected processor does not support `vstr %d7,[r3,#16]' in ARM mode
/tmp/ccvbOWZe.s:1798: Error: selected processor does not support `vstr %s0,[r3,#16]' in ARM mode
/tmp/ccvbOWZe.s:1813: Error: selected processor does not support `vstr %s1,[r3,#16]' in ARM mode
    ...
/tmp/ccvbOWZe.s:2128: Error: selected processor does not support `vstr %d6,[r3,#16]' in ARM mode
/tmp/ccvbOWZe.s:2529: Error: selected processor does not support `vstr %s0,[r5,#16]' in ARM mode
/tmp/ccvbOWZe.s:2553: Error: selected processor does not support `vstr %d0,[r5,#16]' in ARM mode
make: *** [/home/honggyu/src/uftrace/git/arm/uftrace/arch/arm/mcount-support.op] Error 1
make: *** Waiting for unfinished jobs....
  CC FPIC  libtraceevent/event-plugin.o

```

namhyung:
First issue looks like due to using host compiler for checking dependencies.  Does following patch make any difference?

```
diff --git a/configure b/configure
index 4c943ed..0ec7675 100755
--- a/configure
+++ b/configure
@@ -100,6 +100,8 @@ output=${output:-${objdir}/.config}
 MAKEFLAGS=
 MAKEOVERRIDES=
 
+export CC CFLAGS LD LDFLAGS
+
 make -siC ${srcdir}/check-deps check-clean
 make -siC ${srcdir}/check-deps check-build
 
```

[...]

PR #218, opened 2017-11-02
build: Add install-elfutils.sh to build elfutils manually

honggyukim:
Sometimes it'd be better to provide full compilation method especially
in cross compile environment or in some systems that cannot install
libelf properly for some reasons.

This adds "misc/install-elfutils.sh" to download and compile elfutils.
The usage is as follows:

     $ export CROSS_COMPILE=arm-linux-gnueabi-
     $ export ARCH=arm
     $ export CFLAGS="-march=armv7-a"
     $ ./misc/install-elfutils.sh --prefix=`pwd`/build.arm
     $ ./configure --prefix=`pwd`/build.arm --with-elfutils=`pwd`/build.arm

     $ make
     $ make install

Some of environmental variables and prefix is better to be same both for
install-elfutils.sh and configure.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -81,12 +81,23 @@ directories or build directory with this script.
     $ ./configure --help
     Usage: ./configure [<options>]
 
-      --help             print this message
-      --prefix=<DIR>     set install root dir as <DIR>        (default: /usr/local)
-      --bindir=<DIR>     set executable install dir as <DIR>  (default: ${prefix}/bin)
-      --libdir=<DIR>     set library install dir as <DIR>     (default: ${prefix}/lib)
-      --mandir=<DIR>     set manual doc install dir as <DIR>  (default: ${prefix}/share/man)
-      --objdir=<DIR>     set build dir as <DIR>               (default: ${PWD})
+      --help                print this message
+      --prefix=<DIR>        set install root dir as <DIR>        (default: /usr/local)
+      --bindir=<DIR>        set executable install dir as <DIR>  (default: ${prefix}/bin)
+      --libdir=<DIR>        set library install dir as <DIR>     (default: ${prefix}/lib)
+      --mandir=<DIR>        set manual doc install dir as <DIR>  (default: ${prefix}/share/man)
+      --objdir=<DIR>        set build dir as <DIR>               (default: ${PWD})
+      --sysconfdir=<DIR>    override the etc dir as <DIR>
+      --with-elfutils=<DIR> search for elfutils in <DIR>/include and <DIR>/lib
+
+      -p                    preserve old setting
+
+      Some influential environment variables:
+        ARCH           Target architecture    e.g. arm, aarch64, or x86_64
+        CROSS_COMPILE  Specify the compiler prefix during compilation
+                       e.g. CC is overridden by $(CROSS_COMPILE)gcc
+        CFLAGS         C compiler flags
+        LDFLAGS        linker flags
 
 Also you can set the target architecture and compiler options like CC, CFLAGS.
 
@@ -96,4 +107,27 @@ For cross compile, you may want to setup the toolchain something like below:
     $ ./configure ARCH=arm CFLAGS='--sysroot /path/to/sysroot'
 
 This assumes you already installed the cross-built `libelf` on the sysroot
-directory.
+directory.  Otherwise, you can also build it from source (please see below) or
+use it on a different path using `--with-elfutils=<PATH>`.
+
+BUILD WITH ELFUTILS (libelf)
+============================
+
+It may be useful to manually compile libelf for uftrace build if the target
+system doesn't have libelf installed.  `misc/install-elfutils.sh` provides a way
+to download and build libelf, which is one of the libraries in elfutils.
+
+The below is the way to compile uftrace together with libelf.
+
+    $ export CROSS_COMPILE=arm-linux-gnueabi-
+    $ export ARCH=arm
+    $ export CFLAGS="-march=armv7-a"
+    $ ./misc/install-elfutils.sh --prefix=/path/to/install
+    $ ./configure --prefix=/path/to/install --with-elfutils=/path/to/install
+
+    $ make
+    $ make install
+
+`misc/install-elfutils.sh` downloads and builds elfutils and install libelf to
+prefix directory.  The installed libelf can be found using `--with-elfutils` in
+`configure` script.
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,10 @@ COMMON_CFLAGS := -O2 -g -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
 COMMON_CFLAGS +=  -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
 #CFLAGS-DEBUG = -g -D_GNU_SOURCE $(CFLAGS_$@)
 COMMON_LDFLAGS := -lelf -lrt -ldl -pthread $(LDFLAGS)
+ifneq ($(elfdir),)
+  COMMON_CFLAGS +=  -I$(elfdir)/include
+  COMMON_LDFLAGS += -L$(elfdir)/lib
+endif
 
 COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
 
@@ -205,6 +209,13 @@ install: all
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(bindir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(libdir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(etcdir)/bash_completion.d
+ifneq ($(wildcard $(elfdir)/lib/libelf.so),)
+ifeq ($(wildcard $(prefix)/lib/libelf.so),)
+	# install libelf only when it's not in the install directory.
+	$(call QUIET_INSTALL, libelf)
+	$(Q)$(INSTALL) $(elfdir)/lib/libelf.so   $(DESTDIR)$(libdir)/libelf.so
+endif
+endif
 	$(call QUIET_INSTALL, uftrace)
 	$(Q)$(INSTALL) $(objdir)/uftrace         $(DESTDIR)$(bindir)/uftrace
 	$(call QUIET_INSTALL, libmcount)
--- a/configure
+++ b/configure
@@ -20,6 +20,13 @@ usage() {
   --with-elfutils=<DIR> search for elfutils in <DIR>/include and <DIR>/lib
 
   -p                    preserve old setting
+
+  Some influential environment variables:
+    ARCH           Target architecture    e.g. arm, aarch64, or x86_64
+    CROSS_COMPILE  Specify the compiler prefix during compilation
+                   e.g. CC is overridden by \$(CROSS_COMPILE)gcc
+    CFLAGS         C compiler flags
+    LDFLAGS        linker flags
 "
     exit 1
 }
--- a/configure
+++ b/configure
@@ -10,15 +10,16 @@ output=${output:-${objdir}/.config}
 usage() {
     echo "Usage: $0 [<options>]
 
-  --help             print this message
-  --prefix=<DIR>     set install root dir as <DIR>        (default: /usr/local)
-  --bindir=<DIR>     set executable install dir as <DIR>  (default: \${prefix}/bin)
-  --libdir=<DIR>     set library install dir as <DIR>     (default: \${prefix}/lib)
-  --mandir=<DIR>     set manual doc install dir as <DIR>  (default: \${prefix}/share/man)
-  --objdir=<DIR>     set build dir as <DIR>               (default: \${PWD})
-  --sysconfdir=<DIR> override the etc dir as <DIR>
+  --help                print this message
+  --prefix=<DIR>        set install root dir as <DIR>        (default: /usr/local)
+  --bindir=<DIR>        set executable install dir as <DIR>  (default: \${prefix}/bin)
+  --libdir=<DIR>        set library install dir as <DIR>     (default: \${prefix}/lib)
+  --mandir=<DIR>        set manual doc install dir as <DIR>  (default: \${prefix}/share/man)
+  --objdir=<DIR>        set build dir as <DIR>               (default: \${PWD})
+  --sysconfdir=<DIR>    override the etc dir as <DIR>
+  --with-elfutils=<DIR> search for elfutils in <DIR>/include and <DIR>/lib
 
-  -p                 preserve old setting
+  -p                    preserve old setting
 "
     exit 1
 }
@@ -105,8 +106,9 @@ export CC CFLAGS LD LDFLAGS
 make -siC ${srcdir}/check-deps check-clean
 make -siC ${srcdir}/check-deps check-build
 
-if [ ! -e ${srcdir}/check-deps/have_libelf ]; then
+if [ ! -e ${srcdir}/check-deps/have_libelf ] && [ ! -f ${with_elfutils}/lib/libelf.so ]; then
     echo "Error: cannot find 'libelf' (from elfutils): Please install it first."
+    echo "Otherwise, please install it using \"misc/install-elfutils.sh\" script."
     exit 1
 fi
 
@@ -117,6 +119,13 @@ override bindir := $bindir
 override libdir := $libdir
 override mandir := $mandir
 override etcdir := $etcdir
+EOF
+
+if [ ! -z $with_elfutils ]; then
+    echo "override elfdir := $with_elfutils" >> $output
+fi
+
+cat >>$output <<EOF
 
 override ARCH   := $ARCH
 override CC     := $CC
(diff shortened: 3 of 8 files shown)


honggyukim:
Hm.. When I cross compile uftrace, it has some compilation error as follows:
```
perf.s:846: Error: selected processor does not support `dmb ish' in ARM mode
```
It comes from the below code:
```c
void record_perf_data(struct uftrace_perf_writer *perf, int cpu, int sock)
{
        struct perf_event_mmap_page *pc = perf->page[cpu];
        unsigned char *data = perf->page[cpu] + pc->data_offset;
        volatile uint64_t *ptr = (void *)&pc->data_head;
        uint64_t mask = pc->data_size - 1;
        uint64_t old, pos, start, end;
        unsigned long size;
        unsigned char *buf;

        pos = *ptr;
        old = perf->data_pos[cpu];

        /* ensure reading the data head first */
        read_memory_barrier();

        if (pos == old)
                return;
```
`read_memory_barrier` is as follows in compiler.h:
```c
#if defined(__arm__)
# define cpu_relax()            compiler_barrier()
# define full_memory_barrier()  asm volatile("dmb ish" ::: "memory")
# define read_memory_barrier()  asm volatile("dmb ish" ::: "memory")
# define write_memory_barrier() asm volatile("dmb ishst" ::: "memory")
#endif
```
The complete command for that is
```
$ arm-linux-gnueabi-gcc -O2 -g -D_GNU_SOURCE   -iquote /home/honggyu/work/uftrace -iquote /home/honggyu/work/uftrace -iquote /home/honggyu/work/uftrace/arch/x86_64 -I/home/honggyu/work/uftrace/.build/zlib -I/home/honggyu/work/uftrace/.build/elfutils/libelf -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers -DHAVE_CXA_DEMANGLE -DHAVE_PERF_CLOCKID -DHAVE_PERF_CTXSW   -c -o /home/honggyu/work/uftrace/utils/perf.o /home/honggyu/work/uftrace/utils/perf.c
```
Did I miss some architecture related options for that?

namhyung:
> -iquote /home/honggyu/work/uftrace/arch/x86_64 \
> ...
> Did I miss some architecture related options for that?

You need to set `ARCH` variable to "arm".

namhyung:
Also I think it'd be better moving the zlib compilation from configure to make.  And you need to consider cases like "srcdir != objdir".

honggyukim:
I got the same error even if I use `ARCH=arm`, but found that `-march=armv7-a` has to be explicitly specified. I will see if I can find a good way to use it.

honggyukim:
> Error: selected processor does not support `vstr %d7,[r2,#16]' in ARM mode

The above error message can also be disappears if I give `-mfloat-abi=softfp` option. So the problem is that we don't actually specify `-march=armv7-a` and `-mfloat-abi=softfp` options even if we give `ARCH=arm`.

honggyukim:
One more thing I want to make sure is what about doing compilation and installation during `configure --build-elfutils`?
- Anyway, it requires zlib to be compiled to complete elfutils configuration.
- And the path setting for compilation is better to be in installed path rather than directories in `.build/elfutils`.
- If we compile elfutils whenever we do `make clean`, it requires more time for `elfutils` compilation.
- Make build rule check is a bit strange to handle.

So I think that we can first build and install elfutils with `configure --build-elfutils` or `configure --install-elfutils` to the directory by `--prefix`. Then, we can set the include and linker path to the installed directories.

namhyung:
> We need to provide a way to specify those architecture specific options.

Isn't it sufficient to use `CFLAGS`?  And I think it's related to #148.

> One more thing I want to make sure is what about doing compilation and installation during configure --build-elfutils?

Installing to the prefix might require root privilege so I don't want to do it during configure or compile stage.  This `--build-elftuils` thing is tricky so I won't care even if it uses some random directory during build.  Maybe you can ignore deleting it when `make clean` is invoked.

honggyukim:
>> We need to provide a way to specify those architecture specific options.

> Isn't it sufficient to use CFLAGS? And I think it's related to #148.

I found that the users can explicitly specify arch-specific flags as follows:
```
$ CROSS_COMPILE=arm-linux-gnueabi- ARCH=arm CFLAGS="-march=armv7-a -mfloat-abi=softfp" \
       ./configure --prefix=`pwd`/build.arm --build-elfutils
```
Would it be okay for you? I know that [review/arm-softfp-v1](https://github.com/namhyung/uftrace/compare/review/arm-softfp-v1) is needed but it's working even without it anyway using `-mfloat-abit=softfp`.

>> One more thing I want to make sure is what about doing compilation and installation during configure --build-elfutils?

> Installing to the prefix might require root privilege so I don't want to do it during configure or compile stage. This --build-elftuils thing is tricky so I won't care even if it uses some random directory during build. Maybe you can ignore deleting it when make clean is invoked.

I think `make install` eventually requires root privilege if `prefix` is not properly set to its local directory. I already wrote a working version of `--install-elfutils` in [check/install-elfutils](https://github.com/honggyukim/uftrace/compare/check/install-elfutils) before receiving your comment.
It's different from your expectation, but could you please give me some feedback on it?
You can test it using the commands below:
```
$ CROSS_COMPILE=arm-linux-gnueabi- ARCH=arm CFLAGS="-march=armv7-a -mfloat-abi=softfp" \
       ./configure --prefix=`pwd`/build.arm --install-elfutils
$ make
$ make install
```

namhyung:
> Would it be okay for you using CFLAGS outside?

Yes, it should respect the CFLAGS IMHO.

> If you still don't agree on it, I can also change it back to --build-elfutils for only configuration, not configure-time installation.

I'd be supprised if configure script tries to compile something non-trivial.  On the other hand, I sometimes see dependant extra programs are configured during compilation.  So I suggest deferring the build to the make command.

Makefile
```diff
@@ -131,7 +136,7 @@ CFLAGS_$(objdir)/utils/demangle.op = -Wno-unused-value
 MAKEFLAGS = --no-print-directory
 
 
-all: $(objdir)/.config $(TARGETS)
+all: $(objdir)/.config $(srcdir)/check-deps/have_libelf $(TARGETS)
```
honggyukim: @namhyung There is one thing I do care about is elfutils has to be compiled before uftrace linking but I tried to put the below:
```
.NOTPARALLEL: all
```
It makes the order correctly but invalidate all the parallel compilation for uftrace source code. Could you let me know if you know better way?

namhyung: Then why not making uftrace and libmcount depend on libelf?

honggyukim: Hm.. The script is getting more complicated.
```make
HAVE_LIBELF := $(srcdir)/check-deps/have_libelf

$(objdir)/libmcount/libmcount.so: $(LIBMCOUNT_OBJS) $(LIBMCOUNT_UTILS_OBJS) $(LIBMCOUNT_ARCH_OBJS) $(HAVE_LIBELF)
        $(QUIET_LINK)$(CC) -shared -o $@ $^ $(LIB_LDFLAGS)
```
Adding `have_libelf` into dependency list makes it to be compiled together that makes compilation error. I may have to think about it in a different way.

honggyukim: It's getting longer but I just used `filter-out` to fix it.

honggyukim:
I've updated this again. `configure` now only does configuration after downloading and the actual compilation for elfutils is done with `make` command. But `zlib` still requires to be compiled during `configure` time.

namhyung:
I tried to use it but failed during confiure elfutils like below:

```
$ CROSS_COMPILE=$HOME/x-tools/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi- ARCH=arm CFLAGS=-march=armv7-a ./configure --prefix=$HOME/build --build-elfutils
...
checking host system type... Invalid configuration `/home/namhyung/x-tools/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi': machine `/home/namhyung/x-tools/arm-unknown-linux-gnueabi/bin/arm-unknown' not recognized
configure: error: /bin/sh config/config.sub /home/namhyung/x-tools/arm-unknown-linux-gnueabi/bin/arm-unknown-linux-gnueabi failed
```

It was due to your script assumes CROSS_COMPILE variable doesn't have pathname components.  I fixed it with below change but it still failed:
```
  LINK     libmcount/libmcount.so
/home/namhyung/project/uftrace/.build/elfutils/libelf/libelf.so: file not recognized: File format not recognized
collect2: error: ld returned 1 exit status
make: *** [Makefile:176: /home/namhyung/project/uftrace/libmcount/libmcount.so] 오류 1
make: *** 끝나지 않은 작업을 기다리고 있습니다....
```

I managed to build it by passing CC with following change:
```diff
diff --git a/configure b/configure
index 5e70f4a..5f08a7c 100755
--- a/configure
+++ b/configure
@@ -71,8 +71,8 @@ prepare_elfutils() {
     tar xfj $ELFUTILS_TARBALL
     mv $ELFUTILS_NAME elfutils
     cd elfutils
-    HOST=$(echo $CROSS_COMPILE | sed 's/-$//g')
-    ./configure --prefix=$prefix --host=$HOST \
+    HOST=$(basename $CROSS_COMPILE | sed 's/-$//g')
+    ./configure --prefix=$prefix --host=$HOST CC="${CROSS_COMPILE}gcc" \
         CFLAGS="-I${builddir}/zlib" LDFLAGS="-L${builddir}/zlib -lz"
     cd ..
 }
```

misc/install-elfutils.sh
```diff
@@ -0,0 +1,68 @@
...
+
+ELFUTILS_VERSION=0.164
+ELFUTILS_NAME=elfutils-$ELFUTILS_VERSION
+ELFUTILS_TARBALL=$ELFUTILS_NAME.tar.bz2
+ELFUTILS_URL=https://sourceware.org/elfutils/ftp/$ELFUTILS_VERSION/$ELFUTILS_TARBALL
+
+wget -c $ELFUTILS_URL
+tar xfj $ELFUTILS_TARBALL
+rm -rf elfutils
+mv $ELFUTILS_NAME elfutils
```
namhyung: What about doing above only if "elfutils" directory doesn't exist?

misc/install-elfutils.sh
```diff
@@ -0,0 +1,68 @@
...
+cd elfutils
+
+opt_host_cc=""
+if [ ! -z $CROSS_COMPILE ]; then
+    HOST=$(basename $CROSS_COMPILE | sed 's/-$//g')
+    opt_host_cc="--host=$HOST CC=${CROSS_COMPILE}gcc"
+fi
+
+./configure --prefix=$prefix $opt_host_cc
```
namhyung: Likewise, please do this if Makefile is not there.

Makefile
```diff
@@ -52,8 +52,10 @@ export ARCH CC AR LD RM srcdir objdir LDFLAGS
 
 COMMON_CFLAGS := -O2 -g -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
 COMMON_CFLAGS +=  -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
+COMMON_CFLAGS +=  -I$(prefix)/include
 #CFLAGS-DEBUG = -g -D_GNU_SOURCE $(CFLAGS_$@)
-COMMON_LDFLAGS := -lelf -lrt -ldl -pthread $(LDFLAGS)
+COMMON_LDFLAGS := -L$(prefix)/lib
+COMMON_LDFLAGS += -lelf -lrt -ldl -pthread $(LDFLAGS)
```
namhyung: Please minimize the diff by adding "-L" option later.

configure
```diff
@@ -7,6 +7,8 @@ srcdir=$(readlink -f $(dirname $0))
 objdir=$(readlink -f ${objdir:-${PWD}})
 output=${output:-${objdir}/.config}
 
+n_cpus=$(cat /proc/cpuinfo | grep ^processor | wc -l)
+
```
namhyung: Not needed anymore?

honggyukim:
@namhyung I've done the modification as you suggested. Please take a look at it. Thanks!

misc/install-elfutils.sh
honggyukim: We have to check whether configure option is same as before, but I just made it simply check whether Makefile is there or not.

misc/install-elfutils.sh
honggyukim: I removed it.

configure
honggyukim: done!

Makefile
honggyukim: done!

misc/install-elfutils.sh
honggyukim: I just modified it only when configure parameters are the same.

Makefile
```diff
@@ -52,8 +52,10 @@ export ARCH CC AR LD RM srcdir objdir LDFLAGS
 
 COMMON_CFLAGS := -O2 -g -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
 COMMON_CFLAGS +=  -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
+COMMON_CFLAGS +=  -I$(elfdir)/include
 #CFLAGS-DEBUG = -g -D_GNU_SOURCE $(CFLAGS_$@)
 COMMON_LDFLAGS := -lelf -lrt -ldl -pthread $(LDFLAGS)
+COMMON_LDFLAGS += -L$(elfdir)/lib
```
namhyung: What about this?

```
ifneq ($(elfdir),)
  COMMON_CFLAGS  += -I$(elfdir)/include
  COMMON_LDFLAGS += -L$(elfdir)/lib
endif
```

Makefile
```diff
@@ -205,6 +207,12 @@ install: all
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(bindir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(libdir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(etcdir)/bash_completion.d
+ifneq ($(wildcard $(elfdir)/lib/libelf.so),)
```
namhyung: please add a comment that it only installs when it's built but not installed yet.

misc/install-elfutils.sh
```diff
@@ -0,0 +1,76 @@
...
+mkdir -p ${builddir} && cd ${builddir}
+
+ELFUTILS_VERSION=0.164
+ELFUTILS_NAME=elfutils-$ELFUTILS_VERSION
+ELFUTILS_TARBALL=$ELFUTILS_NAME.tar.bz2
+ELFUTILS_URL=https://sourceware.org/elfutils/ftp/$ELFUTILS_VERSION/$ELFUTILS_TARBALL
+
+wget -c $ELFUTILS_URL
+tar xfj $ELFUTILS_TARBALL
+ln -sfn $ELFUTILS_NAME elfutils
```
namhyung: Doesn't it overwrite the source of elfutils everytime?

Makefile
honggyukim: Done. Thanks!

namhyung:
Could you please update INSTALL.md too?

honggyukim:
Sure. I've done it!

INSTALL.md
```diff
@@ -81,12 +81,23 @@ directories or build directory with this script.
...
-      --bindir=<DIR>     set executable install dir as <DIR>  (default: ${prefix}/bin)
-      --libdir=<DIR>     set library install dir as <DIR>     (default: ${prefix}/lib)
-      --mandir=<DIR>     set manual doc install dir as <DIR>  (default: ${prefix}/share/man)
-      --objdir=<DIR>     set build dir as <DIR>               (default: ${PWD})
+      --help                print this message
+      --prefix=<DIR>        set install root dir as <DIR>        (default: /usr/local)
+      --bindir=<DIR>        set executable install dir as <DIR>  (default: \${prefix}/bin)
+      --libdir=<DIR>        set library install dir as <DIR>     (default: \${prefix}/lib)
+      --mandir=<DIR>        set manual doc install dir as <DIR>  (default: \${prefix}/share/man)
+      --objdir=<DIR>        set build dir as <DIR>               (default: \${PWD})
```
namhyung: Please remove the backslash in the output.

INSTALL.md
```diff
@@ -81,12 +81,23 @@ directories or build directory with this script.
...
+      --with-elfutils=<DIR> search for elfutils in <DIR>/include and <DIR>/lib
+
+      -p                    preserve old setting
+
+      Some influential environment variables:
+        ARCH          Target architecture    e.g. arm, aarch64, or x86_64
+        CROSS_COMPILE Specify the compiler prefix during compilation
+                      e.g. CC is overridden by $(CROSS_COMPILE)gcc
+        CFLAGS        C compiler flags
+        LDFLAGS       linker flags
```
namhyung: Could you please add one more space between variable name and decription?  Same for the previous commit too.

INSTALL.md
```diff
@@ -97,3 +108,25 @@ For cross compile, you may want to setup the toolchain something like below:
 
 This assumes you already installed the cross-built `libelf` on the sysroot
 directory.
+
```
namhyung: "Or else you can build it from source (please see below) or use it on a different path using `--with-elfutils=<PATH>`."

honggyukim:
Updated it again. Thanks for pointing out in details :)

PR #219, opened 2017-11-04
doc: Add missing imply rules in man pages.

honggyukim:
Since --nest-libcall implies --force, it updates record and live man
pages.  It also updates record man pages that --kernel-buffer implies
--kernel as well.


honggyukim:
> Does --kernel-buffer imply --kernel?

Ah.. It used to be but now it's not by 41b584d.

honggyukim:
I've updated uftrace-live man page regarding `--kernel-buffer`. Thanks!

PR #223, opened 2017-11-09
doc:Fix example command

JangSoJin:
backtrace trigger is not working with `-F` any more.
So it should be replaced by `-T` option.

Before
~~~
$ uftrace replay -F 'b@backtrace,depth=2'
~~~

After
~~~
$ uftrace replay -T 'b@filter,backtrace,depth=2'
~~~

Thanks !! :)

Signed-off-by: JangSoJin <wkdthwls3@naver.com>


namhyung:
LGTM

Issue #232, opened 2017-11-14
what is testcase 126 intend?


ParkHanbum:
testcase 126 do not passing argument for called function 'alloc1', 
but in testcase compare between first argument and 1.

``` 
    def runcmd(self):
        return '%s -A "alloc*@arg1" %s' % (TestBase.ftrace, 't-allocfree')
```

```
static void * alloc1(void)
{
        return alloc2();
}


int main(int argc, char *argv[])
{
        free1(alloc1());
        return 0;
}
                     
```

```
        TestBase.__init__(self, 'allocfree', """
# DURATION    TID     FUNCTION
   2.417 us [32130] | __monstartup();
   1.535 us [32130] | __cxa_atexit();
            [32130] | main() {
            [32130] |   alloc1(1) {
            [32130] |     alloc2(1) {
            [32130] |       alloc3(1) {
            [32130] |         alloc4(1) {
            [32130] |           alloc5(1) {
   1.850 us [32130] |             malloc(1);

```

testcase 127 seem to similar of this.
would you explain this?

namhyung:
Oh, you're right.  It should pass the argument, could you please send a PR (or do you want me to fix it)?

ParkHanbum:
@namhyung sure 😆

PR #233, opened 2017-11-14
fix for testcase 126, 127. #232

ParkHanbum:
passing int argument to alloc and its sub functions.
volatile is needed, for disable the optimization.


namhyung:
Merged with commit message slightly change: 981f017

Issue #242, opened 2017-11-16
Segmentation Fault when using pthread_exit()


minchae:
i found a segfault error when using pthread_exit()

*  target source

```c++
#include <stdio.h>
#include <pthread.h>
#include <stdlib.h>
#include <unistd.h>

#define NUM_THREAD 4

pthread_t threads[10];
void *thread_main(void *);

int main(void)
{
	int i,rc,status,x[NUM_THREAD]={0,};

	printf("pid=%d\n", (int)getpid());

	for (int i = 0; i < NUM_THREAD; i++){
		x[i] = i;
		pthread_create(&threads[i], NULL, &thread_main, (void *)x[i]);
	}

	for (int i = 0; i < NUM_THREAD; i++){
		rc = pthread_join(threads[i], (void **)&status);
	}
	return 0;
}

inline void *thread_main(void *arg)
{
	float result = 0;
	for (int i = 0; i < 2; i++)
	{
		result = result + (double)random();
	}
	printf("thread: %d, result = %f\n", arg, result);
	pthread_exit((void *)0);
}
```
* result
```
```

* result by uftrace
```
$ uftrace ./target
=================================================
pid=8531
thread: 2, result = 2651220224.000000
process crashed by signal 11: Segmentation fault (si_code: 128)
child terminated by signal: 11: Segmentation fault
# DURATION    TID     FUNCTION
   0.459 us [ 8531] | __monstartup();
   0.387 us [ 8531] | __cxa_atexit();
            [ 8531] | main() {
   0.301 us [ 8531] |   getpid();
   2.710 us [ 8531] |   printf();
  16.553 us [ 8531] |   pthread_create();
   8.896 us [ 8531] |   pthread_create();
  11.488 us [ 8531] |   pthread_create();
   9.841 us [ 8531] |   pthread_create();
            [ 8535] | thread_main() {
   1.270 us [ 8535] |   random();
   0.082 us [ 8535] |   random();
   4.288 us [ 8535] |   printf();
            [ 8535] |   pthread_exit() {
            [ 8536] | thread_main() {
   0.233 us [ 8536] |   random();
   0.061 us [ 8536] |   random();

uftrace stopped tracing with remaining functions
================================================
task: 8531
[0] main

task: 8535
[1] pthread_exit
[0] thread_main

task: 8536
[0] thread_main
```

this is an unusual case when using double and float type and print it.

I think this problem caused by pthread_exit (), which refers to where thread memory is invalied.

[...]

PR #243, opened 2017-11-16
Add a test case for terminated thread by pthread_exit() function

minchae:
Issue : https://github.com/namhyung/uftrace/issues/242

Small source of segfault caused by pthread_exit() function

### DIFF
--- a/tests/s-thread-pthread_exit.c
+++ b/tests/s-thread-pthread_exit.c
@@ -1,0 +1,27 @@
+#include <pthread.h>
+
+#define NUM_THREAD 4
+
+pthread_t threads[NUM_THREAD];
+void *thread_main(void *arg)
+{
+	double result;
+	result = 1.0;
+	printf("%f\n",result);
+	pthread_exit((void *) 0);
+}
+
+int main(void)
+{
+	for (int i = 0; i < NUM_THREAD; i++){	
+		pthread_create(&threads[i], NULL, &thread_main, (void *)i);
+	}
+	for (int i = 0; i < NUM_THREAD; i++){
+		int status;	
+		pthread_join(threads[i], (void **)&status);
+	}
+	return 0;
+}
+
+
+


tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
+#include <pthread.h>
+
```
namhyung: I think you need to add `<stdio.h>` as well.  Also I prefer a short file name like `s-thread-exit.c`.

tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
...
+#define NUM_THREAD 4
```
namhyung: Hmm.. do you really need that number of threads?  Isn't it reproducible if it creates a single thread?

tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
...
+
+pthread_t threads[NUM_THREAD];
+void *thread_main(void *arg)
+{
+	double result;
+	result = 1.0;
+	printf("%f\n",result);
+	pthread_exit((void *) 0);
```
namhyung: I still want to add a return statement at the end.

tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+int main(void)
+{
+	for (int i = 0; i < NUM_THREAD; i++){
```
namhyung: Please put a space before the starting brace ("{").

tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
...
+	double result;
+	result = 1.0;
+	printf("%f\n",result);
+	pthread_exit((void *) 0);
+}
+
+int main(void)
+{
+	for (int i = 0; i < NUM_THREAD; i++){	
+		pthread_create(&threads[i], NULL, &thread_main, (void *)i);
```
namhyung: The `(void *)i` will complain about the cast.  As it's not used in the thread_main(), you can just pass NULL here.  Also please put a space between ")" and "{".

tests/s-thread-pthread_exit.c
```diff
@@ -0,0 +1,27 @@
...
+	}
+	for (int i = 0; i < NUM_THREAD; i++){
+		int status;	
+		pthread_join(threads[i], (void **)&status);
+	}
+	return 0;
+}
+
+
+
```
namhyung: Please don't add a empty line unnecessarily.

namhyung:
Oh, also your branch based on an old code.  Please rebase! :)

namhyung:
Also you missed the Signed-off-by line in the commit message.  Without it, I cannot take your patch.

PR #245, opened 2017-11-17
Update testcase for i386

ParkHanbum:
hello. 
I changed some test cases for i386 support. 

first commit : 
```
    s-exp-int.c
    changed function int_div's first argument type to int from unsigned.
    because, when divide "unsigned type" to "long type",
    it will be unsigned operation during divide in 32bit system.
    we call this "automatic conversion".
    test does not intend unsigned divide, but instead a "signed" divide.

    t043_full_demangle.py,
    t058_arg_int.py, t063_retval.py, t065_arg_order.py
    extended argument option for arm to 32bit.
```

second commit:
```
    since i386 use stack for passing argument,
    previous argument size are affect to next argument position.
    for example when first argument's size are over 32bit,
    second argument have position must be 3.

    the following testcase have been changed to consider for these issue:
    t083_arg_float.py, t084_arg_mixed.py, t086_arg_stack.py
    t087_arg_variadic.py
```

### DIFF
--- a/tests/s-exp-int.c
+++ b/tests/s-exp-int.c
@@ -13,7 +13,7 @@ int int_mul(long long a, int b)
 	return a * b;
 }
 
-int int_div(unsigned a, long b)
+int int_div(int a, long b)
 {
 	return a / b;
 }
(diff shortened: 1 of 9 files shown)


namhyung:
Looks ok but you need to change commit subject to have a proper prefix and to be a complete sentence like below:

1. test: prevent unsigned division on 32-bit systems
2. test: adjust argument offset on 32-bit systems

ParkHanbum:
@namhyung consider it done.

PR #254, opened 2017-11-18
some fix for uftrace

ParkHanbum:
this "PR" is splitted part from "support-i386".
two fix for uftrace.

first commit:
```
trigger: do not use memcpy() for copying trigger

if we use memcpy to copy the &iter->trigger,
depending on the size of the uftrace_trigger *tr
we will ride optimization path to use SSE inside memcpy.
it will change status of SIMD register like mmx.
therefore we lose original value of register.
so, we decide not to use memcpy.
```

second commit:
```
mcount: add mcount_arch_child_idx()

The value of child idx may be different for each architecture.
For this reason, we added the part that calculates child_idx
to be implemented by each architecture.
```

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -701,6 +701,11 @@ static void update_pltgot(struct mcount_thread_data *mtdp,
 	}
 }
 
+__weak unsigned long mcount_arch_child_idx(unsigned long child_idx)
+{
+	return child_idx;
+}
+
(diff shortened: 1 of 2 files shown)


ParkHanbum:
I can not believe it. I'm sososososososo stupid!
Sorry about that.

namhyung:
Do not blame yourself, everyone makes a mistake. :)

namhyung:
Sign off please!

ParkHanbum:
@namhyung OK. it is done!

PR #265, opened 2017-11-22
script: Add trace-memcpy.py script for memcpy info

honggyukim:
trace-memcpy.py example traces only memcpy() call and calculate how
many times memcpy is called and how much bytes are copied in total.

The usage and output looks as follows:

    $ uftrace script -S scripts/trace-memcpy.py --record /bin/ls scripts/
    count.py
    trace-memcpy.py
    replay.py
    simple.py
    9 times memcpy called
    183 bytes copied

The last two lines are printed by trace-memcpy.py

### DIFF
--- a/scripts/trace-memcpy.py
+++ b/scripts/trace-memcpy.py
@@ -1,0 +1,30 @@
+#
+# trace-memcpy.py
+#
+# uftrace-option: --nest-libcall -T memcpy@filter,arg3
+#
+#   void *memcpy(void *dest, const void *src, size_t n);
+#
+
+# Only "memcpy" calls this script and other functions never.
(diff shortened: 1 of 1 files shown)


honggyukim:
Hi @namhyung, it's a very simple example but can be a good example to show how `# uftrace-option:` and `UFTRACE_FUNC = [ ... ]` can be used.

honggyukim:
I think if there's a way to trace or run script only for entry(or vice versa). In this case, calling 'uftrace_end()' is just pure overhead doing nothing.

namhyung:
> I think it'd be better if there's a way to trace or run script only for entry(or vice versa).

I think it's easily doable by checking existence of the function when loading.

PR #266, opened 2017-11-26
utils: Adjust the logic removing old directory

Taeung:
When there is only a 'uftrace.data.old',
```
  $ ls
  hello  uftrace.data.old
```
if recording, innocent old directory can be eliminated.

```
  $ uftrace record hello
  $ ls
  hello  uftrace.data
```
So I change the logic removing old directory.

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -176,16 +176,16 @@ int create_directory(char *dirname)
 
 	xasprintf(&oldname, "%s.old", dirname);
 
-	if (!access(oldname, F_OK)) {
-		if (remove_directory(oldname) < 0) {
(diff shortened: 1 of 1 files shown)


Taeung:
Additional exaplaination:

We can think 4 cases:
 * Case 1) Nothing
 * Case 2) uftrace.data
 * Case 3) uftrace.data.old
 * Case 4) Both

If `uftrace.data` already exists, we need to rename the directory to `.old`.
However, I think we should remove existing old directory in the only `Case 4)`, not `Case 3)`

Taeung:
I got it ! I chaged PR 😄

PR #272, opened 2017-11-27
script: Fix to add global keyword to count variable

honggyukim:
The keyword "global" was missing for "count" variable in uftrace_end
function so this patch fixes it.


namhyung:
I think it's ok since the python interpreter seems to use the global variable for reading when there's no local with same name.  The global keyword is necessary when you write to it.  But I'll merge it for the consistency reason anyway.

https://docs.python.org/2/faq/programming.html#what-are-the-rules-for-local-and-global-variables-in-python

PR #273, opened 2017-11-28
replay: Adjust TID space for tid more than 6 digits

honggyukim:
Some systems are running for a long time and it creates tid more than
6 digits.  In this case, the alignment is a bit broken so this fixes the
problem by giving more spaces for TID field.

Before:
```
  $ uftrace -D 2 tests/t-abc
     1.627 us [142664] | __monstartup();
     0.913 us [142664] | __cxa_atexit();
              [142664] | main() {
     1.536 us [142664] |   a();
     2.670 us [142664] | } /* main */
```
After:
```
  $ uftrace -D 2 tests/t-abc
     1.627 us [142664] | __monstartup();
     0.913 us [142664] | __cxa_atexit();
              [142664] | main() {
     1.536 us [142664] |   a();
     2.670 us [142664] | } /* main */
```

### DIFF
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -59,7 +59,7 @@ static void print_duration(struct ftrace_task_handle *task,
 static void print_tid(struct ftrace_task_handle *task,
 		      struct fstack *fstack, void *arg)
 {
-	pr_out("[%5d]", task->tid);
+	pr_out("[%6d]", task->tid);
 }
 
 static void print_addr(struct ftrace_task_handle *task,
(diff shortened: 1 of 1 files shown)


honggyukim:
Thanks! I found that usual systems have maximum pid is 32768
```
$ cat /proc/sys/kernel/pid_max
32768
$ uname -r
4.10.0-35-generic
```
But some systems have higher number.
```
$ cat /proc/sys/kernel/pid_max 
196608
$ uname -r
4.4.0-97-generic
```
It's not related to the kernel version but may be different from the system configuration.

namhyung:
By default, it's propotional to the number of (possible) CPUs in the system only if it's larger than 32.  My system has 12 cpus.  But it can be changed by the /proc interface as you showed.

```
$ cat /sys/devices/system/cpu/possible
0-11
```

honggyukim:
Thanks for the confirmation!

PR #279, opened 2017-11-29
utils: Recursively remove directory

### LINKED ISSUE #186: utils: Old directory problem when recording
I found a trivial problem like below.
I don't know the exact case this problem occurs,
but I found this situation during many test case..

Even though `uftrace.data.benchSTL/` not exists,
the below uftrace-record not works because of the old directory `uftrace.data.benchSTL.old/`.
After the old directory manually removed, it works fine.

```
$ uftrace record -D 5 -t 1us -d uftrace.data.benchSTL ./benchmark_STL_containers
WARN: removing old directory failed: Is a directory

$ rm -rf uftrace.data.benchSTL.old/

$ uftrace record -D 5 -t 1us -d uftrace.data.benchSTL ./benchmark_STL_containers
```

Taeung:
If uftrace.data.old directory contains
subdirectories, we can see the warning message
when recording a target program:

  $ file uftrace.data.old/sub
  uftrace.data.old/sub: directory

  $ uftrace record hello
  WARN: removing old directory failed: Is a directory

So let remove_directory() able to recursively delete
subdirectories.

Fixes #186.

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -138,6 +138,7 @@ int remove_directory(char *dirname)
 {
 	DIR *dp;
 	struct dirent *ent;
+	struct stat statbuf;
 	char buf[PATH_MAX];
 	int saved_errno = 0;
 	int ret = 0;
@@ -154,9 +155,18 @@ int remove_directory(char *dirname)
 			continue;
 
 		snprintf(buf, sizeof(buf), "%s/%s", dirname, ent->d_name);
-		if (unlink(buf) < 0) {
+		ret = stat(buf, &statbuf);
+		if (ret < 0)
+			goto failed;
+
+		if (S_ISDIR(statbuf.st_mode))
+			ret = remove_directory(buf);
+		else
+			ret = unlink(buf);
+
+		if (ret < 0) {
+failed:
 			saved_errno = errno;
-			ret = -1;
 			break;
 		}
 	}


utils/utils.c
```diff
@@ -154,10 +155,14 @@ int remove_directory(char *dirname)
 			continue;
 
 		snprintf(buf, sizeof(buf), "%s/%s", dirname, ent->d_name);
-		if (unlink(buf) < 0) {
-			saved_errno = errno;
-			ret = -1;
-			break;
+		if (!stat(buf, &statbuf)) {
+			if (S_ISDIR(statbuf.st_mode))
+				ret = remove_directory(buf);
```
namhyung: What if `stat` or (recursive) `remove_directory` failed?

Taeung: Ah, I missed it.. what about the below change ?
(Because of duplication, I used `failed` )
```patch
diff --git a/utils/utils.c b/utils/utils.c
index cb08634..b51b31a 100644
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -156,14 +156,19 @@ int remove_directory(char *dirname)
 
                 snprintf(buf, sizeof(buf), "%s/%s", dirname, ent->d_name);
                 if (!stat(buf, &statbuf)) {
-                        if (S_ISDIR(statbuf.st_mode))
-                                ret = remove_directory(buf);
+                        if (S_ISDIR(statbuf.st_mode) &&
+                            remove_directory(buf) < 0)
+                                goto failed;
                         else if (unlink(buf) < 0) {
                                 saved_errno = errno;
-                                ret = -1;
-                                break;
+                                goto failed;
                         }
                 }
+                else {
+failed:
+                        ret = -1;
+                        break;
+                }
         }
 
         closedir(dp);

```

namhyung: Looks ok, but you also need to save errno for all error cases.

Taeung: @namhyung , I modified this PR 😄 
I'd appreciate it, if you check it.

Taeung: In addition, I found my mistake that is twice elimination tries for same directory.

In the case "it is directory && successfully remove it",
the statement `else if (unlink(buf) < 0)` can run as well.

For example, even though 'uftrace.data.old/test' was already removed
by remove_directory(), unlink() can try to remove it, again.

So we can see the warnning `No such file or directory` like below:
```
  $ mkdir -p uftrace.data.old/test

  $ uftrace record hello
  WARN: removing old directory failed: No such file or directory
```

So, I fixed it as well in new PR.

utils/utils.c
```diff
@@ -154,11 +155,19 @@ int remove_directory(char *dirname)
...
 			ret = -1;
 			break;
 		}
+
+		if (S_ISDIR(statbuf.st_mode)) {
+			if (remove_directory(buf) < 0)
+				goto failed;
+		}
+		else if (unlink(buf) < 0)
+			goto failed;
```
namhyung: I'd like to avoid backward jump (goto) if possible.  Could you please just save the result value and check it at the end of the loop body?

Taeung: Yep, I got it !

Taeung: I changed this PR again !
I'd appreciate it, if you check this 😄 

Additionally if you don't want nested if statement,
change it like below ?
```patch
diff --git a/utils/utils.c b/utils/utils.c
index 67b1903..254dcae 100644
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -138,6 +138,7 @@ int remove_directory(char *dirname)
 {
         DIR *dp;
         struct dirent *ent;
+        struct stat statbuf;
         char buf[PATH_MAX];
         int saved_errno = 0;
         int ret = 0;
@@ -154,9 +155,21 @@ int remove_directory(char *dirname)
                         continue;
 
                 snprintf(buf, sizeof(buf), "%s/%s", dirname, ent->d_name);
-                if (unlink(buf) < 0) {
-                        saved_errno = errno;
+                if (stat(buf, &statbuf)) {
                         ret = -1;
+                        goto failed;
+                }
+
+                if (S_ISDIR(statbuf.st_mode)) {
+                        if (remove_directory(buf) < 0)
+                                ret = -1;
+                }
+                else if (unlink(buf) < 0)
+                        ret = -1;
+
+                if (ret < 0) {
+failed:
+                        saved_errno = errno;
                         break;
                 }
         }

```

namhyung: What about this?

```c
   ret = stat();
   if (ret < 0)
      goto failed;

   if (S_ISDIR(...))
      ret = remove_directory();
   else
      ret = unlink();

   if (ret < 0) {
failed:
       saved_errno = errno;
       break;
   }
```

Taeung: That is better !
I changed this PR based on your code,
so I added your signed-off-by to commit log message. 😄

PR #290, opened 2017-12-14
graph: Use -f/--output-fields to customize its output

honggyukim:
The -f/--output-fields option is to specify fields in the replay output.
This patch implements the same feature for graph output.  Users can
customize the output for their needs like below:
```
  $ uftrace graph -f time,selftime,addr
  # Function Call Graph for 't-sort' (session: e07b5987dd5ee757)
  #   TIME     SELFTIME     ADDRESS     FUNCTION
    10.339 ms                  4004a0 : (1) t-sort
    10.339 ms 134.739 us       4004a0 : (1) main
    53.878 us   2.207 us       4005f1 :  +-(2) foo
    51.671 us  51.671 us       4005c7 :  | (6) loop
                                      :  |
    10.150 ms  57.077 us       400621 :  +-(1) bar
    10.093 ms  10.093 ms       400480 :    (1) usleep
```
The default value is 'time' only.  And it also accepts a special
field name of 'none' which disables the field display and shows function
graph output only.
```
  $ uftrace graph -f none
  # Function Call Graph for 't-sort' (session: e07b5987dd5ee757)
  (1) t-sort
  (1) main
   +-(2) foo
   | (6) loop
   |
   +-(1) bar
     (1) usleep
```
This patch does NOT change the default graph out at all.

### DIFF
--- a/cmd-graph.c
+++ b/cmd-graph.c
@@ -107,6 +107,7 @@ static void print_addr(struct field_data *fd)
 static struct display_field field_total_time= {
 	.id      = GRAPH_F_TOTAL_TIME,
 	.name    = "total-time",
+	.alias   = "total",
 	.header  = "TOTAL TIME",
 	.length  = 10,
 	.print   = print_total_time,
@@ -116,6 +117,7 @@ static struct display_field field_total_time= {
 static struct display_field field_self_time= {
 	.id      = GRAPH_F_SELF_TIME,
 	.name    = "self-time",
+	.alias   = "self",
 	.header  = " SELF TIME",
 	.length  = 10,
 	.print   = print_self_time,
@@ -124,7 +126,8 @@ static struct display_field field_self_time= {
 
 static struct display_field field_addr = {
 	.id      = GRAPH_F_ADDR,
-	.name    = "addr",
+	.name    = "address",
+	.alias   = "addr",
 #if __SIZEOF_LONG == 4
 	.header  = "  ADDR  ",
 	.length  = 8,
--- a/cmd-graph.c
+++ b/cmd-graph.c
@@ -154,11 +154,8 @@ static void print_field(struct graph_node *node)
 		.arg = (struct graph_node*)node,
 	};
 
-	if (!list_empty(&output_fields))
-		pr_out("  ");
-
-	if (print_field_data(&output_fields, &fd))
-		pr_out(": ");
+	if (print_field_data(&output_fields, &fd, 2))
+		pr_out(" : ");
 }
 
 static int create_graph(struct uftrace_session *sess, void *func)
@@ -598,8 +595,8 @@ static void print_graph_node(struct uftrace_graph *graph,
 
 		if (&child->list != node->head.prev) {
 			/* print blank line between siblings */
-			if (print_empty_field(&output_fields))
-				pr_out(": ");
+			if (print_empty_field(&output_fields, 2))
+				pr_out(" : ");
 			pr_indent(indent_mask, indent, false);
 			pr_out("\n");
 		}
@@ -632,7 +629,7 @@ static int print_graph(struct uftrace_graph *graph, struct opts *opts)
 
 	if (graph->root.time || graph->root.nr_edges) {
 		pr_out("========== FUNCTION CALL GRAPH ==========\n");
-		print_header(&output_fields, "# ");
+		print_header(&output_fields, "# ", 2);
 		indent_mask = xcalloc(opts->max_stack, sizeof(*indent_mask));
 		print_graph_node(graph, &graph->root, indent_mask, 0,
 				 graph->root.nr_edges > 1);
--- a/cmd-graph.c
+++ b/cmd-graph.c
@@ -10,7 +10,10 @@
 #include "utils/symbol.h"
 #include "utils/filter.h"
 #include "utils/fstack.h"
+#include "utils/field.h"
 
+
+static LIST_HEAD(output_fields);
 
 struct graph_backtrace {
 	struct list_head list;
@@ -70,6 +73,93 @@ struct task_graph {
 static bool full_graph = false;
 static struct rb_root tasks = RB_ROOT;
 static struct uftrace_graph *graph_list = NULL;
+
+static void print_total_time(struct field_data *fd)
+{
+	struct graph_node *node = (struct graph_node*)fd->arg;
+	uint64_t d;
+
+	d = node->time;
+
+	print_time_unit(d);
+}
+
+static void print_self_time(struct field_data *fd)
+{
+	struct graph_node *node = (struct graph_node*)fd->arg;
+	uint64_t d;
+
+	d = node->time - node->child_time;
+
+	print_time_unit(d);
+}
+
+static void print_addr(struct field_data *fd)
+{
+	struct graph_node *node = (struct graph_node*)fd->arg;
+
+	/* uftrace records (truncated) 48-bit addresses */
+	int width = sizeof(long) == 4 ? 8 : 12;
+
+	pr_out("%*lx", width, node->addr);
+}
+
+static struct display_field field_total_time= {
+	.id      = GRAPH_F_TOTAL_TIME,
+	.name    = "total-time",
+	.header  = "TOTAL TIME",
+	.length  = 10,
+	.print   = print_total_time,
+	.list    = LIST_HEAD_INIT(field_total_time.list),
+};
+
+static struct display_field field_self_time= {
+	.id      = GRAPH_F_SELF_TIME,
+	.name    = "self-time",
+	.header  = " SELF TIME",
+	.length  = 10,
+	.print   = print_self_time,
+	.list    = LIST_HEAD_INIT(field_self_time.list),
+};
+
+static struct display_field field_addr = {
+	.id      = GRAPH_F_ADDR,
+	.name    = "addr",
+#if __SIZEOF_LONG == 4
+	.header  = "  ADDR  ",
+	.length  = 8,
+#else
+	.header  = "   ADDRESS  ",
+	.length  = 12,
+#endif
+	.print   = print_addr,
+	.list    = LIST_HEAD_INIT(field_addr.list),
+};
+
+/* index of this table should be matched to display_field_id */
+static struct display_field *field_table[] = {
+	&field_total_time,
+	&field_self_time,
+	&field_addr,
+};
+
+static void setup_default_field(struct list_head *fields, struct opts *opts)
+{
+	add_field(fields, field_table[GRAPH_F_TOTAL_TIME]);
+}
+
+static void print_field(struct graph_node *node)
+{
+	struct field_data fd = {
+		.arg = (struct graph_node*)node,
+	};
+
+	if (!list_empty(&output_fields))
+		pr_out("  ");
+
+	if (print_field_data(&output_fields, &fd))
+		pr_out(": ");
+}
 
 static int create_graph(struct uftrace_session *sess, void *func)
 {
@@ -484,9 +574,7 @@ static void print_graph_node(struct uftrace_graph *graph,
 
 	symname = symbol_getname(sym, node->addr);
 
-	pr_out(" ");
-	print_time_unit(node->time);
-	pr_out(" : ");
+	print_field(node);
 	pr_indent(indent_mask, indent, needs_line);
 
 	if (full_graph && node == &graph->root)
@@ -510,7 +598,8 @@ static void print_graph_node(struct uftrace_graph *graph,
 
 		if (&child->list != node->head.prev) {
 			/* print blank line between siblings */
-			pr_out("%*s: ", 12, "");
+			if (print_empty_field(&output_fields))
+				pr_out(": ");
 			pr_indent(indent_mask, indent, false);
 			pr_out("\n");
 		}
@@ -539,8 +628,11 @@ static int print_graph(struct uftrace_graph *graph, struct opts *opts)
 		print_backtrace(graph);
 	}
 
+	setup_field(&output_fields, opts, &setup_default_field, field_table, ARRAY_SIZE(field_table));
+
 	if (graph->root.time || graph->root.nr_edges) {
 		pr_out("========== FUNCTION CALL GRAPH ==========\n");
+		print_header(&output_fields, "# ");
 		indent_mask = xcalloc(opts->max_stack, sizeof(*indent_mask));
 		print_graph_node(graph, &graph->root, indent_mask, 0,
 				 graph->root.nr_edges > 1);
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -162,8 +162,9 @@ static void print_field(struct ftrace_task_handle *task,
 		.fstack = fstack,
 		.arg = arg,
 	};
-	if (print_field_data(&output_fields, &fd))
-		pr_out("| ");
+
+	if (print_field_data(&output_fields, &fd, 1))
+		pr_out(" | ");
 }
 
 static void setup_default_field(struct list_head *fields, struct opts *opts)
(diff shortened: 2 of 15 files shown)


cmd-graph.c
```diff
@@ -71,6 +71,173 @@ static bool full_graph = false;
...
+			pr_out("\n");
+			exit(1);
+		}
+
+		p = strtok(NULL, ",");
+	}
+
+	free(str);
+}
+
```
namhyung: This code is a duplication, it'd be nice if you could find a way to share the code (if possible).

cmd-graph.c
```diff
@@ -539,8 +704,12 @@ static int print_graph(struct uftrace_graph *graph, struct opts *opts)
 		print_backtrace(graph);
 	}
 
+	setup_field(opts);
+
 	if (graph->root.time || graph->root.nr_edges) {
 		pr_out("========== FUNCTION CALL GRAPH ==========\n");
+		if (opts->fields != NULL && strcmp(opts->fields, "none"))
```
namhyung: I think it'd be better using `if (!list_empty(&output_fields))`.

honggyukim:
I've updated it as you mentioned. Thanks for suggesting good ideas!

tests/t097_dump_basic.py
```diff
@@ -14,7 +14,7 @@ def __init__(self):
 uftrace file header: endian        = 1 (little)
 uftrace file header: class         = 2 (64 bit)
 uftrace file header: features      = 0x263 (PLTHOOK | TASK_SESSION | SYM_REL_ADDR | MAX_STACK | AUTO_ARGS)
-uftrace file header: info          = 0xbff
+uftrace file header: info          = 0xbfd
```
Taeung: Hi @honggyukim 😄 

I think it'd be better to modify `fixup()` instead of expected output like below:
Because it is related to Build ID.

```patch
diff --git a/tests/t097_dump_basic.py b/tests/t097_dump_basic.py
index 595a251..1a975b9 100644
--- a/tests/t097_dump_basic.py
+++ b/tests/t097_dump_basic.py
@@ -52,5 +52,5 @@ reading 5231.dat
             result = result.replace("2 (64 bit)", "1 (32 bit)")
         p = sp.Popen(['file', 't-' + self.name], stdout=sp.PIPE)
         if 'BuildID' not in p.communicate()[0].decode(errors='ignore'):
-            result = result.replace("0x3ff", "0x3fd")
+            result = result.replace("0xbff", "0xbfd")
         return result
```
First of all, we can check the change history(`0x3ff` -> `0xbff`) on 5e0abb1
(This is related to added new info bit `RECORD_DATE`)

And the difference between `0xbff` and `0xbfd` depends on whether a build id exists like below:
`0xbfd` : 0000 0000 1011 1111 1101
`0xbff` :  0000 0000 1011 1111 1111
So I think it'd be better to change `fixup()`.
(Of cource, you can check the code about info bit on https://github.com/namhyung/uftrace/blob/master/cmd-info.c#L835 )

You can also see the change of `fixup()` due to this situation on e01ae7d

utils/display-field.h
```diff
@@ -0,0 +1,74 @@
...
+	enum display_field_id id;
+	const char *name;
+	const char *header;
+	int length;
+	bool used;
+	void (*print)(struct field_data *fd);
+};
+
+static int print_field_data(struct field_data *fd)
+{
```
Taeung: Hum.. I thought it seems good that a header file contains data structure, variables,
function headers, inline function definitions and etc.

But this header file includes function definitions.
Is it OK on coding convention of uftrace ?

cmd-graph.c
Taeung: I agree that, otherwise the output can be inconsistent like below:
```patch
$ diff <(uftrace graph) <(uftrace graph -f time)
--- /dev/fd/63	2017-12-16 06:53:20.246154501 +0900
+++ /dev/fd/62	2017-12-16 06:53:20.238154395 +0900
@@ -1,5 +1,6 @@
 # Function Call Graph for 't-abc' (session: 368d2cdf6c16a307)
+#   TIME      FUNCTION
    2.882 us : (1) t-abc
    0.685 us :  +-(1) __monstartup
             :  |
```

cmd-graph.c
```diff
@@ -71,6 +72,103 @@ static bool full_graph = false;
...
+
+	/* uftrace records (truncated) 48-bit addresses */
+	int width = sizeof(long) == 4 ? 8 : 12;
+
+	pr_out("%*lx", width, node->addr);
+}
+
+static struct display_field graph_field_total_time= {
+	.id      = GRAPH_F_TOTAL_TIME,
+	.name    = "time",
```
Taeung: I thought the field name "time" can be confused graph total time and replay timestamp
https://github.com/namhyung/uftrace/blob/master/cmd-replay.c#L139-L140

What do you think about it ?

tests/t097_dump_basic.py
honggyukim: Thanks. I've done it in bd25cb1.

utils/display-field.h
honggyukim: I moved it into a new file `utils/field.c`. Thanks!

cmd-graph.c
honggyukim: I thought about changing it `totaltime` but it's too verbose. And the field name is also too long to put in. I think using `time` and `selftime` is enough.

honggyukim:
Thanks @Taeung for your comments. I've updated it again.

cmd-graph.c
honggyukim: I've done it in `utils/field.[ch]`. Thanks!

cmd-graph.c
namhyung: And I prefer having `-` in between.

namhyung: What about each field gets separated by 2 spaces like `uftrace report`?  The time fields already uses a space between time and unit so single space looks not enough IMO.

cmd-replay.c
```diff
@@ -166,64 +166,16 @@ static void print_field(struct ftrace_task_handle *task,
 		pr_out("| ");
 }
 
-static void setup_field(struct opts *opts)
+static void setup_default_field(struct opts *opts)
```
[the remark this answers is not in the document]
honggyukim: I passed it as a function pointer because of the range handle routine as below:
```
static void setup_default_field(struct opts *opts)
{
        if (opts->range.start > 0 || opts->range.stop > 0) {
                if (opts->range.start_elapsed || opts->range.stop_elapsed)
                        add_field(&output_fields, field_table[REPLAY_F_ELAPSED]);
                else
                        add_field(&output_fields, field_table[REPLAY_F_TIMESTAMP]);
        }
        add_field(&output_fields, field_table[REPLAY_F_DURATION]);
        add_field(&output_fields, field_table[REPLAY_F_TID]);
}
```
Please let me know if there is a better way to handle this.

utils/field.h
```diff
@@ -11,7 +11,7 @@ struct field_data {
 	void *arg;
 };
 
-enum replay_field_id {
+enum display_field_id {
 	REPLAY_F_NONE           = -1,
```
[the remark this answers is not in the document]
honggyukim: Hmm.. I added `REPLAY_F_` enum and `GRAPH_F_` enum in here so not sure if it's better to change this.

cmd-replay.c
```diff
@@ -193,38 +193,44 @@ struct replay_field *field_table[] = {
...
 	struct replay_field *field;
-	struct field_data fd = {
-		.task = task,
-		.fstack = fstack,
-		.arg = arg,
-	};
 
 	if (list_empty(&output_fields))
-		return;
+		return 1;
```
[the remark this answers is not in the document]
honggyukim: I've changed it. Thanks!

cmd-replay.c
```diff
@@ -173,13 +173,14 @@ static void print_header(void)
...
+	if (list_empty(&output_fields))
+		return;
+
 	pr_out("#");
 	list_for_each_entry(field, &output_fields, list)
 		pr_out("%s ", field->header);
 
-	if (!list_empty(&output_fields))
-		pr_out(" ");
-
```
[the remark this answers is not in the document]
honggyukim: Right. I missed it. But the `output_fields` is not empty here, so I've just gave one more space before `FUNCTION` to be printed. Thanks!

honggyukim:
@namhyung Strangely, I cannot write on your comment, so copied it here.
```diff
+
+static struct display_field graph_field_total_time= {
+	.id      = GRAPH_F_TOTAL_TIME,
+	.name    = "time",
```
> But I also agree with Taeung. What about add alias support for display fields? So that total-time (or even simply total) can be treated same as time. Likewise, self-time can be just self.

Hmm.. If so, what about the below candidates:
1. using `total` and `self`
```
$ uftrace graph -f total,self
# Function Call Graph for 't-sort' (session: b007f4b7cf792878)
# TOTAL TIME  SELF TIME   FUNCTION
   10.145 ms            : (1) t-sort
   10.145 ms  39.890 us : (1) main
   16.773 us   0.734 us :  +-(2) foo
   16.039 us  16.039 us :  | (6) loop
                        :  |
   10.088 ms  14.740 us :  +-(1) bar
   10.073 ms  10.073 ms :    (1) usleep
```
Maybe we can also provide aliases as you mentioned. `total-time` for `total` and `self-time` for `self`.

utils/field.h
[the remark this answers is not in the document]
namhyung: OK then.

cmd-replay.c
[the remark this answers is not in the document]
namhyung: I think the below looks a bit better:

```
static void setup_default_field(struct list_head *head, struct opts *opts)
{
   ...
}
```

cmd-graph.c
```diff
@@ -127,11 +128,11 @@ static struct display_field field_addr = {
...
 #if __SIZEOF_LONG == 4
-	.header  = "  ADDR  ",
-	.length  = 8,
+	.header  = "   ADDR  ",
+	.length  = 9,
 #else
-	.header  = "   ADDRESS  ",
-	.length  = 12,
+	.header  = "    ADDRESS  ",
+	.length  = 13,
```
namhyung: I think you'd better doing it in the `print_field()`.

cmd-replay.c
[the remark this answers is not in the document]
honggyukim: Ok. Done it.

namhyung:
I pushed my modification to [check/graph-field](https://github.com/honggyukim/uftrace/compare/check/graph-output-fields...namhyung:check/graph-field) branch.  Please check it out.

honggyukim:
@namhyung Thanks for doing this work! I think it looks very good but I changed the man page a bit. But not entirely wrote about aliases. I just wrote `total` and `self`. I pushed it again after changing it.

cmd-graph.c
```diff
@@ -107,6 +107,7 @@ static void print_addr(struct field_data *fd)
 static struct display_field field_total_time= {
 	.id      = GRAPH_F_TOTAL_TIME,
 	.name    = "total-time",
+	.alias   = { "total", "time", NULL },
```
namhyung: Hmm.. this syntax seems to be a GCC extension thus not portable.  I couldn't find any other way initializing a flexible array member though.

honggyukim: What about using only `total` instead of providing multiple aliases?

namhyung: OK, that would make it a lot easier. :)

honggyukim: If you agree, I will modify and update this PR again :)

namhyung: It'd be nice if you could update it.

honggyukim: Updated it. Thanks!

honggyukim:
Hmm.. This changes the default replay output so needs to be changed a bit more.
```
$ uftrace replay
# DURATION     TID    FUNCTION
            [ 20547]| main() {
            [ 20547]|   foo() {
  50.967 us [ 20547]|     loop();
  57.817 us [ 20547]|     loop();
  51.203 us [ 20547]|     loop();
 162.171 us [ 20547]|   } /* foo */
            [ 20547]|   foo() {
  51.223 us [ 20547]|     loop();
  57.617 us [ 20547]|     loop();
  60.487 us [ 20547]|     loop();
 170.082 us [ 20547]|   } /* foo */
            [ 20547]|   bar() {
  10.062 ms [ 20547]|     usleep();
  10.335 ms [ 20547]|   } /* bar */
  11.468 ms [ 20547]| } /* main */
```

cmd-graph.c
```diff
@@ -125,6 +127,7 @@ static struct display_field field_self_time= {
 static struct display_field field_addr = {
 	.id      = GRAPH_F_ADDR,
 	.name    = "addr",
+	.alias   = "address",
```
namhyung: Minor nitpick.. could you please swap the name and alias of 'addr' for consistency?

honggyukim: Sure. I was thinking about it, but thought about the man pages as well. I think we can just simply write aliases only on the man pages for simplicity.

namhyung: ok

namhyung:
How about [the below](https://github.com/namhyung/uftrace/files/1576151/fix-graph-replay-space.patch.txt)?  It could be squashed into 336c92a.

```diff
diff --git a/cmd-replay.c b/cmd-replay.c
index b7ac7a9..00ab2ca 100644
--- a/cmd-replay.c
+++ b/cmd-replay.c
@@ -164,7 +164,7 @@ static void print_field(struct ftrace_task_handle *task,
        };
 
        if (print_field_data(&output_fields, &fd, 1))
-               pr_out("| ");
+               pr_out(" | ");
 }
 
 static void setup_default_field(struct list_head *fields, struct opts *opts)
@@ -217,7 +217,7 @@ static void print_backtrace(struct ftrace_task_handle *task)
                        pr_out(" ");
                }
                if (!list_empty(&output_fields))
-                       pr_out("| ");
+                       pr_out(" | ");
 
                name = symbol_getname(sym, fstack->addr);
                pr_gray("/* [%2d] %s */\n", i, name);
@@ -317,7 +317,7 @@ static void print_task_newline(int current_tid)
 {
        if (prev_tid != -1 && current_tid != prev_tid) {
                if (print_empty_field(&output_fields, 1))
-                       pr_out("| ");
+                       pr_out(" | ");
                pr_out("\n");
        }
 
@@ -785,7 +785,7 @@ out:
 static void print_warning(struct ftrace_task_handle *task)
 {
        if (print_empty_field(&output_fields, 1))
-               pr_out("| ");
+               pr_out(" | ");
        pr_red(" %*s/* inverted time: broken data? */\n",
               (task->display_depth + 1) * 2, "");
 }
diff --git a/utils/field.c b/utils/field.c
index 5162961..164d836 100644
--- a/utils/field.c
+++ b/utils/field.c
@@ -6,18 +6,19 @@ void print_header(struct list_head *output_fields, const char *prefix,
                  int space)
 {
        struct display_field *field;
+       bool first = true;
 
        /* do not print anything if not needed */
        if (list_empty(output_fields))
                return;
 
-       pr_out("%s", prefix);
        list_for_each_entry(field, output_fields, list) {
+               pr_out("%*s", space, first ? prefix : "");
                pr_out("%s", field->header);
-               pr_out("%*s", space, "");
+               first = false;
        }
 
-       pr_out(" FUNCTION\n");
+       pr_out("   FUNCTION\n");
 }
 
 int print_field_data(struct list_head *output_fields, struct field_data *fd,
```

honggyukim:
> How about the below? It could be squashed into 336c92a.

Thanks! I applied your change except for the backtrace hunk below.
```
@@ -217,7 +217,7 @@ static void print_backtrace(struct ftrace_task_handle *task)
                        pr_out(" ");
                }
                if (!list_empty(&output_fields))
-                       pr_out("| ");
+                       pr_out(" | ");
 
                name = symbol_getname(sym, fstack->addr);
                pr_gray("/* [%2d] %s */\n", i, name);
```
It seems that there's no more problems now.

PR #292, opened 2017-12-17
Support old gcc version

ParkHanbum:
some code of supporting i386 have been not supported from the old version gcc.
this PR is the patch to resolve that.

### DIFF
--- a/arch/i386/common.S
+++ b/arch/i386/common.S
@@ -1,0 +1,6 @@
+#include "utils/asm.h"
+
+ENTRY(get_pc_thunk)
+	movl 0(%esp), %eax
+	ret
+END(get_pc_thunk)
--- a/arch/i386/mcount-support.c
+++ b/arch/i386/mcount-support.c
@@ -208,13 +208,15 @@ unsigned long *mcount_arch_parent_location(struct symtabs *symtabs,
 
 		// Assuming that this happens only in main..			
 		bool found_main_ret = false;
+		int stack_index = 0;
+
 		if (!(strcmp(find_main[0], parent_name) || 
 		      strcmp(find_main[1], child_name))) {
 			ret_addr = *parent_loc;
-			for (int i = 1; i < MAX_SEARCH_STACK; i++ ) {
-				search_ret_addr = *(unsigned long *)(parent_loc + i);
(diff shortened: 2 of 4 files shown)


arch/i386/mcount-support.c
```diff
@@ -211,7 +211,8 @@ unsigned long *mcount_arch_parent_location(struct symtabs *symtabs,
 		if (!(strcmp(find_main[0], parent_name) || 
 		      strcmp(find_main[1], child_name))) {
 			ret_addr = *parent_loc;
-			for (int i = 1; i < MAX_SEARCH_STACK; i++ ) {
+			int i;
+			for (i = 1; i < MAX_SEARCH_STACK; i++) {
```
namhyung: I prefer having declarations before statements - I know it's an old style but...

arch/i386/common.S
```diff
@@ -0,0 +1,7 @@
+#include "utils/asm.h"
+
+ENTRY(__x86.get_pc_thunk.ax)
+	movl 0(%esp), %eax
+	ret
+END(__x86.get_pc_thunk.ax)
```
namhyung: Is it ok to have it in uftrace if compiler supports it?

arch/i386/mcount-support.c
```diff
@@ -208,13 +208,14 @@ unsigned long *mcount_arch_parent_location(struct symtabs *symtabs,
 
 		// Assuming that this happens only in main..			
 		bool found_main_ret = false;
+		int stack_index = 0;
```
namhyung: Please keep a blank line between declarations and statements.

arch/i386/common.S
```diff
@@ -0,0 +1,6 @@
+#include "utils/asm.h"
+
+GLOBAL(get_pc_thunk)
```
namhyung: It should be `ENTRY` instead of `GLOBAL`.  The difference is the visibility of the symbol - a `GLOBAL` symbol will be visible from outside of libmcount while `ENTRY` symbol will be visible only in libmcount.

ParkHanbum: @namhyung ok, I see!

PR #299, opened 2017-12-26
misc: Add mmap64() to auto args list

honggyukim:
Some programs use mmap64() instead of mmap(), so it'd be better to add
mmap64() to auto args list.


honggyukim:
I added this because qml engine uses it.

utils/auto-args.h
```diff
@@ -24,6 +24,7 @@ static char *auto_args_list =
 	"calloc@arg1/u,arg2/u;"
 	"realloc@arg1/x,arg2/u;"
 	"mmap@arg1/x,arg2/u,arg3/e:uft_mmap_prot,arg4/e:uft_mmap_flag,arg5,arg6;"
+	"mmap64@arg1/x,arg2/u,arg3/e:uft_mmap_prot,arg4/e:uft_mmap_flag,arg5,arg6;"
```
namhyung: I think the last arg should be `arg6/d64` (for 32-bit systems).

honggyukim: Hmm.. do we have the format `d64`. It's not written in the man page.
```
ARGUMENTS
       The uftrace tool supports recording function arguments and/or return values using the -A/--argument and -R/--retval options respectively.  The syntax is very similar to  that
       of triggers:

              <argument>    :=  <symbol> "@" <specs>
              <specs>       :=  <spec>  | <spec> "," <spec>
              <spec>        :=  ( <int_spec> | <float_spec> | <ret_spec> )
              <int_spec>    :=  "arg" N [ "/" <format> [ <size> ] ] [ "%" ( <reg> | <stack> ) ]
              <float_spec>  :=  "fparg" N [ "/" ( <size> | "80" ) ] [ "%" ( <reg> | <stack> ) ]
              <ret_spec>    :=  "retval" [ "/" <format> [ <size> ] ]
              <format>      :=  "i" | "u" | "x" | "s" | "c" | "f" | "S" | "p"
              <size>        :=  "8" | "16" | "32" | "64"
              <reg>         :=  <arch-specific register name>  # "rdi", "xmm0", "r0", ...
              <stack>       :=  "stack" [ "+" ] <offset>
```

honggyukim: I updated it using `d64` anyway. Thanks!

namhyung: The commit 2352d68 added the 'd' specifier but missed to update the manpage.

honggyukim: I just found it. And here is one more, it seems that `enum` type is not handled in `misc/gen-autoargs.py`. Do you think we should add the below change?
```diff
diff --git a/misc/gen-autoargs.py b/misc/gen-autoargs.py
index e49892b..da9d6ff 100755
--- a/misc/gen-autoargs.py
+++ b/misc/gen-autoargs.py
@@ -184,6 +184,8 @@ def make_uftrace_retval_format(ctype, funcname):
         retval_format += "retval/p"
     elif ctype == "off64_t":
         retval_format += "retval/d64"
+    elif ctype.startswith('enum'):
+        retval_format += "retval/e:%s" % ctype[5:]
     else:
         retval_format += "retval"

```

namhyung: Yes I do

honggyukim: done it!

honggyukim:
@namhyung Please take a look at it again. Thanks!

PR #305, opened 2018-01-02
fstack: Fix the segfault when using --tid

### LINKED ISSUE #304: Segfault when using --tid
uftrace replay crashes when using `--tid` option. Here is the steps to reproduce.
```
$ uftrace record tests/t-fork
$ uftrace replay --tid [one of recorded tid]
```
The crash point and its backtrace are as follows:
```
1694│         else {  /* must be perf event */
1695│                 struct uftrace_perf_reader *perf;
1696│
1697│                 assert(handle->last_perf_idx >= 0);
1698│                 perf = &handle->perf[handle->last_perf_idx];
1699│
1700│                 if (rstack->addr == EVENT_ID_PERF_COMM) {
1701├───────────────────────> memcpy(task->t->comm, perf->u.comm.comm,
1702│                                sizeof(task->t->comm));
1703│                 }
1704│
1705│                 perf->valid = false;
1706│                 handle->last_perf_idx = -1;
1707│         }

(gdb) p task->t
$1 = (struct uftrace_task *) 0x0

(gdb) bt
#0  __fstack_consume (task=task@entry=0x5555557c6500, kernel=kernel@entry=0x0, cpu=cpu@entry=-1) at /home/honggyu/src/uftrace/git/uftrace/utils/fstack.c:1701
#1  0x00005555555827b8 in __read_rstack (handle=handle@entry=0x7fffffffd0a0, taskp=taskp@entry=0x7fffffffd048, consume=consume@entry=true) at /home/honggyu/src/uftrace/git/uftrace/
utils/fstack.c:1817
#2  0x0000555555582d7a in read_rstack (handle=handle@entry=0x7fffffffd0a0, task=task@entry=0x7fffffffd048) at /home/honggyu/src/uftrace/git/uftrace/utils/fstack.c:1840
#3  0x000055555556cef5 in command_replay (argc=<optimized out>, argv=<optimized out>, opts=<optimized out>) at /home/honggyu/src/uftrace/git/uftrace/cmd-replay.c:928
#4  0x000055555555bda5 in main (argc=5, argv=0x7fffffffde48) at /home/honggyu/src/uftrace/git/uftrace/uftrace.c:996

```

Taeung:
Currently the task that is chosen by --tid option
task->t become NULL on setup_task_filter().
Hence this crash can occur.
So properly set task->t regardless --tid option.

Fixes #304.

### DIFF
--- a/utils/fstack.c
+++ b/utils/fstack.c
@@ -189,6 +189,7 @@ setup:
 			task->done = true;
 			task->tid  = tid;
 			task->h    = handle;
+			task->t = find_task(&handle->sessions, tid);
 
 			/* need to read the data to check elapsed time */
 			xasprintf(&filename, "%s/%d.dat", handle->dirname, tid);


honggyukim:
@Taeung Thanks for fixing the bug!

PR #309, opened 2018-01-03
misc: Add a few more functions to auto args

honggyukim:
This PR adds a few more functions to auto args list.


namhyung:
Hmm.. I guess these `_IO_` variants might be used for other functions too?  Also I think it'd be better updating necessary functions at once after you collect them for a while.

honggyukim:
Okay. Then I will not add `_IO_` variants at this time but I need others when analyzing Qt program at the moment. If you're okay with it l will update it again.

misc/prototypes.h
```diff
@@ -218,6 +218,8 @@ enum uft_dlopen_flag {
 	RTLD_NODELETE = 0x1000,
 };
 void *dlopen(const char *filename, enum uft_dlopen_flag flags);
+void *dlsym(void *handle, const char *symbol);
+void *dlvsym(void *handle, char *symbol, char *version);
```
namhyung: Wouldn't it be better adding `dlclose()` and `dlerror()` as well?

misc/prototypes.h
```diff
@@ -44,6 +44,26 @@ void *mmap64(void *addr, size_t length, enum uft_mmap_prot prot, enum uft_mmap_f
...
+    MADV_MERGEABLE   = 12,
+    MADV_UNMERGEABLE = 13,
+    MADV_HUGEPAGE    = 14,
+    MADV_NOHUGEPAGE  = 15,
+    MADV_DONTDUMP    = 16,
+    MADV_DODUMP      = 17,
+    MADV_HWPOISON    = 100,
+};
+int madvise(void *addr, size_t length, enum uft_madvise_advice advice);
+
```
namhyung: There are `posix_madvise` and `posix_fadvise` too.

honggyukim:
Hi @namhyung, I've updated as you mentioned. Please take a look at it!

misc/prototypes.h
```diff
@@ -44,6 +44,45 @@ void *mmap64(void *addr, size_t length, enum uft_mmap_prot prot, enum uft_mmap_f
...
+enum uft_posix_madvise_advice {
```
namhyung: Sorry for bothering you, but the name seems too long.  What about `uft_posix_madvise` or `uft_posix_madvise_adv`?

honggyukim: Ok. I will make it as `uft_posix_madvise_adv` then.

honggyukim: Just updated it. Thanks!

misc/prototypes.h
```diff
@@ -44,6 +44,45 @@ void *mmap64(void *addr, size_t length, enum uft_mmap_prot prot, enum uft_mmap_f
 int munmap(void *addr, size_t length);
 int mprotect(void *addr, size_t len, enum uft_mmap_prot prot);
 
+enum uft_madvise_advice {
```
namhyung: This also can be just `uft_madvise`.

honggyukim: Sure. Also updated it!

PR #313, opened 2018-01-11
Rename 'ftrace' residues -> 'uftrace'

Taeung:
This is trivial patches:
 - Rename 'ftrace' residues on some comments to 'uftrace'
 - Rename TestBase.ftrace to TestBase.uftrace_cmd to be more readable

### DIFF
--- a/tests/arch/arm/Makefile
+++ b/tests/arch/arm/Makefile
@@ -1,6 +1,6 @@
 CC = gcc
 RM = rm -f
-FTRACE = ../../../ftrace -L ../../.. --flat
+UFTRACE = ../../../uftrace -L ../../.. --flat
 
 #CFLAGS = -g -finstrument-functions
 CFLAGS = -g -pg -fno-omit-frame-pointer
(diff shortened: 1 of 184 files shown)


Taeung:
In addition, we need to modify the below two line (Makefile:45, 54) ?
I thought ftrace.data and gmon.out aren't needed anymore.
```
  $ grep ftrace
tests/arch/x86_64/Makefile
45:	$(RM) $(ALL_TARGETS) *.o ftrace.data* gmon.out *-diff

tests/arch/arm/Makefile
54:	$(RM) $(ALL_TARGETS) *.o ftrace.data* gmon.out *-diff
```
And need to modify the below lines because of old 'ftrace' ?
https://github.com/namhyung/uftrace/blob/master/tests/arch/arm/Makefile#L3
https://github.com/namhyung/uftrace/blob/master/tests/arch/x86_64/Makefile#L3

namhyung:
> I thought ftrace.data and gmon.out aren't needed anymore.

Yep, it should be change to `uftrace.data`, but please keep the `gmon.out` as it can be made during some testing when built with `-pg`.

> And need to modify the below lines because of old 'ftrace' ?

Yes please.

Taeung:
Yep, I got it 😄

Taeung:
I changed this PR !

PR #316, opened 2018-01-21
misc: Add qsort and bsearch prototypes to auto args

honggyukim:
This adds qsort and bsearch function prototypes to auto args list.
The function pointers for comparison can be easily identified with it.


honggyukim:
@namhyung I also think that it's better to collect library functions and upload them at once, but I just sometimes get ideas from time to time. If you don't like the way, I will just hold this and wait more. Thanks!

Issue #319, opened 2018-02-07
? option doesn't work in zsh 


qpakzk:
Hello,
I used zsh in Ubuntu 16.04.3 LTS. 
When I use ```-?``` option on the zsh, the result is below.

<img width="348" alt="zsh opt" src="https://user-images.githubusercontent.com/20490285/35921721-d2561fae-0c5e-11e8-8977-77c6a3fcdcd6.png">

In other words, zsh doesn't catch ```-?``` option.

namhyung:
Hmm.. as far as I know it came from argparse in glibc, not from shell.  Does it work on bash?  Does zsh treat it specially?

qpakzk:
Yes. When I use 'uftrace -?' on bash, it works well. Below is the result.

<img width="702" alt="screenshot 2018-02-08 13 46 53" src="https://user-images.githubusercontent.com/20490285/35956096-e32e63ae-0cd6-11e8-98c4-5348037ed091.png">

chneukirchen:
Set `NONOMATCH` or quote the flag: `-\?`.

This is not a uftrace problem.

qpakzk:
It works. Thank you!
<img width="688" alt="screenshot 2018-02-09 04 06 57" src="https://user-images.githubusercontent.com/20490285/35992810-b33edaca-0d4e-11e8-888e-f2ec5ce01c3d.png">

namhyung:
@chneukirchen thanks for pointing that out.

In fact, I'd like to have a short option of the help message as `-h` instead (or both) of `-?`.

qpakzk:
How about replacing ```-h``` with ```-?``` in the code when entering ```-h```?

namhyung:
@qpakzk what about adding explicit `-h` option for help (although duplicated)?

namhyung:
@qpakzk anyway I'd like to close this issue.  If you want to work on the help refactoring, please open another issue for that, thanks!

PR #321, opened 2018-02-20
uftrace: Add -h option

qpakzk:
I solved issue #319 .

The -h option is to print help message. It's the short version of --help
option.

Although -? option has a same role as -h option, the -h option is easier
to assume --help option than the -? option.

So it is useful to add the -h option.

Signed-off-by: Sangwon Hong <qpakzk@gmail.com>

### DIFF
--- a/doc/uftrace.md
+++ b/doc/uftrace.md
@@ -56,6 +56,9 @@ OPTIONS
 -?, \--help
 :   Print help message and list of options with description
 
+-h, \--help
+:   Print help message and list of options with description
+
 \--usage
 :   Print usage string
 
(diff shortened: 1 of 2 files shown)


namhyung:
By the way, is it possible to customize help output using help-filters?  I mean that it'd be nice to print command-related options only.  For example `uftrace record -h` should show options used by the record command only.

qpakzk:
I see. I'll try to do it.

honggyukim:
By the way, why don't we remove the duplicated part as follows?
```diff
diff --git a/doc/uftrace.md b/doc/uftrace.md
index e35e010..7dde7be 100644
--- a/doc/uftrace.md
+++ b/doc/uftrace.md
@@ -53,10 +53,7 @@ script

 OPTIONS
 =======
--?, \--help
-:   Print help message and list of options with description
-
--h, \--help
+-?, -h, \--help
 :   Print help message and list of options with description

 \--usage
```

PR #332, opened 2018-03-27
uftrace: Fix wrong old uftrace data name

Taeung:
Hi @namhyung,

Default old uftrace data name should be `uftrace.data.old`,
not `ftrace.dir`. I found this problem that can find uftrace
old data due to wrong old name below case:
( it is related to https://github.com/namhyung/uftrace/blob/master/utils/data-file.c#L383-L384)

Before:
```
  $ ls
  uftrace.data.old

  $ uftrace replay
  WARN: cannot open record data: uftrace.data: No such file or directory
```
So rename it.

After:
```
  $ ls
  uftrace.data.old

  $ uftrace replay
     1.221 us [ 22468] | __monstartup();
     0.985 us [ 22468] | __cxa_atexit();
              [ 22468] | main() {
              [ 22468] |   printf() {
...
```

### DIFF
--- a/uftrace.h
+++ b/uftrace.h
@@ -18,7 +18,7 @@
 #define UFTRACE_FILE_VERSION  4
 #define UFTRACE_FILE_VERSION_MIN  3
 #define UFTRACE_DIR_NAME     "uftrace.data"
-#define UFTRACE_DIR_OLD_NAME  "ftrace.dir"
+#define UFTRACE_DIR_OLD_NAME "uftrace.data.old"
 
 #define UFTRACE_RECV_PORT  8090
 


Taeung:
Sorry, I confused it with `uftrace.data.old`.. Close this PR ?

namhyung:
Hmm.. I think it's not needed to try the .old file since it's created only when the default file is being created.  Do you think it's needed?

Taeung:
I rethought about trying the .old file as you said. But I think you're right and
it is right that uftrace works with only `uftrace.data` or `ftrace.dir` , not .old file.

It seems that even if there is only `uftrace.data.old`, the user can rename it `uftrace.data` by oneself.

PR #338, opened 2018-04-05
 record: Use {get,put}_libmcount_path()

Taeung:
If we use `{get,put}_libmcount_path()` functions,

First, we can check both the library path the user gave with `-L`
and second path appending "/libmcount"(for test).

Second, we can previously check whether libmcount*.so files exist or no,
to avoid errors that can occur when recording.

How about this patches ?

### DIFF
--- a/cmd-live.c
+++ b/cmd-live.c
@@ -77,7 +77,7 @@ static bool can_skip_replay(struct opts *opts, int record_result)
 static void setup_child_environ(struct opts *opts)
 {
 	char buf[4096];
-	char *old_preload, *old_libpath;
+	char *old_preload, *old_libpath, *libpath;
 
 	if (opts->lib_path) {
 		strcpy(buf, opts->lib_path);
@@ -103,22 +103,19 @@ static void setup_child_environ(struct opts *opts)
 	else
 		setenv("LD_LIBRARY_PATH", buf, 1);
 
-	if (opts->lib_path)
-		snprintf(buf, sizeof(buf), "%s/libmcount/libmcount.so", opts->lib_path);
-	else
-		strcpy(buf, "libmcount.so");
+	libpath = get_libmcount_path(opts);
 
 	old_preload = getenv("LD_PRELOAD");
 	if (old_preload) {
-		size_t len = strlen(buf) + strlen(old_preload) + 2;
+		size_t len = strlen(libpath) + strlen(old_preload) + 2;
 		char *preload = xmalloc(len);
 
-		snprintf(preload, len, "%s:%s", buf, old_preload);
+		snprintf(preload, len, "%s:%s", libpath, old_preload);
 		setenv("LD_PRELOAD", preload, 1);
 		free(preload);
 	}
 	else
-		setenv("LD_PRELOAD", buf, 1);
+		setenv("LD_PRELOAD", libpath, 1);
 }
 
 int command_live(int argc, char *argv[], struct opts *opts)
--- a/cmd-live.c
+++ b/cmd-live.c
@@ -104,6 +104,8 @@ static void setup_child_environ(struct opts *opts)
 		setenv("LD_LIBRARY_PATH", buf, 1);
 
 	libpath = get_libmcount_path(opts);
+	if (libpath == NULL)
+		pr_err_ns("cannot found libmcount.so\n");
 
 	old_preload = getenv("LD_PRELOAD");
 	if (old_preload) {
--- a/cmd-live.c
+++ b/cmd-live.c
@@ -76,32 +76,15 @@ static bool can_skip_replay(struct opts *opts, int record_result)
 
 static void setup_child_environ(struct opts *opts)
 {
-	char buf[4096];
-	char *old_preload, *old_libpath, *libpath;
-
-	if (opts->lib_path) {
-		strcpy(buf, opts->lib_path);
-		strcat(buf, "/libmcount:");
-	} else {
-		/* to make strcat() work */
-		buf[0] = '\0';
-	}
+	char *old_preload, *libpath;
 
 #ifdef INSTALL_LIB_PATH
-	strcat(buf, INSTALL_LIB_PATH);
-#endif
-
-	old_libpath = getenv("LD_LIBRARY_PATH");
-	if (old_libpath) {
-		size_t len = strlen(buf) + strlen(old_libpath) + 2;
-		char *libpath = xmalloc(len);
-
-		snprintf(libpath, len, "%s:%s", buf, old_libpath);
+	if (!opts->lib_path) {
+		libpath = strjoin(getenv("LD_LIBRARY_PATH"), INSTALL_LIB_PATH, ":");
 		setenv("LD_LIBRARY_PATH", libpath, 1);
 		free(libpath);
 	}
-	else
-		setenv("LD_LIBRARY_PATH", buf, 1);
+#endif
 
 	libpath = get_libmcount_path(opts);
 	if (libpath == NULL)
--- a/cmd-record.c
+++ b/cmd-record.c
@@ -111,16 +111,26 @@ char * get_libmcount_path(struct opts *opts)
 	}
 
 	if (opts->lib_path) {
-		snprintf(lib, 4096, "%s/libmcount/", opts->lib_path);
+		snprintf(lib, 4096, "%s/libmcount/%s", opts->lib_path, libmcount);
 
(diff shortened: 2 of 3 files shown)


uftrace.h
```diff
@@ -30,6 +30,8 @@
 #define KB 1024
 #define MB (KB * 1024)
 
+#define BUF_SIZE 4096
```
namhyung: In many cases (usually dealing with a file path) it's more natural to use `PATH_MAX` instead IMO.

cmd-info.c
```diff
@@ -802,7 +802,7 @@ static int read_pattern_type(void *arg)
 {
 	struct ftrace_file_handle *handle = arg;
 	struct uftrace_info *info = &handle->info;
-	char buf[4096];
+	char buf[BUF_SIZE];
```
namhyung: I think it's better to pass a buffer instead of having a local buffer for each function.

cmd-record.c
```diff
@@ -108,8 +174,8 @@ static void setup_child_environ(struct opts *opts, int pfd)
 	old_libpath = getenv("LD_LIBRARY_PATH");
 	if (old_libpath) {
 		size_t len = strlen(buf) + strlen(old_libpath) + 2;
-		char *libpath = xmalloc(len);
 
+		libpath = xmalloc(len);
 		snprintf(libpath, len, "%s:%s", buf, old_libpath);
 		setenv("LD_LIBRARY_PATH", libpath, 1);
 		free(libpath);
```
namhyung: I thought about this logic again, and now I think we don't need to set `LD_LIBRARY_PATH` anymore since it uses a full path for `LD_PRELOAD`.

cmd-record.c
```diff
@@ -87,15 +87,81 @@ static char *build_debug_domain_string(void)
...
+
+	strv_split(&strv, getenv("LD_LIBRARY_PATH"), ":");
+
+	strv_for_each(&strv, libpath, i) {
+		snprintf(lib, 4096, "%s/%s", libpath, libmcount);
+		if (access(lib, F_OK) == 0) {
+			found = true;
+			break;
+		}
+	}
```
namhyung: It should be enough to check just `INSTALL_LIB_PATH` as that's what we know for libmcount.

cmd-record.c
```diff
@@ -87,15 +87,81 @@ static char *build_debug_domain_string(void)
...
+
+	strv_free(&strv);
+
+	if (!found) {
+		pr_warn("didn't you make install ?\n");
```
namhyung: "Didn't you run 'make install'?"

namhyung: But I think maybe we should just return the name of libmcount in this case and let the dynamic linker load it properly.

cmd-record.c
```diff
@@ -245,42 +311,25 @@ static void setup_child_environ(struct opts *opts, int pfd)
 	if (opts->patt_type != PATT_REGEX)
 		setenv("UFTRACE_PATTERN", get_filter_pattern(opts->patt_type), 1);
 
-	if (opts->lib_path)
-		snprintf(buf, sizeof(buf), "%s/libmcount/", opts->lib_path);
-	else
-		buf[0] = '\0';  /* to make strcat() work */
+	libpath = get_libmcount_path(opts);
+	if (libpath == NULL)
+		pr_err("cannot found libmcount.so");
```
namhyung: Please use `pr_err_ns` instead since we don't know the error code at the time.

cmd-info.c
Taeung: Oh, so do I !

cmd-record.c
Taeung: I agree that. I also agonized it because it seems to be duplicated.
I'll fix it based on your comment.

cmd-record.c
Taeung: Hum.. OK !

cmd-record.c
Taeung: Yep !

Taeung:
@namhyung  I changed this PR, I'm too late..

v2
- separate a patch
- rename `BUF_SIZE` to  `PATH_MAX`
- fix setup_child_environ() in cmd-live.c likewise
- cmd-info: pass a buffer instead of having a local buffer
- just check INSTALL_LIB_PATH
- use pr_err_ns() instead of pr_err()
- no set LD_LIBRARY_PATH anymore

cmd-record.c
```diff
@@ -245,46 +286,25 @@ static void setup_child_environ(struct opts *opts, int pfd)
 	if (opts->patt_type != PATT_REGEX)
 		setenv("UFTRACE_PATTERN", get_filter_pattern(opts->patt_type), 1);
 
-	if (opts->lib_path) {
-		snprintf(buf, sizeof(buf), "%s/libmcount/", opts->lib_path);
+	libpath = get_libmcount_path(opts);
+	if (libpath == NULL)
+		pr_err_ns("cannot found libmcount.so\n");
```
namhyung: When does it return NULL?

cmd-record.c
```diff
@@ -87,11 +87,52 @@ static char *build_debug_domain_string(void)
...
+	}
+	else if (opts->libmcount_single && !must_use_multi_thread) {
+		if (can_use_fast_libmcount(opts))
+			libmcount = "libmcount-fast-single.so";
+		else
+			libmcount = "libmcount-single.so";
+	}
+	else {
+		if (must_use_multi_thread && opts->libmcount_single)
+			pr_dbg("--libmcount-single is off because it calls pthread_create()\n");
```
namhyung: I know it's not your change but could you please change the message to `"--libmcount-single is off because it uses pthread\n"`?

cmd-live.c
```diff
@@ -76,32 +76,7 @@ static bool can_skip_replay(struct opts *opts, int record_result)
...
-		strcpy(buf, opts->lib_path);
-		strcat(buf, "/libmcount:");
-	} else {
-		/* to make strcat() work */
-		buf[0] = '\0';
-	}
-
-#ifdef INSTALL_LIB_PATH
-	strcat(buf, INSTALL_LIB_PATH);
-#endif
```
namhyung: Well, I think you don't use a full path unless `-L` option is given, right?  Then `LD_LIBRARY_PATH` is necessary for `INSTALL_LIB_PATH` IMO.

uftrace.h
```diff
@@ -30,6 +30,8 @@
 #define KB 1024
 #define MB (KB * 1024)
 
+#define PATH_MAX 4096
```
namhyung: `PATH_MAX` is defined in `<limits.h>`.

Taeung: Ah.. yep !

cmd-record.c
Taeung: OK !

cmd-live.c
Taeung: You're right ! I'll change it

cmd-record.c
Taeung: Hum.. it can return NULL at 16f0afb
So, I'll move this error handling code to 16f0afb

Taeung:
@namhyung , I changed this PR based on your comments !
will resend the before last two commits as other PR, later !

cmd-live.c
```diff
@@ -76,32 +76,23 @@ static bool can_skip_replay(struct opts *opts, int record_result)
...
+			snprintf(libpath, len, "%s:%s", INSTALL_LIB_PATH, old_libpath);
+			setenv("LD_LIBRARY_PATH", libpath, 1);
+			free(libpath);
+		}
+		else
+			setenv("LD_LIBRARY_PATH", INSTALL_LIB_PATH, 1);
 	}
-	else
-		setenv("LD_LIBRARY_PATH", buf, 1);
+#endif
```
namhyung: I think we can simply do it with `strjoin` now:

```c
libpath = strjoin(getenv("LD_LIBRARY_PATH"), INSTALL_LIB_PATH, ":");
setenv("LD_LIBRARY_PATH", libpath);
free(libpath);
```

It will add the `INSTALL_LIB_PATH` at  the end of the existing `LD_LIBRARY_PATH` but I think it's ok.

Taeung: OK, I'll chage it as you said

Taeung:
I changed it using `strjoin(getenv("LD_LIBRARY_PATH"), INSTALL_LIB_PATH..`.
It make the code more clean !

namhyung:
LGTM

Taeung:
Thanks !!  
(I like your `str` function series in utils/utils.c :smile: )

Issue #324, opened 2018-03-02
make uninstall does not work?


ParkHanbum:
```
m@m:~/git/uftrace$ sudo make install
  INSTALL  uftrace
  INSTALL  libmcount
  INSTALL  bash-completion
Makefile:36: To install man pages, please install 'pandoc'.

m@m:~/git/uftrace$ whereis libmcount.so
libmcount: /usr/local/lib/libmcount.so
m@m:~/git/uftrace$ whereis libmcount-fast.so
libmcount-fast: /usr/local/lib/libmcount-fast.so

m@m:~/git/uftrace$ ls -al /usr/local/lib | grep libmcount
-rwxr-xr-x  1 root root   714344  3월  2 10:17 libmcount-fast-single.so
-rwxr-xr-x  1 root root   720784  3월  2 10:17 libmcount-fast.so
-rwxr-xr-x  1 root root     9552  3월  2 10:17 libmcount-nop.so
-rwxr-xr-x  1 root root   740104  3월  2 10:17 libmcount-single.so
-rwxr-xr-x  1 root root   742424  3월  2 10:17 libmcount.so

m@m:~/git/uftrace$ sudo make uninstall
  REMOVE   uftrace
  REMOVE   libmcount
  REMOVE   bash-completion
Makefile:36: To install man pages, please install 'pandoc'.

m@m:~/git/uftrace$ ls -al /usr/local/lib | grep libmcount
-rwxr-xr-x  1 root root   714344  3월  2 10:17 libmcount-fast-single.so
-rwxr-xr-x  1 root root   720784  3월  2 10:17 libmcount-fast.so
-rwxr-xr-x  1 root root     9552  3월  2 10:17 libmcount-nop.so
-rwxr-xr-x  1 root root   740104  3월  2 10:17 libmcount-single.so
-rwxr-xr-x  1 root root   742424  3월  2 10:17 libmcount.so

m@m:~/git/uftrace$ file /usr/local/lib/libmcount.so
/usr/local/lib/libmcount.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=840370715bc14c5d014051687b62f5a2a498621d, not stripped
```

honggyukim:
@ParkHanbum Can you show me the log by running `sudo make V=1 uninstall`?

ParkHanbum:
@honggyukim  here it is.

```
m@m:~/uftrace$ sudo make V=1 uninstall
[...]

PR #339, opened 2018-04-06
changed the uninstall-script in the makefile to not use the brace expansion.

ParkHanbum:
this PR associated with #324.

we have used brace expansion to uninstall libmcount shared object series.

but it is not supported from other shell environment like dash.
therefore uninstall-script at makefile be changed to that did not use
brace expansion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,7 +259,15 @@ uninstall:
 	$(call QUIET_UNINSTALL, uftrace)
 	$(Q)$(RM) $(DESTDIR)$(bindir)/uftrace
 	$(call QUIET_UNINSTALL, libmcount)
-	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount{,-nop,-fast,-single,-fast-single}.so
+	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount.so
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -259,7 +259,15 @@ uninstall:
 	$(call QUIET_UNINSTALL, uftrace)
 	$(Q)$(RM) $(DESTDIR)$(bindir)/uftrace
 	$(call QUIET_UNINSTALL, libmcount)
-	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount{,-nop,-fast,-single,-fast-single}.so
+	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount.so
+	$(call QUIET_UNINSTALL, libmcount)
```
namhyung: Please change it to show the name of each library properly.

namhyung:
The commit message should look like below

> build: Do not use brace expansion
> 
> We have used brace expansion to uninstall a couple of libmcount*
> shared objects at once but it is not supported by other shell like dash.
> Therefore uninstall-script in the Makefile should be changed to
> not use the brace expansion and remove them separately.

PR #341, opened 2018-04-17
utils: Fix to avoid realloc fail in strjoin

honggyukim:
Since strjoin internally uses realloc passing the @left argument, the
@left must be dynamically allocated buffer.

But the commit 5eb6468 passes a buffer by getting getenv(), which makes
realloc in strjoin fail.

This patch fixes the problem and adds such information to strjoin
function to prevent such misuse.


honggyukim:
You're right. I just fixed it and updated it again. : )

honggyukim:
I've changed it again. Thanks.

PR #346, opened 2018-04-21
info: Update related to 'PATH_NAME'

qpakzk:
Because 'PATH_NAME' is defined in `limits.h`, add a header file `limits.h`
although not an error without it.

And replace 4096 which is a array `buf`'s size with PATH_NAME,
because PATH_NAME is 4096 defined in `limits.h`.

### DIFF
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -16,6 +16,7 @@
 #include <gelf.h>
 #include <argp.h>
 #include <fcntl.h>
+#include <limits.h>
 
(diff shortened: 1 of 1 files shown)


namhyung:
@Taeung worked on that some time ago so I think he should have something to talk. :)

Taeung:
OK, I resent the remaining work as #348 PR just now, @namhyung @qpakzk

namhyung:
Then we can close this, right?

PR #348, opened 2018-04-23
Refactoring the code regarding buffer and PATH_MAX

Taeung:
Hi @namhyung ,

I missed this PR,
I'd appreciate it, if you check this.

v2:
- remove `#define PATH_MAX 4096` in uftrace.h
- rebased on current master

### DIFF
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -25,12 +25,18 @@
 #define BUILD_ID_SIZE 20
 #define BUILD_ID_STR_SIZE (BUILD_ID_SIZE * 2 + 1)
 
+struct read_handler_arg {
+	struct ftrace_file_handle *handle;
+	char buf[PATH_MAX];
+};
+
 struct fill_handler_arg {
 	int fd;
 	int exit_status;
 	struct opts *opts;
 	struct rusage *rusage;
 	char *elapsed_time;
+	char buf[PATH_MAX];
 };
 
 static char *copy_info_str(char *src)
@@ -47,10 +53,9 @@ static char *copy_info_str(char *src)
 static int fill_exe_name(void *arg)
 {
 	struct fill_handler_arg *fha = arg;
-	char buf[4096];
 	char *exename;
 
-	exename = realpath(fha->opts->exename, buf);
+	exename = realpath(fha->opts->exename, fha->buf);
 	if (exename == NULL)
 		exename = fha->opts->exename;
 
@@ -59,11 +64,12 @@ static int fill_exe_name(void *arg)
 
 static int read_exe_name(void *arg)
 {
-	struct ftrace_file_handle *handle = arg;
+	struct read_handler_arg *rha = arg;
+	struct ftrace_file_handle *handle = rha->handle;
 	struct uftrace_info *info = &handle->info;
-	char buf[4096];
+	char *buf = rha->buf;
 
-	if (fgets(buf, sizeof(buf), handle->fp) == NULL)
+	if (fgets(buf, sizeof(rha->buf), handle->fp) == NULL)
 		return -1;
 
 	if (strncmp(buf, "exename:", 8))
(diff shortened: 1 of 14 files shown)


namhyung:
Could you please make sure that we have a single place to include the `<limits.h>`?

cmd-info.c
```diff
@@ -44,10 +44,9 @@ static char *copy_info_str(char *src)
 	return dst;
 }
 
-static int fill_exe_name(void *arg)
+static int fill_exe_name(void *arg, char *buf)
 {
 	struct fill_handler_arg *fha = arg;
```
namhyung: It'd be better if the struct arg have the buffer.  You could define `struct read_handler_arg` as well.

Taeung: I got it !

Taeung:
I changed this PR.

v2:
- the struct arg has `char * buf[PATH_MAX]` variable
- include `<limits.h>` at only `utils/utils.h`

However, I leave `libtraceevent/event-parse.c` as it is.
```bash
$ grep limits.h
utils/utils.h
19:#include <limits.h>

libtraceevent/event-parse.c
33:#include <limits.h>
```

cmd-info.c
Taeung: > You could define struct read_handler_arg as well.

Hum.. @namhyung , I didn't understand your comment:
I just add `char * buf[PATH_MAX]` to both  `struct fill_handler_arg` 
and `struct ftrace_file_handle`. Isn't it what you mean ?

namhyung:
Yes libtraceevent is different, please do not touch it.

And no, instead of using file handle please make a new struct has the handle and the buffer (pointer).

Taeung:
Hum.. like this ?
```patch
diff --git a/cmd-info.c b/cmd-info.c
index 0f7dfff..a4a0f52 100644
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -25,6 +25,11 @@
 #define BUILD_ID_SIZE 20
 #define BUILD_ID_STR_SIZE (BUILD_ID_SIZE * 2 + 1)
 
+struct read_handler_arg {
+        void *arg;
+        char buf[PATH_MAX];
+}
+
 struct fill_handler_arg {
         int fd;
         int exit_status;
```

namhyung:
Yes, but it should have a pointer to handle instead of void.

Taeung:
Yep, but we should also handle both `struct fill_handler_arg` and `struct ftrace_file_handle`, right ?
So I add new struct that has two pointers for them, OK ?
```patch
diff --git a/cmd-info.c b/cmd-info.c
index 0f7dfff..0088083 100644
--- a/cmd-info.c
+++ b/cmd-info.c
@@ -25,6 +25,12 @@
 #define BUILD_ID_SIZE 20
 #define BUILD_ID_STR_SIZE (BUILD_ID_SIZE * 2 + 1)
 
+struct read_handler_arg {
+        struct fill_handler_arg *fha;
+        struct ftrace_file_handle *handle;
+        char buf[PATH_MAX];
+}
+
 struct fill_handler_arg {
         int fd;
         int exit_status;
```

namhyung:
Why does it need the `fill_handler_arg`?

Taeung:
As you know, there are {fill_, read_} functions in cmd-info.c.
`fill_` function use `struct fill_handler_arg`
and `read_` function use `struct ftrace_file_handle`. like below:
```c
 54 static int fill_exe_name(void *arg)
 55 {
 56         struct fill_handler_arg *fha = arg;
 57         char *exename;
 58 
 59         exename = realpath(fha->opts->exename, fha->buf);
 60         if (exename == NULL)
 61                 exename = fha->opts->exename;
 62 
 63         return dprintf(fha->fd, "exename:%s\n", exename);
 64 }
 65 
 66 static int read_exe_name(void *arg)
 67 {
 68         struct ftrace_file_handle *handle = arg;
 69         struct uftrace_info *info = &handle->info;
 70         char *buf = handle->buf;
...

```

Taeung:
Hum.. I understood what you said.
But `fill_` functions also use `char *buf[4096]` for `char *buf[1024]` as a local buffer like `read_` functions
i.e. `fill_exe_name()`, `fill_cmdline()`, `fill_meminfo()` and `fill_osinfo()`.

Unlike `read_` functions (read callbacks),
just change them with `char *buf[PATH_MAX]` ?

namhyung:
I don't understand.  I think you can add the buffer both for read and fill args.

Taeung:
Hum.. sorry I didn't smoothly catch what you mean.
I changed this PR again, just now.

I'm not sure this change is right, but I'd appreciate it, if you check this.

namhyung:
Yes, this is what I wanted.  But now I think it'd be better keeping buffer and its size together (instead of passing `PATH_MAX` separately when using it).  So you could add a bufsize field to the args or just include the buffer (i.e. array, not the pointer) in the args (and use `sizeof(arg->buf)`) as you did earlier.

PR #352, opened 2018-05-04
test: Suppress generating buffer overflow detection calls

honggyukim:
_FORTIFY_SOURCE macro is used by default when optimization is on.  It generates additional buffer overflow detection calls.  So this fixes those kind of useless difference.

This also includes to use `colordiff` instead of `diff` for better comparison.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -23,7 +23,8 @@ class TestBase:
     objdir = 'objdir' in os.environ and os.environ['objdir'] or '..'
     uftrace_cmd = objdir + '/uftrace --no-pager -L' + objdir
 
-    default_cflags = ['-fno-inline', '-fno-builtin', '-fno-omit-frame-pointer']
+    default_cflags = ['-fno-inline', '-fno-builtin',
+                      '-fno-omit-frame-pointer', '-D_FORTIFY_SOURCE=0']
 
     def __init__(self, name, result, lang='C', cflags='', ldflags='', sort='task'):
         self.name = name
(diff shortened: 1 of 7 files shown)


namhyung:
What about using `diff --color=always` instead of `colordiff`?

tests/runtest.py
```diff
@@ -23,7 +23,8 @@ class TestBase:
     objdir = 'objdir' in os.environ and os.environ['objdir'] or '..'
     uftrace_cmd = objdir + '/uftrace --no-pager -L' + objdir
 
-    default_cflags = ['-fno-inline', '-fno-builtin', '-fno-omit-frame-pointer']
+    default_cflags = ['-fno-inline', '-fno-builtin', \
```
namhyung: I'm ok with this but I guess the backslash at the end of the line is unnecessary though..

honggyukim: done

honggyukim:
> What about using diff --color=always instead of colordiff?

Hmm.. my `diff` tool doesn't have `--color=` option. Can you please check it once again?

namhyung:
Hmm.. it's available on my machine.  Maybe it's a recent addition.

honggyukim:
Oh.. it's better now, but my diff tool seems to be a bit older than that.
```
$ diff --version
diff (GNU diffutils) 3.3
```
Which would be better using `colordiff` or `--color=always`?

namhyung:
Is the `colordiff` installed by default?  I cannot find it on my system.  I prefer more popular one but we can try all options - `diff --color` first and `colordiff`, finally `diff`.

honggyukim:
What would be a good way to check if `diff` has `--color=always` or not in python? Using try `Popen` doesn't work in this case.

namhyung:
Hmm.. I don't know.  Just try `colordiff` then.

honggyukim:
Thanks!

PR #358, opened 2018-05-11
several minor fix. 

ParkHanbum:
the 3rd commit contain the patch that support to next feature. 
if you think it never required others, let me know. 
I'll push it next, not this.

### DIFF
--- a/cmd-record.c
+++ b/cmd-record.c
@@ -1854,8 +1854,15 @@ int do_main_loop(int pfd[2], int ready, struct opts *opts, int pid)
 		if (pollfd.revents & POLLIN)
 			read_record_mmap(pfd[0], opts->dirname, opts->bufsize);
 
-		if (pollfd.revents & (POLLERR | POLLHUP))
+		if (pollfd.revents & POLLERR) {
+			pr_dbg("POLLERR\n");
 			break;
+		}
+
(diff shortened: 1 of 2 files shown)


namhyung:
I think it's better to include the first patch when it's really used.  Also I don't see why the second patch is needed.  For the third patch, I think it'd be more natural to see filename before the error string.

Finally you missed to add your Sign-off in the commit logs. :)

namhyung:
When did you see the HUP and what do you want to do?  The patch should describe it instead of just printing debug messages.

ParkHanbum:
@namhyung That's all I want to know. At least it allows me to know what is happening and to find the cause of the incident. it is not about "Handling POLLHUP and POLLERR " but it is aobut "let us know exactly what we've met case."

namhyung:
I don't want to add the current commit since it has no more information.  Please add some logic if you find out the cause of the POLLHUP later.

ParkHanbum:
@namhyung OK, I respect you. but let me ask you just one question. if user met POLLHUP or POLLERR for unspecificied reason, How could user report that issue?

namhyung:
@ParkHanbum sorry for late.  well I don't know.. if an user meet such a case it's hard to know it's because of the failed poll() or something.  Let's leave it now and think about when a report comes up.

PR #360, opened 2018-05-13
implement the attaching to already running process.

ParkHanbum:
this PR contains implementation of TODO that attaching the already running process.

usage :
'./uftrace dynamic -Lpwd --data=pwd/xxx --pid=<target pid>

explain:
use 'ptrace' to load 'libtrigger.so' into the target process. Later 'libtrigger.so' will load 'libmcount-dynamic.so'. Currently 'libmcount-dynamic.so' is almost identical to 'libmcount.so'. The only difference is that the environment variables are read from file which save environment variable. Currently, 'libmcount-dynamic.so' can only perform functions related to plthook.

### DIFF
--- a/arch/x86_64/Makefile
+++ b/arch/x86_64/Makefile
@@ -8,6 +8,7 @@ include $(srcdir)/Makefile.include
 ARCH_ENTRY_SRC = $(wildcard $(sdir)/*.S)
 ARCH_MCOUNT_SRC = $(wildcard $(sdir)/mcount-*.c) $(sdir)/regs.c $(sdir)/symbol.c
 ARCH_UFTRACE_SRC = $(sdir)/cpuinfo.c $(sdir)/regs.c $(sdir)/symbol.c
+ARCH_UFTRACE_SRC += $(sdir)/inject.c
 
 ARCH_MCOUNT_OBJS  = $(patsubst $(sdir)/%.S,$(odir)/%.op,$(ARCH_ENTRY_SRC))
 ARCH_MCOUNT_OBJS += $(patsubst $(sdir)/%.c,$(odir)/%.op,$(ARCH_MCOUNT_SRC))
--- a/arch/x86_64/inject.c
+++ b/arch/x86_64/inject.c
@@ -1,0 +1,296 @@
+/*
+ * Linux module for injecting shared object to the process by using ptrace.
+ *
+ * copied from: https://github.com/gaffe23/linux-inject
+ * Released under the GPL v2+.
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/user.h>
+#include <wait.h>
+#include <sys/time.h>
+#include "utils/inject-utils.h"
+#include "utils/ptrace.h"
+#include "utils/utils.h"
+
+/*
+ * inject_shared_lib()
+ *
+ * This is the code that will actually be injected into the target process.
+ * This code is responsible for loading the shared library into the target
+ * process' address space.  First, it calls malloc() to allocate a buffer to
+ * hold the filename of the library to be loaded. Then, it calls
+ * __libc_dlopen_mode(), libc's implementation of dlopen(), to load the desired
+ * shared library. Finally, it calls free() to free the buffer containing the
+ * library name. Each time it needs to give control back to the injector
+ * process, it breaks back in by executing an "int $3" instruction. See the
+ * comments below for more details on how this works.
+ *
+ */
+
+void inject_shared_lib(long mallocaddr, long freeaddr, long dlopenaddr)
+{
+	// here are the assumptions I'm making about what data will be located
+	// where at the time the target executes this code:
+	//
+	//   rdi = address of malloc() in target process
+	//   rsi = address of free() in target process
+	//   rdx = address of __libc_dlopen_mode() in target process
+	//   rcx = size of the path to the shared library we want to load
+
+	// save addresses of free() and __libc_dlopen_mode() on the stack for later use
+	asm(
+		// rsi is going to contain the address of free(). it's going to get wiped
+		// out by the call to malloc(), so save it on the stack for later
+		"push %rsi \n"
+		// same thing for rdx, which will contain the address of _dl_open()
+		"push %rdx"
+	);
+
+	// call malloc() from within the target process
+	asm(
+		// save previous value of r9, because we're going to use it to call malloc()
+		"push %r9 \n"
+		// now move the address of malloc() into r9
+		"mov %rdi,%r9 \n"
+		// choose the amount of memory to allocate with malloc() based on the size
+		// of the path to the shared library passed via rcx
+		"mov %rcx,%rdi \n"
+		// now call r9; malloc()
+		"callq *%r9 \n"
+		// after returning from malloc(), pop the previous value of r9 off the stack
+		"pop %r9 \n"
+		// break in so that we can see what malloc() returned
+		"int $3"
+	);
+
+	// call __libc_dlopen_mode() to load the shared library
+	asm(
+		// get the address of __libc_dlopen_mode() off of the stack so we can call it
+		"pop %rdx \n"
+		// as before, save the previous value of r9 on the stack
+		"push %r9 \n"
+		// copy the address of __libc_dlopen_mode() into r9
+		"mov %rdx,%r9 \n"
+		// 1st argument to __libc_dlopen_mode(): filename = the address of the buffer returned by malloc()
+		"mov %rax,%rdi \n"
+		// 2nd argument to __libc_dlopen_mode(): flag = RTLD_LAZY
+		"movabs $1,%rsi \n"
+		// call __libc_dlopen_mode()
+		"callq *%r9 \n"
+		// restore old r9 value
+		"pop %r9 \n"
+		// break in so that we can see what __libc_dlopen_mode() returned
+		"int $3"
+	);
+
+	// call free() to free the buffer we allocated earlier.
+	//
+	// Note: I found that if you put a nonzero value in r9, free() seems to
+	// interpret that as an address to be freed, even though it's only
+	// supposed to take one argument. As a result, I had to call it using a
+	// register that's not used as part of the x64 calling convention. I
+	// chose rbx.
+	asm(
+		// at this point, rax should still contain our malloc()d buffer from earlier.
+		// we're going to free it, so move rax into rdi to make it the first argument to free().
+		"mov %rax,%rdi \n"
+		// pop rsi so that we can get the address to free(), which we pushed onto the stack a while ago.
+		"pop %rsi \n"
+		// save previous rbx value
+		"push %rbx \n"
+		// load the address of free() into rbx
+		"mov %rsi,%rbx \n"
+		// zero out rsi, because free() might think that it contains something that should be freed
+		"xor %rsi,%rsi \n"
+		// break in so that we can check out the arguments right before making the call
+		"int $3 \n"
+		// call free()
+		"callq *%rbx \n"
+		// restore previous rbx value
+		"pop %rbx"
+	);
+
+	// we already overwrote the RET instruction at the end of this function
+	// with an INT 3, so at this point the injector will regain control of
+	// the target's execution.
+}
+
+/*
+ * inject_shared_lib_end()
+ *
+ * This function's only purpose is to be contiguous to inject_shared_lib(),
+ * so that we can use its address to more precisely figure out how long
+ * inject_shared_lib() is.
+ *
+ */
+
+void inject_shared_lib_end()
+{
+}
+
+int inject(char* libname, pid_t pid)
+{
+	char* lib_path = realpath(libname, NULL);
+	pid_t target = 0;
+
+	if(!lib_path) {
+		pr_err_ns("can't find file \"%s\"\n", libname);
+	}
+	target = pid;
+
+	int lib_path_length = strlen(lib_path) + 1;
+
+	int mypid = getpid();
+	long mylibcaddr = getlibcaddr(mypid);
+
+	// find the addresses of the syscalls that we'd like to use inside the
+	// target, as loaded inside THIS process (i.e. NOT the target process)
+	long malloc_addr = get_function_addr("malloc");
+	long free_addr = get_function_addr("free");
+	long dlopen_addr = get_function_addr("__libc_dlopen_mode");
+
+	// use the base address of libc to calculate offsets for the syscalls
+	// we want to use
+	long malloc_offset = malloc_addr - mylibcaddr;
+	long free_offset = free_addr - mylibcaddr;
+	long dlopen_offset = dlopen_addr - mylibcaddr;
+
+
+	// get the target process' libc address and use it to find the
+	// addresses of the syscalls we want to use inside the target process
+	long target_libc_addr = getlibcaddr(target);
+	long target_malloc_addr = target_libc_addr + malloc_offset;
+	long target_free_addr = target_libc_addr + free_offset;
+	long target_dlopen_addr = target_libc_addr + dlopen_offset;
+
+	struct user_regs_struct oldregs, regs;
+	memset(&oldregs, 0, sizeof(struct user_regs_struct));
+	memset(&regs, 0, sizeof(struct user_regs_struct));
+
+	ptrace_attach(target);
+
+	ptrace_getregs(target, &oldregs);
+	memcpy(&regs, &oldregs, sizeof(struct user_regs_struct));
+
+	// find a good address to copy code to
+	long addr = freespaceaddr(target) + sizeof(long);
+
+	// now that we have an address to copy code to, set the target's rip to
+	// it. we have to advance by 2 bytes here because rip gets incremented
+	// by the size of the current instruction, and the instruction at the
+	// start of the function to inject always happens to be 2 bytes long.
+	regs.rip = addr + 2;
+
+	// pass arguments to my function inject_shared_lib() by loading them
+	// into the right registers. note that this will definitely only work
+	// on x64, because it relies on the x64 calling convention, in which
+	// arguments are passed via registers rdi, rsi, rdx, rcx, r8, and r9.
+	// see comments in inject_shared_lib() for more details.
+	regs.rdi = target_malloc_addr;
+	regs.rsi = target_free_addr;
+	regs.rdx = target_dlopen_addr;
+	regs.rcx = lib_path_length;
+	ptrace_setregs(target, &regs);
+
+	// figure out the size of inject_shared_lib() so we know how big of a buffer to allocate.
+	size_t inject_shared_lib_size = (intptr_t)inject_shared_lib_end - (intptr_t)inject_shared_lib;
+
+	// also figure out where the RET instruction at the end of
+	// inject_shared_lib() lies so that we can overwrite it with an INT 3
+	// in order to break back into the target process. note that on x64,
+	// gcc and clang both force function addresses to be word-aligned,
+	// which means that functions are padded with NOPs. as a result, even
+	// though we've found the length of the function, it is very likely
+	// padded with NOPs, so we need to actually search to find the RET.
+	intptr_t inject_shared_lib_ret = (intptr_t)find_ret(inject_shared_lib_end) - (intptr_t)inject_shared_lib;
+
+	// back up whatever data used to be at the address we want to modify.
+	char* backup = malloc(inject_shared_lib_size * sizeof(char));
+	ptrace_read(target, addr, backup, inject_shared_lib_size);
+
+	// set up a buffer to hold the code we're going to inject into the
+	// target process.
+	char* newcode = malloc(inject_shared_lib_size * sizeof(char));
+	memset(newcode, 0, inject_shared_lib_size * sizeof(char));
+
+	// copy the code of inject_shared_lib() to a buffer.
+	memcpy(newcode, inject_shared_lib, inject_shared_lib_size - 1);
+	// overwrite the RET instruction with an INT 3.
+	newcode[inject_shared_lib_ret] = INTEL_INT3_INSTRUCTION;
+
+	// copy inject_shared_lib()'s code to the target address inside the
+	// target process' address space.
+	ptrace_write(target, addr, newcode, inject_shared_lib_size);
+
+	// now that the new code is in place, let the target run our injected
+	// code.
+	ptrace_cont(target);
+
+
+	// at this point, the target should have run malloc(). check its return
+	// value to see if it succeeded, and bail out cleanly if it didn't.
+	struct user_regs_struct malloc_regs;
+	memset(&malloc_regs, 0, sizeof(struct user_regs_struct));
+	ptrace_getregs(target, &malloc_regs);
+	unsigned long long target_buf = malloc_regs.rax;
+	if(target_buf == 0) {
+		restore_state_and_detach(target, addr, backup, inject_shared_lib_size, oldregs);
+		free(backup);
+		free(newcode);
+		pr_err_ns("malloc() failed to allocate memory\n");
+	}
+
+	// if we get here, then malloc likely succeeded, so now we need to copy
+	// the path to the shared library we want to inject into the buffer
+	// that the target process just malloc'd. this is needed so that it can
+	// be passed as an argument to __libc_dlopen_mode later on.
+
+	// read the current value of rax, which contains malloc's return value,
+	// and copy the name of our shared library to that address inside the
+	// target process.
+	ptrace_write(target, target_buf, lib_path, lib_path_length);
+
+	// continue the target's execution again in order to call
+	// __libc_dlopen_mode.
+	ptrace_cont(target);
+
+	// check out what the registers look like after calling dlopen.
+	struct user_regs_struct dlopen_regs;
+	memset(&dlopen_regs, 0, sizeof(struct user_regs_struct));
+	ptrace_getregs(target, &dlopen_regs);
+	unsigned long long lib_addr = dlopen_regs.rax;
+
+	// if rax is 0 here, then __libc_dlopen_mode failed, and we should bail
+	// out cleanly.
+	if(lib_addr == 0) {
+		restore_state_and_detach(target, addr, backup, inject_shared_lib_size, oldregs);
+		free(backup);
+		free(newcode);
+		pr_err_ns("__libc_dlopen_mode() failed to load %s\n", libname);
+	}
+
+	// now check /proc/pid/maps to see whether injection was successful.
+	if(checkloaded(target, libname)) {
+		pr_dbg("\"%s\" successfully injected\n", libname);
+	}
+	else {
+		pr_err_ns("could not inject \"%s\"\n", libname);
+	}
+
+	// as a courtesy, free the buffer that we allocated inside the target
+	// process. we don't really care whether this succeeds, so don't
+	// bother checking the return value.
+	ptrace_cont(target);
+
+	// at this point, if everything went according to plan, we've loaded
+	// the shared library inside the target process, so we're done. restore
+	// the old state and detach from the target.
+	restore_state_and_detach(target, addr, backup, inject_shared_lib_size, oldregs);
+	free(backup);
+	free(newcode);
+
+	return 0;
+}
--- a/cmd-dynamic.c
+++ b/cmd-dynamic.c
@@ -1,0 +1,476 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <inttypes.h>
+#include <unistd.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <assert.h>
+#include <dirent.h>
+#include <pthread.h>
+#include <signal.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <sys/ioctl.h>
+#include <sys/eventfd.h>
+#include <sys/resource.h>
+#include <sys/epoll.h>
+#include <fnmatch.h>
+#include <linux/limits.h>
+#include <sys/ptrace.h>
+#include <sys/time.h>
+
+#include "uftrace.h"
+#include "libmcount/mcount.h"
+#include "utils/utils.h"
+#include "utils/symbol.h"
+#include "utils/list.h"
+#include "utils/filter.h"
+#include "utils/kernel.h"
+#include "utils/perf.h"
+#include "utils/compiler.h"
+#include "utils/env-file.h"
+
+
+// define which method to use.
+#define DYNAMIC_TO_PROCESS 0
+
+static bool has_perf_event;
+static char absolute_file_path[PATH_MAX];
+extern int do_main_loop(int pfd[2], int ready, struct opts *opts, int pid);
+extern bool check_linux_schedule_event(char *events,
+				       enum uftrace_pattern_type ptype);
+extern bool check_linux_perf_event(char *events);
+extern char *build_debug_domain_string(void);
+extern int inject(char* libname, pid_t pid);
+
+#define ERROR_WITHOUT_LDPRELOAD \
+"you must specify LD_PRELOAD path by using '-L' option if want use \n"	\
+"\tdynamic tracing to the process.\n"
+
+/*
+ * Create a temporary file to save the environment variable for uftrace.
+ *
+ * in case for dynamic tracing to processes, we use file to passing
+ * environment variable for uftrace because LD_PRELOAD not work.
+ */
+static void make_uftrace_environ_file(struct opts *opts, int pfd)
+{
+	char buf[4096];
+	char *old_preload, *old_libpath;
+	int env_fd;
+	int method = DYNAMIC_TO_PROCESS;
+
+	env_fd = create_env_file();
+	if (env_fd < 0)
+		pr_err_ns("FILE OPEN FAILED\n");
+
+	if (opts->pid) {
+		method = DYNAMIC_TO_PROCESS;
+	}
+
+	if (opts->lib_path) {
+		strcpy(buf, opts->lib_path);
+		strcat(buf, "/libmcount:");
+	} else {
+		if (method == DYNAMIC_TO_PROCESS) {
+			pr_err_ns(ERROR_WITHOUT_LDPRELOAD);
+		}
+		/* to make strcat() work */
+		buf[0] = '\0';
+	}
+
+#ifdef INSTALL_LIB_PATH
+	strcat(buf, INSTALL_LIB_PATH);
+#endif
+
+	old_libpath = getenv("LD_LIBRARY_PATH");
+	if (old_libpath) {
+		size_t len = strlen(buf) + strlen(old_libpath) + 2;
+		char *libpath = xmalloc(len);
+
+		snprintf(libpath, len, "%s:%s", buf, old_libpath);
+		set_env_to_file(env_fd, "LD_LIBRARY_PATH", libpath);
+		free(libpath);
+	}
+	else {
+		set_env_to_file(env_fd, "LD_LIBRARY_PATH", buf);
+	}
+
+	if (opts->filter) {
+		char *filter_str = uftrace_clear_kernel(opts->filter);
+
+		if (filter_str) {
+			set_env_to_file(env_fd, "UFTRACE_FILTER", filter_str);
+			free(filter_str);
+		}
+	}
+
+	if (opts->trigger) {
+		char *trigger_str = uftrace_clear_kernel(opts->trigger);
+
+		if (trigger_str) {
+			set_env_to_file(env_fd, "UFTRACE_TRIGGER", trigger_str);
+			free(trigger_str);
+		}
+	}
+
+	if (opts->retval) {
+		char *retval_str = uftrace_clear_kernel(opts->retval);
+
+		if (retval_str) {
+			set_env_to_file(env_fd, "UFTRACE_RETVAL", retval_str);
+			free(retval_str);
+		}
+	}
+
+	if (opts->auto_args) {
+		set_env_to_file(env_fd, "UFTRACE_AUTO_ARGS", "1");
+	}
+
+	if (opts->patch) {
+		char *patch_str = uftrace_clear_kernel(opts->patch);
+
+		if (patch_str) {
+			set_env_to_file(env_fd, "UFTRACE_PATCH", patch_str);
+			free(patch_str);
+		}
+	}
+
+	if (opts->event) {
+		char *event_str = uftrace_clear_kernel(opts->event);
+
+		if (event_str) {
+			set_env_to_file(env_fd, "UFTRACE_EVENT", event_str);
+			free(event_str);
+		}
+	}
+
+	if (opts->depth != OPT_DEPTH_DEFAULT) {
+		snprintf(buf, sizeof(buf), "%d", opts->depth);
+		set_env_to_file(env_fd, "UFTRACE_DEPTH", buf);
+	}
+
+	if (opts->max_stack != OPT_RSTACK_DEFAULT) {
+		snprintf(buf, sizeof(buf), "%d", opts->max_stack);
+		set_env_to_file(env_fd, "UFTRACE_MAX_STACK", buf);
+	}
+
+	if (opts->threshold) {
+		snprintf(buf, sizeof(buf), "%"PRIu64, opts->threshold);
+		set_env_to_file(env_fd, "UFTRACE_THRESHOLD", buf);
+	}
+
+	if (opts->libcall) {
+		set_env_to_file(env_fd, "UFTRACE_PLTHOOK", "1");
+
+		if (opts->want_bind_not) {
+			/* do not update GOTPLT after resolving symbols */
+			set_env_to_file(env_fd, "LD_BIND_NOT", "1");
+		}
+
+		if (opts->nest_libcall) {
+			set_env_to_file(env_fd, "UFTRACE_NEST_LIBCALL", "1");
+		}
+	}
+
+	if (strcmp(opts->dirname, UFTRACE_DIR_NAME)) {
+		set_env_to_file(env_fd, "UFTRACE_DIR", opts->dirname);
+	}
+
+	if (opts->bufsize != SHMEM_BUFFER_SIZE) {
+		snprintf(buf, sizeof(buf), "%lu", opts->bufsize);
+		set_env_to_file(env_fd, "UFTRACE_BUFFER", buf);
+	}
+
+	if (opts->logfile) {
+		snprintf(buf, sizeof(buf), "%d", fileno(logfp));
+		set_env_to_file(env_fd, "UFTRACE_LOGFD", buf);
+	}
+
+	snprintf(buf, sizeof(buf), "%d", pfd);
+	set_env_to_file(env_fd, "UFTRACE_PIPE", buf);
+	set_env_to_file(env_fd, "UFTRACE_SHMEM", "1");
+
+	if (debug) {
+		snprintf(buf, sizeof(buf), "%d", debug);
+		set_env_to_file(env_fd, "UFTRACE_DEBUG", buf);
+		set_env_to_file(env_fd, "UFTRACE_DEBUG_DOMAIN", build_debug_domain_string());
+	}
+
+	if(opts->disabled) {
+		set_env_to_file(env_fd, "UFTRACE_DISABLED", "1");
+	}
+
+	if (log_color == COLOR_ON) {
+		snprintf(buf, sizeof(buf), "%d", log_color);
+		set_env_to_file(env_fd, "UFTRACE_COLOR", buf);
+	}
+
+	snprintf(buf, sizeof(buf), "%d", demangler);
+	set_env_to_file(env_fd, "UFTRACE_DEMANGLE", buf);
+
+	if ((opts->kernel || has_kernel_event(opts->event)) &&
+	    check_kernel_pid_filter()) {
+		set_env_to_file(env_fd, "UFTRACE_KERNEL_PID_UPDATE", "1");
+	}
+
+	if (opts->script_file) {
+		set_env_to_file(env_fd, "UFTRACE_SCRIPT", opts->script_file);
+	}
+
+	if (opts->lib_path)
+		snprintf(buf, sizeof(buf), "%s/libmcount/", opts->lib_path);
+	else
+		buf[0] = '\0';  /* to make strcat() work */
+
+	strcat(buf, "libmcount-dynamic.so");
+	pr_dbg("using %s library for tracing\n", buf);
+
+	old_preload = getenv("LD_PRELOAD");
+	if (old_preload) {
+		size_t len = strlen(buf) + strlen(old_preload) + 2;
+		char *preload = xmalloc(len);
+
+		snprintf(preload, len, "%s:%s", buf, old_preload);
+		set_env_to_file(env_fd, "LD_PRELOAD", preload);
+		free(preload);
+	}
+	else {
+		set_env_to_file(env_fd, "LD_PRELOAD", buf);
+	}
+
+	// The process ID to inject the shared object.
+	snprintf(buf, sizeof(buf), "%d", getpid());
+	set_env_to_file(env_fd, "UFTRACE_PID", buf);
+
+	// The below code should be placed at the end.
+	set_env_to_file(env_fd, "XRAY_OPTIONS", "patch_premain=false");
+	close(env_fd);
+}
+
+// settings for using dynamic tracing feature.
+void setup_uftrace_environ(struct opts *opts, int pfd)
+{
+	make_uftrace_environ_file(opts, pfd);
+}
+
+
+char* get_libtrigger_path(struct opts *opts)
+{
+	char *lib = xmalloc(PATH_MAX);
+	char *libtrigger = "libtrigger.so";
+        if (opts->lib_path) {
+                snprintf(lib, PATH_MAX, "%s/libmcount/%s", opts->lib_path, libtrigger);
+
+                if (access(lib, F_OK) == 0) {
+                        return lib;
+                }
+                else if (errno == ENOENT) {
+                        snprintf(lib, PATH_MAX, "%s/%s", opts->lib_path, libtrigger);                        if (access(lib, F_OK) == 0)
+                                return lib;
+                }
+                free(lib);
+                return NULL;
+        }
+
+#ifdef INSTALL_LIB_PATH
+        snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libtrigger);
+        if (access(lib, F_OK) != 0 && errno == ENOENT)
+                pr_warn("Didn't you run 'make install' ?\n");
+#endif
+        strcpy(lib, libtrigger);
+	return lib;
+}
+
+
+/*
+ * inject the shared object 'libtrigger.so' to processes.
+ *
+ * inject libtrigger.so using '__libc_dlopen_mode' which export
+ * from 'libc.so'. This is because libc.so is always loaded.
+ * Another option is to use 'dlopen' which exported from libdl.so,
+ * but 'libdl.so' will not always load.
+ *
+ * we must  have to load 'libmcount-dynamic.so' but '__libc_dlopen_mode'
+ * will failed to load it. maybe there is some complex issue.
+ *
+ * therefore, we inject 'libtrigger.so' to the process first.
+ * after it loaded, it will load 'libmcount-dynamic.so' continuly
+ * with using 'dlopen'.
+ */
+void do_inject(int pfd[2], int ready, struct opts *opts, char *argv[])
+{
+	int target_pid = opts->pid;
+	char* libtrigger_path;
+        uint64_t dummy;
+        close(pfd[0]);
+	setup_uftrace_environ(opts, pfd[1]);
+
+        if (read(ready, &dummy, sizeof(dummy)) != (ssize_t)sizeof(dummy))
+                pr_err("waiting for parent failed");
+
+	libtrigger_path = get_libtrigger_path(opts);
+
+	if(libtrigger_path == NULL)
+		pr_err("connot found libtrigger.so at %s\n", libtrigger_path);
+	pr_dbg("libtrigger found at : %s\n", libtrigger_path);
+	inject(libtrigger_path, target_pid);
+}
+
+/*
+find the executable file for the process you want to inject.
+Local symbol informations are required to use dynamic tracing.
+*/
+int find_exefile(struct opts *opts) {
+	DIR *directory = opendir("/proc/");
+	char* exePath;
+	int exePathLen = 0x1000;
+	ssize_t len;
+	pid_t pid = opts->pid;
+
+	if (!directory)
+		return 0;
+	exePath = malloc(exePathLen * sizeof(char));
+	sprintf(exePath, "/proc/%d/exe", pid);
+	exePath[exePathLen-1] = '\0';
+	len = readlink(exePath, absolute_file_path, PATH_MAX - 1);
+	if(len == -1) {
+		free(exePath);
+		return 0;
+	}
+	absolute_file_path[len] = '\0';
+	free(exePath);
+	closedir(directory);
+	opts->exename = absolute_file_path;
+	return 1;
+}
+
+#define UFTRACE_MSG  "Cannot trace '%s': No such executable file\n"	\
+"\tNote that uftrace doesn't search $PATH for you.\n"			\
+"\tIf you really want to trace executables in the $PATH,\n"		\
+"\tplease give it the absolute pathname (like /usr/bin/%s).\n"
+
+#define UFTRACE_ELF_MSG  "Cannot trace '%s': Invalid file\n"		\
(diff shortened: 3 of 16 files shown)


namhyung:
I like the way you did the job - I'd also like to work on attaching to a process at first.  But I don't want to add a new command just for this feature.  I prefer having `-p <pid>` option similar to other tools like `gdb`, `strace` and `perf`.

I didn't review this PR yet but it seems there are many codying style issues.  And you need to ensure to keep the license and copyright notice as is.  It'd be nice if you could minimize the code copy.  I'll review soon anyway.  Many thanks to your great work! :)

honggyukim:
Hi @ParkHanbum, I also appreciate your effort for this work.

But as @namhyung mentioned, this PR contains a lot of style issues. In Linux kernel source tree, there is a tool called `checkpatch.pl` to check this kind of problems.

Here are the error reports of `checkpatch.pl`.
```diff
WARNING: Missing a blank line after declarations
#68: FILE: utils/inject-utils.c:34:
+       char perms[5];
+       sprintf(filename, "/proc/%d/maps", pid);

ERROR: space required before the open parenthesis '('
#70: FILE: utils/inject-utils.c:36:
+       if(fp == NULL)

WARNING: unnecessary whitespace before a quoted newline
#71: FILE: utils/inject-utils.c:37:
+               pr_err("cannot open /proc/%d/maps \n", pid);

ERROR: "foo* bar" should be "foo *bar"
#133: FILE: utils/inject-utils.c:99:
+int checkloaded(pid_t pid, char* libname)

WARNING: line over 80 characters
#449: FILE: utils/ptrace.c:223:
+               pr_dbg("instead of expected SIGTRAP, target stopped with signal %d: %s\n", targetsig.si_signo, strsignal(targetsig.si_signo));

WARNING: unnecessary whitespace before a quoted newline
#88: FILE: arch/x86_64/inject.c:55:
+               "push %r9 \n"

ERROR: Bad function definition - void inject_shared_lib_end() should probably be void inject_shared_lib_end(void)
#163: FILE: arch/x86_64/inject.c:130:
+void inject_shared_lib_end()

        ... many more ...
```

It shows too many errors and warnings so I've attached the full output so please refer to the report and correct the style first so that @namhyung can focus on the important parts other than the style issues.
[checkpatch-for-PR_FOR_DT.txt](https://github.com/namhyung/uftrace/files/2001729/checkpatch-for-PR_FOR_DT.txt)

For more about the style issues, you can read [coding stye](https://github.com/namhyung/uftrace/blob/master/CONTRIBUTING.md#coding-style) doc.

ParkHanbum:
@honggyukim I didn't know about that. thank you for let me know. I'll check that!

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+"\tfor example:\n" 							\
+"\t[BADD]./uftrace dynamic -L/home/m/uftrace --pid=xxxx\n" 		\
+"\t[GOOD]./uftrace dynaimc -L/home/m/uftrace"				\
+"--data=/home/m/uftrace/data --pid=xxxx\n"
+
+#define ERROR_ENVFILE_NOT_EXIST \
+"\tCannot access to the %s which contain environment variable.\n"
+
+
+__attribute__((weak)) void build_debug_domain(char *dbg_domain_str)
```
Taeung: I thought this `build_debug_domain()` can be duplicated with `build_debug_domain()` in libmcount/misc.c
So I thought we can include `internal.h` to this file, no ?

cmd-dynamic.c
```diff
@@ -0,0 +1,473 @@
...
+
+	// The below code should be placed at the end.
+	set_env_to_file(env_fd, "XRAY_OPTIONS", "patch_premain=false");
+	close(env_fd);
+}
+
+// settings for using dynamic tracing feature.
+void setup_uftrace_environ(struct opts *opts, int pfd)
+{
+	make_uftrace_environ_file(opts, pfd);
```
Taeung: I thought it is enough that if there is one of the two functions `setup_uftrace_environ` and `make_uftrace_environ_file`, no ?

libmcount/mcount-dynamic.c
```diff
@@ -0,0 +1,91 @@
...
+
+/*
+ * previous mcount_startup() the contructor.
+ */
+void pre_startup(void)
+{
+	int fd_envfile;
+
+	fd_envfile = open_env_file();
+	set_env_from_file(fd_envfile);
```
Taeung: After `pre_startup()` does `setenv()`, `mcount_startup()` does `getenv()` to pass various uftrace option values.
But I think we can skip `setenv()` and `getenv()`.
To sum up, after reading "envfile", we can directly set uftrace option values. What do you think about this ?

honggyukim:
I cannot review this PR now but can you rebase it based on the current master? It has some conflicts.

Taeung:
Hi @ParkHanbum 

I think this job is really great !! Thanks for doing this.
I comment several things after briefly reviewing this PR.
I'd appreciate it, if you check my comments.

In addition, It seems that you need to add `libtrigger.so` to .gitignore file,
or rename it to `libmcount-trigger.so`.
How about that ?

Because I found the below git output after building this PR source code:
```bash
On branch PR_FOR_DT
Your branch is up to date with 'origin/PR_FOR_DT'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	libtrigger.so

nothing added to commit but untracked files present (use "git add" to track)
```

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+#include "utils/env-file.h"
+#include "utils/utils.h"
+
+
+#define ERROR_LIBRARY_NOT_EXIST \
+"%s does not exist libmcount shared object.\n"				\
+"\tmake sure that the file exists.\n\n"					\
+"\t[notification]\n"							\
+"\twhen you use dynamic tracing to the process, you must specify the\n"	\
+"\tabsolute path of 'libmcount-dynamic.so' not the relative path.\n\n" 	\
```
namhyung: Where comes this restriction from?  Isn't it just a matter of calling `realpath()`?

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+	outfp = stdout;
+	logfp = stderr;
+
+	/*
+	 * the Process output is delayed for unknown reasons when using
+	 * dynamic tracing. cannot found the reason of delay but below code
+	 * can mitigation the symptom.
+	 */
+	setvbuf(stdout, NULL, _IONBF, 1024);
+	setvbuf(stderr, NULL, _IONBF, 1024);
```
namhyung: But this changes the behavior of the target process which is not desired.

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+		pr_dbg("DATA-DIR EXIST : %s\n", data_dir_path);
+	else
+		pr_err_ns(ERROR_DATADIR_NOT_EXIST, data_dir_path);
+
+	if (stat(preload_library_path, &file) == 0)
+		pr_dbg("LIBRARY EXIST : %s\n", preload_library_path);
+	else
+		pr_err_ns(ERROR_LIBRARY_NOT_EXIST, preload_library_path);
+
+	handle = dlopen(preload_library_path, RTLD_LAZY);
```
namhyung: Hmm.. so do you want to load libmcount.so again?  Why not just doing it in this library?

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+
+	pr_dbg("LIBRARY PATH : %s\n", preload_library_path);
+	pr_dbg("DATADIR PATH : %s\n", data_dir_path);
+
+	if (stat(data_dir_path, &file) == 0)
+		pr_dbg("DATA-DIR EXIST : %s\n", data_dir_path);
+	else
+		pr_err_ns(ERROR_DATADIR_NOT_EXIST, data_dir_path);
+
+	if (stat(preload_library_path, &file) == 0)
```
namhyung: `LD_PRELOAD` can have a list of libraries separated by colons..

libmcount/trigger.c
```diff
@@ -0,0 +1,114 @@
...
+		build_debug_domain(getenv("UFTRACE_DEBUG_DOMAIN"));
+	}
+
+	pr_dbg("LIBRARY PATH : %s\n", preload_library_path);
+	pr_dbg("DATADIR PATH : %s\n", data_dir_path);
+
+	if (stat(data_dir_path, &file) == 0)
+		pr_dbg("DATA-DIR EXIST : %s\n", data_dir_path);
+	else
+		pr_err_ns(ERROR_DATADIR_NOT_EXIST, data_dir_path);
```
namhyung: It can fail with various reasons so you'd better checking the error code.

honggyukim:
@ParkHanbum Sorry but I just made a change to move all `cmd-*.c` files into `cmds` directory so please update your PR as well.

ParkHanbum:
@namhyung thank you for your reviewing! I'll check the source code that you commented. 
@honggyukim nope. it is not problem.

libmcount/trigger.c
ParkHanbum: @Taeung if it is possible, I would have done that. "internal.h" have many dependency with source file which located at libmcount. but libtrigger does not need other codes, need only 'build_debug_domain'. so, I have no choice. lately, I'll refactoring this code by decoupling 'build_debug_domain' from 'internal.h' if it need

cmd-dynamic.c
ParkHanbum: @Taeung you're right. I'm considering that remove setup_uftrace_environ.

libmcount/mcount-dynamic.c
ParkHanbum: @Taeung you're right, but I'm considering that dynamic tracing feature need more time to stable. for that reason, I want to split dynamic tracing feature from uftrace. it is the reason what I have wrote code this way. if you have more idea about isolate the dynamic tracing feature, let me know please.

arch/x86_64/Makefile
```diff
@@ -8,6 +8,7 @@ include $(srcdir)/Makefile.include
 ARCH_ENTRY_SRC = $(wildcard $(sdir)/*.S)
 ARCH_MCOUNT_SRC = $(wildcard $(sdir)/mcount-*.c) $(sdir)/regs.c $(sdir)/symbol.c
 ARCH_UFTRACE_SRC = $(sdir)/cpuinfo.c $(sdir)/regs.c $(sdir)/symbol.c
+ARCH_UFTRACE_SRC += $(sdir)/inject.c
```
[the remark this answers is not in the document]
ParkHanbum: I was not thoughtful enough. as you say, I'll change this commit to include commit which about inject.c .

ParkHanbum:
how about chagne the name "libtrigger.so" to  "libmcount-loader.so" .

utils/inject-utils.c
```diff
@@ -0,0 +1,172 @@
+#include <stdio.h>
```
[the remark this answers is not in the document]
ParkHanbum: OK~! it will be.

arch/x86_64/inject.c
```diff
@@ -0,0 +1,335 @@
...
+	pid_t target = 0;
+
+	if (!lib_path)
+		pr_err_ns("can't find file \"%s\"\n", libname);
+
+	target = pid;
+
+	int lib_path_length = strlen(lib_path) + 1;
+	int mypid = getpid();
+	long mylibcaddr = getlibcaddr(mypid);
```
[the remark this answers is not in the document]
ParkHanbum: hmm... how about now? it was implemented at previouse commit.

utils/inject-utils.c
```diff
@@ -0,0 +1,172 @@
...
+	char filename[30];
+	char line[850];
+	long addr;
+	char str[20];
+	char perms[5];
+
+	sprintf(filename, "/proc/%d/maps", pid);
+	fp = fopen(filename, "r");
+	if (fp == NULL)
+		pr_err("cannot open /proc/%d/maps\n", pid);
```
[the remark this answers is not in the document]
ParkHanbum: OH I SEE!! I'll remove them all :D

libmcount/trigger.c
ParkHanbum: yes, it may. but uftrace could not print any output without that. what choice do I have?

libmcount/trigger.c
ParkHanbum: because libmcount could not loaded by using __libc_dlopen_mode but it can be possible by using dlopen. I don't know what cause this. see below backtrace. 
```
   0x000000000040062a <+19>:    callq  0x400520 <__libc_dlopen_mode@plt>
   0x000000000040062f <+24>:    mov    $0x40071c,%edi
   0x0000000000400634 <+29>:    callq  0x4004f0 <puts@plt>
   0x0000000000400639 <+34>:    mov    0x200458(%rip),%rax        # 0x600a98 <so_name>
   0x0000000000400640 <+41>:    mov    $0x2,%esi
   0x0000000000400645 <+46>:    mov    %rax,%rdi
   0x0000000000400648 <+49>:    callq  0x400510 <dlopen@plt>
   0x000000000040064d <+54>:    mov    $0x0,%eax
   0x0000000000400652 <+59>:    pop    %rbp
   0x0000000000400653 <+60>:    retq
End of assembler dump.
(gdb) b *main+49
Breakpoint 1 at 0x400648
(gdb) r
Starting program: /home/m/git/a.out

Program received signal SIGSEGV, Segmentation fault.
_dl_map_object_from_fd (name=name@entry=0x4006e8 "/home/m/git/_uftrace/libmcount/libmcount-dynamic.so", origname=origname@entry=0x0,
    fd=<optimized out>, fbp=fbp@entry=0x7fffffffdba0, realname=<optimized out>, loader=loader@entry=0x0, l_type=2, mode=268435458,
    stack_endp=0x7fffffffdb98, nsid=0) at dl-load.c:1337
1337    dl-load.c: No such file or directory.
(gdb) bt
#0  _dl_map_object_from_fd (name=name@entry=0x4006e8 "/home/m/git/_uftrace/libmcount/libmcount-dynamic.so", origname=origname@entry=0x0,
    fd=<optimized out>, fbp=fbp@entry=0x7fffffffdba0, realname=<optimized out>, loader=loader@entry=0x0, l_type=2, mode=268435458,
    stack_endp=0x7fffffffdb98, nsid=0) at dl-load.c:1337
#1  0x00007ffff7ddfc27 in _dl_map_object (loader=0x0, loader@entry=0x7ffff7ffe168,
    name=name@entry=0x4006e8 "/home/m/git/_uftrace/libmcount/libmcount-dynamic.so", type=type@entry=2, trace_mode=trace_mode@entry=0,
    mode=mode@entry=268435458, nsid=<optimized out>) at dl-load.c:2498
#2  0x00007ffff7dec577 in dl_open_worker (a=a@entry=0x7fffffffe160) at dl-open.c:237
#3  0x00007ffff7de7564 in _dl_catch_error (objname=objname@entry=0x7fffffffe150, errstring=errstring@entry=0x7fffffffe158,
    mallocedp=mallocedp@entry=0x7fffffffe14f, operate=operate@entry=0x7ffff7dec4d0 <dl_open_worker>, args=args@entry=0x7fffffffe160)
    at dl-error.c:187
#4  0x00007ffff7debda9 in _dl_open (file=0x4006e8 "/home/m/git/_uftrace/libmcount/libmcount-dynamic.so", mode=2,
    caller_dlopen=0x40062f <main+24>, nsid=-2, argc=<optimized out>, argv=<optimized out>, env=0x7fffffffe4a8) at dl-open.c:660
#5  0x00007ffff794c5ad in do_dlopen (ptr=ptr@entry=0x7fffffffe380) at dl-libc.c:87
#6  0x00007ffff7de7564 in _dl_catch_error (objname=0x7fffffffe370, errstring=0x7fffffffe378, mallocedp=0x7fffffffe36f,
    operate=0x7ffff794c570 <do_dlopen>, args=0x7fffffffe380) at dl-error.c:187
#7  0x00007ffff794c664 in dlerror_run (args=0x7fffffffe380, operate=0x7ffff794c570 <do_dlopen>) at dl-libc.c:46
#8  __GI___libc_dlopen_mode (name=<optimized out>, mode=<optimized out>) at dl-libc.c:163
#9  0x000000000040062f in main ()
```

libmcount/trigger.c
ParkHanbum: yes, this logic is not enough as you say.

libmcount/trigger.c
ParkHanbum: if we can assume that uftrace has been installed, we can get path of libmcount-dynamic.so. otherwise, libtrigger.so has the path as default from attached process located at. so, we can get the absolute path of libmcount-dynamic.so.

libmcount/trigger.c
ParkHanbum: sure, I'll update this.

utils/inject.c
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * args:
+ * - pid_t pid: pid of process to inspect
+ *
+ * returns:
+ * - a long containing the address of an executable region of memory inside the
+ *   specified process' address space.
+ *
+ */
+long freespaceaddr(pid_t pid)
```
namhyung: Please follow the rules in the other function comments.  It looks like:

```
/**
 * free_space_addr() - find executable region of memory
 * @pid: pid of process to inspect
 *
 * Search the target process' /proc/pid/maps entry and find an executable
 * region of memory that we can use to run code in.
 * This function returns the address of an executable region of memory
 * inside the specified process' address space.
 */

File: utils/inject.c [hunk 14]
```diff
@@ -0,0 +1,172 @@
...
+{
+	FILE *fp;
+	char filename[30];
+	char line[850];
+	long addr;
+	char str[20];
+	char perms[5];
+
+	sprintf(filename, "/proc/%d/maps", pid);
```
namhyung: Please use `snprintf()` instead of `sprintf()`.

cmds/record.c
```diff
@@ -1511,8 +1511,8 @@ static void check_binary(struct opts *opts)
 	close(fd);
 }
 
-static bool check_linux_schedule_event(char *events,
-				       enum uftrace_pattern_type ptype)
+bool check_linux_schedule_event(char *events,
+		enum uftrace_pattern_type ptype)
```
namhyung: Please check the indentation.

ParkHanbum:
@namhyung I think dynamic tracing feature need more time to stable. until then, How about separating the dynamics from the old ones?

namhyung:
I'm ok with separating dynamic tracing feature for now.  And then we can just focus on the attaching itself.  I think you don't need to carry related patches.  Instead let's finish the design of `libtrigger.so` first.

ParkHanbum:
@namhyung "Let's finish the design of libtrigger.so.".  What does it mean? I think you are not talk about naming only.

ParkHanbum:
I have updated suggestions what talked from last discussion. 
thank you guys, for precious reviewing and teaching.
muchas gracias!

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+#include "utils.h"
+
+void ptrace_attach(pid_t target)
+{
+	int waitpidstatus;
+
+	if (ptrace(PTRACE_ATTACH, target, NULL, NULL) == -1)
+		pr_err("ptrace(PTRACE_ATTACH) failed");
+
+	if (waitpid(target, &waitpidstatus, WUNTRACED) != target)
```
namhyung: I'm not sure `WUNTRACED` flag is needed.  Also it seems you need to check the status (`WSTOPSIG(status) == SIGSTOP`).

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+
+void ptrace_getregs(pid_t target, struct REG_TYPE *regs)
+{
+	if (ptrace(PTRACE_GETREGS, target, NULL, regs) == -1)
+		pr_err("ptrace(PTRACE_GETREGS) failed");
+}
+
+void ptrace_cont(pid_t target)
+{
+	struct timespec *sleeptime = malloc(sizeof(struct timespec));
```
namhyung: Looks like a memory leak.  Why not using a local variable?

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+
+	sleeptime->tv_sec = 1;
+	sleeptime->tv_nsec = 0000000;
+
+	if (ptrace(PTRACE_CONT, target, NULL, NULL) == -1)
+		pr_err("ptrace(PTRACE_CONT) failed");
+	nanosleep(sleeptime, NULL);
```
namhyung: Why is this needed?  I think we can drop it and split the waiting part below..

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+	if (ptrace(PTRACE_GETREGS, target, NULL, regs) == -1)
+		pr_err("ptrace(PTRACE_GETREGS) failed");
+}
+
+void ptrace_cont(pid_t target)
+{
+	struct timespec *sleeptime = malloc(sizeof(struct timespec));
+
+	sleeptime->tv_sec = 1;
+	sleeptime->tv_nsec = 0000000;
```
namhyung: Single 0 will be enough.

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+	checktargetsig(target);
+}
+
+void ptrace_setregs(pid_t target, struct REG_TYPE *regs)
+{
+	if (ptrace(PTRACE_SETREGS, target, NULL, regs) == -1)
+		pr_err("ptrace(PTRACE_SETREGS) failed");
+}
+
+siginfo_t ptrace_getsiginfo(pid_t target)
```
namhyung: Hmm.. It'd be nice if you can pass it as an argument IMO.

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+{
+	int bytesRead = 0;
+	int i = 0;
+	long word = 0;
+	long *ptr = (long *) vptr;
+
+	while (bytesRead < len) {
+		word = ptrace(PTRACE_PEEKTEXT, pid, addr + bytesRead, NULL);
+		if (word == -1) {
+			pr_err("ptrace(PTRACE_PEEKTEXT) failed %lx", addr + bytesRead, bytesRead);
```
namhyung: Please wrap long lines.

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+ * Use ptrace() to write to the target process' address space.
+ *
+ * args:
+ * - int pid: pid of the target process
+ * - unsigned long addr: the address to start writing to
+ * - void *vptr: a pointer to a buffer containing the data to be written to the
+ *   target's address space
+ * - int len: the amount of data to write to the target
+ */
+void ptrace_write(int pid, unsigned long addr, void *vptr, int len)
```
namhyung: Please follow the style of function comments.

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+	while (byteCount < len) {
+		memcpy(&word, vptr + byteCount, sizeof(word));
+		word = ptrace(PTRACE_POKETEXT, pid, addr + byteCount, word);
+		if (word == -1)
+			pr_err("ptrace(PTRACE_POKETEXT) failed");
+		byteCount += sizeof(word);
+	}
+}
+
+void checktargetsig(int pid)
```
namhyung: `check_target_sig()` ?

utils/ptrace.c
```diff
@@ -0,0 +1,129 @@
...
+{
+	// check the signal that the child stopped with.
+	siginfo_t targetsig = ptrace_getsiginfo(pid);
+
+	// if it wasn't SIGTRAP, then something bad happened (most likely a
+	// segfault).
+	if (targetsig.si_signo != SIGTRAP) {
+		pr_dbg("instead of expected SIGTRAP, target stopped with signal %d: %s\n", targetsig.si_signo, strsignal(targetsig.si_signo));
```
namhyung: Please wrap long lines.

utils/ptrace.c
namhyung: And I think you'd be better stopping other threads first.  As I read the ptrace manual, after `PTRACE_ATTACH` you'll need to inject SIGSTOP with `PTRACE_CONT`.

arch/x86_64/inject.c
```diff
@@ -0,0 +1,205 @@
...
+	long target_dlopen_addr, addr;
+	unsigned long long target_buf, lib_addr;
+	char *newcode, *backup;
+	char *lib_path = realpath(libname, NULL);
+	pid_t target = 0;
+
+	// figure out the size of contents to be injected.
+	// so, we know how big of a buffer to allocate.
+	uint32_t inject_content_size = (uintptr_t)&inject_content_end - (uintptr_t)inject_content;
+	size_t inject_shared_lib_size = inject_content_size;
```
namhyung: No need to keep 2 variables for a same value IMO.  Also how about using a shorter name like "inject_size"?

arch/x86_64/inject.c
```diff
@@ -0,0 +1,205 @@
...
+	ptrace_getregs(target, &oldregs);
+	memcpy(&regs, &oldregs, sizeof(struct user_regs_struct));
+
+	/*
+	 * since ELF align each section size by paging size, many section
+	 * have padding at its tail. code be load shared object is
+	 * injected to here. and substract word size because ptrace read
+	 * word size at once.
+	 */
+	addr = get_inject_code_addr(target) - inject_content_size - sizeof(long);
```
namhyung: I don't understand why you subtracted a word size.  But it'd be better align it to a word size though.

arch/x86_64/inject-content.S
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	mov %rcx,%rdi
+	// now call r9; malloc()
+	callq *%r9
+	// after returning from malloc(),
+	// pop the previous value of r9 off the stack
+	pop %r9
+	// break in so that we can see what malloc() returned
+	int $3
+	// get the address of dlopen() off of
```
namhyung: Please add a blank line and some more comments so that it can be noticed it's executed in a separate phase.

arch/x86_64/inject-content.S
```diff
@@ -0,0 +1,82 @@
...
+	// filename = the address of the buffer returned by malloc()
+	mov %rax,%rdi
+	// 2nd argument to dlopen(): flag = RTLD_LAZY
+	movabs $1,%rsi
+	// call dlopen()
+	callq *%r9
+	// restore old r9 value
+	pop %r9
+	// break in so that we can see what dlopen() returned
+	int $3
```
namhyung: Ditto.

arch/x86_64/inject-content.S
```diff
@@ -0,0 +1,82 @@
...
+	// save previous rbx value
+	push %rbx
+	// load the address of free() into rbx
+	movq %rsi,%rbx
+	// zero out rsi, because free() might think that it contains
+	// something that should be freed
+	xor %rsi,%rsi
+	// break in so that we can check out the arguments right
+	// before making the call
+	int $3
```
namhyung: Same here.

arch/x86_64/inject.c
```diff
@@ -0,0 +1,205 @@
...
+
+	// at this point, the target should have run malloc().
+	// check its return value to see if it succeeded, and
+	// bail out cleanly if it didn't.
+	memset(&malloc_regs, 0, sizeof(struct user_regs_struct));
+	ptrace_getregs(target, &malloc_regs);
+	target_buf = malloc_regs.rax;
+
+	if (target_buf == 0) {
+		restore_state_and_detach(target, addr, backup, inject_shared_lib_size, oldregs);
```
namhyung: Please wrap long lines.

arch/x86_64/inject-content.S
honggyukim: And please use `/* */` style comment instead of `//` to make the code look same as others.

utils/ptrace.c
ParkHanbum: @namhyung  what is mean IMO?

namhyung: It's an acronym for 'In My Opinion'..

utils/ptrace.c
ParkHanbum: @namhyung PTRACE_ATTACH sends SIGSTOP to target. target process and its childs will stop when it have recieved SIGSTOP. so, I think this code reasonable.

namhyung: Yes `PTRACE_ATTACH` will send it but after signal-delivery-stop you didn't inject it to the target process.  Please read "Attaching and detaching" section in the ptrace man page.

namhyung:
I tested signal-delivery-stop after `PTRACE_ATTACH` with below code.  It seems you need to inject `SIGSTOP` manually if you want to stop all threads in a process.

```c
$ cat two-thread.c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <pthread.h>

volatile int n;

void *thread_fn(void *arg)
{
  while (1) {
    n++;
    sleep(1);
  }
  return NULL;
}

{
  pthread_t thid;

  printf("pid: %d\n", getpid());
  printf("addr: %p\n", &n);

  pthread_create(&thid, NULL, thread_fn, NULL);
  while (1) {
    n++;
    sleep(1);
  }
  pthread_join(thid, NULL);
  return 0;
}
```

The variable `n` is increased by two threads.  If we stop the two threads at once, the value should not change.  And following code is to attach to the above process.

```c
$ cat attach.c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/wait.h>
#include <sys/ptrace.h>
#include <sys/syscall.h>

{
  int pid = 0;
  unsigned long addr = 0;
  long data;
  int stop = 0;

  if (argc > 1)
    pid = atoi(argv[1]);
  if (argc > 2)
    addr = strtoul(argv[2], NULL, 16);
  if (argc > 3)
    stop = atoi(argv[3]);

  ptrace(PTRACE_ATTACH, pid, 0, 0);
  waitpid(pid, NULL, 0);

  if (stop) {
    ptrace(PTRACE_CONT, pid, 0, SIGSTOP);
    waitpid(pid, NULL, 0);
  }

  data = ptrace(PTRACE_PEEKDATA, pid, addr, 0);
  printf("data = %ld\n", data);

  sleep(2);

  data = ptrace(PTRACE_PEEKDATA, pid, addr, 0);
  printf("data = %ld\n", data);

  if (stop)
    syscall(SYS_tgkill, pid, pid, SIGCONT);

  ptrace(PTRACE_DETACH, pid, 0, 0);
  return 0;
}
```

The third argument is whether it should inject SIGSTOP or not.  When I tested it prints a different value after sleeping 2 seconds when not injecting the signal.

```
$ gcc -o two-thread two-thread.c -pthread
$ gcc -o attach attach.c

$ ./two-thread
pid: 2539
addr: 0x5569f4e0a04c

# in another terminal
$ ./attach 2539 0x5569f4e0a04c
data = 30
data = 32

$ ./attach 2539 0x5569f4e0a04c 1
data = 44
data = 44
```

arch/x86_64/inject-content.S
honggyukim: Ditto means he wants to repeat the same comment for this. Please see above. He's saying as follows: 
> "Please add a blank line and some more comments so that it can be noticed it's executed in a separate phase."

honggyukim: And I mentioned about "//" because it is not used in Linux kernel coding style.

ParkHanbum:
@namhyung I'm understand what you say now. you mean it is need to send SIGSTOP to target manually after doing ptrace cont because 'ptrace cont' release all child thread and make them could be execute.

namhyung:
Why do you want to use single-step?  It still allows other threads run (i.e. same effect with ptrace cont when it comes to the execution of other threads).  Note that ptrace attach is just for single thread.

Yes ptrace attach sends `SIGSTOP` but it's caught (before other threads see it) by wait() and ignored unless you pass it to the process with one of ptrace restart (cont, single-step, syscall, ...).

ParkHanbum:
problem we faced is, make only specified process to be execute. use ptrace cont with SIGSTOP will stop process not execute. we need to execute the code which injected by uftrace.

ParkHanbum:
@namhyung I think this PR is ready to merge!!

namhyung:
I don't think so, we have way many things to discuss.  Anyway I think we're not gonna have the 'dynamic' command - it's a kind of feature, not command.

ParkHanbum:
@namhyung 
I think it's a small part of this issue.
Tell me where you want this PR to be located.
and if you want discuss the part of technically, 
Tell me when you can do it.

arch/x86_64/inject.c
```diff
@@ -0,0 +1,305 @@
...
+	 * rsi, rdx, rcx, r8, and r9.
+	 */
+	regs.rdi = target_malloc_addr;
+	regs.rsi = target_free_addr;
+	regs.rdx = target_dlopen_addr;
+	regs.rcx = lib_path_length;
+	ptrace_setregs(target, &regs);
+
+	// back up whatever data used to be at the address we want to modify.
+	backup = malloc(inject_size * sizeof(char));
```
namhyung: `xmalloc()` please. and `* sizeof(char)` is not needed.

arch/x86_64/inject.c
```diff
@@ -0,0 +1,305 @@
...
+	ptrace_read(target, addr, backup, inject_size);
+
+	/*
+	 * set up a buffer to hold the code we're going to inject
+	 * into the target process.
+	 */
+	injected_area = malloc(inject_size * sizeof(char));
```
namhyung: ditto

arch/x86_64/inject.c
```diff
@@ -0,0 +1,305 @@
...
+	ptrace_write(target, addr, injected_area, inject_size);
+
+	/*
+	 * ptrace continue will make child thread get wake-up.
+	 * this can be harmful during injection.
+	 * therefore, make all child threads enter the loop
+	 * until injecting done.
+	 */
+	thread_count = get_tids(pid, &tids);
+	suspend_child_threads(target, addr, &tids, thread_count);
```
namhyung: Where do you release (or detach) the child threads?

ParkHanbum:
I'm trying to attaching node.js. here is log.
```
$ node index.js
server is listening on 3000
symbol: no symtab, using dynsyms instead
symbol: loading symbols from /usr/bin/node (offset: 0x5558c16e0000)
symbol: loaded 27412 symbols
symbol: removed 2800 duplicates
symbol: updating symbol name using dynamic symbols
mcount: update module id to 0x5558c3c42de0
mcount: mcount setup done
mcount: new session started: df703aaceadad30b: node
mcount: preparing shmem buffers: tid = 85180
mcount: [mod: 5558c3c42de0, idx: 785] enter 5558c1ae78e0: uv_tty_reset_mode
mcount: [mod: 5558c3c42de0, idx: 661] enter 5558c1ae7120: raise
```

ParkHanbum:
I'm trying to attaching the python also. version 2.7 here is the result.

```
python app.py
 * Serving Flask app "app" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://0.0.0.0:8000/ (Press CTRL+C to quit)
mcount: ignore invalid pipe fd: 4
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /usr/lib/locale/locale-archive: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /usr/lib/locale/locale-archive: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
dwarf: prepare debug info
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /usr/lib/locale/locale-archive: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
mcount: setup PLT hooking 
mcount: mcount setup done
mcount: new session started: 1e4f084b729e803a: python2.7
221.148.248.130 - - [03/Nov/2018 05:18:56] "GET / HTTP/1.1" 304 -
...
...
...
mcount: [mod: 563b2d3d7300, idx: 106] enter 563b2a91f3d0: free
filter: filter match: free
mcount: [mod: 563b2d3d7300, idx: 287] enter 563b2a91ff20: fflush
mcount: [mod: 563b2d3d7300, idx: 287] enter 563b2a91ff20: fflush
mcount: shmem_finish: tid: 87529 seqnum = 2 curr = 0, nr_buf = 2 max_buf = 2
mcount: releasing all shmem buffers for task 87529
mcount: destroy plthook special function index
symbol: unload symbol tables
mcount: exit from libmcount
```

and this is part of uftrace data.
```
   0.669 us [ 87547] | strlen("WSGIRequestHandler") = 18;
   0.166 us [ 87547] | __ctype_b_loc();
   0.389 us [ 87547] | strlen("environ") = 7;
   0.156 us [ 87547] | __ctype_b_loc();
   0.473 us [ 87547] | strlen("WSGIRequestHandler") = 18;
   0.160 us [ 87547] | __ctype_b_loc();
   0.347 us [ 87547] | strlen("environ") = 7;
   0.612 us [ 87547] | strchr("|O:str", ':') = ":str";
   0.176 us [ 87547] | __ctype_b_loc();
```
It seemed to be success to attaching. but not alwyas.

ParkHanbum:
@namhyung this is why it add 2 to rip. 
```
regs.rip = addr;

[injected address] 0x400fb0


instead of expected SIGTRAP, target stopped with signal 11: Segmentation fault


(gdb) x/10i 0x400fae
[0x7f602fd0173c] [130037]                                                     │=> 0x400fae:    add    %al,(%rax)
[0x7f602fd0173c] [150037]                                                     │   0x400fb0:    push   %rsi
[0x7f602fd0173c] [110037]                                                     │   0x400fb1:    push   %rdx
[0x7f602fd0173c] [140037]                                                     │   0x400fb2:    push   %r9
[0x7f602fd0173c] [160038]                                                     │   0x400fb4:    mov    %rdi,%r9
[0x7f602fd0173c] [110038]                                                     │   0x400fb7:    mov    %rcx,%rdi



regs.rip = addr + 1;

[injected address] 0x400fb0


instead of expected SIGTRAP, target stopped with signal 11: Segmentation fault


(gdb) x/10i 0x400faf
[0x7f33db36c73c] [140013]                                                     │=> 0x400faf:    add    %dl,0x52(%rsi)
[0x7f33db36c73c] [160013]                                                     │   0x400fb2:    push   %r9
[0x7f33db36c73c] [110014]                                                     │   0x400fb4:    mov    %rdi,%r9
[0x7f33db36c73c] [140014]                                                     │   0x400fb7:    mov    %rcx,%rdi
[0x7f33db36c73c] [150014]                                                     │   0x400fba:    callq  *%r9
[0x7f33db36c73c] [130014]                                                     │   0x400fbd:    pop    %r9
[0x7f33db36c73c] [120014]                                                     │   0x400fbf:    int3
[0x7f33db36c73c] [160014]                                                     │   0x400fc0:    pop    %rdx
[0x7f33db36c73c] [110015]                                                     │   0x400fc1:    push   %r9
[0x7f33db36c73c] [140015]                                                     │   0x400fc3:    mov    %rdx,%r9

```

ParkHanbum:
@namhyung @honggyukim @Taeung 
This PR has been updated. would you review this aaaagain? thanks.

ParkHanbum:
@namhyung did you see my request?

namhyung:
Sorry for being late, I'll take a look.

utils/env-file.c
```diff
@@ -0,0 +1,86 @@
...
+ * UFTRACE_PLTHOOK=1
+ */
+void set_env_to_file(int fd, char *key, const char *value)
+{
+	int res;
+	char buf[256];
+
+	sprintf(buf, "%s", key);
+	sprintf(buf + strlen(buf), "%s", "=");
+	sprintf(buf + strlen(buf), "%s\n", value);
```
namhyung: Please do not use `sprintf()` as it's know to vulnerable.  Also why not doing it at once with "%s=%s\n" ?

utils/env-file.c
```diff
@@ -0,0 +1,86 @@
...
+	bool flag = true;
+	int i, count;
+
+	count = read(fd, buf, 4096);
+	if (count < 0)
+		pr_err(ERROR_READ_FILE);
+	close(fd);
+
+	for (i = 0; buf[i] != '\0' ; i++) {
+		if (buf[i] == '\n') {
```
namhyung: Why not using `fgets()` + `sscanf()`?

utils/env-file.h
```diff
@@ -0,0 +1,29 @@
...
+"Error occurred while open file.\n"
+
+#define ERROR_READ_FILE \
+"Error occurred while read file.\n"
+
+#define ERROR_WHILE_WRITE \
+"Error occurred while write file.\n"				\
+"[Trouble shooting]\n"						\
+"Check a file named 'uftrace_environ_file' is existed "		\
+"under '/tmp'.\n"
```
namhyung: Please move these messages to .c file.

cmds/record.c
```diff
@@ -134,6 +134,8 @@ char * get_libmcount_path(struct opts *opts)
 	snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
 	if (access(lib, F_OK) != 0 && errno == ENOENT)
 		pr_warn("Didn't you run 'make install' ?\n");
+	else
+		return lib;
```
namhyung: Then you should return only if `access()` returns 0.

cmds/record.c
```diff
@@ -263,8 +263,10 @@ static void setup_child_environ(struct opts *opts, int argc, char *argv[])
 			setenv("UFTRACE_NEST_LIBCALL", "1", 1);
 	}
 
-	if (strcmp(opts->dirname, UFTRACE_DIR_NAME))
-		setenv("UFTRACE_DIR", opts->dirname, 1);
+	char *path = realpath(opts->dirname, NULL);
+	if (mode == METHOD_ATTACH && path == NULL)
```
namhyung: I don't like declaration in the middle of code.  Also what's the `METHOD_ATTACH`?  You need to free the path as well.

libmcount/mcount.c
```diff
@@ -1185,6 +1185,9 @@ int mcount_entry(unsigned long *parent_loc, unsigned long child,
 	struct mcount_ret_stack *rstack;
 	struct uftrace_trigger tr;
 
+	// store errno from target context.
+	int _saved_errno = errno;
+
```
namhyung: Is it always needed?  Or just for init time?  Also I think the comment is meaningless - it only says what it does not why.  Please just remove them.

libmcount/mcount.h
```diff
@@ -74,6 +74,14 @@ enum shmem_buffer_flags {
...
 };
 
+#define SAVE_ERRNO(block)			\
+do {						\
+	int _saved_errno = errno;		\
+	do { block; } while (0);		\
+	errno = _saved_errno;			\
+}						\
+while (0);
+
```
namhyung: Do you consider using this macro in other place?  If not, I think open coding is better.

libmcount/mcount.c
```diff
@@ -1849,6 +1850,14 @@ UFTRACE_ALIAS(__cyg_profile_func_exit);
 static void __attribute__((constructor))
 mcount_init(void)
 {
+	int env_fd;
+	char *has_environ = getenv("UFTRACE_ENVIRON");
+
+	if (!has_environ) {
+		env_fd = open_env_file();
+		set_env_from_file(env_fd);
+	}
```
namhyung: I think `errno` can be changed here already.

libmcount/mcount.h
```diff
@@ -74,6 +74,11 @@ enum shmem_buffer_flags {
 	SHMEM_FL_RECORDING	= (1U << 2),
 };
 
+/* the methods used to tracing */
+#define METHOD_PRELOAD			1
+#define METHOD_ATTACH			2
+
```
namhyung: Why do you define them here?   Also please do not add blank lines unnecessarily.

uftrace.c
```diff
@@ -815,15 +823,12 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
 			argp_usage(state);
 
 		if (opts->exename == NULL) {
-			switch (opts->mode) {
-			case UFTRACE_MODE_RECORD:
-			case UFTRACE_MODE_LIVE:
+			if (opts->mode == UFTRACE_MODE_LIVE
+			    || (opts->mode == UFTRACE_MODE_RECORD
+				&& opts->pid <= 0)) {
```
namhyung: The operator should be put on the above line like `opts->mode == UFTRACE_MODE_RECORD &&`.  Also why not keep switch statement and just adding the pid check?

uftrace.c
```diff
@@ -815,15 +823,12 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
...
 				argp_usage(state);
-				break;
-			default:
-				/* will be set after read_ftrace_info() */
-				break;
 			}
+			break;
```
namhyung: seems not needed.

uftrace.c
```diff
@@ -985,6 +990,7 @@ int main(int argc, char *argv[])
 		.depth		= OPT_DEPTH_DEFAULT,
 		.max_stack	= OPT_RSTACK_DEFAULT,
 		.port		= UFTRACE_RECV_PORT,
+		.pid		= 0,
```
namhyung: You don't need to zero-initialize here - that's the thing the compiler will do for us.

cmds/record.c
```diff
@@ -2010,20 +2037,74 @@ int do_child_exec(int ready, struct opts *opts,
 	abort();
 }
 
+void do_inject(int ready, struct opts *opts, int argc, char** argv)
+{
+	int target_pid = opts->pid;
```
namhyung: No need to create a variable.

cmds/record.c
```diff
@@ -2010,20 +2037,74 @@ int do_child_exec(int ready, struct opts *opts,
...
+	char *libmcount_path;
+
+	setup_child_environ(opts, argc, argv);
+	libmcount_path = get_libmcount_path(opts);
+
+	if (libmcount_path == NULL)
+		pr_err("connot found shared object from %s", libmcount_path);
```
namhyung: typo

cmds/record.c
```diff
@@ -2010,20 +2037,74 @@ int do_child_exec(int ready, struct opts *opts,
...
+
+	inject(libmcount_path, target_pid);
+	sleep(2);
```
namhyung: Could you please add some comment why it's needed?

cmds/record.c
```diff
@@ -28,9 +28,12 @@
 #include "utils/filter.h"
 #include "utils/kernel.h"
 #include "utils/perf.h"
+#include "utils/env-file.h"
 
 #define SHMEM_NAME_SIZE (64 - (int)sizeof(struct list_head))
 
+extern int inject(char *libname, pid_t pid);
```
namhyung: Why not adding `#include "utils/inject.h"`?

namhyung:
@ParkHanbum it's hard to review the whole patchset at once.  It'd be much nicer if you could reply each comment that I left above if possible.

ParkHanbum:
@namhyung sorry, it because I refactoring whole of codes to according your advise. I will comment to you when its ready.

ParkHanbum:
@namhyung change summary
- removed codes to keep errno
- removed codes to change thread model to initial-exec

honggyukim:
Hmm.. I tested it to attach already running `top` command, but it doesn't work as follows:
```
$ uftrace record -la --pid 8485
attach: /mnt/data1/home3/honggyu/work/uftrace/git/uftrace/utils/inject.c:209:inject
 ERROR: could not found shared object from libmcount-fast.so: No such file or directory
```

ParkHanbum:
@honggyukim checked. The cause of the problem is that INSTALL_LIB_PATH is defined only at run "make install". so you can test it if you run "make install" only. as I wrote to commit, it is requirement that the full path of libmcount when using attaching.

ParkHanbum:
@honggyukim note that #702 must be applied.

honggyukim:
Hmm.. It doesn't work as well. I tested it to attach already running `top` process as follows:
```
$ sudo uftrace record -la -vvv --logfile=attach.log --pid 13384
(... got stuck ...)
```
Right after running the command above, it got stuck and doesn't show anything. In the other terminal that is running `top` command, it shows the following error message.
```
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: Permission denied
plthook: destroy plthook special function index
symbol: unload symbol tables
mcount: exit from libmcount
```
So I just pressed Ctrl-C to stop recording and it shows as follows:
```
$ sudo uftrace record -la -vvv --logfile=attach.log --pid 13384
(... got stuck ...)
^Cchild terminated by signal: 6: Aborted
```
Here is the log message in `attach.log`.
```
uftrace: /usr/bin/top
uftrace: checking binary /usr/bin/top
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
attach: [66e048 - 66e114] size of code to inject 205
attach: Library Path : /home/honggyu/usr/lib/libmcount.so
uftrace: creating 10 thread(s) for recording
uftrace: start writer thread 0
uftrace: start writer thread 2
uftrace: start writer thread 1
uftrace: start writer thread 3
uftrace: start writer thread 4
uftrace: start writer thread 5
uftrace: start writer thread 6
uftrace: start writer thread 7
uftrace: start writer thread 8
uftrace: start writer thread 9
uftrace: stop writer thread 6
uftrace: fill header (metadata) info in uftrace.data/info
uftrace: stop writer thread 1
uftrace: stop writer thread 8
uftrace: stop writer thread 9
uftrace: stop writer thread 5
uftrace: stop writer thread 7
uftrace: stop writer thread 4
uftrace: stop writer thread 3
uftrace: stop writer thread 0
uftrace: stop writer thread 2
uftrace: /home/honggyu/work/uftrace/cmds/record.c:1443:save_session_symbols
 ERROR: cannot find map files: No such file or directory
```
@ParkHanbum Thanks for your effort on this challenging work, but please take a look at it again.

ParkHanbum:
@honggyukim please, do not use sudo when using attaching. if you run uftrace with root permission then target program could not access to uftrace data directory that created from uftrace because uftrace data directory require root permission.

honggyukim:
I tested it again without `sudo`, but it shows that I cannot attach to the process as follows:
```
$ uftrace record -la -vvv --pid 26444
uftrace: /usr/bin/top
uftrace: checking binary /usr/bin/top
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
attach: [66e048 - 66e114] size of code to inject 205
attach: Library Path : /home/honggyu/usr/lib/libmcount.so
attach: /home/honggyu/work/uftrace/git/uftrace/new/go/utils/ptrace.c:18:ptrace_attach
 ERROR: ptrace(PTRACE_ATTACH) failed: Operation not permitted
uftrace: skipping perf event due to error: No such process
uftrace: creating 10 thread(s) for recording
uftrace: start writer thread 1
uftrace: start writer thread 3
uftrace: start writer thread 0
uftrace: start writer thread 5
uftrace: start writer thread 2
uftrace: start writer thread 6
uftrace: start writer thread 4
uftrace: start writer thread 7
uftrace: start writer thread 8
uftrace: start writer thread 9
```

honggyukim:
Could you show me how to test it with a simple sequence of commands?

ParkHanbum:
@honggyukim for now, please refer top article of this issue.

```
# another console
$ ps -ef | grep node
m        123218   1540  0 19:56 pts/1    00:00:00 ./node_g simple_server.js
m        123250   1577  0 19:57 pts/2    00:00:00 grep --color=auto node

# for use ptrace
$ echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope

# without force option, it will be cause the error about mcount symbol.
$ uftrace record -v --pid=123218 --force
uftrace: /home/m/git/node/out/Debug/node
uftrace: skipping perf event due to error: Permission denied
uftrace: please check /proc/sys/kernel/perf_event_paranoid
uftrace: using /usr/local/lib/libmcount.so library for tracing
uftrace: size of code to inject 209
uftrace: Path of Shared object to be injected : /usr/local/lib/libmcount.so
uftrace: Return address : 7f41f419a94e
uftrace: RIP Register : 308cf2a
```

honggyukim:
Wow! I test it for a simple `top` command and works fine. But I pressed Ctrl-C in uftrace side, but it also finishes attached `top` program as well. Is it possible to detach it as well?

ParkHanbum:
@honggyukim for now, only attaching was implemented. to implement detach we need channel to communicate between uftrace and libmcount.

honggyukim:
Hmm.. I tested it to `node`, but it gets crashed right after attaching.
**in tty 1**
```
$ node
> require('process').pid
28606

> (... waiting for input ...)
```
**in tty 2**
```
$ uftrace record --force -v --pid 28606
```
If both are ready, then uftrace attach command runs in tty2.  However, it gets crashed right after attaching it as follows.
**in tty 2**
```
$ uftrace record --force -v --pid 28606
uftrace: /home/honggyu/usr/bin/node
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
uftrace: creating 10 thread(s) for recording
child terminated by signal: 6: Aborted
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-12b15a0abdeb2069-28606-000
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Bad file descriptor
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Bad file descriptor
symbol: ELF error when loading symbols: invalid `Elf' handle
```
**in tty 1**
```
$ node
> require('process').pid
28606

> mcount: initializing mcount library
demangle: demangle failed: _ZTCN2v88internal8compiler12_GLOBAL__N_113TurboJsonFileE0_St14basic_ofstreamIcSt11char_traitsIcEE
demangle: demangle failed: _ZTCN2v88internal8compiler12_GLOBAL__N_112TurboCfgFileE0_St14basic_ofstreamIcSt11char_traitsIcEE
demangle: demangle failed: _ZTVN10__cxxabiv121__vmi_class_type_infoE@@CXXABI_1.3
demangle: demangle failed: _ZTVNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEEE@@GLIBCXX_3.4.21
demangle: demangle failed: _ZTVN10__cxxabiv117__class_type_infoE@@CXXABI_1.3
demangle: demangle failed: _ZTVNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEE@@GLIBCXX_3.4.21
demangle: demangle failed: _ZTVNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEE@@GLIBCXX_3.4.21
demangle: demangle failed: _ZTTNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEE@@GLIBCXX_3.4.21
demangle: demangle failed: _ZTVN10__cxxabiv120__si_class_type_infoE@@CXXABI_1.3
demangle: demangle failed: _ZTTNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEE@@GLIBCXX_3.4.21
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
symbol: ELF error when loading symbols: invalid `Elf' handle
dwarf: prepare debug info
symbol: ELF error when loading symbols: invalid `Elf' handle
symbol: error during open symbol file: /etc/ld.so.cache: Success
plthook: setup PLT hooking "/home/honggyu/usr/bin/node"
mcount: mcount setup done
mcount: new session started: 12b15a0abdeb2069: node
process crashed by signal 6: Aborted (si_code: -6)
Backtrace from uftrace:
=====================================
[0] (abort[862d10] <= uv__io_poll[1602238])
Aborted (core dumped)
```

namhyung:
My personal opinion on this is that it's a kind of future work since most use-cases will require (full) dynamic tracing anyway.  Could we focus on finishing it first and then comeback to this?

ParkHanbum:
@honggyukim  you got a right way. it is need #704 to attaching node. but it is work fine with python without #704 .

ParkHanbum:
this PR has been refreshed.

ParkHanbum:
statistic:
test with simple http server.
nodejs : success
pytohn : success
rust : success
go : particle success

daemon
apache : failed 
[[Thu Dec 05 22:08:04.646853 2019] [core:notice] [pid 28921] AH00060: seg fault or similar nasty error detected in the parent process

etc
tomcat : success

bernhardkaindl:
@ParkHanbum Is it possible to re-start this feature on attaching to running process, but now using full dynamic tracing?

Thanks, Bernhard

namhyung:
@ParkHanbum are you talking about #1643?  I'm not sure but I don't think it would help avoiding any existing bugs.  Doing the work at runtime would make thing more complicated rather than making life easier. :)

senderista:
This feature would change my life. I cannot find any good alternative to `uftrace` but I need to attach to running processes.

Issue #353, opened 2018-05-06
data option need more precision.


ParkHanbum:
when user specify the data directory with --data option, if path as given argument with --data option is exist then it will be copied to same path with '.old' prefix. but sometime it is not enough. if there is some file or path exist which could not handle it because permission, hidden, and so on, we just move files which we can move it. another files are gone. 

I am a victim for this issue. my '.git' was removed and blowing up stuff that I did not push yet. anyway, I have found this issue.

namhyung:
I don't understand what's the problem exactly.  Why was your .git directory affected?  Could you please provide a solution then?

ParkHanbum:
ok, I'll tell you about my mistake.
I had want to save the uftrace data at xxx directory under uftrace. but I had mistake. correctly commands is,

```
./uftrace --data=`pwd`/xxx
```
but I had run wrong commands. like this,
```
./uftrace --data=`pwd`
```
so, my uftrace directory that currently I located at was moved to same path with adding 'old' suffix. but the '.git' directory was not moved.

honggyukim:
Oh.. It could happen. So if someone gives the current working directory to `--data`, such embarrassing situation can be happened.

honggyukim:
I think we have to restrict this with the following options:

1. first check if the target directory is same as `pwd`. If so, do not proceed.
2. check if the target directory is empty, otherwise have to check if the directory is uftrace recorded directory. If not, do not proceed.

For 2, we can check if `task.txt` or `info` file exists or not but not the perfect solution.

namhyung:
I think we should prevent using a directory which is an ancestor (including itself) of the cwd.

namhyung:
And for your information, you almost always don't need to use `pwd` part as `$PWD/xxx` is same as just `xxx`.

namhyung:
I'm seeing repeated `"WARN: creating directory failed: File exists"` messages during test (maybe from test 151).  Could you please take a look?

PR #363, opened 2018-05-19
utils: Prevent selecting path where uftrace is located.

ParkHanbum:
uftrace can not guarantee that uftrace will run correctly if user chooses
a path where the uftrace is located to store data as argument. For that
reason, must prevent that choose a path where the uftrace is located.

Signed-off-by: HanbumPark <kese111@gmail.com>


ParkHanbum:
Associated issue : #353

honggyukim:
It seems that it only compares if the target directory is cwd or not.  Please consider @namhyung's comment. https://github.com/namhyung/uftrace/issues/353#issuecomment-387640576

ParkHanbum:
@honggyukim right, I'll add more validation logic.

honggyukim:
In this case, you can force updated it to the same pull request. I think you better reopen and update it here.

namhyung:
Agreed with @honggyukim , please reuse this PR number.

ParkHanbum:
the path of uftrace binary be changed after install it. in this case, current implementation haven't meaning. maybe make it can move all of file in previous path is more simple solution. how do you think?
```
m@m:~/git/prevent$ `which uftrace` record -v --data=`pwd`/xxx tests/t-abc
uftrace: dirname : /home/m/git/prevent/xxx, /usr/local/bin/uftrace, /home/m/git/prevent/xxxuftrace: /home/m/git/prevent/utils/utils.c:222:create_directory
 ERROR:  DO NOT REACH HERE!: No such file or directory
m@m:~/git/prevent$
```

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
 	return ret;
 }
 
+#define PATH_SEPARATOR '/'
+
+int check_bin_under_path(char *bin_path, char *path)
```
namhyung: Please specify what it returns..

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
+                if (bin_path[index] != path[index])
+                        return -1;
+                index++;
+        }
+
+        if (bin_path[index] == PATH_SEPARATOR || bin_path[index] == '\00')
+                return 0;
+
+	return -1;
+}
```
namhyung: I'm not sure checking the path of uftrace is meaningful here.  It'll usually be in a read-only directory so I don't think we need to check it.  But I think we should check the given (data) directory is an ancestor of the **current** directory.

honggyukim: @ParkHanbum We should consider its ancestor.  It will make a problem if you record a target program with uftrace as follows:
```
$ pwd
/home/honggyu/work/myprog
$ gcc -pg hello.c
```
If its ancestor directory is given with `-d` option, it looks the record is done without a problem.
```
$ uftrace record -d /home/honggyu/work a.out
Hello, World!
$ pwd
/home/honggyu/work/myprog
```
But if you go back to the parent directory, then you will see the directory name is change from `work` to `work.old`.
```
$ cd ..
$ pwd
/home/honggyu/work.old
```
We're thinking that it's the problematic situation, so please consider this as well.

honggyukim: And previously existing `work.old` directory contains full of uftrace generated files, which looks very confusing to users.

namhyung: Now I think that we need to check it's actually a uftrace data directory first.

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
 }
 
+#define PATH_SEPARATOR '/'
+#define DATAPATH_ROOT -2
+#define NOT_RELAVANT -1
+#define PATH_UNDER_DATAPATH 0
+
+int check_path_under_datapath(char *path, char *data_path)
+{
+        int data_path_leng = strlen(data_path);
```
namhyung: I prefer shorter name like just 'len'.

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
+        int index = 0;
+
+        if (data_path_leng == 1 && data_path[0] == PATH_SEPARATOR)
+		return DATAPATH_ROOT;
```
namhyung: Why did you add this special case?

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
 	return ret;
 }
 
+#define PATH_SEPARATOR '/'
+#define DATAPATH_ROOT -2
+#define NOT_RELAVANT -1
+#define PATH_UNDER_DATAPATH 0
+
+int check_path_under_datapath(char *path, char *data_path)
```
namhyung: Why not calling it `is_ancestor()`?

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
+        while (index < data_path_leng) {
+                if (path[index] != data_path[index])
+			return NOT_RELAVANT;
+                index++;
+        }
+
+        if (path[index] == PATH_SEPARATOR || path[index] == '\00')
+                return PATH_UNDER_DATAPATH;
+
+	return NOT_RELAVANT;
```
namhyung: Hmm.. isn't it same as below?

```c
len = strlen(data_path);
if (strncmp(path, data_path, len)
    return false;

return (path[len] == '/' || path[len] == '\0');
```

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
 	char *oldname = NULL;
+	char *cwd_path = NULL;
+	char resolved_data_path[PATH_MAX], current_working_dir[PATH_MAX];
 
 	xasprintf(&oldname, "%s.old", dirname);
+	realpath(dirname, resolved_data_path);
+	cwd_path = getcwd(current_working_dir, PATH_MAX);
+
+	if (!check_path_under_datapath(cwd_path, resolved_data_path)) {
+		pr_err("Do not select the path which uftrace located or its parent path to save uftrace data.");
```
namhyung: Some people hate long lines, please add a newline :)

utils/utils.c
```diff
@@ -776,4 +809,47 @@ TEST_CASE(utils_strv)
...
+{
+	char *cwd_path;
+	char *data_path;
+	int result;
+
+	cwd_path = "/home/usr/uftrace";
+
+	data_path = "/";
+	result = check_path_under_datapath(cwd_path, data_path);
+	TEST_EQ(result, -2);
```
namhyung: Why not making an oneliner?

```c
TEST_EQ(check_path_under_datapapth(cwd_path, "/"), -2);
```

utils/utils.c
```diff
@@ -178,12 +178,45 @@ int remove_directory(char *dirname)
...
 	xasprintf(&oldname, "%s.old", dirname);
+	realpath(dirname, resolved_data_path);
+	cwd_path = getcwd(current_working_dir, PATH_MAX);
+
+	if (!check_path_under_datapath(cwd_path, resolved_data_path)) {
+		pr_err("Do not select the path which uftrace located or its parent path to save uftrace data.");
+	}
 
 	if (!access(dirname, F_OK)) {
 		if (!access(oldname, F_OK) && remove_directory(oldname) < 0) {
```
namhyung: In addition, I think it's better checking the current and old directories are actually uftrace data directory (or empty directory).  You could check it has `info` and/or `task.txt` files at least.

namhyung:
Well, I don't want to keep the data unnecessarily.  If you really want to keep the data, you'd be better doing it by yourself.

ParkHanbum:
@namhyung OK then, let's finish this issue!

utils/utils.c
```diff
@@ -133,6 +133,27 @@ int writev_all(int fd, struct iovec *iov, int count)
...
+#define FTRACE_SIGNATURE "Ftrace!"
+bool is_uftrace_dir(char *path)
+{
+	int fd;
+	char info_path[PATH_MAX] = {0,};
+	char sig[8] = {0,};
+
+	// <uftrace data dir>/info always be exist.
+	strcat(info_path, path);
+	strcat(info_path, "/info");
```
namhyung: You can use `snprintf()` or `xasprintf()` for this.

utils/utils.c
```diff
@@ -133,6 +133,27 @@ int writev_all(int fd, struct iovec *iov, int count)
 	return 0;
 }
 
+#define FTRACE_SIGNATURE "Ftrace!"
```
namhyung: Why not using `UFTRACE_MAGIC_STR`?

utils/utils.c
```diff
@@ -148,6 +169,9 @@ int remove_directory(char *dirname)
 
 	pr_dbg("removing %s directory\n", dirname);
 
+	if(!is_uftrace_dir(dirname))
+		return -1;
```
namhyung: I think it's better moving this check out of the function.  Like this:

```c
  if (is_uftrace_dir(dirname))
     remove_directory(dirname);
```

utils/utils.c
```diff
@@ -133,6 +133,27 @@ int writev_all(int fd, struct iovec *iov, int count)
...
+	if (fd == -1) {
+		return false;
+	}
+
+	// read string signature "Ftrace!"
+	read(fd, sig, 8); 
+	close(fd);
+	return !strncmp(sig, "Ftrace!", 8);
+}
+
```
namhyung: Could you please split `is_uftrace_dir()` part into a separate commit?

utils/utils.c
```diff
@@ -178,12 +202,34 @@ int remove_directory(char *dirname)
...
+	return (path[len] == '/' || path[len] == '\0');
+}
+
 int create_directory(char *dirname)
 {
 	int ret = -1;
 	char *oldname = NULL;
+	char *cwd_path = NULL;
+	char resolved_data_path[PATH_MAX], current_working_dir[PATH_MAX];
```
namhyung: I know it's just preference, but I'd rather break it into two lines.

utils/utils.c
```diff
@@ -776,4 +822,38 @@ TEST_CASE(utils_strv)
 
 	return TEST_OK;
 }
+
+TEST_CASE(utils_check_datapath)
```
namhyung: Thanks for providing a unit test!

ParkHanbum:
@namhyung I was updated testcase for testing uftrace signature from info file.
could you review this again, plz?

namhyung:
Please update commit logs!

utils/utils.c
```diff
@@ -191,6 +191,25 @@ bool is_ancestor(char *path, char *data_path)
...
+	char sig[8] = {0,};
+
+	// <uftrace data dir>/info always be exist.
+	xasprintf(&info_path, "%s/info", path);
+	fd = open(info_path, O_RDONLY);
+	if (fd == -1) {
+		return false;
+	}
+
+	// read string signature "Ftrace!"
```
namhyung: I don't want to expose this detail in the comment or commit logs.  It can be changed anytime..

utils/utils.c
```diff
@@ -191,6 +191,25 @@ bool is_ancestor(char *path, char *data_path)
...
 
+bool is_uftrace_dir(char *path)
+{
+	int fd;
+	char *info_path = NULL;
+	char sig[8] = {0,};
+
+	// <uftrace data dir>/info always be exist.
+	xasprintf(&info_path, "%s/info", path);
+	fd = open(info_path, O_RDONLY);
```
namhyung: You'd better free `info_path` after using it.

utils/utils.c
```diff
@@ -178,12 +178,35 @@ int remove_directory(char *dirname)
...
+	char *cwd_path = NULL;
+	char resolved_data_path[PATH_MAX];
+	char current_working_dir[PATH_MAX];
 
 	xasprintf(&oldname, "%s.old", dirname);
+	realpath(dirname, resolved_data_path);
+	cwd_path = getcwd(current_working_dir, PATH_MAX);
+
+	if (is_ancestor(cwd_path, resolved_data_path)) {
+		pr_err("Do not select the path which uftrace located\n"\
```
namhyung: `pr_err` adds error string based on the value of `errno` at the end.  In this case it's not a system error, it's an user's fault.

And cwd is not necessarily a directory of uftrace source/binary.  It's just a current working directory and uftrace should not remove it.  (i.e. do not run `rm -rf .` or `rm -rf ..`)

utils/utils.c
```diff
@@ -209,14 +228,25 @@ int create_directory(char *dirname)
...
-		if (!access(oldname, F_OK) && remove_directory(oldname) < 0) {
+		if (!access(oldname, F_OK)) {
+			if (is_uftrace_dir(oldname)) {
+				if (remove_directory(oldname) < 0)
+					pr_warn("removing old directory "\
+							"failed: %m\n");
+			}
+			else {
+				pr_warn("%s is not a uftrace data path.\n");
+			}
```
namhyung: Hmm.. I don't like a deeply nested if statements.  What about something like this?

```c
if (access(OK) {
   if (!is_uftrace_dir)
      warn_and_out
   if (remove_directory < 0)
      warn_and_out
}
```

utils/utils.c
```diff
@@ -209,14 +228,25 @@ int create_directory(char *dirname)
...
 			pr_warn("removing old directory failed: %m\n");
 			goto out;
```
namhyung: isn't it a duplicate warn?

utils/utils.c
```diff
@@ -209,14 +228,25 @@ int create_directory(char *dirname)
...
 		}
 
-		if (rename(dirname, oldname) < 0) {
-			pr_warn("rename %s -> %s failed: %m\n", dirname, oldname);
-			goto out;
+		if (is_uftrace_dir(dirname)) {
+			if (rename(dirname, oldname) < 0) {
```
namhyung: these two if statements can be written in a single line.

utils/utils.c
```diff
@@ -833,4 +863,87 @@ TEST_CASE(utils_check_datapath)
...
+	if (fd < 0) {
+		pr_err("Error while trying to create file '%s' [%d] \n%m\n",
+				path, fd);
+	}
+	write(fd, text, sizeof(text));
+
+	return path;
+}
+
+TEST_CASE(utils_is_uftrace_dir)
```
namhyung: Well, I don't think it's needed testing with a fake directory.  Let' just get rid of the test case for this.

ParkHanbum:
@namhyung consider it done~!

ParkHanbum:
@namhyung sorry, but review this PR again please.

namhyung:
Again, please *remove* the signature details (= "Ftrace!") from the **commit message**!

PR #365, opened 2018-05-19
uftrace: Move cmd-*.c files to a new commands dir

honggyukim:
Since uftrace now has many commands, it'd be better to have a new
directory for them.

This patch adds a new directory "commands" and move command related
files to it.


honggyukim:
I've just changed `commands` dir to `cmds` as you mentioned. Thanks.

Issue #374, opened 2018-06-06
out of memory on an invalid task filter


ParkHanbum:
I have meet some trouble while dumping specific data from uftrace.data.

```
[179290.373666] Out of memory: Kill process 118807 (uftrace) score 912 or sacrifice child
[179290.374223] Killed process 118807 (uftrace) total-vm:7471228kB, anon-rss:5768024kB, file-rss:0kB, shmem-rss:0kB
```

what command I have used.
```
uftrace dump -d noin_noret --tid=cc1
```

and this is my machine specification. 
```
m@m:~$ inxi -Fx
System:    Host: m Kernel: 4.13.0-43-generic x86_64 (64 bit gcc: 5.4.0) Console: tty 21
           Distro: Ubuntu 16.04 xenial
Machine:   System: VMware product: VMware Virtual Platform
           Mobo: Intel model: 440BX Desktop Reference Platform Bios: Phoenix v: 6.00 date: 07/02/2015
CPU(s):    2 Single core Intel Core i3-7100Us (-HT-SMP-) cache: 6144 KB
           flags: (lm nx sse sse2 sse3 sse4_1 sse4_2 ssse3) bmips: 9600
           clock speeds: max: 2400 MHz 1: 2400 MHz 2: 2400 MHz
Graphics:  Card: VMware SVGA II Adapter bus-ID: 00:0f.0
           Display Server: N/A drivers: vmware (unloaded: fbdev,vesa)
           tty size: 174x45 Advanced Data: N/A out of X
Network:   Card: Intel 82545EM Gigabit Ethernet Controller (Copper)
           driver: e1000 v: 7.3.21-k8-NAPI port: 2000 bus-ID: 02:01.0
           IF: ens33 state: up speed: 1000 Mbps duplex: full mac: 00:0c:29:b0:04:48
Drives:    HDD Total Size: 42.9GB (74.4% used) ID-1: /dev/sda model: VMware_Virtual_S size: 42.9GB temp: 0C
Partition: ID-1: / size: 38G used: 28G (79%) fs: ext4 dev: /dev/sda1
           ID-2: swap-1 size: 2.14GB used: 0.40GB (18%) fs: swap dev: /dev/sda5
RAID:      No RAID devices: /proc/mdstat, md_mod kernel module present
Sensors:   System Temperatures: cpu: 100.0C mobo: N/A
           Fan Speeds (in rpm): cpu: N/A
[...]

PR #375, opened 2018-06-06
uftrace: Check if wrong tid value or not

Taeung:
If given wrong tid value with --tid option,
some problems can occur at setup_task_filter().
So check given tid value. Hanbum Park reported this
problem with the dump subcommand.

Fixes: #374

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -507,7 +507,10 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
 		break;
 
 	case OPT_tid_filter:
-		opts->tid = opt_add_string(opts->tid, arg);
+		if (strtol(arg, NULL, 0) <= 0)
+			pr_use("invalid thread id: %s\n", arg);
+		else
+			opts->tid = opt_add_string(opts->tid, arg);
 		break;
 
 	case OPT_no_merge:


Taeung:
In addition, I tested another case with two `--tid` option values:
```bash
$ uftrace info | grep task
# number of tasks     : 3
# task list           : 4577(bash), 4582(sudo), 4583(sudo)

$ uftrace dump --tid=4582 --tid=4583
uftrace file header: magic         = 4674726163652100
uftrace file header: version       = 4
uftrace file header: header size   = 40
uftrace file header: features      = 0x367 (PLTHOOK | TASK_SESSION | KERNEL | 
uftrace file header: info          = 0x1bff

reading 4577.dat
reading 4582.dat
44966.128202199   4582: [exit ] <41e2b0>(41e2b0) depth: 0
44966.128324317   4582: [entry] <41d890>(41d890) depth: 0
...
```
But I thought that "reading 4577.dat" shouldn't be needed
So how about the below patch  ?
```patch
diff --git a/cmds/dump.c b/cmds/dump.c
index e91417e..bb733b3 100644
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1202,6 +1202,9 @@ static void do_dump_file(struct uftrace_dump_ops *ops, struct opts *opts,
 
                 prev_time = 0;
 
+                if (task->done)
+                        continue;
+
                 ops->task_start(ops, task);
 
                 while (!read_task_ustack(handle, task) && !uftrace_done) {
```

Taeung:
And, I thought that if the user gives --tid option values, we should show the output related with the only tids.
So what about excluding reading `kernel-*.dat` and `perf-*.dat` on the below case ? @namhyung 
```bash
$ uftrace dump --tid=4582 --tid=4583 | grep reading
reading 4582.dat
reading 4583.dat
reading kernel-cpu0.dat
reading kernel-cpu1.dat
reading kernel-cpu2.dat
reading kernel-cpu3.dat
reading perf-cpu0.dat
reading perf-cpu1.dat
reading perf-cpu2.dat
reading perf-cpu3.dat

```

Taeung:
Yep, I got it !

PR #376, opened 2018-06-07
session: Fix missing kernel base addr problem

Taeung:
Based on current master branch,
I found this problem in the hello world case with `-k` option
like below:
```
  $ sudo uftrace -k ./hello
  hello
     1.107 us [ 25464] | __monstartup();
     0.994 us [ 25464] | __cxa_atexit();
              [ 25464] | main() {
              [ 25464] |   puts() {
     0.537 us [ 25464] |     <ffff81003860>();
     1.690 us [ 25464] |     <ffff81215420>();
     0.425 us [ 25464] |     <ffff81003620>();
     7.092 us [ 25464] |     <ffff810692b0>();
    16.544 us [ 25464] |   } /* puts */
  ...
```
This problem is related with The commit de7387f
("session: Don't create a map for main executable")

The commit excludes `[stack]` in a map, when skipping special
mappings like `[heap]`, however, `[stack]` is still needed
to get kernel base addr. If skipping `[stack]` mapping,
we aren't able to properly find kernel base addr at `set_kernel_base()`.
It is the reason that the above output cannot show kernel symbol name.

For information,
the second refactoring commit ("symbol: Refactor storing kernel base addr") is
related with d841004 by @iamwonseok

### DIFF
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -868,6 +868,8 @@ void record_proc_maps(char *dirname, const char *sess_id,
 	if (ofp == NULL)
 		pr_err("cannot open for writing maps file");
 
+	symtabs->kernel_base = -1ULL;
+
 	while (fgets(buf, sizeof(buf), ifp)) {
 		unsigned long start, end;
 		char prot[5];
@@ -885,8 +887,10 @@ void record_proc_maps(char *dirname, const char *sess_id,
 		 * but [stack] is still needed to get kernel base address.
 		 */
 		if (path[0] == '[') {
-			if (strncmp(path, "[stack", 6) == 0)
+			if (strncmp(path, "[stack", 6) == 0) {
+				symtabs->kernel_base = get_kernel_base(buf);
 				goto next;
+			}
 			else
 				continue;
 		}
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -880,9 +880,16 @@ void record_proc_maps(char *dirname, const char *sess_id,
 			   &start, &end, prot, path) != 4)
 			continue;
 
-		/* skip special mappings like [heap], [stack] */
-		if (path[0] == '[')
-			continue;
+		/*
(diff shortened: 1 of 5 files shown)


libmcount/record.c
```diff
@@ -880,8 +880,11 @@ void record_proc_maps(char *dirname, const char *sess_id,
 			   &start, &end, prot, path) != 4)
 			continue;
 
-		/* skip special mappings like [heap], [stack] */
-		if (path[0] == '[')
+		/*
+		 * skip special mappings like [heap], [vdso] etc.
+		 * but [stack] is needed when getting kernel base address later.
+		 */
+		if (path[0] == '[' && path[1] != 's')
```
namhyung: How about using `strncmp(path, "[stack", 6)`?

Taeung: OK, Got it!

Taeung:
@namhyung , I thought  the checking of '[' in `load_module_symtabs()`,
but how about no making `uftrace_mmap map` for [stack] in `read_session_map()` like changed PR?

I thought we can skip making the 'map' for [stack] before `load_module_symtabs()`.
If the 'map' for [stack]  is omitted in `read_session_map()` and `record_proc_maps()`,
`symtabs->maps` don't have the 'map' for [stack] anymore.

I'd appreciate it, if you review v2 PR ! :smile:

Taeung:
In addition, how about storing kernel base addr in `read_session_map()` and `record_proc_maps()` ?
when finding [stack] by the if statements `strcmp(path, "[stack", 6)` or `path[0] == '['`,
in order not to reading "sid-*.map" file  twice.

namhyung:
I think your approach is better.  Please add a commit to change `set_kernel_base` is called when reading maps.

Taeung:
OK, I changed this PR.
I'd appreciate it, if you check v3

PR #383, opened 2018-06-19
record: Do not read uftrace.data/info with --nop

honggyukim:
This fixes the below error message when --nop option is enabled.

    $ uftrace --nop -vvv abc
    uftrace: checking binary abc
    symbol: check trace functions in abc
    uftrace: creating 0 thread(s) for recording
    uftrace: using libmcount-nop.so library for tracing
    uftrace: all process/thread exited
    uftrace: child terminated with exit code: 0
    uftrace: fill header (metadata) info in /tmp/uftrace-live-1O7bAj/info
    uftrace: /home/honggyu/work/uftrace/cmds/record.c:377:fill_file_header
     ERROR: cannot open info file: No such file or directory

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1764,15 +1764,18 @@ static void finish_writers(struct writer_data *wd, struct opts *opts)
 	int i;
 	char *elapsed_time = get_child_time(&wd->ts1, &wd->ts2);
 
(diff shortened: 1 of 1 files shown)


namhyung:
One of the reasons using `--nop` option was to compare elapsed time with `--time` option.  So I think it should print the times before `fill_file_header()`.

honggyukim:
Thanks for the comment. I just updated it.

PR #393, opened 2018-07-06
info: Change elapsed time format in info

honggyukim:
This patch changes elapsed time format more like replay output.

Before:
  $ uftrace info
          ...
  # elapsed time        : 0.003219479 sec
  # cpu time            : 0.000 / 0.003 sec (sys / user)
          ...

After:
  $ uftrace info
          ...
  # elapsed time        : 7.74 ms
  # cpu time            : 0.004 / 0.000 sec (sys / user)
          ...


cmds/record.c
```diff
@@ -1382,19 +1382,14 @@ static void save_session_symbols(struct opts *opts)
 
 static char *get_child_time(struct timespec *ts1, struct timespec *ts2)
 {
-#define SEC_TO_NSEC  (1000000000ULL)
-
-	char *elapsed_time = NULL;
+	static char *elapsed_fmt_time = NULL;
```
namhyung: Why did you make it static?

honggyukim:
I see. What about showing it as follows then?
```
# elapsed time        : 0.004970269 sec  (4.970 ms)
```

cmds/record.c
honggyukim: Right, it's not needed. Thanks.

PR #395, opened 2018-07-10
doc: Add missing tui command

qpakzk:
tui command is missing in uftrace man page so update it.

### DIFF
--- a/doc/uftrace.md
+++ b/doc/uftrace.md
@@ -50,6 +50,8 @@ graph
 script
 :   Run a script for recorded function trace
 
+tui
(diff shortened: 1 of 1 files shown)


honggyukim:
@qpakzk I think we should also add `tui` to the `SYNOPSIS` in the same file. Anyway, welcome back :)

qpakzk:
@honggyukim Thank you. @namhyung  Then, may I  change a commit which was merged by adding `SYNOPSIS`?

qpakzk:
OK!

Issue #398, opened 2018-07-12
Finding a child with longest execution time in tui mode


honggyukim:
In `tui` mode, `c` is very useful by collapsing nested children. I also think that it'd be very useful to provide a short key to find the child that has longest execution time among all the other children.

Maybe we can use `l` key for this feature.

honggyukim:
It may be very useful when finding the critical path among all the other execution path of each child.

honggyukim:
If the current function is folded, then we can first open and collapse its children for preparation. Then we can find the longest child.

This can be used as follows:
1. fold the the callgraph at the root.
2. press `l` button the find the most executed child.
3. repeat pressing `l` button

PR #399, opened 2018-07-14
tui: Support to find longest executed child in tui graph

honggyukim:
In tui mode, 'c' key is very useful by collapsing nested children.
It'd also be very useful to provide a short key to find the child that
has longest execution time among all the other children.

The 'l' key for longest child performs as below:

  1. unfold the current node
  2. find a child that has the longest execution time
  3. collapse all the other children except for the longest child
  4. move to the longest child

It'd be useful when finding the performance critical execution path.

Resolved: #398

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -2093,8 +2093,7 @@ static void tui_main_loop(struct opts *opts, struct ftrace_file_handle *handle)
 			break;
 		case KEY_ENTER:
 		case '\n':
-			if (tui_window_enter(win, win->curr))
-				full_redraw = true;
+			full_redraw = tui_window_enter(win, win->curr);
 
 			if (win == &session->win) {
 				struct tui_list_node *cmd = win->curr;
@@ -2168,12 +2167,10 @@ static void tui_main_loop(struct opts *opts, struct ftrace_file_handle *handle)
 			}
 			break;
 		case 'c':
-			if (tui_window_collapse(win))
-				full_redraw = true;
+			full_redraw = tui_window_collapse(win);
 			break;
 		case 'e':
-			if (tui_window_expand(win))
-				full_redraw = true;
+			full_redraw = tui_window_expand(win);
 			break;
 		case 'p':
 			tui_window_move_prev(win);
@@ -2297,7 +2294,7 @@ int command_tui(int argc, char *argv[], struct opts *opts)
 
 int command_tui(int argc, char *argv[], struct opts *opts)
 {
-	pr_warn("TUI is not implemented (libncursesw.so is missing)");
+	pr_warn("TUI is unsupported (libncursesw.so is missing)\n");
 	return 0;
 }
 
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -66,6 +66,7 @@ struct tui_window_ops {
 	void (*footer)(struct tui_window *win, struct ftrace_file_handle *handle);
 	void (*display)(struct tui_window *win, void *node);
 	bool (*search)(struct tui_window *win, void *node, char *str);
+	bool (*longest_child)(struct tui_window *win, void *node);
 };
 
 struct tui_window {
@@ -121,6 +122,8 @@ static const struct tui_window_ops report_ops;
 static const struct tui_window_ops info_ops;
 static const struct tui_window_ops session_ops;
 
+static void tui_window_move_down(struct tui_window *win);
+
 #define FIELD_SPACE  2
 #define FIELD_SEP  " :"
 
@@ -143,6 +146,7 @@ static const char *help[] = {
 	"c/e           Collapse/Expand graph",
 	"n/p           Next/Prev sibling",
 	"u             Move up to parent",
+	"l             Move to the longest executed child",
 	"j/k           Move down/up",
 	"/             Search",
 	"</>/N/P       Search next/prev",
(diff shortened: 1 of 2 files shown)


honggyukim:
Thanks for the review. Do you want me to move `win_longest_child_graph` inside `tui_window_longest_child` and remove `longest_child` callback in `graph_ops`?

namhyung:
Yes

honggyukim:
Just updated it as you mentioned. Thanks!

honggyukim:
I understand that I should make this routine work only in graph mode. 

> You may add win->value() callback to get a value of a node.

Could you explain more about it? I'm not sure what `value()` callback does.

namhyung:
Well, I thought about it again and ended up not adding complexity unnecessarily.  Just go with your first version that implements it in the graph code.

honggyukim:
Okay. I'm back to the first version and made a few minor modifications.

honggyukim:
I just added one more commit that makes `u` key find parent and collapse its children. This can be a good counter part with `l` key that finds longest child function.

honggyukim:
I just made a few more commits that makes the behaviors of `n` and `p` keys similar to `l` and `u` keys.

honggyukim:
I just added one more refactoring commit.

cmds/tui.c
```diff
@@ -2032,6 +2008,48 @@ static bool tui_window_expand(struct tui_window *win)
...
+	fold_graph_node(curr, true);
+
+	while (win->curr != prev)
+		tui_window_move_up(win);
+
+	/* collapse the current node after moving to the previous sibling */
+	tui_window_collapse(win);
+
+	return true;
+}
```
namhyung: What about this?

```c
static bool tui_window_move_prev(win) {
   prev = win->ops->sibling_prev(win, win->curr);
   parent = win->ops->parent(win, win->curr);

   if (prev == NULL)
      return false;

   if (parent != NULL)
      win->ops->collapse(win, parent);

   while (win->curr != prev)
      tui_window_move_up(win);

   return tui_window_collapse(win);
}
```

cmds/tui.c
```diff
@@ -2297,7 +2294,7 @@ int command_tui(int argc, char *argv[], struct opts *opts)
 
 int command_tui(int argc, char *argv[], struct opts *opts)
 {
-	pr_warn("TUI is not implemented (libncursesw.so is missing)");
+	pr_warn("TUI is not implemented (libncursesw.so is missing)\n");
```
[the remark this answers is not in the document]
honggyukim: Okay.

cmds/tui.c
```diff
@@ -2032,6 +2008,48 @@ static bool tui_window_expand(struct tui_window *win)
 	return win->ops->expand(win, win->curr);
 }
 
+/* move to the previous sibling */
+static bool tui_window_move_prev(struct tui_window *win)
+{
+	struct tui_graph_node *curr = win->curr;
```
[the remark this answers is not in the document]
honggyukim: Okay. I see.

cmds/tui.c
honggyukim: Hmm.. It makes the time on the left disappears.

namhyung: What about returning `true` instead?

honggyukim: Okay, I will test it. By the way, why do we collapse the parent in this change? I think we discussed that we don’t touch other nodes except for prev and next.

honggyukim: I just tested but it has another bugs..

namhyung: Maybe you can change `win->ops->collapse()` and `win->ops->expand()` to pass `depth` argument and use it in the generic layer.

honggyukim:
I just made the change to simply collapse and expand with depth argument in `fold_graph_node`.
Please take a look at it again. Thanks!

cmds/tui.c
```diff
@@ -916,48 +916,40 @@ static bool win_enter_graph(struct tui_window *win, void *node)
...
+static int fold_graph_node(struct tui_graph_node *node, bool fold, int depth)
 {
 	struct tui_graph_node *child;
 	int count = 0;
+	bool curr_fold = fold;
+
+	if (depth < 0)
+		return 0;
+	else if (depth > 0)
+		curr_fold = false;
```
namhyung: Ah, this is somewhat different than I thought, but it's also ok.

cmds/tui.c
```diff
@@ -2002,15 +1994,17 @@ static bool tui_window_collapse(struct tui_window *win)
 	if (win->ops->collapse == NULL)
 		return false;
 
-	return win->ops->collapse(win, win->curr);
+	/* fold all the directly children */
+	return win->ops->collapse(win, win->curr, 1);
```
namhyung: This changes the current behavior - do you think it's worth changing when it comes to usability?

cmds/tui.c
```diff
@@ -2028,6 +2004,50 @@ static bool tui_window_expand(struct tui_window *win)
...
+
+	/* fold the current node before moving to the previous sibling */
+	if (win->ops->collapse)
+		count = win->ops->collapse(win, win->curr, 0);
+
+	while (win->curr != prev)
+		tui_window_move_up(win);
+
+	/* collapse the current node after moving to the previous sibling */
+	count += tui_window_collapse(win);
```
namhyung: Instead of moving the code (in order to use `tui_window_collapse` I guess), you can call `win->ops->collapse(depth = 1)` directly.  I believe it's better to understand what it does actually.

cmds/tui.c
honggyukim: It changes to unfold the current node first then folds all its children. I think it's very much needed because previous behavior doesn't make any visible change when the current node is already folded.

cmds/tui.c
honggyukim: Okay. I will do that. Thanks!

PR #411, opened 2018-07-22
tui: Apply various enhancements for tui command

honggyukim:
This PR makes some changes on TUI as follows:

1. Set the focus on the current function when switching from graph to report window
2. Make 'R' to go to the top of report window
3. Set to the top when switched from session window
4. Cancel search when cursor moves
5. Add 'z' key to set current line to the center of screen
6. Update the description of Enter key
7. Show source location info in report footer

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -2332,7 +2332,21 @@ static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
 		case 'R':
 		case 'r':
 			if (tui_window_change(win, &report->win)) {
+				struct tui_report_node *func;
+				struct tui_graph_node *graph_curr = win->curr;
+
+				func = (void *)report_find_node(&report->name_tree,
+					       graph_curr->n.name);
+				if (func == NULL)
+					break;
+
+				/* change to report window */
 				win = &report->win;
+
+				/* move focus on the same function */
+				tui_window_move_home(win);
+				tui_window_set_middle_next(win, func);
+
 				full_redraw = true;
 			}
 			break;
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -133,7 +133,7 @@ static const char *help[] = {
 	"ARROW         Navigation",
 	"PgUp/Dn",
 	"Home/End",
-	"Enter         Select/Fold",
+	"Enter         Fold/unfold graph or Select session",
 	"G             Show (full) call graph",
 	"g             Show call graph for this function",
 	"R             Show uftrace report",
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -2219,6 +2219,12 @@ static void tui_window_help(void)
 	delwin(win);
 }
 
+static inline void cancel_search()
+{
+	free(tui_search);
+	tui_search = NULL;
+}
+
 static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
 {
 	int key = 0;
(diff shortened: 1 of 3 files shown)


honggyukim:
I would like to set the window position to the center so please let me know if there's a way to do it.

honggyukim:
Hmm.. Just found that it's not an exact match but just a substring match.

`v8::internal::Compiler::Compile` is searched `v8::internal::Compiler::CompileOptimized` in report at first so have to go to find the next one.

cmds/tui.c
```diff
@@ -2215,8 +2215,23 @@ static void tui_main_loop(struct opts *opts, struct ftrace_file_handle *handle)
 			break;
 		case 'R':
 		case 'r':
+			if (win == &graph->win) {
+				struct tui_graph_node *curr = win->curr;
+
+				free(tui_search);
+				tui_search = xstrdup(curr->n.name);
+			}
```
namhyung: Maybe we could similar for graph.  'R' key is to go to the root of the report, and 'r' key is to go to the current node.

cmds/tui.c
```diff
@@ -2167,6 +2167,7 @@ static void tui_main_loop(struct opts *opts, struct ftrace_file_handle *handle)
 				switch ((long)cmd->data) {
 				case TUI_SESS_REPORT:
 					win = &report->win;
+					tui_window_move_home(win);
```
namhyung: You'd better doing it consistently - the same should go to 'graph' and 'info' as well.

honggyukim: I applied the same thing to graph and info as well.

cmds/tui.c
honggyukim: I added 'R' key for doing that.

honggyukim:
> Please check the current master I added tui_window_set_middle_prev/next(). I think you should not use the search interface. Instead find a target node using find_report_node().

It took quite some time for this PR, but I finally updated it. I used `report_find_node` as you suggested to find the node in report window. Thanks a lot for your help!

cmds/tui.c
```diff
@@ -133,7 +133,7 @@ static const char *help[] = {
 	"ARROW         Navigation",
 	"PgUp/Dn",
 	"Home/End",
-	"Enter         Select/Fold",
+	"Enter         Graph fold/unfold or Select session",
```
namhyung: Wouldn't it be better saying "Fold/unfold graph" instead?

cmds/tui.c
```diff
@@ -1274,10 +1274,21 @@ static void win_footer_report(struct tui_window *win, struct uftrace_data *handl
...
-			 handle->dirname, report->nr_sess, report->nr_func);
+		if (dloc != NULL && dloc->file != NULL) {
+			snprintf(buf, COLS, "uftrace report: %s [line:%d]",
+				 dloc->file->name, dloc->line);
+		}
+		else {
+			struct tui_report *report = (struct tui_report *)win;
+
+			snprintf(buf, COLS,
+				 "uftrace report: %s (%d sessions, %d functions)",
```
namhyung: I think in this case you'd better putting it in the above line.  A notable exception of 80 column rule is the string constant.

cmds/tui.c
honggyukim: I updated it to "Fold/unfold graph or Select session". Thanks.

cmds/tui.c
honggyukim: Done.

honggyukim:
I just updated it based on your comment. Thanks for the review!

PR #421, opened 2018-07-28
uftrace: Remove recover trigger for simplicity

honggyukim:
Since auto-recover is on by 00340e5(Merge branch 'auto-recover'),
'recover' trigger is no longer needed.

This patch removes the historic 'recover' trigger to make it simpler.


namhyung:
We should at least support to parse it for backward compatibility.

honggyukim:
Right. I will update it later. Thanks!

namhyung:
Also I found that auto recover caused crash on ARM.

honggyukim:
Hmm.. I will just hold this PR then.

honggyukim:
Sure. No problem then.

PR #422, opened 2018-07-28
uftrace: Make -a option synonym for --auto-args

honggyukim:
It'd be better to provide a short option for --auto-args, so this patch
adds -g so that we can simply see arguments and return values.


honggyukim:
I just updated it to use `-a` instead of `-g`. Thanks!

PR #427, opened 2018-07-29
chrome: Don't use parenthesis for return value

honggyukim:
Currently, the arguments are wraped by parenthesis and it looks like
tuple.  But return value is also represented in the same way.

As return value only contains a single value, so it'd be better to show
it without parenthesis.

Before:
  { ...,"args":{"arguments":"(0xa9ad68, -1)"}},
  { ...,"args":{"retval":"(2)"}},

After:
  { ...,"args":{"arguments":"(0xa9ad68, -1)"}},
  { ...,"args":{"retval":"2"}},

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -857,7 +857,7 @@ static void print_chrome_task_rstack(struct uftrace_dump_ops *ops,
 	char ph;
 	char spec_buf[1024];
 	struct uftrace_record *frs = task->rstack;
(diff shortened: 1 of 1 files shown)


namhyung:
Is there any problem with the parenthesis?

honggyukim:
I think it's useless to represent return value with parenthesis. For example, `"10"` is better than `"(10)"` and uses less characters.

Issue #412, opened 2018-07-22
reading symbols will be fail if target has no section header.


ParkHanbum:
while reading symbols from each modules, iterating section header to find its string tab. when this time, if target module has no section header or string table, like 'ld.so.cache'. it will be crashed. 
```
 55 void elf_get_strtab(struct uftrace_elf_data *elf,
 56                     struct uftrace_elf_iter *iter,
 57                     int shidx)
 58 {
 59         Elf_Shdr *shdr = elf->file_map + elf->ehdr.e_shoff;
 60
 61         iter->strtab = elf->file_map + shdr[shidx].sh_offset;
 62 }
```

```
 91 #define elf_for_each_shdr(elf, iter)                                    \
 92         for (elf_get_strtab((elf), (iter), (elf)->ehdr.e_shstrndx),     \
 93                      (iter)->i = 0, (iter)->nr = (elf)->ehdr.e_shnum;   \
 94              (iter)->i < (iter)->nr &&                                  \
 95                      memcpy(&(iter)->shdr,                              \
 96                             (elf)->file_map + (elf)->ehdr.e_shoff +     \
 97                             (iter)->i * (elf)->ehdr.e_shentsize,        \
 98                             (elf)->ehdr.e_shentsize);                   \
 99              (iter)->i++)
```

it can be fixed simply by add simple validation logic.
how do you think?

namhyung:
Sure, please send a PR.  Does this problem exist only in the raw ELF code?

ParkHanbum:
@namhyung I think it is occured in raw-elf only, because when I test it in another system that has installed libelf did not crashed. but I'll review it again, and notice to you.

namhyung:
OK, fot testing you can run `configure --without-libelf` to disable it.

ParkHanbum:
```
Core was generated by `./uftrace dynamic -L/home/m/git/attach -d /home/m/git/attach/test --pid=2511'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000417da2 in elf_get_strtab (elf=elf@entry=0x7ffd995eeb10,
    iter=iter@entry=0x7ffd995eeb70, shidx=0)
    at /home/m/git/attach/utils/symbol-rawelf.c:61
61              iter->strtab = elf->file_map + shdr[shidx].sh_offset;
[...]

PR #432, opened 2018-08-04
symbol: validate ELF format before using it

ParkHanbum:
sometime uftrace should reference the address (eg, section offset) that
the ELF field points to. Since section headers are not essential, they
can refer to insecure memory and cause errors such as segment faults.
so, add logic to validate ELF before using it.

### DIFF
--- a/utils/symbol-rawelf.c
+++ b/utils/symbol-rawelf.c
@@ -13,6 +13,83 @@
 #include "utils/utils.h"
 #include "utils/symbol-rawelf.h"
 
+/*
+ *  ELF File Header validation logic.
+ */
+int elf_validation(struct uftrace_elf_data *elf)
+{
+	Elf_Ehdr *ehdr;
+	int eclass, data, version;
+	long size, offset;
+
+	ehdr = &elf->ehdr;
+	elf->has_shdr = false;
+
+	// validate ELF Magic.
+	if (memcmp(ehdr, ELFMAG, SELFMAG)) {
+		pr_dbg2("ELF Signature not matched\n");
+		return -1;
+	}
+
+	// validate some field of elf header.
+	eclass = (int) ehdr->e_ident[EI_CLASS];
+	data = (int) ehdr->e_ident[EI_DATA];
+	version = (int) ehdr->e_ident[EI_VERSION];
+
+	if (!(eclass > ELFCLASSNONE && eclass < ELFCLASSNUM)
+		|| eclass != get_elf_class()) {
+		pr_dbg2("Invalid eclass : [%d]\n", eclass);
+		return -1;
+	}
+
+	if (!(data > ELFDATANONE && data < ELFDATANUM)
+		|| data != get_elf_endian()) {
+		pr_dbg2("Invalid endian : [%d]\n", data);
+		return -1;
+	}
+
+	if (!(version > EV_NONE && version < EV_NUM)) {
+		pr_dbg2("Invalid ELF version : [%d]\n", version);
+		return -1;
+	}
+
+	if (ehdr->e_phnum == 0 || ehdr->e_phentsize == 0) {
+		pr_dbg2("Invalid Program header. Num:[%d] Size:[%d]\n",
+				ehdr->e_phnum, ehdr->e_phentsize);
+		return -1;
+	}
+
+	if (ehdr->e_shnum > 0 && ehdr->e_shentsize == 0) {
+		pr_dbg2("Section Header entry size cannot be 0.\n");
+		return -1;
+	}
+
+	// validate program header offset.
+	size = (long)elf->file_size;
+	offset = ehdr->e_phoff + ehdr->e_phnum * ehdr->e_phentsize;
+
+	if (offset > size) {
+		pr_dbg2("Invalid Program Header offset."\
+				"offset:[%lu], size:[%lu]\n"\
+				,offset, size);
+		return -1;
+	}
+
+	// validate sectio header offset.
+	offset = ehdr->e_shoff + ehdr->e_shnum * ehdr->e_shentsize;
+
+	if (offset > size) {
+		pr_dbg2("Invalid Program Header offset."\
+				"offset:[%lu], size:[%lu]\n"\
+				,offset, size);
+		elf->has_shdr = false;
+	}
+	else
+		elf->has_shdr = true;
+
+	return 0;
+}
+
 int elf_init(const char *filename, struct uftrace_elf_data *elf)
 {
 	struct stat stbuf;
@@ -32,6 +109,9 @@ int elf_init(const char *filename, struct uftrace_elf_data *elf)
 		goto err_close;
 
 	memcpy(&elf->ehdr, elf->file_map, sizeof(elf->ehdr));
+
+	if (elf_validation(elf) < 0)
+		goto err;
 
 	return 0;
 
@@ -60,9 +140,10 @@ void elf_get_strtab(struct uftrace_elf_data *elf,
 		    struct uftrace_elf_iter *iter,
 		    int shidx)
 {
-	Elf_Shdr *shdr = elf->file_map + elf->ehdr.e_shoff;
-
-	iter->strtab = elf->file_map + shdr[shidx].sh_offset;
+	if (elf->has_shdr) {
+		Elf_Shdr *shdr = elf->file_map + elf->ehdr.e_shoff;
+		iter->strtab = elf->file_map + shdr[shidx].sh_offset;
+	}
 }
 
 void elf_get_secdata(struct uftrace_elf_data *elf,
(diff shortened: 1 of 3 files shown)


ParkHanbum:
```
symbol: ELF Signature not matched
symbol: eclass out of range. it must be inside range between 0 and 3, but Yours is [111]
symbol: data out of range. it must be inside range between 0 and 3, but Yours is [45]
symbol: version out of range. it must be inside range between 0 and 2, but Yours is [49]
symbol: Program Header of ELF Header is not normal.
Program Header: Num:[12256] Size:[0]
symbol: Section Header of ELF Header size cannot be 0.
symbol: Section Header address must not over the file size that mapped
mapped address : [7f079e993000 - 7f079e99ea74]but Yours is : [aebd9e993303]
symbol: Section Header address must not over the file size that mapped
mapped address : [7f079e993000 - 7f079e99ea74]but Yours is : [ae9d9e995f8c]
symbol: error during open symbol file: /etc/ld.so.cache: Bad file descriptor
symbol: ELF Signature not matched
symbol: eclass out of range. it must be inside range between 0 and 3, but Yours is [111]
symbol: data out of range. it must be inside range between 0 and 3, but Yours is [45]
symbol: version out of range. it must be inside range between 0 and 2, but Yours is [49]
symbol: Program Header of ELF Header is not normal.
Program Header: Num:[12256] Size:[0]
symbol: Section Header of ELF Header size cannot be 0.
symbol: Section Header address must not over the file size that mapped
mapped address : [7f079e987000 - 7f079e992a74]but Yours is : [aebd9e987303]
symbol: Section Header address must not over the file size that mapped
mapped address : [7f079e987000 - 7f079e992a74]but Yours is : [ae9d9e989f8c]
symbol: error during open symbol file: /etc/ld.so.cache: Bad file descriptor
symbol: ELF Signature not matched
symbol: eclass out of range. it must be inside range between 0 and 3, but Yours is [111]
```

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+	int eclass, data, version;
+	int res = 0;
+	unsigned long mapped_saddr, mapped_eaddr, mapped_size, addr;
+
+	ehdr = &elf->ehdr;
+
+	// validate ELF Magic.
+	if (memcmp(ehdr, ELFMAG, SELFMAG)) {
+		pr_dbg("ELF Signature not matched\n");
+		res = -1;
```
namhyung: there's no need to proceed.  I think we could return here.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+	}
+
+	// validate some field of elf header.
+	eclass = (int) ((unsigned char *) ehdr)[EI_CLASS];
+	data = (int) ((unsigned char *) ehdr)[EI_DATA];
+	version = (int) ((unsigned char *) ehdr)[EI_VERSION];
```
namhyung: you can use `ehdr->e_ident[]`.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+
+	if (!(eclass > ELFCLASSNONE && eclass < ELFCLASSNUM)) {
+		pr_dbg("eclass out of range. it must be inside range "\
+				"between %d and %d, but Yours is [%d]\n",
+				ELFCLASSNONE, ELFCLASSNUM, eclass);
```
namhyung: generally I don't like multi-line strings especially in debug level 1.  Also I think this verbose info is not very useful to users.  What about just like below?

```
   pr_dbg("invalid ELF class: %d\n", eclass);
```

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+				ELFDATANONE, ELFDATANUM, data);
+		res = -1;
+	}
+
+	if (!(version > EV_NONE && version < EV_NUM)) {
+		pr_dbg("version out of range. it must be inside range "\
+				"between %d and %d, but Yours is [%d]\n",
+				EV_NONE, EV_NUM, version);
+		res = -1;
+	}
```
namhyung: ditto.

Also it should check the class and endian is same for the host.  IOW raw-ELF code cannot handle 64-bit ELF on 32-bit systems and vice versa.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+
+	// validate ELF file size.
+	mapped_saddr = (long)elf->file_map;
+	mapped_size = elf->file_size;
+	mapped_eaddr = mapped_saddr + mapped_size;
```
namhyung: These variables seem not needed except for printing.

utils/symbol-rawelf.c
```diff
@@ -33,6 +123,10 @@ int elf_init(const char *filename, struct uftrace_elf_data *elf)
 
 	memcpy(&elf->ehdr, elf->file_map, sizeof(elf->ehdr));
 
+	if (elf_validation(elf) < 0) {
+		goto err;
+	}
```
namhyung: You don't have to put parenthesis on this simple case.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,96 @@
...
+		ehdr->e_shnum * ehdr->e_shentsize;
+
+	if (addr > mapped_eaddr) {
+		pr_dbg("Section Header address must not over the " \
+				"file size that mapped\n" \
+				"mapped address : [%llx - %llx]" \
+				"but Yours is : [%llx]\n",
+				mapped_saddr, mapped_eaddr, addr);
+		res = -1;
+	}
```
namhyung: I was not sure it'd better failing out when it cannot find section headers.  It's legal to access ELF binary using program headers only.

utils/symbol-rawelf.c
ParkHanbum: Yes, but this message appears to be valuable to the developer. 
So, how about leave a message and just raise the level to 2 or 3 instead?

utils/symbol-rawelf.c
ParkHanbum: I'll move the logic that validates the ELF section header to another function and change the function that refers to the section header to call this function before reference it.

utils/symbol-rawelf.c
namhyung: That's why I printed the class number.  I guess relavant developers know what the value means and can check it's invalid easily.

ParkHanbum: maybe, but I think it cost a little bit.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,86 @@
...
+ *  ELF File Header validation logic.
+ */
+int elf_validation(struct uftrace_elf_data *elf)
+{
+	Elf_Ehdr *ehdr;
+	int eclass, data, version;
+	unsigned long mapped_saddr, addr;
+
+	ehdr = &elf->ehdr;
+	elf->have_section = false;
```
namhyung: I'd like to change it to `has_shdr` as we used the third-person form and section (content) itself is there - only section headers are missing.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,86 @@
...
+
+	// validate ELF Magic.
+	if (memcmp(ehdr, ELFMAG, SELFMAG)) {
+		pr_dbg3("ELF Signature not matched\n");
```
namhyung: let's make it `pr_dbg2()` for (detailed) error messages.

utils/symbol-rawelf.c
```diff
@@ -13,6 +13,86 @@
...
+	mapped_saddr = (long)elf->file_map;
+	addr = mapped_saddr + ehdr->e_phoff +
+		ehdr->e_phnum * ehdr->e_phentsize;
+
+	if (addr > mapped_saddr + elf->file_size) {
+		pr_dbg3("Invalid Program Header address. "\
+				"mapped address : [%llx - %llx]" \
+				"but phdr address is : [%llx]\n",
+				mapped_saddr,
+				mapped_saddr + elf->file_size, addr);
```
namhyung: I don't think the address is meaningful.  Also `mapped_saddr` is not needed for the check - we can only check the offsets.

utils/symbol-rawelf.h
```diff
@@ -91,7 +92,7 @@ struct uftrace_elf_iter {
 #define elf_for_each_shdr(elf, iter)					\
 	for (elf_get_strtab((elf), (iter), (elf)->ehdr.e_shstrndx),	\
 		     (iter)->i = 0, (iter)->nr = (elf)->ehdr.e_shnum;	\
-	     (iter)->i < (iter)->nr &&					\
+		     (iter)->i < (iter)->nr && (elf)->have_section &&	\
```
namhyung: Please don't change the indentation - it's there to identify the start of second part of the for statement.

namhyung:
In addition, please update the commit message that it fixes the issue #412 like below:

```
message here..

Fixed: #412

Signed-off-by: ...
```

ParkHanbum:
@namhyung uftrace has a field to know cpu info that include endian or bitsize?

Issue #437, opened 2018-08-12
failing dlopen lead to create of directory that has named randomize unintentionally.


ParkHanbum:
```
m@m:~/git/uftrace$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        libtraceevent/.event-parse.d.97952
        libtraceevent/.event-parse.d.97997
        libtraceevent/.event-parse.d.98237
        "tests/\f\204\001"
        "tests/!\204\001"
        "tests/d\203\001"
        "tests/y\203\001"
        "tests/\216\203\001"
        "tests/\243\203\001"
        "tests/\270\203\001"
        "tests/\315\203\001"
        "tests/\342\203\001"
        "tests/\367\203\001"
```

```
m@m:~/git/uftrace$ ls -al tests/
total 5292
-r-xrws--T  1 m m       0 Aug 12 02:00 '!'$'\204\001'
drwxrwxr-x  3 m m   24576 Aug 12 02:00  .
drwxrwxr-x 16 m m    4096 Aug 12 01:53  ..
-r-xrws--T  1 m m       0 Aug 12 01:59 ''$'\342\203\001'
-r-xrws--T  1 m m       0 Aug 12 01:59 '̓'$'\001'
-r--r-x---  1 m m       0 Aug 12 01:59 ''$'\243\203\001'
-r-xrws--T  1 m m       0 Aug 12 01:59 ''$'\367\203\001'
-rwxr-x---  1 m m       0 Aug 12 01:59 ''$'\270\203\001'
-rwSrwx---  1 m m       0 Aug 12 01:59 ''$'\216\203\001'
-r-xrws--T  1 m m       0 Aug 12 01:59 ''$'\f\204\001'
drwxrwxr-x  4 m m    4096 Aug 12 01:53  arch
-rws--s---  1 m m       0 Aug 12 01:59 'd'$'\203\001'
```

why these files has been created?

namhyung:
Hmm.. I cannot reproduce.  Could you please tell me more about the your environment?  Is this 100% reproducible?

ParkHanbum:
```
m@m:~/git$ git clone https://github.com/namhyung/uftrace.git test
Cloning into 'test'...
remote: Counting objects: 13174, done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 13174 (delta 1), reused 1 (delta 0), pack-reused 13166
[...]

PR #440, opened 2018-08-14
test: update tests that relate with dlopen

ParkHanbum:
recently loader does not allow load shared object from unspecified path.
therefore, when load shared object by using dlopen then must specify the
path where to load shared object.

Fixed: #437

### DIFF
--- a/tests/t200_lib_dlopen2.py
+++ b/tests/t200_lib_dlopen2.py
@@ -28,3 +28,6 @@ class TestCase(TestBase):
             return TestBase.TEST_BUILD_FAIL
         return TestBase.build_libmain(self, name, 's-dlopen2.cpp', ['libdl.so'],
                                       cflags, ldflags)
+
+    def runcmd(self):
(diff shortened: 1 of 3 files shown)


tests/t200_lib_dlopen2.py
```diff
@@ -28,3 +28,6 @@ def build(self, name, cflags='', ldflags=''):
             return TestBase.TEST_BUILD_FAIL
         return TestBase.build_libmain(self, name, 's-dlopen2.cpp', ['libdl.so'],
                                       cflags, ldflags)
+
+    def runcmd(self):
+        return 'LD_LIBRARY_PATH=. %s -F a %s' % (TestBase.uftrace_cmd, 't-dlopen2')
```
namhyung: What about test case 131 then?  Doesn't it need `LD_LIBRARY_PATH` too?

namhyung: Ah, ok.  thanks for the explanation.

PR #443, opened 2018-08-19
doc: Prefer relative link than absolute link in README.md

taeguk:
Currently, some links in README.md point things that are in 'master' branch of 'upstream' repository.
But, these links couldn't work as expected in different branch or forked repository.
So, I think that using relative link than absolute link is better.

Related Material: https://help.github.com/articles/about-readmes/#relative-links-and-image-paths-in-readme-files


taeguk:
@honggyukim There is already ```Signed-off-by``` in my commit. Can you check one more time, please?

PR #446, opened 2018-08-21
script: Add missing argument in 'uftrace_begin'

taeguk:
https://github.com/namhyung/uftrace/blob/419e65024ad19ac85d120f30123802be694a0967/utils/script-python.c#L508
As you can see above, ```uftrace_begin``` requires one argument.
But there are no arguments in ```uftrace_begin``` of some scripts and documents.
So, I add missing argument in 'uftrace_begin'.


honggyukim:
Nice catch! We didn't update these yet :)

Issue #441, opened 2018-08-16
memory leak in Multiple functions


gy741:
Hello.

I found a memory leak bug.

In the load_debug_file, read_events_file, read_task_txt_file, read_session function.

For the first parameters used in the getline() function, free() function is not used.

So, I think there can be a memory leak.

What do you think?

Thanks.

Example:

1.Problem Code : utils/dwarf.c:1309 
https://github.com/namhyung/uftrace/blob/957332a69716679d01a9c227709b43575e06ae99/utils/dwarf.c#L1277-L1375

2.Good Code : uftrace.c:936
https://github.com/namhyung/uftrace/blob/f2537c8f9bf164f8a0f3898e8f3cff8ec8a0e3ea/uftrace.c#L881-L938

honggyukim:
Hi @gy741, could you please send a pull request instead of writing it as an issue?

gy741:
@honggyukim 

Hello,

I sent a pull request.

Thanks.

PR #447, opened 2018-08-21
Fix memory leak in Multiple functions

gy741:
Hello,

The memory allocated by getline() is not freed.

So, It should free the allocated memory.

Associated issue : #441

Thanks.

### DIFF
--- a/misc/symbols.c
+++ b/misc/symbols.c
@@ -159,6 +159,7 @@ static int read_session(struct uftrace_session_link *link, char *dirname)
 	if (count > 1)
 		needs_session = true;
 
+	free(line);
 	fclose(fp);
 	free(fname);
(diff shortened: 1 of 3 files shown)


honggyukim:
Thanks for the PR but I think you'd better merge all those commits into a single commit in this case because all of them make the same change.

gy741:
@honggyukim 

Hello,

Thank you for your reply.

If created a single commit, What should I write about the commit title?

Thanks.

honggyukim:
I could be anything but I would write it as `uftrace: Fix memory leaks allocated by getline` if you ask me the title.

gy741:
@honggyukim 

Hello,

I modified the commit.

Please check.

Thanks.

PR #448, opened 2018-08-21
build: Check if 'pkg-config' is installed when performing 'make' for check-deps

taeguk:
If ```pkg-config``` is not installed in a system and performing ```make``` for check-deps, the ```make``` command runs with no errors but it couldn't run as expected.
So, add check if 'pkg-config' is installed in a system to the build target related to check-deps.

### before
```./configure``` runs with no errors, but it doesn't run as expected. (Expected Result : ON for libncursesw)
image

### after
If 'pkg-config' is not installed in a system, the ```./configure``` and ```make``` fail with the message "pkg-config is not available. please install pkg-config".
image


namhyung:
I don't want to make it fail to build just because of the missing `pkg-config`.  As an optional dependency we can skip some features if not detected for some reason.  And I believe valid build environments already installed the `pkg-config`.

namhyung:
You'd better mention installing `pkg-config` in the INSTALL.md

taeguk:
@namhyung Okay, I close this PR. And I opened new PR that your advice is applied.

Issue #452, opened 2018-08-27
meaningless condition check in utils/symbol.c


gy741:
Hello.

I think, the if conditional statement in utils/symbol.c on line 1301 is useless.

The if conditional statement is always true.

what do you think?

Thanks.

https://github.com/namhyung/uftrace/blob/fe5e6dcea5be5cacf92b61b02ea56bed65b9fab3/utils/symbol.c#L1263-L1310

namhyung:
Indeed, please send a PR. :)

PR #455, opened 2018-08-28
symbol: Remove dead code in save_module_symbol()

gy741:
A part of conditional expression is always true.

Fixed: #452

### DIFF
--- a/utils/symbol.c
+++ b/utils/symbol.c
@@ -1298,13 +1298,12 @@ static void save_module_symbol(struct symtab *stab, const char *symfile,
 		fprintf(fp, "%016"PRIx64" %c %s\n", sym->addr - offset,
 			(char) sym->type, sym->name);
 	}
-	if (i > 0) {
-		struct sym *prev = &stab->sym[i - 1];
+	
+	struct sym *last = &stab->sym[stab->nr_sym - 1];
 
-		fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
-			prev->addr + prev->size - offset,
-			(char) prev->type, prev_was_plt ? "dyn" : "");
-	}
+	fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
+		last->addr + last->size - offset,
+		(char) last->type, prev_was_plt ? "dyn" : "");
 
 	fclose(fp);
 }


utils/symbol.c
```diff
@@ -1298,13 +1298,12 @@ static void save_module_symbol(struct symtab *stab, const char *symfile,
 		fprintf(fp, "%016"PRIx64" %c %s\n", sym->addr - offset,
 			(char) sym->type, sym->name);
 	}
-	if (i > 0) {
-		struct sym *prev = &stab->sym[i - 1];
+	
+	struct sym *prev = &stab->sym[i - 1];
```
namhyung: While at it, I think it's better renaming this to 'last'.

utils/symbol.c
```diff
@@ -1298,13 +1298,12 @@ static void save_module_symbol(struct symtab *stab, const char *symfile,
...
 
-		fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
-			prev->addr + prev->size - offset,
-			(char) prev->type, prev_was_plt ? "dyn" : "");
-	}
+	fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
+	prev->addr + prev->size - offset,
+	(char) prev->type, prev_was_plt ? "dyn" : "");
```
namhyung: Please indent function arguments properly.

gy741: oops, I did not think of it as a parameter of fprintf.
I'll fix it.

utils/symbol.c
gy741: Like this way?

```
struct sym *last = &stab->sym[i - 1];

fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
	last->addr + last->size - offset,
	(char) last->type, last_was_plt ? "dyn" : "");
```

namhyung: Right but excluding the last one - 'prev_was_plt' which is updated in the loop.

gy741: Aha, Like this way?
```
struct sym *last = &stab->sym[i - 1];

fprintf(fp, "%016"PRIx64" %c __%ssym_end\n",
	last->addr + last->size - offset,
	(char) last->type, prev_was_plt ? "dyn" : "");
```

namhyung: Yep.  While it's effectively same, I slightly prefer this:

```c
struct sym *last = &stab->sym[stab->nr_sym - 1];

...
```

gy741: Okay.
I'll fix it.

PR #466, opened 2018-08-31
doc: Add -fxray-instrument description to --patch option

honggyukim:

### DIFF
--- a/doc/uftrace-live.md
+++ b/doc/uftrace-live.md
@@ -121,7 +121,9 @@ OPTIONS
 :   Show kernel functions only without user functions.
 
 -P *FUNC*, \--patch=*FUNC*
-:   Patch FUNC dynamically.  This is only applicable binaries built with `-pg -mfentry -mnop-mcount` on x86_64.  This option can be used more than once.  See *DYNAMIC TRACING*.
+:   Patch FUNC dynamically.  This is only applicable binaries built by
+    gcc with `-pg -mfentry -mnop-mcount` or clang with `-fxray-instrument`.
+    This option can be used more than once.  See *DYNAMIC TRACING*.
 
 -E *EVENT*, \--event=*EVENT*
 :   Enable event tracing.  The event should be available on the system.
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -99,7 +99,9 @@ OPTIONS
 :   Set kernel tracing buffer size.  The default value (in the kernel) is 1408k.
 
 -P *FUNC*, \--patch=*FUNC*
-:   Patch FUNC dynamically.  This is only applicable binaries built with `-pg -mfentry -mnop-mcount` on x86_64.  This option can be used more than once.  See *DYNAMIC TRACING*.
+:   Patch FUNC dynamically.  This is only applicable binaries built by
+    gcc with `-pg -mfentry -mnop-mcount` or clang with `-fxray-instrument`.
+    This option can be used more than once.  See *DYNAMIC TRACING*.
 
 -E *EVENT*, \--event=*EVENT*
 :   Enable event tracing.  The event should be available on the system.


honggyukim:
I also think it's needed.  But even if we break down the line at around 80 column, adding a few words inside the description makes the following lines to be updated. Hmm.. I'm not sure if there's better way.

honggyukim:
I will reorganize the pages first, then apply this change.

honggyukim:
We have too many options so users might got confused among the large set of options. So I think it'd be better to categorize them into a few sections.
How do you think about the below output of `uftrace record` man page?
```
UFTRACE-RECORD(1)                                                            UFTRACE-RECORD(1)

NAME
       uftrace-record - Run a command and record its trace data

SYNOPSIS
       uftrace record [options] COMMAND [command-options]

DESCRIPTION
       This  command  runs  COMMAND and gathers function trace data from it, and saves it into
       files under the uftrace data directory - without displaying anything.

       This data can then be inspected later on, using uftrace replay or uftrace report.

FILTER OPTIONS
       -F FUNC, --filter=FUNC
              Set filter to trace selected functions only.  This option can be used more  than
              once.  See FILTERS.

       -N FUNC, --notrace=FUNC
              Set  filter  not to trace selected functions (or the functions called underneath
              them).  This option can be used more than once.  See FILTERS.

       -T TRG, --trigger=TRG
              Set trigger on selected functions.  This option can be used more than once.  See
              TRIGGERS.

       -D DEPTH, --depth=DEPTH
              Set global trace limit in nesting level.

       -t TIME, --time-filter=TIME
              Do not show functions which run under the time threshold.  If some functions ex‐
              plicitly have 'trace' trigger applied, those are always traced regardless of ex‐
              ecution time.

ARGUMENTS OPTIONS
       -A SPEC, --argument=SPEC
              Record  function  arguments.  This option can be used more than once.  See ARGU‐
              MENTS.

       -R SPEC, --retval=SPEC
              Record function return values.  This option can be used more than once.  See AR‐
              GUMENTS.

       -a, --auto-args
              Automatically  record arguments and return values of known functions.  These are
              usually functions in standard (C language or system) libraries but if debug info
              is available it includes functions in the user program.

TRACING TARGET OPTIONS
       --nest-libcall
              Trace function calls between libraries.  By default, uftrace only record library
              call from the main executable.  Implies --force.

       --no-libcall
              Do not record library function invocations.  Library calls are  normally  traced
              by hooking the dynamic linker's resolve function in the PLT.  One can disable it
              with this option.

       --no-pltbind
              Do not bind dynamic symbol address.  This option uses the  LD_BIND_NOT  environ‐
              ment variable to trace library function calls which might be missing due to con‐
              current (first) accesses.  It is not meaningful to  use  this  option  with  the
              --no-libcall option.

       -k, --kernel
              Trace  kernel functions as well as user functions.  Only kernel entry/exit func‐
              tions will be traced by default.  Use  the  --kernel-depth  option  to  override
              this.

       -K DEPTH, --kernel-depth=DEPTH
              Set kernel max function depth separately.  Implies --kernel.

       --kernel-buffer=SIZE
              Set kernel tracing buffer size.  The default value (in the kernel) is 1408k.

       -E EVENT, --event=EVENT
              Enable event tracing.  The event should be available on the system.

       --no-event
              Disable  event  recording  which  is  used by default.  Note that explicit event
              tracing by --event option is not affected by this.

OTHER OPTIONS
       -P FUNC, --patch=FUNC
              Patch FUNC dynamically.  This is only applicable binaries built with  -pg -mfen‐
              try -mnop-mcount on x86_64.  This option can be used more than once.  See DYNAM‐
              IC TRACING.

       -S SCRIPT_PATH, --script=SCRIPT_PATH
              Add a script to do additional work at the entry and exit of function.  The  type
              of script is detected by the postfix such as '.py' for python.

       -L PATH, --library-path=PATH
              Load necessary internal libraries from this path.  This is for testing.
...
```

namhyung:
That would be great, but let's do it later with other manuals all together.

honggyukim:
Okay. I will just leave it as of now.

namhyung:
Yep, please break lines in the `--patch` option only.

honggyukim:
I just updated it. Thanks!

PR #470, opened 2018-09-03
mcount: Remove dead code in is_arm_machine()

gy741:
Hello.

I think, the if conditional statement in utils/filter.c on line 388 is useless.

The if conditional statement is always true.

what do you think?

Thanks.

problem code:
https://github.com/namhyung/uftrace/blob/fc9ba76fcde0121af1fcc59307ca7ac13db2a69a/utils/filter.c#L384-L396


namhyung:
Nop, it's a "static" variable.

gy741:
@namhyung  oops, sorry. clsoe.

namhyung:
That's totally ok, don't worry. :)

PR #477, opened 2018-09-04
plthook: Remove dead code in update_pltgot()

gy741:
Hello.

I think, the if conditional statement in line 684 is useless.

The if conditional statement is always true.

what do you think?

Thanks.

problem code:
https://github.com/namhyung/uftrace/blob/919dce7471951b39830b692042397e082fbe3aaa/libmcount/plthook.c#L677-L694


gy741:
@namhyung Aha, Okay. 
I will close PR.

PR #479, opened 2018-09-05
script: Fix useless use of cat and typo

gy741:
Hello.

I modified the UUOC(useless use of cat) and typo.

what do you think?

Thanks.

### DIFF
--- a/misc/install-elfutils.sh
+++ b/misc/install-elfutils.sh
@@ -6,7 +6,7 @@ prefix=/usr/local
 objdir=$(readlink -f ${objdir:-${PWD}})
 builddir=${objdir}/.build
 
-n_cpus=$(cat /proc/cpuinfo | grep ^processor | wc -l)
+n_cpus=$(grep -c ^processor /proc/cpuinfo)
 
 usage() {
     echo "Usage: $0 [<options>]
--- a/misc/bash-completion.sh
+++ b/misc/bash-completion.sh
@@ -1,5 +1,5 @@
 _uftrace () {
-    local cur prev subcmd options uftrace_comp
+    local cur prev subcmds options uftrace_comp
 
     cur=${COMP_WORDS[COMP_CWORD]}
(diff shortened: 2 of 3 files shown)


misc/install-elfutils.sh
```diff
@@ -6,7 +6,7 @@ prefix=/usr/local
 objdir=$(readlink -f ${objdir:-${PWD}})
 builddir=${objdir}/.build
 
-n_cpus=$(cat /proc/cpuinfo | grep ^processor | wc -l)
+n_cpus=$(grep ^processor /proc/cpuinfo | wc -l)
```
namhyung: Actually `wc` also isn't needed like below:

```
$(grep -c ^processor /proc/cpuinfo)
```

namhyung: Looks OK.  But I think the subject line should start with "misc" rather than "script:" which might be confused with "uftrace script" command.

gy741: @namhyung  Okay. I will change the subject.
And, How about this source code?
The use of $(...) appears to be preferred.
If you like, I'll include it in PR.

```diff
diff --git a/misc/version.sh b/misc/version.sh
index fac546d..013814a 100755
--- a/misc/version.sh
+++ b/misc/version.sh
-if test -d .git -a -n "`git --version 2>/dev/null`"; then
+if test -d .git -a -n "$(git --version 2>/dev/null)"; then
     # update current version using git tags
-    GIT_VERSION=`git describe --tags --abbrev=4 --match="v[0-9].[0-9]*" 2>/dev/null`
+    GIT_VERSION=$(git describe --tags --abbrev=4 --match="v[0-9].[0-9]*" 2>/dev/null)
     CURR_VERSION=${GIT_VERSION}
 fi
```
Reference:
http://mywiki.wooledge.org/BashFAQ/082

namhyung: Looks ok

PR #485, opened 2018-09-07
chrome: Update perf event before getting task name

honggyukim:
It was missing to call update_perf_task_comm to get the finally updated
comm name.

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1431,6 +1431,8 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
 	while (!read_rstack(handle, &task) && !uftrace_done) {
 		struct uftrace_record *frs = task->rstack;
 
+		task->timestamp_last = frs->time;
+
 		if (!check_task_rstack(task, opts))
 			continue;
 
@@ -1442,8 +1444,6 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
 			dump_replay_event(ops, task);
 		else
 			dump_replay_func(ops, task);
-
-		task->timestamp_last = frs->time;
 	}
 
 	/* add duration of remaining functions */
@@ -1455,7 +1455,7 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
 		if (task->stack_count == 0)
 			continue;
 
-		last_time = task->rstack->time;
+		last_time = task->timestamp_last;
 
 		if (handle->time_range.stop && handle->time_range.stop < last_time)
 			last_time = handle->time_range.stop;
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -831,23 +831,26 @@ static void print_chrome_header(struct uftrace_dump_ops *ops,
 {
 	struct uftrace_chrome_dump *chrome = container_of(ops, typeof(*chrome), ops);
 	struct uftrace_info *info = &handle->info;
-	struct ftrace_task_handle *task;
+	struct uftrace_task *task;
 	int tid;
 	int i;
 
+	if (handle->hdr.feat_mask & PERF_EVENT)
+		update_perf_task_comm(handle);
+
 	pr_out("{\"traceEvents\":[\n");
 	for (i = 0; i < info->nr_tid; i++) {
-		task = &handle->tasks[i];
 		tid = info->tids[i];
+		task = find_task(&handle->sessions, tid);
 
 		pr_out("{\"ts\":0,\"ph\":\"M\",\"pid\":%d,"
 		       "\"name\":\"process_name\","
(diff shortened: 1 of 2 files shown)


honggyukim:
Hmm.. It shows task name properly but seems to hide `linux:schedule` events in `dump --chrome` output.

honggyukim:
Thanks. It works fine but I have one more problem. It shows `linux:schedule` in `dump --chrome` output but it doesn't properly show the end in chrome trace viewer as follows:
screen shot 2018-09-10 at 12 08 16 pm

Without `linux:schedule`, it shows the end correctly in `as` process.
screen shot 2018-09-10 at 12 08 29 pm

honggyukim:
Hmm.. It seems that timestamp for last exit related functions are postponed for some reasons with perf rewind fix.
```diff
--- orig.json              2018-09-10 12:06:09.910590685 +0900
+++ perf-rewind-fix.json   2018-09-10 12:24:12.843054161 +0900
@@ -1197,10 +1197,10 @@
 {"ts":16008308206320.322,"ph":"E","pid":165469,"name":"unlink"},
 {"ts":16008308206323.966,"ph":"B","pid":165469,"name":"unlink"},
 {"ts":16008308206346.203,"ph":"E","pid":165469,"name":"unlink"},
-{"ts":16008308084788.394,"ph":"E","pid":165481,"name":"exit"},
-{"ts":16008308084788.394,"ph":"E","pid":165481,"name":"xexit"},
-{"ts":16008308204303.787,"ph":"E","pid":165483,"name":"exit"},
-{"ts":16008308204303.787,"ph":"E","pid":165483,"name":"xexit"}
+{"ts":16008308206904.798,"ph":"E","pid":165481,"name":"exit"},
+{"ts":16008308206904.798,"ph":"E","pid":165481,"name":"xexit"},
+{"ts":16008308206904.798,"ph":"E","pid":165483,"name":"exit"},
+{"ts":16008308206904.798,"ph":"E","pid":165483,"name":"xexit"}
 ], "displayTimeUnit": "ns", "metadata": {
 "version":"uftrace v0.8.3-430-g83db ( dwarf python tui perf sched )",
 "recorded_time":"Fri Sep  7 20:29:25 2018",
```

namhyung:
I'm not sure what's the correct result.  But it seems original output had timestamp in an wrong order..  could you please check it in the raw dump?

honggyukim:
I'm wondering if those records are auto generated to make return records.

And here is the recorded data for reference.
[uftrace.data.issue485.tar.gz](https://github.com/namhyung/uftrace/files/2365092/uftrace.data.issue485.tar.gz)

honggyukim:
And I think the previous result is correct because `as` doesn't have to be executed or waited until linking is finished. So it seems that adding schedule event here makes some difference between them.

namhyung:
You're right.  It's uftrace added the exit records for remaining functions after task died.  And it uses a timestamp from other task for some reason.

namhyung:
OK, rstack for perf was shared across all tasks so it saw the timestamp of the last event.  I believe the [below patch](https://github.com/namhyung/uftrace/files/2365142/dump-Use-correct-timestamp-for-unfinished-functions.patch.txt) will fix it:

```diff
diff --git a/cmds/dump.c b/cmds/dump.c
index 53d8247e..082c8b63 100644
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1428,6 +1428,8 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
        while (!read_rstack(handle, &task) && !uftrace_done) {
                struct uftrace_record *frs = task->rstack;

+               task->timestamp_last = frs->time;
+
                if (!check_task_rstack(task, opts))
                        continue;

@@ -1439,8 +1441,6 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
                        dump_replay_event(ops, task);
                else
                        dump_replay_func(ops, task);
-
-               task->timestamp_last = frs->time;
        }

        /* add duration of remaining functions */
@@ -1452,7 +1452,7 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
                if (task->stack_count == 0)
                        continue;

-               last_time = task->rstack->time;
+               last_time = task->timestamp_last;

                if (handle->time_range.stop && handle->time_range.stop < last_time)
                        last_time = handle->time_range.stop;
```

namhyung:
Anyway could you please apply my patches separately (as I made them formal commits) since those fix separate problems?

honggyukim:
`--chrome` output looks good now. Thanks a lot! And could you please write a simple commit message? I will include your commit separetely.

honggyukim:
Otherwise, you can commit it to master yourself first. Then I will make this PR as a single commit.

namhyung:
The attached patches already have it.

honggyukim:
Oh.. I didn't open the files you attached. Just updated including them. Thanks!

PR #487, opened 2018-09-08
include dlfcn.h for RTLD_DEFAULT

kraj:
Fixes
plthook.c:128:41: error: use of undeclared identifier 'RTLD_DEFAULT'

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -5,6 +5,7 @@
 #include <sys/mman.h>
 #include <pthread.h>
 #include <assert.h>
+#include <dlfcn.h>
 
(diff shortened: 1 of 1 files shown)


kraj:
i  see it failing with same message with clang/musl toolchain

namhyung:
Yes, but that's for 'libmcount/plthook.c' not for 'debug/utils.c', right?  I don't see a dl function or macro in it.

kraj:
OK then we can drop this hunk. Do you want me to redo it ?

namhyung:
Yes please.

PR #502, opened 2018-09-16
replay: Fix format string mismatch in get_argspec_string()

gy741:
Hello.

In my opinion, The format string is mismatched.

"%hhx" is the correct format for unsigned char.

what do you think?

Thanks.

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -578,7 +578,7 @@ void get_argspec_string(struct ftrace_task_handle *task,
 					else if (isprint(c))
 						print_args("%c", c);
 					else
-						print_args("\\\\x%02hhx", c);
+						print_args("\\\\x%02hhx", (unsigned char)c);
 				}
 				print_args("\\\"");
 			}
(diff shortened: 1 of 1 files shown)


namhyung:
Well I don't think it's necessary as it doesn't change real behavior.

gy741:
Okay. :) Thank you for answer.

I have additional questions.

How about this problem(1213 line)?

"%ld" is the correct format for long.

https://github.com/namhyung/uftrace/blob/cd9027a56cc7edfa24393f43bfe3114cfc1fe030/utils/dwarf.c#L1190-L1215

namhyung:
Do you wanna change it to `%lu`?  Actually I don't care, but I'm ok if you want.

honggyukim:
Hmm.. I think we better print it in hexa-decimal instead of `%lu`.

namhyung:
hex for line numbers?

honggyukim:
I mean actual data, not line number.

Issue #371, opened 2018-05-26
check the shared memory space before start tracing


ParkHanbum:
my shared memory space was fulled before using uftrace. at this situation, I'm tried to use uftrace. deservedly, my trying was failed. but I could not found the reason immediatly. because it just printed the 'Bus error (core dumped)' which not real reason. 

```
m@m:~/git/uftrace$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            2.9G     0  2.9G   0% /dev
tmpfs           594M   24M  571M   4% /run
/dev/sda1        38G   31G  4.6G  87% /
tmpfs           2.9G  2.9G     0 100% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           2.9G     0  2.9G   0% /sys/fs/cgroup
tmpfs           594M   28K  594M   1% /run/user/108
tmpfs           594M     0  594M   0% /run/user/1000
m@m:~/git/uftrace$ ./uftrace -L.  ../test_pointer_arg
Bus error (core dumped)
```
if uftrace check shared memory before start tracing then can assist more helpful guide to user.

namhyung:
It's a kind of exceptional case so I'm not sure we can check every possible error case before running.  Also it's hard to predict how much memory uftrace will use..

ParkHanbum:
@namhyung it is right, I agree. 
I think this case can be solved easily. 
how about check shared memory to verify it was fulled before write it.

namhyung:
You can send a PR for it.

ParkHanbum:
@namhyung I forgot this issue, could you assign this issue to me?

namhyung:
Sure.

PR #518, opened 2018-09-21
uftrace: check freepace of shared memory before run

ParkHanbum:
When the shared memory space becomes full, running uftrace to trace
something will cause the "Bus error". This is rare case and is not
problem of uftrace.

because this case is rare, it will take time to find the actual problem
of an error. So, added a minimized logic that tells the user if the
shared memory is full.

Issue: #371

### DIFF
--- a/arch/x86_64/symbol.c
+++ b/arch/x86_64/symbol.c
@@ -35,7 +35,7 @@ int arch_load_dynsymtab_noplt(struct symtab *dsymtab,
 			pr_dbg2("found RELA section: %s\n",
 				elf_get_name(elf, &sec_iter, sec_iter.shdr.sh_name));
 
-			reloc_start = rel_iter.shdr.sh_offset + offset;
+			reloc_start = rel_iter.shdr.sh_addr + offset;
 			reloc_entsize = rel_iter.shdr.sh_entsize;
 		}
 		else if (sec_iter.shdr.sh_type == SHT_DYNSYM) {


uftrace.c
```diff
@@ -333,6 +338,21 @@ static bool has_time_unit(const char *str)
...
+	long space;
+
+	if (statvfs(SHMEM_PATH, &stat) != 0) {
+		pr_warn("Cannot access shared memory!!\n");
+		return;
+	}
+
+	space = stat.f_bsize * stat.f_bavail;
+	if (space < MIN_SHMEM_SPACE)
+		pr_warn("Freespace of shared memory too small!!\n");
```
namhyung: It should be "... memory is too small".

uftrace.c
```diff
@@ -994,6 +1014,7 @@ int main(int argc, char *argv[])
 	outfp = stdout;
 
 	argp_parse(&argp, argc, argv, ARGP_IN_ORDER, NULL, &opts);
+	check_shmem_freespace();
```
namhyung: You need to move it to `command_record()` as replay path doesn't need a shared memory.

uftrace.c
```diff
@@ -335,6 +340,21 @@ static bool has_time_unit(const char *str)
 		return false;
 }
 
+static void check_shmem_freespace()
+{
+	struct statvfs stat;
+	long space;
+
+	if (statvfs(SHMEM_PATH, &stat) != 0) {
+		pr_warn("Cannot access shared memory.\n");
```
namhyung: I think you can use `pr_err()` here.

uftrace.c
```diff
@@ -335,6 +340,21 @@ static bool has_time_unit(const char *str)
...
+		return;
+	}
+
+	space = stat.f_bsize * stat.f_bavail;
+	if (space < MIN_SHMEM_SPACE)
+		pr_warn("Freespace of shared memory is too small.\n");
+}
```
namhyung: Could you please move this function into `cmds/record.c`?

uftrace.c
```diff
@@ -335,6 +340,21 @@ static bool has_time_unit(const char *str)
...
+	long space;
+
+	if (statvfs(SHMEM_PATH, &stat) != 0) {
+		pr_warn("Cannot access shared memory.\n");
+		return;
+	}
+
+	space = stat.f_bsize * stat.f_bavail;
+	if (space < MIN_SHMEM_SPACE)
+		pr_warn("Freespace of shared memory is too small.\n");
```
namhyung: What do you want to do here?  Show warning and proceed?

uftrace.c
```diff
@@ -1057,6 +1077,7 @@ int main(int argc, char *argv[])
 
 	switch (opts.mode) {
 	case UFTRACE_MODE_RECORD:
+		check_shmem_freespace();
```
namhyung: To keep consistency I'd like to leave call to `command_record()` here.  Would you move this **into** the `command_record()`?

uftrace.c
namhyung: I was just asking you intended to proceed in this case rather than exit.  I'm ok with it and there's a chance that kernel can reclaim pages to make some available shmem.

namhyung:
Please change the commit message:
```
Issue: #371 
```
to
```
Fixed: #371 
```

ParkHanbum:
@namhyung thank you! have Updated!

cmds/record.c
```diff
@@ -59,6 +63,20 @@ static int thread_ctl[2];
...
 static bool has_sched_event;
 
+static void check_shmem_freespace()
+{
+	struct statvfs stat;
+	long space;
+
+	if (statvfs(SHMEM_PATH, &stat) != 0) {
+		pr_err("Cannot access shared memory.\n");
+		return;
```
namhyung: You don't need to add '\n' in the `pr_err()`.  And 'return' is not necessary too since it calls `exit()` internally.

cmds/record.c
```diff
@@ -59,6 +63,20 @@ static int thread_ctl[2];
 static bool has_perf_event;
 static bool has_sched_event;
 
+static void check_shmem_freespace()
```
namhyung: Please add 'void' in the function argument prototype.

ParkHanbum:
@namhyung consider it done.

ParkHanbum:
@namhyung I will happy if you review this again.

Issue #519, opened 2018-09-22
doc : Need to use different Package name when installing the uftrace.


ClaudiaJKang:
When installing the uftrace, 
RHEL, Fedora, and CentOS machines need to install different Package name. 

- Package names 

  |Ubuntu|RHEL, CentOS, Fedora|
  |---|---|
  |libdw-dev|elfutils-devel|
  |libpython2.7|python2-devel|
  |pandoc|pandoc|
  |libncursesw|ncurses-devel|
  |pkg-config|pkgconf-pkg-config|

How about adding this contents in INSTALL.md?

ParkHanbum:
@ClaudiaJKang is apt-get available in non-ubuntu linux package?

ClaudiaJKang:
@ParkHanbum Sorry. It's my mistake. apt-get is not available in non-ubuntu Linux. I fixed command in PR.

PR #520, opened 2018-09-22
doc : Add to use different package name when installing the uftrace

ClaudiaJKang:
This commit adds different package name in INSTALL.md

It is RHEL, CentOS, Fedora based on system install package name and command.

Fixed: #519

Signed-off-by: Claudia J. Kang <claudiajkang@gmail.com>


seungrye:
As far as I know, Red Hat (CentOS, Fedora) distributions use `yum` or `dnf` instead of `apt-get` when installing packages.

A more detailed description can be found at https://www.maketecheasier.com/install-software-in-various-linux-distros/

ClaudiaJKang:
@seungrye Thank you for your advice. I fixed command in PR.

INSTALL.md
```diff
@@ -12,6 +12,16 @@ source.
     $ make
     $ sudo make install
 
+On RHEL, Fedora and CentOS machines, following commands will build and install uftrace from source. 
+
+    $ sudo yum install elfutils-devel	# for debug info       (optional)
+    $ sudo yum install pandoc		# for man pages        (optional)
+    $ sudo yum install python2-devel	# for python scripting (optional)
+    $ sudo yum install ncurses-devel	# for TUI              (optional)
+    $ sudo yum install pkgconf-pkg-config	# for configure        (optional)
```
namhyung: Does this work all for recent Fedora, RHEL and CentOS?  I'm wondering `dnf` is not working on RHEL yet.

honggyukim: I just tested it (using dnf) on Fedora 28 (Twenty Eight) and it works fine. But not yet confirmed for other distros.

INSTALL.md
```diff
@@ -12,6 +12,16 @@ source.
     $ make
     $ sudo make install
 
+On RHEL, CentOS and Fedora machines, following commands will build and install uftrace from source. 
+
+    $ sudo yum install elfutils-devel	        # for debug info       (optional)
+    $ sudo yum install pandoc		            # for man pages        (optional)(Fedora only)
+    $ sudo yum install python2-devel	        # for python scripting (optional)
+    $ sudo yum install ncurses-devel	        # for TUI              (optional)
+    $ sudo yum install pkgconfig	            # for configure        (optional)(Fedora use pkgconf-pkg-config)
```
namhyung: Could please align the comments so that they can start from the same position?  Also I'd prefer saying single parenthesis separated by comma like "(optional, Fedora only)".

ClaudiaJKang: @namhyung Thanks. I fixed it.

namhyung:
Basically I'm ok with the above change in the script assuming you tested it works on the systems.

namhyung:
Maybe the quick guide just mention the script not the individual packages.

namhyung:
Great! thanks for your work.

ClaudiaJKang:
@namhyung Sorry, I don't understand. Could you please be more specific about "Maybe the quick guide just mention the script not the individual packages."?

namhyung:
I mean that instead of calling package managers for each distro, it could be replaced to call the script above.

honggyukim:
@ClaudiaJKang You're script looks good. But is it okay to write the distro name in lower cases? It seems that my ubuntu machine shows the name as "Ubuntu" in `/etc/os-release`.

I also think that we better just write to call the script above rather than writing the package install commands for each distros because quick guide is better to be simple.

namhyung:
Something like this?

```diff
diff --git a/INSTALL.md b/INSTALL.md
index daf15036..f66ea665 100644
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,11 +4,7 @@ QUICK GUIDE
 On Ubuntu machines, following commands will build and install uftrace from
 source.

-    $ sudo apt-get install libdw-dev            # for debug info       (optional)
-    $ sudo apt-get install pandoc               # for man pages        (optional)
-    $ sudo apt-get install libpython2.7-dev     # for python scripting (optional)
-    $ sudo apt-get install libncursesw5-dev     # for TUI              (optional)
-    $ sudo apt-get install pkg-config           # for configure        (optional)
+    $ sudo misc/install-deps.sh             # optional
     $ make
     $ sudo make install

```

ClaudiaJKang:
@namhyung @honggyukim Thank you for guiding me. I fixed command in PR. Also, `pandoc` installation logic of RHEL, CentOS were updated to the script.

@honggyukim You are right. The `name` value is "Ubuntu" in `/etc/os-release` in Ubuntu machine. 

But, according to [os-release parameter](http://0pointer.de/public/systemd-man/os-release.html#ID=),
> **NAME=**    
> A string identifying the operating system, without a version component, and suitable for presentation to the user. If not set, defaults to "NAME=Linux". Example: "NAME=Fedora" or "NAME="Debian GNU/Linux"".

> **ID=**    
> A lower-case string (no spaces or other characters outside of 0-9, a-z, ".", "_" and "-") identifying the operating system, excluding any version information and suitable for processing by scripts or usage in generated filenames. If not set, defaults to "ID=linux". Example: "ID=fedora" or "ID=debian".

So I selected the `ID` in `/etc/os-release`'s options due to lower-case string.

honggyukim:
Hi @ClaudiaJKang,

> So I selected the ID in /etc/os-release's options due to lower-case string.

Good. I think we don't have to run `sed` command to get distro name if so as follows:
```sh
distro=$(cat /etc/os-release | grep "^ID=" | cut -d'=' -f2)
```
I think we better give new lines for each case statements for better indentation and slightly modified the `*` case message.
```sh
case $distro in
    "ubuntu")
        apt-get install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config ;;
    "fedora")
        dnf install pandoc elfutils-devel python2-devel ncurses-devel pkgconf-pkg-config ;;
    "rhel" | "centos")
        rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
        yum install pandoc elfutils-devel python2-devel ncurses-devel pkgconfig ;;
    *) # we can add more install command for each distros.
       echo "\"$distro\" is not supported distro, so please install packages manually.";;
esac
```
Everything is good to me now. Thanks for your effort!

honggyukim:
Ah.. I have one more request, please remove `misc/ubuntu-install-deps.sh` or rename it to `misc/install-deps.sh` in this PR because your script includes it.

ClaudiaJKang:
@honggyukim Thanks! I fixed. 

If `sed` removes in the command to getting distro name, the result of CentOS and RHEL contains a specific character *"* such as "centos" and "rhel". Thus, if a specific character " existed, switch case condition to check distro name will not be satisfied.

So I kept it unchanged.

INSTALL.md
```diff
@@ -1,14 +1,10 @@
...
-On Ubuntu machines, following commands will build and install uftrace from
+On linux distros machines, following commands will build and install uftrace from
 source.
 
-    $ sudo apt-get install libdw-dev            # for debug info       (optional)
-    $ sudo apt-get install pandoc               # for man pages        (optional)
-    $ sudo apt-get install libpython2.7-dev     # for python scripting (optional)
-    $ sudo apt-get install libncursesw5-dev     # for TUI              (optional)
-    $ sudo apt-get install pkg-config           # for configure        (optional)
+    $ sudo misc/install-deps.sh                 # optional
```
honggyukim: Thanks. but I think it'd be better to change `# optional` to `# optional for special features`.

ClaudiaJKang: @honggyukim Thanks for the feedback! It's done.

PR #528, opened 2018-10-04
doc: Show uftrace.html in the front page

honggyukim:


honggyukim:
I've just created a sample page to increase uftrace.html document access. It doesn't have to be merged right now but I would like to get some feedback. Thanks!

honggyukim:
Please take a look at the below page
- https://github.com/honggyukim/uftrace/tree/doc/promote-uftrace.html

namhyung:
While it's great I guess the animated GIF was better for (lazy) users. :)  How about adding just a link below it?

honggyukim:
Understand that. It's sad that github doesn't allow embedding the page in an iframe.

honggyukim:
What about leave the site link at the top of our github page instead?
It's written as "Function (graph) tracer for user-space", but didn't set the Website yet.

namhyung:
That might be possible.  But I feel that the website should be a kind of official place to get all of the information about uftrace.  To do that, I think we should at least make an index page other than just slide.

honggyukim:
Would it be okay if I create an `index.html` on base directory of https://uftrace.github.io/ and makes it redirect to https://uftrace.github.io/slide/ until the official page is ready?

honggyukim:
I made a redirection from https://uftrace.github.io/ to https://uftrace.github.io/slide/ so that we can use  the main page for the other purpose later on.

I think we can add a link to https://uftrace.github.io/slide/ in the top of the page just besides "Function (graph) tracer for user-space" if you think it's okay. It can be changed to the main page when it's ready in the future.

namhyung:
> I think we can add a link to https://uftrace.github.io/slide/ in the top of the page just besides "Function (graph) tracer for user-space" if you think it's okay. It can be changed to the main page when it's ready in the future.

OK, let's do it.

honggyukim:
Thanks!

namhyung:
Done.

PR #534, opened 2018-10-08
utils: Move the definition of `demangle_full()`

taeguk:
`demangle_full()` is only used in `utils/demangle.c`.  So, move the definition of `demangle_full()` from `utils/symbol.h` to `utils/demangle.c`.

### DIFF
--- a/utils/demangle.c
+++ b/utils/demangle.c
@@ -1571,6 +1571,12 @@ static char *demangle_full(char *str)
 
 	return symname;
 }
+#else
+static inline char *demangle_full(char *str)
+{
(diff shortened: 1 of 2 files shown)


taeguk:
@namhyung Hmm.. but, `demangle_full()` is not used except in `utils/demangle.c`. If it is also used in another codes, the technique you said is right, of course.
Keeping it in the header file is still proper even though it is only used in `demangle_full()` ?

namhyung:
I have something in mind which calling `demangle_full()` is necessary, but not having time to play with it yet.

taeguk:
Okay, I close this PR for now. Thanks.

PR #538, opened 2018-10-09
Misc: Fetch latest elfutils version and install

DanielTimLee:
This commit allows fetching latest elfutils version (tag)
by using git ls-remote and install it.

### DIFF
--- a/misc/install-elfutils.sh
+++ b/misc/install-elfutils.sh
@@ -42,7 +42,9 @@ shift $((OPTIND - 1))
 
 mkdir -p ${builddir} && cd ${builddir}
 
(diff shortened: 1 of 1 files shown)


honggyukim:
@DanielTimLee Thanks for your work, but I'm not so sure if there's a reason to use the latest elfutils version. I actually used somewhat old version to minimize the compilation target. I remember that I had to compile more targets that we don't have to use in higher version. And I think we better use only tested version for this because I worry that if the latest version changes something that makes our script fail.

DanielTimLee:
Oh. I see. 
I didn't realize there was a specific purpose with versioning.

PR #542, opened 2018-10-12
build: Extract build scripts to build other programs

ParkHanbum:
The uftrace makefile contains build scripts that build misc programs.
This commit extracts these scripts into a new makefile under the misc
subdirectory to simpify uftrace makefile.


Makefile
```diff
@@ -120,23 +114,11 @@ LIBMCOUNT_TARGETS += libmcount/libmcount-single.so libmcount/libmcount-fast-sing
 
 _TARGETS := uftrace libtraceevent/libtraceevent.a
 _TARGETS += $(LIBMCOUNT_TARGETS) libmcount/libmcount-nop.so
-_TARGETS += misc/demangler misc/symbols
```
namhyung: This removes them from default make target so it changes the behavior.  Please keep it unchanged.

misc/Makefile
```diff
@@ -0,0 +1,131 @@
...
+
+COMMON_CFLAGS := -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
+COMMON_CFLAGS += -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
+COMMON_LDFLAGS := -lrt -ldl -pthread $(LDFLAGS)
+ifneq ($(elfdir),)
+  COMMON_CFLAGS  += -I$(elfdir)/include
+  COMMON_LDFLAGS += -L$(elfdir)/lib
+endif
+
+COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
```
namhyung: Do these need to be repeated?  I think you can export them in the main Makefile and just use it here.

misc/Makefile
```diff
@@ -0,0 +1,131 @@
...
+SYMBOLS_SRCS := $(srcdir)/misc/symbols.c
+SYMBOLS_OBJS := $(srcdir)/misc/symbols.o
+
+MISC_DEP_SRCS := $(srcdir)/utils/session.c
+MISC_DEP_SRCS += $(srcdir)/utils/demangle.c $(srcdir)/utils/rbtree.c
+MISC_DEP_SRCS += $(srcdir)/utils/utils.c $(srcdir)/utils/debug.c
+MISC_DEP_SRCS += $(srcdir)/utils/filter.c $(srcdir)/utils/dwarf.c
+MISC_DEP_SRCS += $(srcdir)/utils/auto-args.c
+MISC_DEP_SRCS += $(wildcard $(srcdir)/utils/symbol*.c)
+MISC_DEP_OBJS := $(patsubst $(srcdir)/utils/%.c,$(objdir)/misc/%.o,$(MISC_DEP_SRCS))
```
namhyung: Why did you separate it?  Isn't it only used by the `symbols`?

Makefile
ParkHanbum: @namhyung it can be same effect by adding "@ $ (MAKE) -C $ (srcdir) / misc" under 'all:' command. like below:

misc/Makefile
ParkHanbum: @namhyung I thought these objects would be used by other misc programs.

Makefile
namhyung: Why not adding `misc` to `TARGETS`?  I don't know it's good to have separate Makefile.

misc/Makefile
namhyung: You can separate it when it's actually used.

Makefile
ParkHanbum: @namhyung This is because each time you add a program to misc you need to add the relevant script to the makefile. I want to add a new program to misc, but before that I want to clean up the Makefile.

ParkHanbum:
@namhyung do you ok with this?

Makefile
namhyung: Anyway you still need to change misc/Makefile.

namhyung:
I don't like the duplications but otherwise I'm ok with the idea.

ParkHanbum:
@namhyung I have updated this pr. remove duplication of each makefiles, keep the order even when user has specified multi-thread by '-j' option. it has tested in my environments like ubuntu 18.04 with make 4.1, manjaro with make 4.2.1 and it is OK to me. but I think I have change the makefile too many, so I'm worried about it work well without any problems.

ParkHanbum:
@namhyung would you review this again, please?
this was tested with ubuntu-18.04 and recently manjaro. and test was succeed.

namhyung:
You want to do 3 different things in a commit,  I'd like to suggest to break them into 3 commits.  Also I don't aware of the problem in #564.  How often does it reproduce?  Do you always see the failure with `-j 8`?

Also I don't want to run configure script everytime I build it.  The intention was to run it once and use the (previous) result until there's a change in the dependencies.

ParkHanbum:
It will always occur in make 4.21 and will be the same in later versions. this issue did not make build fail. it is done the build but does not include libraries correctly like libdw, libelf. 

even if user installed these libraries, will be ignored. for more info about this issue, refer #564.

namhyung:
Does it occur only when `configure` script is not run before `make`, right?

PR #551, opened 2018-10-16
misc: Fix useless use of cat

gy741:
Hello.

I modified the UUOC(useless use of cat).

what do you think?

Thanks.


gy741:
Oops, My mistake. I will add.

PR #559, opened 2018-10-20
 doc: Add description of -a option

Simhyunjeong16:
This PR is the addition of infomation about -a in uftrace-record.md

### DIFF
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -122,6 +122,7 @@ OPTIONS
 :   Automatically record arguments and return values of known functions.
     These are usually functions in standard (C language or system) libraries
     but if debug info is available it includes functions in the user program.
+    You must compile the target program with debug info (using -g option).
 
 \--num-thread=*NUM*
 :   Use NUM threads to record trace data.  Default is 1/4 of online CPUs (but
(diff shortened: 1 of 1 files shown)


doc/uftrace-record.md
```diff
@@ -512,6 +513,24 @@ all (supported) functions.  It shows two arguments for "main" and a single
 string argument for "puts".  If you simply want to see all arguments and
 return values of every functions (if supported), use -a/\--auto-args option.
 
+For example,
+
+    $ gcc -g -pg ../tests/s-abc.c -o abc
```
ClaudiaJKang: I think that line 502~509's example is the better than this PR's example. 
Because example about `-a` option means that simply want to see all arguments and return values of every function about "puts" example.


> If you simply want to see all arguments and return values of every functions (if supported), use -a/\--auto-args option..

So I suggest the following example.

```sh
$ uftrace record -a . ./hello
Hello world

$ uftrace replay
....result......
```

doc/uftrace-record.md
```diff
@@ -122,6 +122,7 @@ OPTIONS
 :   Automatically record arguments and return values of known functions.
     These are usually functions in standard (C language or system) libraries
     but if debug info is available it includes functions in the user program.
+    You must compile with the -g option.
```
namhyung: You must compile the target program with debug info (using `-g` option).

doc/uftrace-record.md
```diff
@@ -512,6 +513,24 @@ all (supported) functions.  It shows two arguments for "main" and a single
 string argument for "puts".  If you simply want to see all arguments and
 return values of every functions (if supported), use -a/\--auto-args option.
 
+For example,
+
+    $ gcc -g -pg ../tests/s-autoargs.c -o abc
+    $ uftrace record -a ./abc hello
```
namhyung: You'd better make the program name same as the source name to avoid confusion.  And I assume the base directory of the source so it should be:

```
$ gcc -g -pg -o autoargs tests/s-autoargs.c
$ uftrace record -a ./autoargs hello
```

doc/uftrace-record.md
```diff
@@ -512,6 +513,24 @@ all (supported) functions.  It shows two arguments for "main" and a single
...
+       1.663 us [ 24332] | __monstartup();
+       1.313 us [ 24332] | __cxa_atexit();
+                [ 24332] | main(2, 0x7fff61c8cbb8) {
+     165.122 us [ 24332] |   strlen("autoargs test") = 13;
+       0.421 us [ 24332] |   calloc(1, 14) = 0x562f52fdfe00;
+       0.330 us [ 24332] |   free(0x562f52fdfe00);
+       0.852 us [ 24332] |   strcmp("hello", "hello") = 0;
+      26.911 us [ 24332] |   puts("hello") = 6;
+     198.805 us [ 24332] | } = 0; /* main */
+
```
namhyung: And for completeness, you need to add it to the doc of `live` command as well.

PR #561, opened 2018-10-20
script : Add arg.py

LeeYeEun2:
This patch adds arg.py.

The output of this script is as follows : 

```
$ uftrace record -a -S scripts/arg.py tests/abc
main has args
foo has args
bar has args
strcmp has args
strcmp has retval
bar has retval
bar has args
strcmp has args
strcmp has retval
bar has retval
bar has args
strcmp has args
strcmp has retval
bar has retval
foo has retval
many has args
many has retval
pass has args
check has args
check has retval
pass has retval
main has retval
```

Signed-off-by: leeyeeun <dldpdms0229@duksung.ac.kr>

### DIFF
--- a/scripts/arg.py
+++ b/scripts/arg.py
@@ -1,0 +1,10 @@
+def uftrace_entry(ctx):
+        if "args" in ctx:
+        	print(ctx["name"] + " has args")
+	        
+def uftrace_exit(ctx):
+	if "retval" in ctx:
+    		print(ctx["name"] + " has retval")
+
+				
+


scripts/arg.py
```diff
@@ -0,0 +1,10 @@
+def uftrace_entry(ctx):
+        if "args" in ctx:
```
honggyukim: Please keep the indentation to 4 white spaces.

scripts/arg.py
```diff
@@ -0,0 +1,10 @@
...
+        	print(ctx["name"] + " has args")
+	        
+def uftrace_exit(ctx):
+	if "retval" in ctx:
+    		print(ctx["name"] + " has retval")
```
honggyukim: Likewise, please keep the indentation.

scripts/arg.py
```diff
@@ -0,0 +1,10 @@
...
+
+				
+
```
honggyukim: Remove all those 3 empty lines.

namhyung:
I don't want to add this kind of simple scripts anymore.  There are some example scripts already for curious users.

PR #563, opened 2018-10-20
script : Add main_errono.py

LeeYeEun2:
This patch adds main_errono.py.
This script explains the meaning of the value returned when the program exits. 

The output of this script is as followes : 
```
$ uftrace record -a -S scripts/errono.py tests/abc
completed successfully.

```

Signed-off-by: leeyeeun <lucy0229@naver.com>

### DIFF
--- a/scripts/main_errono.py
+++ b/scripts/main_errono.py
@@ -1,0 +1,79 @@
+def uftrace_exit(ctx):
+	main_error(ctx)
+
+def main_error(ctx):
+	if ctx["name"] == "main" :
+		if "retval" in ctx:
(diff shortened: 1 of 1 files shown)


honggyukim:
Hmm.. I'm sorry to say this but I don't understand the point of this script.

ParkHanbum:
This script seems to describe what the program's return value means.

namhyung:
Return value of the main function doesn't mean the error number.  And you can easily see it with the `$?` variable in the shell.  I don't think it's worth having a script for this info.

Issue #566, opened 2018-10-21
test: s-forkexec based test need dependency of t-abc binary.


DanielTimLee:
`s-forkexec` script expect to execute `t-abc`,
but ` tests/t088_graph_session.py` doesn't include build command for `t-abc`.

```bash
=========== original ===========
# Function Call Graph for 't-forkexec' (session: 436d2dda591bc426)
========== FUNCTION CALL GRAPH ==========
# TOTAL TIME   FUNCTION
  491.093 us : (1) t-forkexec
  491.093 us : (1) main
  126.341 us :  +-(1) fork
             :  | 
  353.045 us :  +-(1) waitpid
             :  | 
   20.940 us :  +-(1) execl
             :  | 
             :  +-(1) exit


===========  result  ===========
 (1) t-forkexec
 (1) main
  +-(1) fork
  | 
  +-(1) waitpid
  | 
  +-(1) execl
  | 
  +-(1) exit
=========== expected ===========
 (1) t-abc
 (1) main
 (1) a
 (1) b
 (1) c
 (1) getpid
 (1) t-forkexec
 (1) main
  +-(1) fork
  | 
  +-(1) waitpid
  | 
  +-(1) execl
```

PR #567, opened 2018-10-21
TEST: s-forkexec based test need dependency of t-abc binary.

DanielTimLee:
This PR fixes s-forkexec based test's missing dependency of `t-abc`

Fixed: #566


honggyukim:
Looks good, but can you combine those two commits into a single one? Those make the same changes so better to be a single commit.

DanielTimLee:
Done!

PR #568, opened 2018-10-21
script: Add elapsed_time to script_info

pengin7384:
Current script_info is a bit lack of data, so I add elapsed_time
to script_info. And I add print(ctx["elapsed_time"]) to info.py.

Signed-off-by: pengin7384 <pengin7384@gmail.com>

### DIFF
--- a/cmds/script.c
+++ b/cmds/script.c
@@ -168,8 +168,7 @@ int command_script(int argc, char *argv[], struct opts *opts)
 
 	ret = open_data_file(opts, &handle);
 	uftrace_info = &handle.info;
-	info.elapsed_time = xstrdup(uftrace_info->elapsed_time);
-
+	info.elapsed_time = atof(uftrace_info->elapsed_time);
 	if (ret < 0) {
 		pr_warn("cannot open record data: %s: %m\n", opts->dirname);
 		return -1;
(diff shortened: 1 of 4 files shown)


pengin7384:
It is true that elapsed time information can not be obtained at record time, but I think it is good information to use when doing script -S. If you do record -S and ctx ['elapsed_time'], you get null, but it looks like there is no problem. What do you think?

pengin7384:
pengin@ubuntu:~/Desktop/project/uftrace/scripts$ uftrace record -S info.py a.outTrue
v0.9-100-g6bf4d ( dwarf python tui perf sched )
('a.out',)
0.0

pengin@ubuntu:~/Desktop/project/uftrace/scripts$ uftrace script -S info.py
False
v0.9-100-g6bf4d ( dwarf python tui perf sched )
()
0.06264812

I changed elapsed_time to numeric type. Now, if it is in record, it will output 0 automatically. If it is not in record, time will be output.

namhyung:
Rather than double, I'd like it to be int (or long) in nano-second unit.

pengin7384:
(Double) 0.06264812 -> (int)6264812
Like this?

namhyung:
It's nanosecond so you need to convert it to a proper number.  I think you should do like:

  0.06264812 -> 62648120
  1.23456 -> 1234560000

pengin7384:
I am modifying. I think if the elapsed time exceeds 3 seconds, overflow will occur. What should I do in this case?

namhyung:
Sorry for being late.  You can use a 64-bit integer.

PR #581, opened 2018-10-25
slide: Add more compile option to foobar.c example

ClaudiaJKang:
This PR updates some points in uftrace.github.io's foobar.c example. 

   - Replace command `$` to `$ cat foobar.c`.
   - Add more compile option to foobar.c example

Before:

After: 

Signed-off-by: Claudia J. Kang <claudiajkang@gmail.com>


honggyukim:
I would like to put `cat` and `gcc` commands just above the contents if so. Can you change the above slide as follows?

image

I'm okay with `-finstrument-functions` example. Thanks for fixing the slides!

ClaudiaJKang:
@honggyukim I fixed it! 

_home_claudia_uftrace_doc_uftrace html

PR #587, opened 2018-11-08
build: adding version.h dependency manually

AnsBal:
When running make with multiple threads, sometimes the script.o target fail to build because of missing version.h
It happens because script.c is including version.h but this dependency is missing in the Makefile.

Signed-off-by: Anas Balboul anasbalbo@gmail.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,9 @@ TARGETS  := $(patsubst %,$(objdir)/%,$(_TARGETS))
 UFTRACE_SRCS := $(srcdir)/uftrace.c $(wildcard $(srcdir)/cmds/*.c $(srcdir)/utils/*.c)
 UFTRACE_OBJS := $(patsubst $(srcdir)/%.c,$(objdir)/%.o,$(UFTRACE_SRCS))
 
+UFTRACE_OBJS_VERSION := $(objdir)/cmds/script.o $(objdir)/cmds/tui.o
+UFTRACE_OBJS_VERSION += $(objdir)/cmds/dump.o $(objdir)/cmds/info.o
+
 DEMANGLER_SRCS := $(srcdir)/misc/demangler.c $(srcdir)/utils/demangle.c $(srcdir)/utils/debug.c
 DEMANGLER_OBJS := $(patsubst $(srcdir)/%.c,$(objdir)/%.o,$(DEMANGLER_SRCS))
 
@@ -194,6 +197,8 @@ config: $(srcdir)/configure
 $(LIBMCOUNT_UTILS_OBJS): $(objdir)/libmcount/%.op: $(srcdir)/utils/%.c $(LIBMCOUNT_DEPS)
 	$(QUIET_CC_FPIC)$(CC) $(LIB_CFLAGS) -c -o $@ $<
 
+$(objdir)/libmcount/mcount.op: $(objdir)/version.h 
+
 $(LIBMCOUNT_OBJS): $(objdir)/%.op: $(srcdir)/%.c $(LIBMCOUNT_DEPS)
 	$(QUIET_CC_FPIC)$(CC) $(LIB_CFLAGS) -c -o $@ $<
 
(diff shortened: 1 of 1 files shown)


AnsBal:
Thank you for the review !

After investigating other dependencies of version.h, I found that it is included in symbols, demangler, uftrace, count, tui, script, and dump. The Makefile contained already 3 of these dependencies (uftrace, symbols and demangler). The rest should be added manually I think. 
Is it possible for the compiler to make the dependency files ?
 
I usually use 32 CPU cores/threads. The problem happens infrequently with 8 cores and less.

Makefile
```diff
@@ -126,6 +126,9 @@ TARGETS  := $(patsubst %,$(objdir)/%,$(_TARGETS))
 UFTRACE_SRCS := $(srcdir)/uftrace.c $(wildcard $(srcdir)/cmds/*.c $(srcdir)/utils/*.c)
 UFTRACE_OBJS := $(patsubst $(srcdir)/%.c,$(objdir)/%.o,$(UFTRACE_SRCS))
 
+UFTRACE_OBJS_VERSION := $(objdir)/cmds/script.o $(objdir)/cmds/tui.o
+UFTRACE_OBJS_VERSION += $(srcdir)/cmds/dump.o $(srcdir)/cmds/info.o
```
namhyung: Hmm.. shouldn't it be `$(objdir)`?

Makefile
```diff
@@ -139,7 +142,7 @@ SYMBOLS_OBJS := $(patsubst $(srcdir)/%.c,$(objdir)/%.o,$(SYMBOLS_SRCS))
 
 UFTRACE_ARCH_OBJS := $(objdir)/arch/$(ARCH)/uftrace.o
 
-UFTRACE_HDRS := $(wildcard $(srcdir)/*.h $(srcdir)/utils/*.h) $(srcdir)/version.h
+UFTRACE_HDRS := $(filter-out $(srcdir)/version.h,$(wildcard $(srcdir)/*.h $(srcdir)/utils/*.h))
```
namhyung: It'd be nice if you leave it as is and get rid of the first commit.

Makefile
```diff
@@ -242,7 +248,10 @@ $(objdir)/misc/demangler.o: $(srcdir)/misc/demangler.c $(objdir)/version.h $(COM
 $(objdir)/misc/symbols.o: $(srcdir)/misc/symbols.c $(objdir)/version.h $(COMMON_DEPS)
 	$(QUIET_CC)$(CC) $(SYMBOLS_CFLAGS) -c -o $@ $<
 
-$(filter-out $(objdir)/uftrace.o,$(UFTRACE_OBJS)): $(objdir)/%.o: $(srcdir)/%.c $(COMMON_DEPS)
+$(UFTRACE_OBJS_VERSION): $(objdir)/%.o: $(srcdir)/%.c $(objdir)/version.h $(COMMON_DEPS)
+	$(QUIET_CC)$(CC) $(UFTRACE_CFLAGS) -c -o $@ $<
+
+$(filter-out $(objdir)/uftrace.o $(UFTRACE_OBJS_VERSION), $(UFTRACE_OBJS)): $(objdir)/%.o: $(srcdir)/%.c $(COMMON_DEPS)
```
namhyung: I'm not sure but just adding following line is not enough?

```Makefile
$(UFTRACE_OBJS_VERSION): $(objdir)/version.h
```

Makefile
AnsBal: Yes, object files should be $(objdir).

Makefile
AnsBal: Yes, it is enough. I tried it and it works. It avoids redundant dependencies and compile command.

Makefile
```diff
@@ -194,7 +197,10 @@ config: $(srcdir)/configure
 $(LIBMCOUNT_UTILS_OBJS): $(objdir)/libmcount/%.op: $(srcdir)/utils/%.c $(LIBMCOUNT_DEPS)
 	$(QUIET_CC_FPIC)$(CC) $(LIB_CFLAGS) -c -o $@ $<
 
-$(LIBMCOUNT_OBJS): $(objdir)/%.op: $(srcdir)/%.c $(LIBMCOUNT_DEPS)
+$(objdir)/libmcount/mcount.op: $(srcdir)/libmcount/mcount.c $(LIBMCOUNT_DEPS) $(objdir)/version.h 
+	$(QUIET_CC_FPIC)$(CC) $(LIB_CFLAGS) -c -o $@ $<
+	
+$(filter-out $(objdir)/libmcount/mcount.op, $(LIBMCOUNT_OBJS)): $(objdir)/%.op: $(srcdir)/%.c $(LIBMCOUNT_DEPS)
```
namhyung: Please do the same for the libmcount as well:

```
$(objdir)/libmcount/mcount.op: $(objdir)/version.h
```

PR #595, opened 2018-11-13
uftrace: Add --no-args option to hide args/retval

honggyukim:
It needs to provide a way to hide args/retval to reduce the amount of
dump output.  It has to be implemented for replay command later on.

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -95,6 +95,7 @@ enum options {
 	OPT_diff_policy,
 	OPT_event_full,
 	OPT_record,
+	OPT_no_args,
 	OPT_libname,
 	OPT_match_type,
 	OPT_no_randomize_addr,
@@ -179,6 +180,7 @@ __used static const char uftrace_help[] =
 "                             regex)\n"
 "      --max-stack=DEPTH      Set max stack depth to DEPTH (default: "
 	stringify(OPT_RSTACK_MAX) ")\n"
+"      --no-args              Do not show arguments and return value\n"
 "      --no-comment           Don't show comments of returned functions\n"
 "      --no-event             Disable (default) events\n"
 "      --no-sched             Disable schedule events\n"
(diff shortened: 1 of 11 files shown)


honggyukim:
It's a draft implementation and it has to be implemented for replay command as well.

namhyung:
> It's a draft implementation and it has to be implemented for replay command as well.

Well, you could simply pass the `opts->no_args` to `get_argspec_string()` and make it return an empty string.

honggyukim:
> Well, you could simply pass the opts->no_args to get_argspec_string() and make it return an empty string.

Right. I just updated it that handles both replay and dump output. Thanks!

honggyukim:
Okay. I will make it as `print_args`.

honggyukim:
There is an internal macro `print_args` in `cmds/replay.c`, so I just updated it using `show_args` as you suggested. Thanks!

namhyung:
Maybe we could make it global variable.  But some code is shared between uftrace binary and libmcount so it might be confusing if some setting has different value for some reason.

namhyung:
And I'm hesitating to add new option everytime we introduce new features..

honggyukim:
As I wrote in #599, it can be combined into `--args="off"` option.

honggyukim:
If the option name `--no-args` is not sure yet, then would it be better to disable argument info in `dump --chrome` output as of now? If some arguments contains not printable characters, then chrome browser fails to load the `.json` file.  

It's already handled in 58ef456a, but it's not clearly fixed for some reasons.

honggyukim:
@namhyung I just rebased and updated this PR again. Please have a look. Thanks!

uftrace.c
```diff
@@ -130,6 +131,7 @@ __used static const char uftrace_help[] =
 "      --avg-self             Show average/min/max of self function time\n"
 "      --avg-total            Show average/min/max of total function time\n"
 "  -a, --auto-args            Show arguments and return value of known functions\n"
+"      --no-args              Do not show arguments and return value\n"
```
namhyung, line 134: I know why you placed it here, but I think we should maintain it in alphabetical order.

namhyung (commented):
Looks ok, one minor nit..

uftrace.c
honggyukim: Right. I just reordered it. Thanks.

namhyung (approved):
LGTM

PR #582, opened 2018-10-25
uftrace: Change option --no-comment to --comment=<value>

BbChip0103:
Hello, This PR is to change option --no-comment to --comment=\<value>

This patch makes it possible to see function names and arguments together.
It is more efficient to read recursive functions.
To add this option, I store arguments string in 'fstack' structure(utils/fstack.h).

How do you think this?

Fixed: #533

Signed-off-by: bbchip0103 <bbchip13@gmail.com>

```
$ uftrace -A .@Arg1 --comment=with-args tests/a.out
   1.324 us [ 32579] | __monstartup(0x400540);
   0.598 us [ 32579] | __cxa_atexit(0x7f3325a6b360);
            [ 32579] | main(1) {
            [ 32579] |   fib(5) {
            [ 32579] |     fib(4) {
            [ 32579] |       fib(3) {
   1.593 us [ 32579] |         fib(2);
   0.114 us [ 32579] |         fib(1);
   2.479 us [ 32579] |       } /* fib(3) */
   0.109 us [ 32579] |       fib(2);
   3.001 us [ 32579] |     } /* fib(4) */
            [ 32579] |     fib(3) {
   0.105 us [ 32579] |       fib(2);
   0.142 us [ 32579] |       fib(1);
   0.721 us [ 32579] |     } /* fib(3) */
   4.133 us [ 32579] |   } /* fib(5) */
   4.680 us [ 32579] | } /* main(1) */
```

PR #599, opened 2018-11-17
replay: Make arguments and retvals colorful

honggyukim:
Automatic argument display is great but it sometimes looks too verbose
with too many arguments and return values.

This patch makes arguments and return values more colorful if they are
some special types.  The color scheme is as follows:

  - string types     : magenta
  - char types       : magenta
  - enum types       : blue
  - function pointer : cyan
  - parens and comma : bold

This would make users more comfortable to identify those values.

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -482,9 +482,9 @@ void get_argspec_string(struct uftrace_task_reader *task,
 	assert(arg_list && !list_empty(arg_list));
 
 	if (needs_paren)
-		print_args("(");
+		print_args("%s(%s", color_bold, color_reset);
 	else if (needs_assignment)
-		print_args(" = ");
+		print_args("%s = %s", color_bold, color_reset);
 
 	list_for_each_entry(spec, arg_list, list) {
 		char fmtstr[16];
@@ -498,7 +498,7 @@ void get_argspec_string(struct uftrace_task_reader *task,
 			continue;
 
 		if (i > 0)
-			print_args(", ");
+			print_args("%s, %s", color_bold, color_reset);
 
 		memset(val.v, 0, sizeof(val));
 		fmt = ARG_SPEC_CHARS[spec->fmt];
@@ -583,8 +583,11 @@ void get_argspec_string(struct uftrace_task_reader *task,
 				}
 				print_args("\\\"");
 			}
-			else
+			else {
+				print_args("%s", color_string);
 				print_args("\"%.*s\"", slen + newline, str);
+				print_args("%s", color_reset);
+			}
 
 			/* std::string can be represented as "TEXT"s from C++14 */
 			if (spec->fmt == ARG_FMT_STD_STRING)
@@ -597,8 +600,11 @@ void get_argspec_string(struct uftrace_task_reader *task,
 			char c;
 
 			memcpy(&c, data, 1);
-			if (isprint(c))
+			if (isprint(c)) {
+				print_args("%s", color_string);
 				print_args("'%c'", c);
+				print_args("%s", color_reset);
+			}
 			else
 				print_args("'\\x%02hhx'", c);
 			size = 1;
@@ -637,8 +643,11 @@ void get_argspec_string(struct uftrace_task_reader *task,
 						 task->rstack->time,
 						 (uint64_t)val.i);
 
-			if (sym)
+			if (sym) {
+				print_args("%s", color_fptr);
 				print_args("&%s", sym->name);
+				print_args("%s", color_reset);
+			}
 			else
 				print_args("%p", val.p);
 		}
@@ -660,10 +669,12 @@ void get_argspec_string(struct uftrace_task_reader *task,
 
 			memcpy(val.v, data, spec->size);
 			estr = get_enum_string(&dinfo->enums, spec->enum_str, val.i);
+			print_args("%s", color_enum);
 			if (strlen(estr) >= len)
 				print_args("<ENUM>");
 			else
 				print_args("%s", estr);
+			print_args("%s", color_reset);
 			free(estr);
 		}
 		else {
@@ -693,8 +704,7 @@ void get_argspec_string(struct uftrace_task_reader *task,
 	}
 
 	if (needs_paren) {
-		args[n] = ')';
-		args[n+1] = '\0';
+		print_args("%s)%s", color_bold, color_reset);
 	} else {
 		if (needs_semi_colon)
 			args[n++] = ';';
--- a/utils/debug.c
+++ b/utils/debug.c
@@ -33,6 +33,14 @@ enum color_setting log_color;
 enum color_setting out_color;
 int dbg_domain[DBG_DOMAIN_MAX];
 
+/* colored output for argspec display */
+const char *color_reset   = TERM_COLOR_RESET;
+const char *color_bold    = TERM_COLOR_BOLD;
+const char *color_string  = TERM_COLOR_MAGENTA;
+const char *color_fptr    = TERM_COLOR_CYAN;
+const char *color_enum    = TERM_COLOR_BLUE;
+const char *color_enum_or = TERM_COLOR_RESET TERM_COLOR_BOLD "|" TERM_COLOR_BLUE;
+
 static const struct color_code {
 	char		code;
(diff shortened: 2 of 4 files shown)


honggyukim:
Here is the output with this PR.  It highlights strings and enum types.

screen shot 2018-11-17 at 11 45 16 am

honggyukim:
Please let me know if you have any other color preference. :)

honggyukim:
I just made function pointer types with cyan color as well.

honggyukim:
You can test it with `tests/t-dwarf1` as follows:

screen shot 2018-11-18 at 11 24 58 am

honggyukim:
Updated for char types in magenta and enum splitter in white.
screen shot 2018-11-18 at 1 31 33 pm

namhyung:
Do you want to make it unconditional?  I think we can have a single option to control behaviors of arguments/return-values.  It could enable color, hide them (#595) or show args in the comment of the function return (#582).  What about "args-policy : color/no-color, hide/show, arg-return" or something?

honggyukim:
> Do you want to make it unconditional?

Yes. I think it will make arguments clearly distinguishable especially with dwarf info.

> I think we can have a single option to control behaviors of arguments/return-values. It could enable color, hide them (#595) or show args in the comment of the function return (#582). What about "args-policy : color/no-color, hide/show, arg-return" or something?

What about making `--args` option?
```
--args=MODE
        Set a display mode for arguments and return value.
        Possible MODEs are one of "on"(default), "off" for hiding,
        and "full" for showing arguments at the exit of function in comments as well.
```

I think the color control can be easily done with `--color=no`.

namhyung:
> What about making --args option?

`--args` can be confusing as we already have `--argument`


> I think the color control can be easily done with --color=no.

ok then.

honggyukim:
> --args can be confusing as we already have --argument

What about ‘--args-mode’ then?

honggyukim:
Hmm.. As you said, `--args-policy=MODE` would be better and the `MODE` can be "on"(default), "off", and "full".

namhyung:
I think the name is extensible meaning that it could deal with additional changes in the future.

honggyukim:
So.. You mean we better use "show", "hide", and "arg-return" instead of "on", "off", and "full"?

honggyukim:
Hmm.. thinking it again, I'm not sure if we really have to make it more complicated with `--args-policy` to support printing arguments at the exit of function in comments. It may look too much verbose to see.

`--no-args` can be really needed when dumping in chrome format. I found that it gets easily broken because of various types of arguments and return values when it's used with dwarf support. And our default behavior is to print them, so I think we better provide a very simple way to hide those maybe with `--no-args` or even shorter way than `--args-policy="hide"`.

honggyukim:
If printing arguments at the exit of function in comments, why don't we merge the feature into a new option `--comment-policy` and its values can be "show", "show-args", and "hide". "hide" can replace current `--no-comment` option. I think we can remove `--no-comment` option with it as I don't think it's widely used as of now.

namhyung:
Basically I worry two things:

 * it's hard to maintain ever-increasing options - yes, it's short and easy for the job.  But in the long term, users will complain there are too many options to know.
 * but I don't want to remove existing options at my (or other's) discretion.  Probably no one cares, but it's possible someone used it in her script.  The change will break the script and she will hate me. ;-p  We'd better having some grace period to obsolete the option before removing it at least.

cmds/replay.c
```diff
@@ -1077,6 +1101,13 @@ int command_replay(int argc, char *argv[], struct opts *opts)
 	if (!opts->flat && peek_rstack(&handle, &task) == 0)
 		print_header(&output_fields, "#", 1);
 
+	if (out_color != COLOR_ON) {
+		paren_open  = "(";
+		paren_close = ")";
+		args_comma  = ", ";
+		args_retval = " = ";
+	}
+
```
namhyung: I think it's better making it a function to set those colors and call it in the common code (like in uftrace.c).  As you know, it's used in dump command too and record command doesn't care the color since it won't call `get_argspec_string()` anyway.

cmds/replay.c
```diff
@@ -583,8 +589,13 @@ void get_argspec_string(struct uftrace_task_reader *task,
 				}
 				print_args("\\\"");
 			}
-			else
+			else {
+				if (out_color == COLOR_ON)
+					print_args(TERM_COLOR_MAGENTA);
 				print_args("\"%.*s\"", slen + newline, str);
+				if (out_color == COLOR_ON)
+					print_args(TERM_COLOR_RESET);
```
namhyung: I suggest you add a new variable to save color for string (start and end) like args_comma and so on.  If you'd initialize them properly, the code will simply look like:

```
color_init() {
   if (COLOR_ON)
      string_start = TERM_COLOR_MAGENTA;
   else
      string_start = "";
   ...
}

get_argspec_string() {
   ...
   print_args("%s\"%.*s\"%s", string_start, slen + newline, str, start_end);
   ...
}
```

honggyukim: I just updated this PR but didn't write it in a single line. I think single call is a bit more efficient, but it hurts readability, so made it as below:
```
get_argspec_string() {
   ...
   print_args("%s", string_start);
   print_args("\"%.*s\"", slen + newline, str);
   print_args("%s", start_end);
   ...
}
```

cmds/replay.c
honggyukim: Done it. Thanks!

tests/runtest.py
```diff
@@ -21,7 +21,7 @@ class TestBase:
     TEST_SUCCESS_FIXED = -8
 
     objdir = 'objdir' in os.environ and os.environ['objdir'] or '..'
-    uftrace_cmd = objdir + '/uftrace --no-pager --no-event -L' + objdir
+    uftrace_cmd = objdir + '/uftrace --no-pager --no-event --color=no -L' + objdir
```
namhyung: Hmm..? I don't think it's needed since we run test with `popen()` redirecting stdout/stderr to a pipe.  The init code should detect it not being a terminal and disable color.

cmds/replay.c
namhyung: Why not doing it in `setup_color()`?

cmds/dump.c
```diff
@@ -1584,6 +1584,7 @@ int command_dump(int argc, char *argv[], struct opts *opts)
 			},
 		};
 
+		disable_argspec_color();
```
namhyung: If you call it in `setup_color()`, I guess this change is not needed since we turn off color when redirecting to a file.

honggyukim:
Ah.. please wait. I will make simpler for color variables, then update it very soon.

honggyukim:
I just simplified color variables. Please take a look at it. Thanks!

utils/debug.c
```diff
@@ -22,6 +22,14 @@ enum color_setting log_color;
...
 int dbg_domain[DBG_DOMAIN_MAX];
 
+/* colored output for argspec display */
+const char *color_reset   = TERM_COLOR_RESET;
+const char *color_bold    = TERM_COLOR_BOLD;
+const char *color_string  = TERM_COLOR_MAGENTA;
+const char *color_fptr    = TERM_COLOR_CYAN;
+const char *color_enum    = TERM_COLOR_BLUE;
+const char *color_enum_or = TERM_COLOR_RESET TERM_COLOR_BOLD "|" TERM_COLOR_BLUE;
+
```
namhyung: Because you moved these to this file, I guess the first commit is not necessary anymore.  Otherwise looks ok.

honggyukim: You're right.

namhyung:
@honggyukim I noticed `color_enum_or` has `TERM_COLOR_BOLD`.  Why did you add it?  I think it should be removed.

honggyukim:
I made it on purpose to better distinguish or separated enum values.

namhyung:
But color already made it distinguishable enough IMO, will remove.

namhyung:
Hmm.. is it the default configuration on Ubuntu?  Or you customized the colors?  Anyway we need to add `TERM_COLOR_RESET` after the `|` character then.

PR #602, opened 2018-11-22
misc: Fix properly exit in misc/install-deps.sh

honggyukim:
Since return statement can only be used in a function routine, it has to
be changed to exit statement instead.


honggyukim:
I updated the commit message. Thanks.

PR #605, opened 2018-11-26
misc: Update library functions in prototypes.h

honggyukim:
This PR adds more c library functions including below:
```
      void exit(int status);

      int unlink(const char *pathname);
      int unlinkat(int dirfd, const char *pathname, int flags);
      int mkdir(const char *pathname, mode_t mode);

      int raise(enum uft_signal sig);

      char *gettext (const char * msgid);
      char *dgettext (const char * domainname, const char * msgid);
      char *dcgettext (const char * domainname, const char * msgid, int category);
```
It also changes argspec of sprintf and snprintf to avoid useless input buffer string read.
Before:
```
      int sprintf(char *str, const char *format, ...);
      int snprintf(char *str, size_t size, const char *format, ...);
```
After:
```
      int sprintf(void* dest, const char *format, ...);
      int snprintf(void* dest, size_t size, const char *format, ...);
```


namhyung:
I have some more suggestions:

```c
void memcpy(void *dest, const void *src, size_t n);
void memset(void *s, int c, size_t n);
void memmove(void *dest, const void *src, size_t n);
```
no need to return dest pointer again.

```c
char *fgets(void *s, int size, FILE *stream);
```
the input string will be overwritten so don't care.

honggyukim:
@namhyung I just updated with your suggestion. Thanks very much!

Issue #606, opened 2018-11-26
argspec_parse_enum unittest is failed


honggyukim:
`argspec_parse_enum` unittest is failed. Please see below:
```
$ ./unittest -v
Running 43 test cases
======================
Testing unittest_framework...
[001] unittest_framework            : PASS
Testing option_parsing5...
[002] option_parsing5               : PASS
Testing option_parsing4...
[003] option_parsing4               : PASS
Testing option_parsing3...
[004] option_parsing3               : PASS
Testing option_parsing2...
[005] option_parsing2               : PASS
Testing option_parsing1...
[006] option_parsing1               : PASS
Testing argspec_parse_enum...
test failed at /home/honggyu/work/uftrace/utils/auto-args.c:870: (str) == ("CCC|BBB")
[007] argspec_parse_enum            : FAIL
Testing argspec_extract...
[008] argspec_extract               : PASS
Testing argspec_auto_args...
[009] argspec_auto_args             : PASS
    ...
```

namhyung:
Yes, I noticed this too.  This probably is because of the enum color change..

honggyukim:
Ah.. you're right.

honggyukim:
Please see #608. Thanks.

PR #608, opened 2018-11-26
test: Turn off color before running unittest

honggyukim:
Some unittests might be failed because of color settings, so it'd be
better to turn off color before running unittest.

Fixed: #606

### DIFF
--- a/tests/unittest.c
+++ b/tests/unittest.c
@@ -218,6 +218,8 @@ int main(int argc, char *argv[])
 	size_t i, test_num = 0;
 	int c;
 
+	setup_color(COLOR_OFF);
+
 	if (setup_unit_test(&test_cases, &test_num) < 0) {
 		printf("Cannot run unit tests - failed to load test cases\n");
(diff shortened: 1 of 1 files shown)


tests/unittest.c
```diff
@@ -218,6 +218,8 @@ int main(int argc, char *argv[])
 	size_t i, test_num = 0;
 	int c;
 
+	setup_color(COLOR_OFF);
+
```
namhyung: I don't want to handle it like this.  And unittest has an option (-n) to turn off color.

Instead, I think `get_enum_string()` should return a plain string without color code.  It's the job of `get_argspec_string()` to make up the fancy colors.  Maybe it's a bit more complex but I think it should work like that.

namhyung:
And please run `make test` before sending PR. :)

namhyung:
I changed it in 19d77d28

honggyukim:
Thanks for fixing it.

PR #609, opened 2018-11-26
uftrace: Add -n option for --nest-libcall

honggyukim:
Since --nest-libcall is often used when tracing prebuilt binaries, it'd
be better to have its own short option name.

It's often used with --auto-args, so the short option for using both is
simply become -na or -an.

  $ uftrace -na /usr/bin/uptime


namhyung:
I'm generally skeptical to add a short option as it'll suffer from a shortage of characters and hard to change later.  And `--nest-libcall` often showed some problems.  Do you really think it's valuable?

honggyukim:
I know that `--nest-libcall` sometimes have problem, but `--nest-libcall` is by far the most frequent used option compared to other short lower case options - `-b`, `-d`, `-f`, `-k`, and `-r` except for `-a`.

I think it'd be better to have a short option for most frequently used options. I also know that we better not to drop the existing options but I think we don't really need short options for `-b`/`--buffer` and -`r`/`--time-range`.

namhyung:
I often use `-d` and `-k` and sometimes `-f`.  I admit that `-b` and `-r` are not used frequently but I still don't like to remove them.

Anyway it **feels** like `-n` is not good for the `--nest-libcall` since it (n) seems somewhat related to number or something negative IMO.  What about using `-l` instead?  It gives me a feeling that it's about library thing..

honggyukim:
> Anyway it feels like -n is not good for the --nest-libcall since it (n) seems somewhat related to number or something negative IMO.

I agree.

> What about using -l instead? It gives me a feeling that it's about library thing..

So it can be used as `-la` in short for `--nest-liball --auto-args`. It reminds me of [latrace](http://people.redhat.com/jolsa/latrace/index.shtml).

If you're okay, then I will update this PR using `-l` instead of `-n`.

namhyung:
Please do!

honggyukim:
Update done. Thanks!

PR #610, opened 2018-11-27
test: Add -pg and -g flags to tests/Makefile

honggyukim:
unittest is a useful tool to be a target of uftrace itself, so it'd be
useful to be compiled with -pg and -g flags unconditionally so that it
can be directly traced to understand uftrace internal.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,4 +1,5 @@
 TEST_CFLAGS  += -include $(srcdir)/tests/unittest.h -Wno-sign-compare
+TEST_CFLAGS  += -pg -g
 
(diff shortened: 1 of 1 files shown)


tests/Makefile
```diff
@@ -1,4 +1,5 @@
 TEST_CFLAGS  += -include $(srcdir)/tests/unittest.h -Wno-sign-compare
+TEST_CFLAGS  += -pg -g
```
namhyung: The only downside I can think of is it'll generate gmon.out when not traced by uftrace.  Maybe no one cares...

honggyukim: I also think so.

PR #621, opened 2018-12-06
mcount: rename ambiguous argument name

DanielTimLee:
Function allocate_shmem_buffer() first argument's name "buf"
seems pretty ambiguous.

Paired to cmds/record.c's record_mmap_file(), both of the function
uses shared memory by leveraging mmap() through sess_id.

static void record_mmap_file(const char *dirname, char *sess_id, int bufsize)

And from this viewpoint, since it doesn't mean anything,
argument name "buf" seems meaningless.

### DIFF
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -23,25 +23,25 @@
 
 #define ARG_STR_MAX	98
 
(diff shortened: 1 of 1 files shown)


DanielTimLee:
Subject updated.

PR #624, opened 2018-12-08
build: Add a build option ASAN=1 to use address sanitizer

honggyukim:
It'd be better to have a switch for address sanizitzer.  Due to
preloading problem, it's not applied to libmcount*.so.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,14 @@ ifeq ($(COVERAGE), 1)
   LIB_LDFLAGS   += --coverage
 endif
 
+ifeq ($(ASAN), 1)
+  UFTRACE_CFLAGS    += -O0 -g -fsanitize=address
+  DEMANGLER_CFLAGS  += -O0 -g -fsanitize=address
+  SYMBOLS_CFLAGS    += -O0 -g -fsanitize=address
+  TRACEEVENT_CFLAGS += -O0 -g -fsanitize=address
+  TEST_CFLAGS       += -O0 -g -fsanitize=address
+endif
+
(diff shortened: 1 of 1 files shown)


honggyukim:
We can enable address sanitizer build, then upload some asan report so that new comers can fix the problem by tagging good first issue.

Makefile
```diff
@@ -102,6 +102,14 @@ ifeq ($(COVERAGE), 1)
   LIB_LDFLAGS   += --coverage
 endif
 
+ifeq ($(ASAN), 1)
+  UFTRACE_CFLAGS    += -O0 -g -fsanitize=address
+  DEMANGLER_CFLAGS  += -O0 -g -fsanitize=address
+  SYMBOLS_CFLAGS    += -O0 -g -fsanitize=address
+  TRACEEVENT_CFLAGS += -O0 -g -fsanitize=address
+  TEST_CFLAGS       += -O0 -g -fsanitize=address
+endif
```
namhyung: I think you can just set `COMMON_CFLAGS` (and same for "TRACE" and "COVERAGE" too).

honggyukim: Please see below:
- There's a conflict when preloading `libasan.so` and `libmcount.so` together if asan build is applied to `libmcount` so didn't use `COMMON_CFLAGS`.
- It's not also possible for "TRACE" because it applies to `libmcount*.so` that is not reenterant.
- `COMMON_CFLAGS` is already used for "COVERAGE".

namhyung: I ok then.

PR #631, opened 2018-12-11
utils: enable @k as an abbreviated form of @kernel

DanielTimLee:
this commit allows passing '@k' for kernel option as an abbreviated form of '@kernel'

created a util function for strstr(buf, '@k')

// before
$uftrace -K 2 -F sys_openat@kernel --force ./echo-sync
// after
$uftrace -K 2 -F sys_openat@k --force ./echo-sync

### DIFF
--- a/libmcount/event.c
+++ b/libmcount/event.c
@@ -160,7 +160,7 @@ int mcount_setup_events(char *dirname, char *event_str,
 		if (sep) {
 			*sep++ = '\0';
 
-			kernel = strstr(sep, "@kernel");
+			kernel = has_kernel_filter(sep);
 			if (kernel)
 				continue;
 
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -263,6 +263,31 @@ static inline bool host_is_lp64(void)
 	return get_elf_class() == ELFCLASS64;
 }
 
+static inline char* has_kernel_opt(char *buf)
+{
+	int idx = 0;
+
+	if (!strncasecmp(buf, "kernel", 6))
+		idx = 6;
+	else if (!strncasecmp(buf, "k", 1))
+		idx = 1;
+
+	if (idx && (buf[idx] == '\0' || buf[idx] == ','))
+		return buf;
+
+	return NULL;
+}
+
+static inline char* has_kernel_filter(char *buf)
+{
+	char *opt = strchr(buf, '@');
+
+	if (opt && has_kernel_opt(opt + 1))
+		return opt;
+
+	return NULL;
+}
+
 struct uftrace_time_range {
 	uint64_t first;
 	uint64_t start;
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -960,7 +960,7 @@ static void setup_trigger(char *filter_str, struct symtabs *symtabs,
 			goto next;
 
 		/* skip unintended kernel symbols */
-		if (module && !strcasecmp(module, "kernel") &&
+		if (module && has_kernel_opt(module) &&
 		    !setting->allow_kernel)
 			goto next;
 
@@ -997,7 +997,7 @@ static void setup_trigger(char *filter_str, struct symtabs *symtabs,
 							&symtabs->dinfo,
 							setting);
 			}
-			else if (!strcasecmp(module, "kernel")) {
+			else if (has_kernel_opt(module)) {
 				ret = add_trigger_entry(root, get_kernel_symtab(),
 							&patt, &tr,
 							&symtabs->dinfo,
@@ -1195,13 +1195,13 @@ char * uftrace_clear_kernel(char *filter_str)
 	if (filter_str == NULL)
 		return NULL;
 
-	if (strstr(filter_str, "@kernel") == NULL)
+	if (has_kernel_filter(filter_str) == NULL)
 		return xstrdup(filter_str);
 
 	strv_split(&filters, filter_str, ";");
 
 	strv_for_each(&filters, pos, j) {
-		if (strstr(pos, "@kernel") == NULL)
+		if (has_kernel_filter(pos) == NULL)
 			ret = strjoin(ret, pos, ";");
 	}
 	strv_free(&filters);
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -339,7 +339,7 @@ static void build_kernel_filter(struct uftrace_kernel_writer *kernel,
 	strv_for_each(&strv, name, j) {
 		struct uftrace_pattern patt;
 
-		pos = strstr(name, "@kernel");
+		pos = has_kernel_filter(name);
 		if (pos == NULL)
 			continue;
 		*pos = '\0';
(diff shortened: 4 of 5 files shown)


utils/utils.h
```diff
@@ -262,6 +262,14 @@ static inline bool host_is_lp64(void)
 	return get_elf_class() == ELFCLASS64;
 }
 
+static inline char* has_kernel_opt(char *buf)
+{
+    char* res = strstr(buf, "@kernel");
+    if (!res)
+        res =  strstr(buf, "@k");
```
namhyung: You also need to check whether it's a whole word.  For example it should not allow something like "foo@kernel-program" or "bar@kbs".  I know the original code had the same problem, please fix it.

DanielTimLee:
I'm really sorry for the late response.


I might not following you, but the thing
> change `setup_trigger()` function to handle kernel filters/symbols separately.

at here, the `kernel filter/symbols` you're mentioning about is it related to `module` in `setup_trigger:utils/fitler.c:931`? 

https://github.com/namhyung/uftrace/blob/2aac509417319e17c701a264e5e6e507d6134b36/utils/filter.c#L981-L1017

(ps. I've updated code with 'comparing whole word' )

namhyung:
Yes, `module` is a name of executable or library.  The module name should come first after the `@` sign but it can also have other parts like trigger actions.

libmcount/event.c
```diff
@@ -160,7 +160,7 @@ int mcount_setup_events(char *dirname, char *event_str,
 		if (sep) {
 			*sep++ = '\0';
 
-			kernel = strstr(sep, "@kernel");
+			kernel = has_kernel_opt(sep);
```
namhyung: I like this part.

utils/utils.h
```diff
@@ -262,6 +262,17 @@ static inline bool host_is_lp64(void)
...
+static inline char* has_kernel_opt(char *buf)
+{
+	char *opt = strchr(buf, '@');
+
+	if (opt != NULL)
+		if ((strcmp(opt, "@k") && strcmp(opt, "@kernel")) == 0)
+			return opt;
+
+	return NULL;
+}
```
namhyung: As I said, the filter/trigger string can have multiple actions after module name like `-T abc@kernel,filter`.  So you'd better checking with `strncmp()` and the next character being `,` or `\0`.

honggyukim:
It's not because of this change, but there is a different behavior between `-F sys_write@kernel` and `-T sys_write@kernel,filter` as follows:
```
$ uftrace -K 2 -F sys_write@kernel --force /usr/bin/uptime
 23:40:15 up 30 days, 12:51, 41 users,  load average: 0.34, 0.32, 0.19
 124.261 us [ 12374] | setlocale();
   2.400 us [ 12374] | bindtextdomain();
   1.667 us [ 12374] | textdomain();
   1.150 us [ 12374] | __cxa_atexit();
  15.070 us [ 12374] | getopt_long();
   2.416 ms [ 12374] | print_uptime();
   1.590 us [ 12374] | __fpending();
   1.086 us [ 12374] | ferror();
            [ 12374] | fclose() {
            [ 12374] |   sys_write() {
   0.870 us [ 12374] |     __fdget_pos();
  15.964 us [ 12374] |     vfs_write();
  33.307 us [ 12374] |   } /* sys_write */
  39.217 us [ 12374] | } /* fclose */
   0.207 us [ 12374] | __fpending();
   0.230 us [ 12374] | ferror();
   1.690 us [ 12374] | fclose();
```
Unlike `-F` option, `-T sys_write@kernel,filter` hides user functions as well. 
```
$ uftrace -K 2 -T sys_write@kernel,filter --force /usr/bin/uptime
 23:40:28 up 30 days, 12:51, 41 users,  load average: 0.37, 0.32, 0.19
            [ 12409] | sys_write() {
   0.877 us [ 12409] |   __fdget_pos();
  12.154 us [ 12409] |   vfs_write();
  18.764 us [ 12409] | } /* sys_write */
```

utils/utils.h
honggyukim: How about this?
```diff
diff --git a/utils/utils.h b/utils/utils.h
index b56a5a0..dd325d5 100644
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -262,6 +262,22 @@ static inline bool host_is_lp64(void)
        return get_elf_class() == ELFCLASS64;
 }

+static inline char* has_kernel_opt(char *buf)
+{
+       char *opt = strchr(buf, '@');
+
+       if (opt) {
+               int idx = 0;
+               if (!strncmp(opt, "@kernel", 7))
+                       idx = 7;
+               else if (!strncmp(opt, "@k", 2))
+                       idx = 2;
+               if (idx && (opt[idx] == '\0' || opt[idx] == ','))
+                       return opt;
+       }
+       return NULL;
+}
+
 struct uftrace_time_range {
        uint64_t first;
        uint64_t start;
```

DanielTimLee: Looks great. 
Is it okay to apply your code?

DanielTimLee: looks great!
Do you mind if I apply your code in this commit?

DanielTimLee: and also maybe using `strncasecmp` instead of `strncmp` might be more better?

I'm not sure which one to follow, but below here, it compares without case insensivity.

https://github.com/namhyung/uftrace/blob/2aac509417319e17c701a264e5e6e507d6134b36/utils/filter.c#L1000-L1004

Which one should I use?

honggyukim: > Do you mind if I apply your code in this commit?

Sure. Please use it.

> and also maybe using strncasecmp instead of strncmp might be more better?

I personally do not want to give multiple usage options, but I'm not sure how @namhyung thinks about it.

namhyung:
Yes, because the trigger also works for replay.  Maybe it's better to remove `reset_live_opts()`?

utils/utils.h
namhyung: For consistency, please use `strcasecmp()`.  Also please add a blank line before the if statements.

DanielTimLee:
updated.

honggyukim:
> Yes, because the trigger also works for replay. Maybe it's better to remove reset_live_opts()?

The replay output looks same if `remove_live_opts()` is not called as follows:
```
$ uftrace -K 2 -F sys_write@kernel --force /usr/bin/uptime
 14:54:18 up 34 days,  4:05, 46 users,  load average: 0.41, 0.18, 0.12
            [175126] | sys_write() {
   0.907 us [175126] |   __fdget_pos();
  20.483 us [175126] |   vfs_write();
  36.657 us [175126] | } /* sys_write */

$ uftrace -K 2 -T sys_write@kernel,filter --force /usr/bin/uptime
 14:54:26 up 34 days,  4:05, 46 users,  load average: 0.42, 0.19, 0.12
            [175156] | sys_write() {
   0.704 us [175156] |   __fdget_pos();
  22.287 us [175156] |   vfs_write();
  28.407 us [175156] | } /* sys_write */
```

honggyukim:
I applied @DanielTimLee's update and the result looks good.

PR #635, opened 2018-12-26
tui: Add C/E keys for nested collapse/expand

honggyukim:
Current c/e works for collapse/expand its direct children, but it
doesn't fold/unfold in a nested manner.

This patch adds C/E keys for nested collapse/expand.


honggyukim:
@namhyung I just did a bit of refactoring on this branch to reduce the amount of changes. Please take a look at it when you have time on it. Thanks.

namhyung:
I'm not sure the name 'nested' is good but I cannot come up with a better name.

honggyukim:
I was thinking "deep" as well but chose "nested". What about "recursive" if "nested" is not good?

namhyung:
Hmm.. "all" as a suffix?

honggyukim:
Okay. I will change "nested" to "all" then.

PR #637, opened 2018-12-29
record: write the previous memory area to file when reach special mapped area

ParkHanbum:
it will ignore the previous map area if we continue when reach special
mapped area like [stack]. because typically the [stack] area that
specially mapped is end of memory space. therefore, modify codes of
record_proc_maps function to save the previous memory area to file when
reach special area.

### DIFF
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -1065,6 +1065,16 @@ int record_trace_data(struct mcount_thread_data *mtdp,
 	return 0;
 }
 
+static void write_map(FILE *out, struct uftrace_mmap *map,
+		      unsigned char major, unsigned char minor,
+		      uint32_t ino, uint64_t off)
+{
+	/* write prev_map when it finds a new map */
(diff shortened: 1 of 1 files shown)


libmcount/record.c
```diff
@@ -1065,6 +1065,18 @@ int record_trace_data(struct mcount_thread_data *mtdp,
 	return 0;
 }
 
+static void write_map(FILE *out, struct uftrace_mmap *map,
+		      unsigned char major, unsigned char minor,
+		      uint32_t ino, uint64_t off)
+{
+	/* write prev_map when it finds a new map */
+	fprintf(out, "%"PRIx64"-%"PRIx64" %.4s %08"
+			PRIx64" %02x:%02x %-26u %s\n",
```
namhyung: Please do not break the format string into two lines.  I prefer one line for string constants (unless it has newlines in it).

PR #651, opened 2019-01-17
tui: Add a loading message before main screen shown

honggyukim:
Since it may take some time to build graph when tui is launched, users
may get confused if it's currently working or just stuck.

To inform that tui is working, this patch adds a kind message at the
center of screen showing "Building graph for TUI...".


namhyung:
Still it's better than nothing, applied..

PR #652, opened 2019-01-17
doc: Add --no-libcall to man pages of analysis commands

honggyukim:
The option --no-libcall is added to many analysis commands, but didn't
update the man pages.  This patch adds the explanation to each analysis
command.

It also updates some description in script command because the meaning
of "show" doesn't make sence for script execution.

### DIFF
--- a/doc/uftrace-script.md
+++ b/doc/uftrace-script.md
@@ -33,9 +33,9 @@ OPTIONS
     See 'uftrace-replay' for details.
 
 -t *TIME*, \--time-filter=*TIME*
-:   Do not show functions which run under the time threshold.  If some functions
-    explicitly have the 'trace' trigger applied, those are always traced
-    regardless of execution time.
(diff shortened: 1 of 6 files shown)


doc/uftrace-script.md
```diff
@@ -46,11 +46,10 @@ OPTIONS
 :   Set trace limit in nesting level.
 
 -r *RANGE*, \--time-range=*RANGE*
-:   Only show functions executed within the time RANGE.  The RANGE can be
-    \<start\>~\<stop\> (separated by "~") and one of \<start\> and \<stop\> can
-    be omitted.  The \<start\> and \<stop\> are timestamp or elapsed time if
-    they have \<time_unit\> postfix, for example '100us'.  The timestamp or
-    elapsed time can be shown with `-f time` or `-f elapsed` option respectively.
```
honggyukim: The description of `-f` option is discarded as well because it's not for script execution.

PR #653, opened 2019-01-19
Reorganize options in man pages

honggyukim:
This PR reorganizes options for each command as follows:
- Separate target command only options, common options, and common analysis options.
- Fix the order of options same across different commands.
- Update/remove/add some incorrect options.

By having different option categories, users may be able to pick useful options for each command.


honggyukim:
Hi @namhyung, this makes a huge changes, so it may be better to review it using diff tools across each command rather than examining diff changes.

doc/uftrace-live.md
```diff
@@ -50,6 +46,39 @@ OPTIONS
...
+    a `trace_on` trigger.
+
+
+LIVE OPTIONS
+============
+\--list-event
+:   Show available events in the process.
+
+\--report
+:   Show live-report before replay.
```
honggyukim: I'm not so sure if they have to be in another category, but just wanted to talk about it what to do.

doc/uftrace-record.md
```diff
@@ -21,12 +21,87 @@ This data can then be inspected later on, using `uftrace replay` or
...
+:   Set kernel max function depth separately.  Implies `--kernel`.
 
+-H *HOST*, \--host=*HOST*
+:   Send trace data to given host via the network, not writing to files.
+    The `uftrace recv` command should be run on the destination host to receive
+    the data.
+
+\--port=*PORT*
+:   When sending data to the network (with `-H`), use the given port instead of
+    the default (8090).
```
honggyukim: I'm not sure yet, but I think `--host 127.0.0.1` and `--port 8888` options can be easily used as `--host 127.0.0.1:8888`. We can consider it later.

doc/uftrace-live.md
namhyung: I'm fine with the current state.

doc/uftrace-record.md
```diff
@@ -375,16 +376,11 @@ The `traceon` and `traceoff` actions (the `_` can be omitted from `trace_on`
...
 The 'recover' trigger is for some corner cases in which the process accesses the
-callstack directly.  During tracing of the v8 javascript engine, for example, it
-kept getting segfaults in the garbage collection stage.  It was because v8
-incorporates the return address into compiled code objects(?).  The `recover`
-trigger restores the original return address at the function entry point and
-resets to the uftrace return hook address again at function exit.  I was managed
-to work around the segfault by setting the `recover` trigger on the related
-function (specifically `ExitFrame::Iterate`).
+callstack directly.  For now it's not necessary to call it as uftrace does the
+job automatically.
```
namhyung: This applies to x86 only.  It's hard to do the same on ARM/AArch64 at least currently..

doc/uftrace-live.md
```diff
@@ -39,21 +39,21 @@ COMMON OPTIONS
...
 :   Do not record library function invocations.  Library calls are normally
-    traced by hooking the dynamic linker's resolve function in the PLT.  One can
-    disable it with this option.
+    traced by hooking the dynamic linker's resolve function in the PLT.
+    One can disable it with this option.
 
 \--no-event
 :   Disable event recording which is used by default.  Note that explicit event
-    tracing by \--event option is not affected by this.
+    tracing by `--event` option is not affected by this.
```
namhyung: I think it's better to use "\\" for escape rather than enclosing with "\`"

honggyukim: I will take a look later for many other places.

doc/uftrace-record.md
namhyung: Yes, that should be possible too

PR #654, opened 2019-01-19
tui: Add 'o' key to open editor using ctags

honggyukim:
Sometimes it may not be easy to have debug info from the recorded data.
It'd be useful to provide another way to open editor using ctags.

It's especially useful when opening kernel functions because there's no
way to handle them as of now.


honggyukim:
I would like to handle other tags such as `global` if it's possible.

namhyung:
I don't want to add another key for tag files.  How about making 'O' key to try DWARF first and fall back to tags if failed?

honggyukim:
I also thought about using tag open as a fall back routine, but worried that people may get confused whether the location is exact or not.  If this concern is not an issue, then it's also possible.

namhyung:
I don't think we can guarantee 100% correctness anyway.  Users can see the source location so she can determine whether it's reliable..

honggyukim:
One more geeky idea is that we can also show the output of `objdump -d` if the target binary can be found and its architecture is same.

We can first get the address range from its symbol file.
```
$ cat uftrace.data/node.sym
    ...
0000000001017900 w v8::internal::Deserializer<v8::internal::BuiltinDeserializerAllocator>::Rehash()
0000000001017960 w v8::internal::Deserializer<v8::internal::BuiltinDeserializerAllocator>::ReadExternalReferenceCase(v8::internal::SerializerDeserializer::HowToCode, void**, unsigned long)
    ...
```
In TUI mode, if user press a key (for example, `D` key), then it executes `objdump` command to show its disassembly output.
```asm
$ objdump -C -d --start-address=0x1017900 --stop-address=0x01017960 node 

node:     file format elf64-x86-64


Disassembly of section .text:

0000000001017900 <v8::internal::Deserializer<v8::internal::BuiltinDeserializerAllocator>::Rehash()>:
 1017900:       55                      push   %rbp
 1017901:       48 89 e5                mov    %rsp,%rbp
 1017904:       41 55                   push   %r13
 1017906:       41 54                   push   %r12
 1017908:       53                      push   %rbx
 1017909:       48 83 ec 08             sub    $0x8,%rsp
 101790d:       ff 15 7d 66 36 01       callq  *0x136667d(%rip)        # 237df90 <mcount@GLIBC_2.2.5>
 1017913:       48 8b 9f 48 01 00 00    mov    0x148(%rdi),%rbx
 101791a:       4c 8b af 50 01 00 00    mov    0x150(%rdi),%r13
 1017921:       4c 39 eb                cmp    %r13,%rbx
 1017924:       74 20                   je     1017946 <v8::internal::Deserializer<v8::internal::BuiltinDeserializerAllocator>::Rehash()+0x46>
 1017926:       49 89 fc                mov    %rdi,%r12
 1017929:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
 1017930:       48 8b 3b                mov    (%rbx),%rdi
 1017933:       49 8b 74 24 50          mov    0x50(%r12),%rsi
 1017938:       48 83 c3 08             add    $0x8,%rbx
 101793c:       e8 cf 8c e1 ff          callq  e30610 <v8::internal::HeapObject::RehashBasedOnMap(v8::internal::Isolate*)>
 1017941:       4c 39 eb                cmp    %r13,%rbx
 1017944:       75 ea                   jne    1017930 <v8::internal::Deserializer<v8::internal::BuiltinDeserializerAllocator>::Rehash()+0x30>
 1017946:       48 83 c4 08             add    $0x8,%rsp
 101794a:       5b                      pop    %rbx
 101794b:       41 5c                   pop    %r12
 101794d:       41 5d                   pop    %r13
 101794f:       5d                      pop    %rbp
 1017950:       c3                      retq   
 1017951:       66 2e 0f 1f 84 00 00    nopw   %cs:0x0(%rax,%rax,1)
 1017958:       00 00 00 
 101795b:       0f 1f 44 00 00          nopl   0x0(%rax,%rax,1)
```

namhyung:
Or we can use capstone as it's already linked :)

honggyukim:
That could be a good option. I was just inspired by perf annotate feature, so we could also use the code base of perf maybe.

honggyukim:
Let's talk about this on Tuesday.

PR #663, opened 2019-01-28
Factor out struct definitions to header files

honggyukim:
Some struct definitions are included inside .c files so those cannot be used outside.
This PR factors out those definitions to header files.

ChangeLog:
- Move struct cu_files and debug_entry to utils/dwarf.h
- Rename gen-autoargs.py to gen-auto-args-list.py
- Rename auto-args.h to auto-args-list.h
- Move struct enum_def and enum_val to newly created auto-args.h


honggyukim:
This is a work in the middle of creating a tool `misc/dbgdump` that dumps `.dbg` file from an ELF file, but it looks more updates are required so no need to take a look at the moment.

PR #666, opened 2019-02-06
tui: Implement LEFT and RIGHT keys

honggyukim:
This PR adds KEY_LEFT and KEY_RIGHT to navigate screen.
It'd be useful when the output is too long to fit in the screen.


honggyukim:
The first patch is to fix the below broken output.
screen shot 2019-02-06 at 9 29 45 pm
And the second patch implements left and right keys.

honggyukim:
@namhyung It's not that urgent, but please take a look when you have some time. Thank you!

cmds/tui.c
```diff
@@ -212,6 +217,7 @@ static void print_graph_total(struct field_data *fd)
 	d = node->time;
 
 	print_time(d);
+	field_width += 10;
```
namhyung: What about making these `print_*` functions to return width/characters it consumed like `printf()` does?

cmds/tui.c
```diff
@@ -23,11 +23,16 @@
...
-#define BLANK  32
+#define BLANK       32
+
+#define MAX_LINE_WIDTH 8192
 
 static bool tui_finished;
 static bool tui_debug;
 
+static char linebuf[MAX_LINE_WIDTH];
+static int field_width;
```
namhyung: I'd like to avoid these static variables for this case.

cmds/tui.c
```diff
@@ -1106,10 +1116,60 @@ static void print_graph_empty(void)
...
+	/*
+	 * simply check whether the given character is a box character or not.
+	 *   ├ = 0xe2949c, │ = 0xe29482, └ = 0xe29494,
+	 *   ─ = 0xe29480, ▶ = 0xe296b6
+	 */
+	if (p >= (unsigned char)0x80 && p <= (unsigned char)0xb6)
+		return true;
+	else
+		return false;
+}
```
namhyung: I don't understand the logic.

cmds/tui.c
```diff
@@ -1106,10 +1116,60 @@ static void print_graph_empty(void)
...
+static void println(const char *fmt, ...)
+{
+	va_list ap;
+	va_start(ap, fmt);
+	char *apstr = NULL;
+	char *linestr = NULL;
+
+	xvasprintf(&apstr, fmt, ap);
+	xasprintf(&linestr, "%s%s", linebuf, apstr);
+	strncpy(linebuf, linestr, sizeof(linebuf));
```
namhyung: So is this for adding new formatted strings to the linebuf?

cmds/tui.c
```diff
@@ -1106,10 +1116,60 @@ static void print_graph_empty(void)
...
+	int cols = COLS - field_width;
+	unsigned char *p = (unsigned char*)linebuf;
+	int width = 0;
+
+	for (int i = 0; p && i < cols; i++) {
+		do {
+			p++;
+			width++;
+		} while (is_boxchar(*p));
+	}
```
namhyung: If only reason of checking it is the box characters in `print_graph_indent()` why not doing it there?

cmds/tui.c
```diff
@@ -1121,16 +1181,16 @@ static void print_graph_indent(struct tui_graph *graph,
...
 
 		if (i < depth - 1 || single_child)
-			printw("  │");
+			println("  │");
 		else if (is_last_child(parent, node))
-			printw("  └");
+			println("  └");
 		else
-			printw("  ├");
+			println("  ├");
```
namhyung: It'd be simpler if we could return number of characters printed and account each special case.

cmds/tui.c
```diff
@@ -2401,6 +2410,20 @@ static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
...
+					full_redraw = true;
+				}
+			}
+			break;
+		case KEY_RIGHT:
+			if (win == &graph->win || win == &partial_graph.win) {
+				leftcut += 1;
+				full_redraw = true;
+			}
+			break;
```
namhyung: So is this only for graph window?  I guess it should be easy (and useful) to apply to other windows too.

honggyukim:
@namhyung Thanks for your feedback during holiday. I will update this when I have some more time. Thanks!

cmds/tui.c
honggyukim: Hmm.. It requires to modify all the print functions in `replay.c`, `graph.c` and `tui.c`. Do you prefer to modify all over the places?

cmds/tui.c
honggyukim: It requires to modify the prototype of `print` field in `struct display_field` to avoid this.

cmds/tui.c
honggyukim: Please refer to the link below.
- https://docs.python.org/3/library/curses.ascii.html#curses.ascii.ismeta

I think I should modify this only to check if it's above `0x80`.

cmds/tui.c
honggyukim: Right.

cmds/tui.c
honggyukim: Hmm.. I thought this place is easier to handle.

cmds/tui.c
honggyukim: Right, but I thought it might not be practically needed for other windows.

honggyukim:
I just rebased it without notable change.

cmds/tui.c
honggyukim: The following python code explains it as well.

https://github.com/python/cpython/blob/283de2b9c18e38c9a573526d6c398ade7dd6f8e9/Lib/curses/ascii.py#L69

https://github.com/python/cpython/blob/283de2b9c18e38c9a573526d6c398ade7dd6f8e9/Python/bltinmodule.c#L1637-L1683

namhyung:
This is [my patch for the problem 1](https://github.com/namhyung/uftrace/files/4143763/tui-Fix-line-wrapping.patch.txt).  For LEFT/RIGHT keys, let me think about it more.

```diff
diff --git a/cmds/tui.c b/cmds/tui.c
index 0f8f0e64..17c15c46 100644
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -1086,7 +1086,7 @@ static void win_footer_graph(struct tui_window *win,
        graph->disp_update = false;
 }
 
-static void print_graph_field(struct uftrace_graph_node *node)
+static void print_graph_field(struct uftrace_graph_node *node, int width)
 {
        struct display_field *field;
        struct field_data fd = {
@@ -1097,18 +1097,29 @@ static void print_graph_field(struct uftrace_graph_node *node)
                return;
 
        list_for_each_entry(field, &graph_output_fields, list) {
-               printw("%*s", FIELD_SPACE, "");
-               field->print(&fd);
+               if (width >= FIELD_SPACE) {
+                       printw("%*s", FIELD_SPACE, "");
+                       width -= FIELD_SPACE;
+               }
+
+               if (width >= field->length) {
+                       field->print(&fd);
+                       width -= field->length;
+               }
        }
-       printw(FIELD_SEP);
+
+       if (width >= FIELD_SPACE)
+               printw(FIELD_SEP);
 }
 
-static void print_graph_empty(void)
+static void print_graph_empty(struct tui_graph *graph, int width)
 {
        struct display_field *field;
 
        if (list_empty(&graph_output_fields))
                return;
+       if (graph->width > width)
+               return;
 
        list_for_each_entry(field, &graph_output_fields, list)
                printw("%*s", field->length + FIELD_SPACE, "");
@@ -1118,12 +1129,18 @@ static void print_graph_empty(void)
 
 static void print_graph_indent(struct tui_graph *graph,
                               struct tui_graph_node *node,
-                              int depth, bool single_child)
+                              int width, int depth, bool single_child)
 {
        int i;
        struct tui_graph_node *parent = (void *)node->n.parent;
 
        for (i = 0; i < depth; i++) {
+               if (width < 3) {
+                       printw("%*.*s", width, width, "   ");
+                       break;
+               }
+               width -= 3;
+
                if (!graph->disp_mask[i]) {
                        printw("   ");
                        continue;
@@ -1150,8 +1167,8 @@ static void win_display_graph(struct tui_window *win, void *node)
        int width;
 
        if (node == NULL) {
-               print_graph_empty();
-               print_graph_indent(graph, graph->disp, d, true);
+               print_graph_empty(graph, COLS);
+               print_graph_indent(graph, graph->disp, COLS - w, d, true);
                return;
        }
 
@@ -1166,8 +1183,8 @@ static void win_display_graph(struct tui_window *win, void *node)
                        fold_sign = " ";
        }
 
-       print_graph_field(&curr->n);
-       print_graph_indent(graph, curr, d, single_child);
+       print_graph_field(&curr->n, COLS);
+       print_graph_indent(graph, curr, COLS - w, d, single_child);
 
        width = d * 3 + w;
 
@@ -1179,14 +1196,22 @@ static void win_display_graph(struct tui_window *win, void *node)
        else {
                char buf[32];
 
-               /* 4 = fold_sign(1) + parenthesis92) + space(1) */
-               width += snprintf(buf, sizeof(buf),
-                                 "%d", curr->n.nr_calls) + 4;
-               width = COLS - width;
+               if (width < COLS) {
+                       w = COLS - width;
+                       width += snprintf(buf, sizeof(buf), "%s(%d) ",
+                                         fold_sign, curr->n.nr_calls);
+
+                       /* handle UTF-8 character length */
+                       if (strcmp(fold_sign, " ")) {
+                               width -= 2;
+                               w += 2;
+                       }
+                       printw("%.*s", w, buf);
+               }
 
-               if (width > 0) {
-                       printw("%s(%d) %-*.*s", fold_sign, curr->n.nr_calls,
-                              width, width, curr->n.name);
+               if (width < COLS) {
+                       w = COLS - width;
+                       printw("%-*.*s", w, w, curr->n.name);
                }
        }
 }
```

honggyukim:
Thanks. The above fix also works fine for the first problem.

Issue #677, opened 2019-02-22
uftrace recv doesn't receive .dbg files


honggyukim:
`uftrace recv` mostly works fine, but it doesn't work if it's recorded with `--auto-args`.
It's because of missing `.dbg` files.

PR #678, opened 2019-02-22
record: Send .dbg files for -H/--host option

honggyukim:
It was missing to send .dbg files when -H/--host option is used, so it
was fail to replay the received data.

This patch sends .dbg files to fix such problem.

Fixed: #677


honggyukim:
Sure, I can add it but I'm just wondering if `extern.dat` is created during record time. As far as I understand, it's manually added by users after checking their timestamp.

namhyung:
Ah, right, never mind..

Issue #679, opened 2019-02-22
node doesn't work with GC in ARM arch


honggyukim:
node is traced without a problem when it runs an empty command with `-e ''`
```
root@raspberrypi3:~# uftrace record -H xx.xxx.xxx.xxx -v -D 3 ./node -e ''
uftrace: creating 1 thread(s) for recording
uftrace: using libmcount.so library for tracing
mcount: initializing mcount library
demangle: demangle failed: _ZTCN2v88internal8compiler12_GLOBAL__N_112TurboCfgFileE0_St14basic_ofstreamIcSt11char_traitsIcEE
dwarf: prepare debug info
mcount: setup PLT hooking
mcount: mcount setup done
mcount: new session started: bfe595d081821542: node
mcount: shmem_finish: tid: 4800 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
mcount: shmem_finish: tid: 4801 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
mcount: shmem_finish: tid: 4803 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
mcount: shmem_finish: tid: 4802 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
mcount: mcount trace finished
mcount: exit from libmcount
uftrace: child terminated with exit code: 0
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-bfe595d081821542-4796-000
uftrace: flushing /uftrace-bfe595d081821542-4799-000
uftrace: removing uftrace.data directory
```
But it doesn't work normally if garbage collection is performed as follows:
```
root@raspberrypi3:~# uftrace record -H xx.xxx.xxx.xxx -v -D 3 ./node --expose-gc -e 'gc()'
uftrace: creating 1 thread(s) for recording
uftrace: using libmcount.so library for tracing
mcount: initializing mcount library
demangle: demangle failed: _ZTCN2v88internal8compiler12_GLOBAL__N_112TurboCfgFileE0_St14basic_ofstreamIcSt11char_traitsIcEE
dwarf: prepare debug info
mcount: setup PLT hooking
mcount: mcount setup done
mcount: new session started: d2c31bce01ef9b01: node
```
It gets stuck here. So pressed `Ctrl+C`, then it's finished.
```
^Cuftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-d2c31bce01ef9b01-4805-000
[...]

PR #688, opened 2019-03-06
arm: Handle struct return type by keeping more regs on stack

honggyukim:
Current implementation for mcount_return and plthook_return only reserve
r0 and r1 registers assuming the size of return value is enough with 8
bytes.

But some functions may have a return type in struct and its size is more
than 8 bytes as following example:
```
  struct { int64_t quot, int64_t rem}
         __aeabi_ldivmod(int64_t numerator, int64_t denominator) {
    int64_t rem, quot;
    quot = __divmoddi4(numerator, denominator, &rem);
    return {quot, rem};
  }
```
In this case, we might partially lose return values because the return
value is stored in r0, r1, r2, and r3 registers to keep 16 bytes in
the return type of struct { int64_t quot, int64_t rem}.

So this patch fixes to keep r0-r3 registers to prevent such problems.

Fixed: #679

### DIFF
--- a/arch/arm/mcount.S
+++ b/arch/arm/mcount.S
@@ -62,18 +62,18 @@ END(__gnu_mcount_nc)
 
 
 ENTRY(mcount_return)
-	push 	{r0-r1, lr, pc}  /* ensure 8-byte alignment */
+	push 	{r0-r3, lr, pc}  /* ensure 8-byte alignment */
 	mov	r0, sp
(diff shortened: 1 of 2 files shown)


honggyukim:
Thanks for coming back! I just updated to keep `d1` register as well as you suggested. Tested with node garbage collection example for both arm and thumb. They both work fine. Thanks!

PR #691, opened 2019-03-08
mcount: Fix a few bugs in libmcount

honggyukim:
This PR contains some bug fixes in libmcount as follows:
- Fix to properly add SINGLE_THREAD macros
- Fix mtd access before mcount_prepare()
- Fix missing comma in skip_libs
- Fix a segfault while checking hard float in arm

### DIFF
--- a/arch/arm/mcount-support.c
+++ b/arch/arm/mcount-support.c
@@ -297,7 +297,7 @@ int check_float_abi_cb(struct dl_phdr_info *info, size_t size, void *data)
 		const Elf32_Phdr *phdr = info->dlpi_phdr + i;
 
 		if (phdr->p_type == PT_LOAD) {
-			Elf32_Ehdr *ehdr = (void *)phdr->p_vaddr;
+			Elf32_Ehdr *ehdr = (void *)info->dlpi_addr + phdr->p_vaddr;
 			use_hard_float = ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD;
 			break;
 		}
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -447,15 +447,15 @@ static int setup_mod_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
 		"libmcount-fast-single.so",
 		/* system base libraries */
 		"libc.so.6",
-		"libc-2.*.so"
+		"libc-2.*.so",
 		"libgcc_s.so.1",
 		"libpthread.so.0",
-		"libpthread-2.*.so"
+		"libpthread-2.*.so",
 		"linux-vdso.so.1",
 		"linux-gate.so.1",
 		"ld-linux-*.so.*",
 		"libdl.so.2",
-		"libdl-2.*.so"
+		"libdl-2.*.so",
 	};
 	size_t k;
 	static bool exe_once = true;


honggyukim:
I have another concern not directly related to this PR.  If `mcount_prepare()` is called even if in a recursion inside `atfork_child_handler()`, should we still allocate `mtdp->rstack` once again?
```diff
iff --git a/libmcount/mcount.c b/libmcount/mcount.c
index 52e860e0..7040f6c9 100644
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -760,7 +760,7 @@ struct mcount_thread_data * mcount_prepare(void)

        mcount_filter_setup(mtdp);
        mcount_watch_setup(mtdp);
-       mtdp->rstack = xmalloc(mcount_rstack_max * sizeof(*mtd.rstack));
+       mtdp->rstack = xmalloc(mcount_rstack_max * sizeof(*mtd.rstack));  // ?

        pthread_once(&once_control, mcount_init_file);
        prepare_shmem_buffer(mtdp);
```
There are two cases one for single threaded and the other is multi threaded as follows:
```c
#ifdef SINGLE_THREAD
# define TLS
# define get_thread_data()  &mtd
# define check_thread_data(mtdp)  (mtdp->rstack == NULL)
#else
# define TLS  __thread
# define get_thread_data()  pthread_getspecific(mtd_key)
# define check_thread_data(mtdp)  (mtdp == NULL)
#endif
```
In single threaded version, it makes sense to allocate `mtdp->rstack`, but in multi-threaded it may already have allocated buffer. It may not be a problem, but leave a comment just in case.

honggyukim:
Hmm.. I just read the gitter discussion regarding mcount recursion, so 7124884 has to be fixed anyway. Please review the commits except for that. I'm testing uftrace to chrome on ARM board, but I got stuck because of this mcount recursion issue.

honggyukim:
As @ParkHanbum analyzed, `mcount_prepare()` calls `__tls_get_addr()` inside. So it makes circular recursion, and never finished until it goes to stack overflow.
```
mcount_entry -> mcount_prepare -> __tls_get_addr -> mcount_entry -> ...
```

arch/arm/mcount-support.c
```diff
@@ -296,7 +296,7 @@ int check_float_abi_cb(struct dl_phdr_info *info, size_t size, void *data)
 	for (i = 0; i < info->dlpi_phnum; i++) {
 		const Elf32_Phdr *phdr = info->dlpi_phdr + i;
 
-		if (phdr->p_type == PT_LOAD) {
+		if (phdr->p_type == PT_LOAD && phdr->p_vaddr) {
 			Elf32_Ehdr *ehdr = (void *)phdr->p_vaddr;
```
namhyung: Hmm.. I guess the check should remain as is but the ehdr pointer should be changed like below:

```c
if (phdr->p_type == PT_LOAD) {
    Elf32_Ehdr *ehdr = (void *)info->dlpi_addr + phdr->p_vaddr;
```

libmcount/mcount.c
```diff
@@ -1745,8 +1751,10 @@ static void mcount_cleanup(void)
 	mcount_finish();
 	destroy_dynsym_indexes();
 
+#ifndef SINGLE_THREAD
 	pthread_key_delete(mtd_key);
 	mtd_key = -1;
+#endif
```
namhyung: Yes these changes are correct but it hurts readability of the code IMO.  It has no runtime cost in the fast path and only consumes a tiny bit of memory so why not leave it as is?

arch/arm/mcount-support.c
honggyukim: Hmm.. I need to test it again. Could you explain about this new change?

libmcount/mcount.c
honggyukim: Okay. I will remove it.

arch/arm/mcount-support.c
namhyung: The dlpi_addr contains address of the module actually loaded.  I guess it was usually 0 for normal executables.

ParkHanbum:
>  inside. So it makes circular recursion, and never finished until it goes to stack overflow.

This will be fixed by my patch. But I do not know when it will be merged. 

refer follow link 
https://github.com/namhyung/uftrace/pull/360/commits/dd7eec05287a27a69d1d2a43e1757df7a027d40c

honggyukim:
@ParkHanbum Thanks for the work. I've tested it before but didn't work on my workload. It may be a different issue. It seems that the recursion path is a bit different from my thought.

honggyukim:
It seems that the recursion was not made because of `__tls_get_addr`. It seems that the recursion path was like below:
```
plthook_entry -> mcount_prepare -> plthook_entry -> ...
```
`__tls_get_addr` was not shown here.

ParkHanbum:
> 
> 
> It seems that the recursion was not made because of `__tls_get_addr` and the recursion path was like below:
> 
> ```
> plthook_entry -> mcount_prepare -> plthook_entry -> ...
> ```
> 
> `__tls_get_addr` was not shown after observing more about it.

if then, there is another reason to make recursive. maybe it called same function again and again. could you let me know what function is called?

arch/arm/mcount-support.c
honggyukim: I mean that the segfault is occurred at `ehdr->e_flags` in line number 10.
```c
     1  int check_float_abi_cb(struct dl_phdr_info *info, size_t size, void *data)
     2  {
     3          unsigned i;
     4
     5          for (i = 0; i < info->dlpi_phnum; i++) {
     6                  const Elf32_Phdr *phdr = info->dlpi_phdr + i;
     7
     8                  if (phdr->p_type == PT_LOAD) {
     9                          Elf32_Ehdr *ehdr = (void *)phdr->p_vaddr;                                                                                                                                                                                         
    10                          use_hard_float = ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD;
    11                          break;
    12                  }
    13          }
    14          float_abi_checked = true;
    15
    16          return 1;
    17  }
```
> The dlpi_addr contains address of the module actually loaded. I guess it was usually 0 for normal executables.

If both `info->dlpi_addr` and `phdr->p_vaddr` are zero, then `ehdr->e_flags` gets crashed.
Please correct me if I'm wrong.

honggyukim:
> if then, there is another reason to make recursive. maybe it called same function again and again. could you let me know what function is called?

Sorry for the late answer but I still couldn't find the exact problem. I will tell you if I find something.

honggyukim:
The problem is shown in `pthread_once` at a first glance.
```
   3.802 us [  2417] | extensions::ShellMainDelegate::ShellMainDelegate();
            [  2417] | content::ContentMain() {
            [  2417] |   operator new() {
   0.781 us [  2417] |     pthread_self();
   0.417 us [  2417] |     pthread_setspecific();
   7.656 us [  2417] |   } /* operator new */
   1.146 us [  2417] |   std::set_new_handler();
            [  2417] |   setenv() {
   0.625 us [  2417] |     pthread_self();
   0.416 us [  2417] |     pthread_setspecific();
   0.417 us [  2417] |     pthread_self();
   0.417 us [  2417] |     pthread_setspecific();
   0.416 us [  2417] |     pthread_self();
   1.510 us [  2417] |     pthread_setspecific();
  43.073 us [  2417] |   } /* setenv */
            [  2417] |   operator new() {
   0.416 us [  2417] |     pthread_self();
   0.417 us [  2417] |     pthread_setspecific();
   3.749 us [  2417] |   } /* operator new */
            [  2417] |   operator new() {
   0.469 us [  2417] |     pthread_self();
   0.417 us [  2417] |     pthread_setspecific();
   4.427 us [  2417] |   } /* operator new */
   0.677 us [  2417] |   strlen();
            [  2417] |   std::__cxx11::basic_string::_M_create() {
            [  2417] |     <726fe124>() {
   0.469 us [  2417] |       pthread_self();
   0.417 us [  2417] |       pthread_setspecific();
   4.010 us [  2417] |     } /* <726fe124> */
   7.552 us [  2417] |   } /* std::__cxx11::basic_string::_M_create */
   0.417 us [  2417] |   memcpy();
            [  2417] |   std::vector::emplace_back() {
            [  2417] |     std::vector::_M_emplace_back_aux() {
            [  2417] |       operator new() {
   0.468 us [  2417] |         pthread_self();
   0.417 us [  2417] |         pthread_setspecific();
   3.802 us [  2417] |       } /* operator new */
   6.197 us [  2417] |     } /* std::vector::_M_emplace_back_aux */
   8.333 us [  2417] |   } /* std::vector::emplace_back */
...
   1.458 us [  2417] |   getenv();
   6.094 us [  2417] |   readlink();
  25.208 us [  2417] |   strstr();
            [  2417] |   localtime_r() {
            [  2417] |     pthread_once() {
   7.916 us [  2417] |       dlsym();
  24.687 us [  2417] |       dlsym();
   3.697 us [  2417] |       dlsym();
   2.083 us [  2417] |       dlsym();
  49.583 us [  2417] |     } /* pthread_once */
   0.365 us [  2417] |     pthread_once();
   0.417 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.365 us [  2417] |     pthread_once();
   0.417 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
...
   0.365 us [  2417] |     pthread_once();
   0.416 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
   0.416 us [  2417] |     pthread_once();
   0.416 us [  2417] |     pthread_once();
   0.416 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
   0.364 us [  2417] |     pthread_once();
 166.873 us [  2417] |     /* linux:schedule */
 480.309 us [  2417] |     /* linux:schedule */
   9.919 ms [  2417] |     /* linux:schedule */
            [  2417] |     /* linux:task-exit */

```

honggyukim:
I just updated with following changes:
- Fix to properly add SINGLE_THREAD macros - **dropped**
- Fix mtd access before mcount_prepare() - **dropped**
- Fix missing comma in skip_libs
- Fix a segfault while checking hard float in arm

libmcount/plthook.c
```diff
@@ -447,10 +447,10 @@ static int setup_mod_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
...
 		"libc.so.6",
-		"libc-2.*.so"
+		"libc-2.*.so",
 		"libgcc_s.so.1",
 		"libpthread.so.0",
-		"libpthread-2.*.so"
+		"libpthread-2.*.so",
 		"linux-vdso.so.1",
 		"linux-gate.so.1",
 		"ld-linux-*.so.*",
```
namhyung: To prevent similar bug in the future, please add the comma after the last entry ("libdl-2.*.so) too.

arch/arm/mcount-support.c
namhyung: I don't think the both can be zero at the same time.  Did you see the case?

libmcount/plthook.c
```diff
@@ -455,7 +455,13 @@ static int setup_mod_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
...
 		"libdl.so.2",
-		"libdl-2.*.so"
+		"libdl-2.*.so",
+		/* additional misc libraries */
+		"libstdc++.so.*",
+		"libelf.so.*",
+		"libdw.so.*",
+		"libSegFault.so",
+		"libunwind*"
```
namhyung: That means you cannot see PLT calls from these libraries.  I guess it depends on the use-case unless the library is a very primitive one.  At least we want to see C++ implementation details, right?

arch/arm/mcount-support.c
honggyukim: Yes, I saw it was crashed so added this check routine.

libmcount/plthook.c
honggyukim: I saw some broken addresses which don't matched with symbols while tracing chrome and `libstdc++.so` is always loaded because of our full demangler even for C programs. I think we better make it safe first, then we can opt them out if we really need to see their functions in a safe manner.

libmcount/plthook.c
honggyukim: Sure, I will

honggyukim:
I just updated with a comma at the last element of `skip_libs` array.

libmcount/plthook.c
honggyukim: One more thing is that most of STL implementation is in its header files, not in the library so I don't find any problem analyzing C++ standard libraries even without `libstdc++.so`.

arch/arm/mcount-support.c
namhyung: Do you mean `info->dlpi_addr + phdr->p_vaddr == 0`?

honggyukim: Yes. It was zero so crashed.

libmcount/plthook.c
honggyukim: @namhyung You're right. `libstdc++.so` shouldn't be skipped. I will remove it from the skip list.

honggyukim:
Updated again removing `libstdc++.so`, `libelf.so` and `libdw.so` libraries from the skip list.

arch/arm/mcount-support.c
honggyukim: Here is the gdb log for this crash.
```
292│ int check_float_abi_cb(struct dl_phdr_info *info, size_t size, void *data)
293│ {
294│         unsigned i;
295│
296│         for (i = 0; i < info->dlpi_phnum; i++) {
297│                 const Elf32_Phdr *phdr = info->dlpi_phdr + i;
298│
299│                 if (phdr->p_type == PT_LOAD) {
300│                         Elf32_Ehdr *ehdr = (void *)phdr->p_vaddr;
301├───────────────────────> use_hard_float = ehdr->e_flags & EF_ARM_ABI_FLOAT_HARD;
302│                         break;
303│                 }
304│         }
305│         float_abi_checked = true;
306│   
307│         return 1;
308│ }
```
```
Thread 2.1 "app_shell" received signal SIGSEGV, Segmentation fault.
[Switching to process 4571]
0x76fc04a8 in check_float_abi_cb (info=<optimized out>, size=<optimized out>, data=<optimized out>) at /usr/src/debug/uftrace/0.9.2-r0/git/arch/arm/mcount-support.c:301
(gdb) bt
#0  0x76fc04a8 in check_float_abi_cb (info=<optimized out>, size=<optimized out>, data=<optimized out>) at /usr/src/debug/uftrace/0.9.2-r0/git/arch/arm/mcount-support.c:301
#1  0x7297339c in dl_iterate_phdr () from /lib/libc.so.6
#2  0x54d53000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

(gdb) p ehdr
$1 = (Elf32_Ehdr *) 0x0
```

ParkHanbum:
@honggyukim nice work. I want more digging this bug. could you help me to reproduce it from my pc?

honggyukim:
@ParkHanbum Could you tell me which bug you want to reproduce?

ParkHanbum:
> 
> 
> @ParkHanbum Could you tell me which bug you want to reproduce?

infinity recursion

honggyukim:
I only see the problem in arm so don't know how to reproduce in x86_64 so far. It would be a lot better if we can reproduce it in the host machine for easier debugging and testing.

arch/arm/mcount-support.c
namhyung: But you only used `phdr->p_vaddr` here..

libmcount/plthook.c
```diff
@@ -456,6 +456,9 @@ static int setup_mod_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
 		"ld-linux-*.so.*",
 		"libdl.so.2",
 		"libdl-2.*.so",
+		/* additional misc libraries */
+		"libSegFault.so",
+		"libunwind*",
```
namhyung: I'm not sure it's worth adding them.  Did you see a problem from those?

arch/arm/mcount-support.c
honggyukim: Actually, I don't see this crash after a few more commits added, but saw it before. So I couldn't test `info->dlpi_addr + phdr->p_vaddr == 0` check.

> The dlpi_addr contains address of the module actually loaded. I guess it was usually 0 for normal executables.

But as you mentioned above, if `dlpi_addr` is normally zero, then there might be some possibility of being crashed because `phdr->p_vaddr` can also be zero. If you're not sure about it, I will drop this, then add this change if I see the same problem next time.

libmcount/plthook.c
honggyukim: I can say so, but they might be unsafe for our use cases. I would like to avoid the possible problems from those libraries. Moreover, I don't think that it's useful to see the functions inside them in most cases.

honggyukim: They do segmentation fault handling and backtrace handling, but those behaviors may have some conflicts to uftrace.

arch/arm/mcount-support.c
namhyung: No, I think they should not be zero at the same time and it's the right solution IMO.  For PIE `phdr->p_vaddr` could be zero but then `info->dlpi_addr` has the actual load address.  For non-PIE it'll go the other way around.

honggyukim: Thank you for the explanation. Then I will drop this although I'm just wondering why it was crashed as shown in the gdb report. I will think about it again when I see the same problem next time. Please never mind this time. Thanks.

libmcount/plthook.c
namhyung: I agree that they are not the target of trace most cases.  But the problem is that once it's added here there's no way to trace them.  On some rare case, user might want to trace them and that's the power of uftrace IMO.

If a function in those libraries is called from the uftrace/libmcount, hopefully the recursion guard will protect us. :)

arch/arm/mcount-support.c
namhyung: No no no... please change it to use `info->dlpi_addr + phdr->p_vaddr`.

libmcount/plthook.c
honggyukim: Okay. I will drop this too if you think so. I just wanted to reduce some possibility when tracing chrome on arm. It has a lot of problems so wanted to narrow down the problems.

Maybe the correct solution for this is to provide a way to pass the black listed libraries from the user. I think the way in address sanitizer looks good.
```sh
$ cat foo.c
#include <stdlib.h>
void bad_foo() {
  int *a = (int*)malloc(40);
  a[10] = 1;
}
int main() { bad_foo(); }

$ cat blacklist.txt
# Ignore reports from bad_foo function.
fun:bad_foo

$ clang -fsanitize=address foo.c ; ./a.out
# AddressSanitizer prints an error report.

$ clang -fsanitize=address -fsanitize-blacklist=blacklist.txt foo.c ; ./a.out
# No error report here.
```

honggyukim:
@namhyung I just updated it. It's become a very light PR now :)
Thanks for your help!

libmcount/plthook.c
namhyung: Looks reasonable.

namhyung:
Please include the patch for arm hard float check.

honggyukim:
Right. I just added it. Thanks.

Issue #625, opened 2018-12-09
Scroll bar or percenatge on tui?


DanielTimLee:
GNU less supports displaying the current line number, **total lines and percentage** as well as shown below:
(when pressing ctrl+g)

lines 51-100/185 byte 3228/5886 54% (press RETURN)

Any plans to support similar functionality in tui?

To search function inside tui or replay, it has to jump here and there and it's kinda distracting.
but what if there is a scroll-bar in tui, it could help to visualize the previous flow (before searching), and easily track the tracing progress.

(win_footer_graph)

namhyung:
We can add it to the lower right corner.  But you need to be careful when source line number and current line number have to overlap due to screen width.

DanielTimLee:
In `tui`, is there any logic or variable that tracks total number of nodes (which is not folded)?

When I see the code, I can't really find such thing exists.

In my guess, the only way to get number of unfolded node is 
just iterate all over the node to figure out it is folded or not.

Is it right?

(I'm trying to figure out the last node index (considering with fold) in `tui_main_loop`, 
but I can't find nicer way than just looping through all nodes.)

honggyukim:
Hi @DanielTimLee, @namhyung is on a travel now so may take some time to answer. I'm not very familiar with tui code based, but leave some comments based on my limited understanding.

> In tui, is there any logic or variable that tracks total number of nodes (which is not folded)?

I think it's not.

I would rather suggest you to take a look at `graph_prev_node` and `graph_next_node` functions.  They already handle whether the prev and next node is folded or not. When up and down keys are pressed, they reply on those functions.

IMHO, you can calculate the current position based on how many times prev and next move can be successful in the current graph (without actually moving the cursor). It may have to be updated if some graph is folded.

Please be aware that the change has to be applied to both graph and report windows. Please see `struct tui_window_ops`.

DanielTimLee:
Is there any way to figure out which type of window is currently shown in `struct tui_window *win`?

honggyukim:
This code snippet will be helpful to you.
```c
static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
{
    struct tui_graph *graph   = tui_graph_init(opts);
    struct tui_report *report = tui_report_init(opts);
    struct tui_list *info     = tui_info_init(opts, handle);
    struct tui_list *session  = tui_session_init(opts);
    struct tui_window *win;
        ...
    while (true) {
        switch (key) {
            ...
        case 'g':
[...]

PR #693, opened 2019-03-11
tui: added current position to footer

DanielTimLee:
This pull request adds tui current position to lower right corner.

By counting the last_index of tui_window at each _init,
it is possible to show position of current index in percentage.
(tui_*_init)

On graph_fold (collapse, expand), longest_child, Enter Key events,
tui will recalculate the last_index to adjust position.

By default, when width(COLS) is not enough to show message at footer,
current index position will be hide automatically.

```
$ uftrace tui

    0.256 us :     └─(1) fflush

uftrace graph: top: 232 depth: 6, curr: 252 depth: 0 last: 252             100%

$ uftrace tui

  149.399 us :  └─(1) main
  148.174 us :     ├─(1) printf
uftrace graph: /home/vault19/workspace/uftrace-data/simple-test/hello.c [line:3
```

Fixed: #625

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -23,6 +23,7 @@
 #include "utils/dwarf.h"
 
 #define KEY_ESCAPE  27
+#define BLANK  32
 
 static bool tui_finished;
 static bool tui_debug;
@@ -797,7 +798,6 @@ static struct tui_graph_node * graph_next_node(struct tui_graph_node *node,
 				indent_mask[*depth] = false;
 		}
 	}
-
 	return NULL;
 }
 
@@ -988,6 +988,18 @@ out:
 	memcpy(graph->disp_mask, graph->top_mask, graph->mask_size);
 }
 
+static void win_footer(struct tui_window *win, char *msg)
+{
+	int msg_len = strlen(msg);
+	char footer[COLS + 1];
+
+	memset(footer, BLANK, sizeof(footer));
+	memcpy(footer, msg, COLS < msg_len ? COLS : msg_len);
+	footer[COLS] = '\0';
+
+	printw("%-*s", COLS, footer);
+}
+
 static void win_footer_graph(struct tui_window *win,
 			     struct uftrace_data *handle)
 {
@@ -1026,10 +1038,8 @@ static void win_footer_graph(struct tui_window *win,
 				 SESSION_ID_LEN, sess->sid, sess->exename);
 		}
 	}
-	buf[COLS] = '\0';
 
-	printw("%-*s", COLS, buf);
-
+	win_footer(win, buf);
 	graph->disp_update = false;
 }
 
@@ -1312,9 +1322,8 @@ static void win_footer_report(struct tui_window *win, struct uftrace_data *handl
 				 handle->dirname, report->nr_sess, report->nr_func);
 		}
 	}
-	buf[COLS] = '\0';
 
-	printw("%-*.*s", COLS, COLS, buf);
+	win_footer(win, buf);
 }
 
 static void win_display_report(struct tui_window *win, void *node)
@@ -1422,7 +1431,7 @@ static struct tui_list * tui_info_init(struct opts *opts,
 	process_uftrace_info(handle, opts, build_info_node, &tui_info);
 
 	tui_window_init(&tui_info.win, &info_ops);
-			
+
 	return &tui_info;
 }
 
@@ -1451,11 +1460,8 @@ static void win_footer_info(struct tui_window *win,
 			    struct uftrace_data *handle)
 {
 	char buf[256];
-	size_t sz = sizeof(buf);
-	int len = 0;
-
-	print_buf("uftrace version: %s", UFTRACE_VERSION);
-	printw("%-*.*s", COLS, COLS, buf);
+	snprintf(buf, COLS, "uftrace version: %s", UFTRACE_VERSION);
+	win_footer(win, buf);
 }
 
 static void win_display_info(struct tui_window *win, void *node)
@@ -1551,7 +1557,7 @@ static void win_footer_session(struct tui_window *win,
 		break;
 	}
 
-	printw("%-*s", COLS, buf);
+	win_footer(win, buf);
 }
 
 static struct tui_graph * get_current_graph(struct tui_list_node *node,
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -73,6 +73,7 @@ struct tui_window {
 	void *old;
 	int top_index;
 	int curr_index;
+	int last_index;
 	int search_count;
 };
 
@@ -123,6 +124,8 @@ static void tui_window_move_down(struct tui_window *win);
 
 #define FIELD_SPACE  2
 #define FIELD_SEP  " :"
+
+#define POS_SIZE 5
 
 #define C_NORMAL  0
 #define C_HEADER  1
@@ -554,6 +557,24 @@ static void copy_graph_node(struct uftrace_graph_node *dst,
 	}
 }
 
+static int tui_last_index(struct tui_window *win)
+{
+	int count = win->curr_index;
+	void *next, *prev = win->curr;
+
+	while (true) {
+		next = win->ops->next(win, prev, false);
+		if (next == NULL)
+			return count;
+
+		count++;
+		if (win->ops->needs_blank(win, prev, next))
+			count++;
+
+		prev = next;
+	}
+}
+
 static void tui_window_init(struct tui_window *win,
 			    const struct tui_window_ops *ops)
 {
@@ -563,6 +584,7 @@ static void tui_window_init(struct tui_window *win,
 	win->top = top;
 	win->curr = win->old = top;
 	win->top_index = win->curr_index = 0;
+	win->last_index = tui_last_index(win);
 }
 
 static struct tui_graph * tui_graph_init(struct opts *opts)
@@ -595,10 +617,10 @@ static struct tui_graph * tui_graph_init(struct opts *opts)
 	partial_graph.top_mask  = xzalloc(graph->mask_size);
 	partial_graph.disp_mask = xmalloc(graph->mask_size);
 
-	tui_window_init(&partial_graph.win, &graph_ops);
-
 	INIT_LIST_HEAD(&partial_graph.ug.root.head);
 	INIT_LIST_HEAD(&partial_graph.ug.special_nodes);
+
+	tui_window_init(&partial_graph.win, &graph_ops);
 
 	/* select first session */
 	partial_graph.ug.sess = graph->ug.sess;
@@ -901,6 +923,7 @@ static bool win_enter_graph(struct tui_window *win, void *node)
 		return false;
 
 	curr->folded = !curr->folded;
+	win->last_index = tui_last_index(win);
 	return true;
 }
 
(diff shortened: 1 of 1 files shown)


cmds/tui.c
```diff
@@ -23,6 +23,7 @@
 #include "utils/dwarf.h"
 
 #define KEY_ESCAPE  27
+#define KEY_SPACE  32
```
namhyung: It'd be better using other name than `KEY_` if it's not used to process keyboard inputs.

cmds/tui.c
```diff
@@ -988,6 +988,22 @@ static void win_header_graph(struct tui_window *win,
 	memcpy(graph->disp_mask, graph->top_mask, graph->mask_size);
 }
 
+static void win_footer(struct tui_window *win, char *msg)
+{
+	char footer[COLS + 1];
+	char buf[COLS + 1];
```
namhyung: Do we really need two buffers for this?

cmds/tui.c
```diff
@@ -1043,7 +1049,7 @@ static void win_footer_graph(struct tui_window *win,
 		}
 	}
 
-	win_footer(win, buf);
+	win_footer(win, buf, true);
```
namhyung: Instead you can skip calling it for graph.

cmds/tui.c
DanielTimLee: I was trying to just use the same argument (`char *msg`) instead of creating new buffer, 

But in next commit, for the debug message
prior to `snprintf`, I call memset to remove NULL character.
```
	if (!is_graph && tui_debug) {
		/* for removing null character */
		memset(buf, KEY_SPACE, COLS);
		snprintf(buf, COLS, "uftrace debug: top: %d, curr: %d",
			 win->top_index, win->curr_index);
	}
```

Well, using the same string buffer with pointer is okay, 
but I thought that it seems weird to memset the pointer buffer without checking the size.

From the point of `win_footer`'s view, it doesn't know how size of the `char *msg` buffer has, 
(well, we know it has [COLS+1] byte since we've defined it)

So copying the `*msg` to a new buffer seems more secure to me.

cmds/tui.c
DanielTimLee: Are you saying that this line has to be removed completely from this pull request? 
Or do I have to move this line to the next commit?

**Cause at next commit, `win_footer` is used to print the `position` at lower right corner.**

If what you meant was not comfortable with adding extra argument `bool is_graph`, I think the same way. 
_I really don't want to add that one for only one purpose which is just comparing which type of window is it._

So what I've thought is, adding type attribute to `tui_window` 
```
#define TUI_SESS_REPORT    1
#define TUI_SESS_INFO      2
#define TUI_SESS_HELP      3
#define TUI_SESS_QUIT      4
#define TUI_SESS_DUMMY_NR  4

struct tui_window {
	const struct tui_window_ops *ops;
	void *top;
	void *curr;
	void *old;
        int type; // <- for storing the type of windows
	int top_index;
	int curr_index;
	int search_count;
};
```

and with this, we don't have to pass extra `is_graph` argument.
What do you think?

cmds/tui.c
```diff
@@ -988,8 +1017,14 @@ static void win_header_graph(struct tui_window *win,
 	memcpy(graph->disp_mask, graph->top_mask, graph->mask_size);
 }
 
+static int win_curr_position(struct tui_window *win)
+{
+	return ((float)win->curr_index / (float)win->last_index) * 100;
```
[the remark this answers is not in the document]
DanielTimLee: I think it's great. 
I'll apply right away.

cmds/tui.c
DanielTimLee: Changed to `KEY_SPACE -> BLANK`

DanielTimLee:
ping?

cmds/tui.c
namhyung: I cannot see the change.

cmds/tui.c
namhyung: Why not doing something like below?

```c
if (debug)
    snprintf(buf, len, "uftrace debug: ...", ...);
else
    strncpy(buf, len, "%s", msg);
```

cmds/tui.c
namhyung: Actually I don't think it's not that useful for info or session window to display debug message or percent (of position) since it doesn't have many lines usually.

DanielTimLee: In that case, remove debug message from `win_footer' and roll-back debug message to graph and report will solve this and upper issue.

DanielTimLee: Shall I proceed with roll-back to each footer? (Graph, report)

cmds/tui.c
DanielTimLee: I didn't update the change, since there might be additional change with the review.

cmds/tui.c
```diff
@@ -695,6 +718,7 @@ static void build_partial_graph(struct tui_report_node *root_node,
 	}
 
 	tui_window_init(&graph->win, &graph_ops);
+	graph->win.last_index = tui_last_index(&graph->win);
```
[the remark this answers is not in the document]
DanielTimLee: At `tui_graph_init`, 
it calls `tui_window_init` for `normal graph` and `partial graph`.

since at this point, there isn't any selected node for `partial graph`, 
so `tui_last_index()` will cause an error.

```
static void tui_window_init(struct tui_window *win,
                            const struct tui_window_ops *ops)
{
        void *top = ops->top(win, true);

        win->ops = ops;
        win->top = top;
        win->curr = win->old = top;
        win->top_index = win->curr_index = 0;
        win->last_index = tui_last_index(&graph->win);
}
```

DanielTimLee:
updated

DanielTimLee:
Ping?

cmds/tui.c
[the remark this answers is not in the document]
namhyung: I don't see what the problem is.  In the last statement you should pass `win` not the `&graph->win`.

cmds/tui.c
```diff
@@ -554,6 +557,25 @@ static void copy_graph_node(struct uftrace_graph_node *dst,
...
+            if (next == NULL)
+                return count;
+
+            count++;
+            if (win->ops->needs_blank(win, prev, next))
+                count++;
+
+            prev = next;
+        }
+}
```
namhyung: Looks like broken indentation (w/ expanded TAB)

DanielTimLee:
```bash
+static int tui_last_index(struct tui_window *win)$
+{$
+        int count = win->curr_index;$
+        void *next, *prev = win->curr;$
+$
+        while(true)$
+        {$
+            next = win->ops->next(win, prev, false);$
+            if (next == NULL)$
+                return count;$
+$
+            count++;$
+            if (win->ops->needs_blank(win, prev, next))$
+                count++;$
+$
+            prev = next;$
+        }$
+}$
+$
```
Apologize for m,y misunderstanding,
I'm not pretty sure which is the broken indentaion

(shown in `cat -A`)

honggyukim: You have to use "TAB" instead of whitespaces. Please replace spaces to TABs.

honggyukim: If your vim setting has `set expandtab` then remove the line in `.vimrc` or use `set noexpandtab`.

cmds/tui.c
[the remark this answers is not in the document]
DanielTimLee: It's my bad with upper code snippet. (Mistyped `win` as `&graph-win`)

[tui_graph_last_index_move_to_window_init.patch.txt](https://github.com/namhyung/uftrace/files/3029632/tui_graph_last_index_move_to_window_init.patch.txt)

If you apply this patch, you can find out it will cause an SIGSEGV error. (which is segment fault)

```
Building graph for TUI...fish: “uftrace tui” terminated by signal SIGSEGV (Address boundary error)
```

This is caused by `partial graph`. 

https://github.com/namhyung/uftrace/blob/master/cmds/tui.c#L567-L606

At line 597, `partial graph` will call tui_window_init.
With the upper patch applied, `tui_last_index` will be called for `partial graph` without any node selected.

And this will eventually cause an segment fault.

DanielTimLee:
updated.

cmds/tui.c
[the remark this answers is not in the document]
namhyung: Could you please make it not crashed?  Maybe you need to call `INIT_LIST_HEAD()` before `tui_window_init()`.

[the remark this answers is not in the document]
DanielTimLee: Seems working!

honggyukim:
@DanielTimLee Can you please rebase it again?

cmds/tui.c
```diff
@@ -988,6 +1017,27 @@ static void win_header_graph(struct tui_window *win,
...
+static void win_footer(struct tui_window *win, char *msg)
+{
+	int pos_start = COLS - POS_SIZE;
+	int msg_len = strlen(msg);
+	char footer[COLS + 1];
+
+	memset(footer, BLANK, sizeof(footer));
+	memcpy(footer, msg, COLS < msg_len ? COLS : msg_len);
+	if (pos_start > msg_len)
+		snprintf(footer + pos_start, POS_SIZE, "%d%%", win_pos_percent(win));
```
honggyukim: What about using "%3d%%"?

honggyukim: I think the right one looks better.
```
  As-Is       To-Be
   9%            9%
   10%          10%
   100%        100%
```

DanielTimLee: Looks great!

DanielTimLee:
rebased to master.

honggyukim:
Thanks for the update, but I have one more tiny thing to tell you.

Here are the commits of this PR.
- tui: win_footer_* print logic wrap to win_footer  …
- tui: added current position to footer  …

We usually use the commit title that begins with a verb so I would like to ask you to change the title as follows:
- tui: Wrap win_footer_* print logic to win_footer
- tui: Add current position to footer

It depends on the author's preference, but it'd be better to have common style in a single project.

DanielTimLee:
commit message updated

honggyukim:
Thanks!

cmds/tui.c
```diff
@@ -1514,7 +1558,7 @@ static struct tui_list * tui_session_init(struct opts *opts)
 		list_add_tail(&node->list, &tui_session.head);
 	}
 
-	tui_window_init(&tui_session.win, &session_ops);
+	tui_window_init(win, &session_ops);
```
namhyung: Please try to minimize your patch size.  Changes here and in `tui_info_init()` are not needed anymore.

DanielTimLee: I'm not sure what the `minimize` the change in here?

Does the upper line refactor from `&tui_session.win` to `win` is unnecessary?
(for keeping consistency of code with first argument of `tui_window_init` with only `win` rather than `&<session>.win` 
at `tui_report_init`, `tui_session_init` and `tui_info_init`)

Or does the `tui_window_init` has to be removed?

namhyung: I mean the first one.  You don't need to create a local variable `win` just for single callsite IMO.

DanielTimLee:
applied!

DanielTimLee:
ping?

honggyukim:
@DanielTimLee Thanks for your work. It looks very good now!

PR #701, opened 2019-03-17
script: Add script/strings.py to print runtime strings

honggyukim:
argstr.py is to display unique string collected from arguments and
return values.

It displays all the strings collected at runtime as follows:
```
  $ uftrace script -S scripts/strings.py --record /usr/bin/uptime
   00:05:01 up 37 days, 13:16, 12 users,  load average: 4.94, 4.85, 4.90
  "procps-ng"
  ---
  "C"
  ---
  "phsV"
  ---
  "/proc/uptime"
  ---
  "00:05:01 up 37 days, 13:16, 12 users,  load average: 4.94, 4.85, 4.90"
  ---
  "en_US.UTF-8"
  ---
  "NULL"
  ---
  "/usr/share/locale"
  ---
  "/proc/loadavg"
  ---
```
In many cases, same strings are repeatedly collected during funtion
tracing so this tool can be useful when observing their unique set.

### DIFF
--- a/scripts/strings.py
+++ b/scripts/strings.py
@@ -1,0 +1,32 @@
+#
+# strings.py - print the unique strings of runtime function arguments and return values.
+#
+# uftrace-option: --nest-libcall --auto-args
+#
+
+strset = set()
+
+def uftrace_entry(ctx):
+    global strset
+    if "args" in ctx:
+        args = ctx["args"]
+        for arg in args:
+            if isinstance(arg, str):
+                arg = arg.strip()
+                if arg is not "":
+                    strset.add(arg)
+
+def uftrace_exit(ctx):
+    global strset
+    if "retval" in ctx:
+        ret = ctx["retval"]
+        if isinstance(ret, str):
+            ret = ret.strip()
+            if ret is not "":
+                strset.add(ret)
+
+def uftrace_end():
+    global strset
+    for strval in strset:
+        print('"%s"' % strval)
+        print("---")


honggyukim:
This script can be useful for library tracing, but it can also be used for strings collected from user functions.

honggyukim:
I changed the name of script from `argstr.py` to `strings.py` to make it as a counter part of `strings` tool that prints the strings in files.

`strings.py` prints the unique set of runtime strings collected from function arguments and return values.

scripts/strings.py
```diff
@@ -0,0 +1,32 @@
...
+
+strset = set()
+
+def uftrace_entry(ctx):
+    global strset
+    if "args" in ctx:
+        args = ctx["args"]
+        for arg in args:
+            if isinstance(arg, str):
+                arg = arg.rstrip().lstrip()
```
namhyung: Simply `arg.strip()` won't work?

scripts/strings.py
```diff
@@ -0,0 +1,32 @@
...
+                if arg is not "":
+                    strset.add(arg)
+
+def uftrace_exit(ctx):
+    global strset
+    if "retval" in ctx:
+        ret = ctx["retval"]
+        if isinstance(ret, str):
+            ret = ret.rstrip().lstrip()
```
namhyung: ditto

scripts/strings.py
```diff
@@ -0,0 +1,32 @@
...
+            if ret is not "":
+                strset.add(ret)
+
+def uftrace_end():
+    global strset
+    for strval in strset:
+        print('"%s"' % strval)
+        print("---")
```
namhyung: For similarity to `strings` util, I think it's better to remove the "---".

scripts/strings.py
honggyukim: Right. I thought that it also removes internal whitespaces. Will change it.

scripts/strings.py
honggyukim: I added it for strings in multiple lines, for example `ps` command, as follows:
```
$ uftrace script -S scripts/strings.py --record /bin/ps
   PID TTY          TIME CMD
136005 pts/87   00:00:00 bash
160745 pts/87   00:00:00 uftrace
160746 pts/87   00:00:00 uftrace
160747 pts/87   00:00:00 ps
":      vi
State:  T (stopped)
Tgid:   110256
Ngid:   0
Pid:    110256
PPid:   109880
TracerPid:      0
Uid:    1011    10..."
---
"NSpid: 229
NSpgid: 0
NSsid:  0
Threads:        1
SigQ:   0/256000
SigPnd: 0000000000000000
ShdPnd: 00000..."
---
"VmRSS:     1788 kB
VmData:     1668 kB
VmStk:       132 kB
VmExe:       412 kB
VmLib:      5744 k..."
---
":           152 kB
VmLib:      2092 kB
VmPTE:        44 kB
VmPMD:        12 kB
VmSwap:      332 kB
Huge..."
---
...
```

honggyukim:
I just updated it. Thanks for your feedback.

scripts/strings.py
namhyung: ok then

PR #702, opened 2019-03-18
record: Add the missed returnning code

ParkHanbum:
This code seems to have attempted to return the full installation path.
However, the return code is missing. If it does not return here, strcpy
will only return the name of libmcount only from below line. therefore,
added returning code.

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -132,8 +132,13 @@ char * get_libmcount_path(struct opts *opts)
 
 #ifdef INSTALL_LIB_PATH
 	snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
-	if (access(lib, F_OK) != 0 && errno == ENOENT)
+	if (access(lib, F_OK) == 0)
+		return lib;
+
+	if (errno == ENOENT)
 		pr_warn("Didn't you run 'make install' ?\n");
+
+	/* fall through */
 #endif
 	strcpy(lib, libmcount);
 	return lib;


cmds/record.c
```diff
@@ -134,6 +134,8 @@ char * get_libmcount_path(struct opts *opts)
 	snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
 	if (access(lib, F_OK) != 0 && errno == ENOENT)
 		pr_warn("Didn't you run 'make install' ?\n");
+	else
+		return lib;
```
namhyung: But in this case, it may return invalid lib path if the errno is not ENOENT?

honggyukim: How about this?
```diff
diff --git a/cmds/record.c b/cmds/record.c
index 8b499b6..295399c 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -132,11 +132,14 @@ char * get_libmcount_path(struct opts *opts)
 
 #ifdef INSTALL_LIB_PATH
        snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
-       if (access(lib, F_OK) != 0 && errno == ENOENT)
-               pr_warn("Didn't you run 'make install' ?\n");
-#endif
+       if (access(lib, F_OK) == 0)
+               return lib;
+       free(lib);
+       return NULL;
+#else
        strcpy(lib, libmcount);
        return lib;
+#endif
 }
 
 void put_libmcount_path(char *libpath)
```

namhyung: How about this?

```diff
diff --git a/cmds/record.c b/cmds/record.c
index 8b499b6..3885759 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -132,8 +132,12 @@ char * get_libmcount_path(struct opts *opts)
 
 #ifdef INSTALL_LIB_PATH
        snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
-       if (access(lib, F_OK) != 0 && errno == ENOENT)
+       if (access(lib, F_OK) == 0)
+               return lib;
+
+        if (errno == ENOENT)
                pr_warn("Didn't you run 'make install' ?\n");
+        /* fall through */
 #endif
        strcpy(lib, libmcount);
        return lib;
```

ParkHanbum:
thank you for help! I added some message. 
```
$ uftrace t-abc
WARN: Library does not exist in /usr/local/lib. Didn't you run 'make install' ?
ERROR: ld.so: object 'libmcount-fast.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```

cmds/record.c
```diff
@@ -132,8 +132,14 @@ char * get_libmcount_path(struct opts *opts)
...
 #ifdef INSTALL_LIB_PATH
 	snprintf(lib, PATH_MAX, "%s/%s", INSTALL_LIB_PATH, libmcount);
-	if (access(lib, F_OK) != 0 && errno == ENOENT)
-		pr_warn("Didn't you run 'make install' ?\n");
+	if (access(lib, F_OK) == 0)
+		return lib;
+
+	if (errno == ENOENT)
+		pr_warn("Library does not exist in %s. Didn't you run 'make install' ?\n"
+			, INSTALL_LIB_PATH);
```
namhyung: Just a style issue - please put the comma after the string literal.

honggyukim: I also think that this message doesn't look good to me.

PR #703, opened 2019-03-18
mcount: Add code to store the errno from each entry

ParkHanbum:
Currently, each entry function changes the value of errno when
initializing mcount thread data. The problem is that since errno is
shared with the target program, it affects the execution of the target
program. therefore, added some code to solve this problem and to prevent
this happening never again.

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -1184,8 +1184,8 @@ mcount_arch_parent_location(struct symtabs *symtabs, unsigned long *parent_loc,
 }
 #endif
 
-int mcount_entry(unsigned long *parent_loc, unsigned long child,
-		 struct mcount_regs *regs)
+static int __mcount_entry(unsigned long *parent_loc, unsigned long child,
+			  struct mcount_regs *regs)
 {
 	enum filter_result filtered;
 	struct mcount_thread_data *mtdp;
@@ -1253,7 +1253,17 @@ int mcount_entry(unsigned long *parent_loc, unsigned long child,
 	return 0;
 }
 
-unsigned long mcount_exit(long *retval)
+int mcount_entry(unsigned long *parent_loc, unsigned long child,
+		 struct mcount_regs *regs)
+{
+	int saved_errno = errno;
+	int ret = __mcount_entry(parent_loc, child, regs);
+
+	errno = saved_errno;
+	return ret;
+}
+
+static unsigned long __mcount_exit(long *retval)
 {
 	struct mcount_thread_data *mtdp;
 	struct mcount_ret_stack *rstack;
@@ -1302,7 +1312,16 @@ unsigned long mcount_exit(long *retval)
 	return retaddr;
 }
 
-static int cygprof_entry(unsigned long parent, unsigned long child)
+unsigned long mcount_exit(long *retval)
+{
+	int saved_errno = errno;
+	unsigned long ret = __mcount_exit(retval);
+
+	errno = saved_errno;
+	return ret;
+}
+
+static int __cygprof_entry(unsigned long parent, unsigned long child)
 {
 	enum filter_result filtered;
 	struct mcount_thread_data *mtdp;
@@ -1380,7 +1399,16 @@ static int cygprof_entry(unsigned long parent, unsigned long child)
 	return 0;
 }
 
-static void cygprof_exit(unsigned long parent, unsigned long child)
+static int cygprof_entry(unsigned long parent, unsigned long child)
+{
+	int saved_errno = errno;
+	int ret = __cygprof_entry(parent, child);
+
+	errno = saved_errno;
+	return ret;
+}
+
+static void __cygprof_exit(unsigned long parent, unsigned long child)
 {
 	struct mcount_thread_data *mtdp;
 	struct mcount_ret_stack *rstack;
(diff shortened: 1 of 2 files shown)


namhyung:
I think return path (ex. `mcount_return`) should retain the `errno` as well.  Also it seems every code flow in a function (like returning -1) should retain it too.

ParkHanbum:
@namhyung how about this? 
```
unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
                            unsigned long module_id, struct mcount_regs *regs)
{
       unsigned long result;
        SAVE_ERRNO(result = plthook_entry_impl(ret_addr, child_idx, module_id ,regs));
       return result;
}
```

namhyung:
I suspect it might cause another infinite recursion of `__errno_location()`.

ParkHanbum:
@namhyung why do you think so? we used it already, aren't we?

namhyung:
Hmm.. nevermind, I forgot `libmcount.so` doesn't hook PLT calls from itself.

namhyung:
Anyway I don't like the macro.  As it doesn't save much why not open-coding?  Also I'm not sure we need to keep errno for entry functions.

ParkHanbum:
@namhyung then How about adding a test code that checks for errno after each entry is called?

namhyung:
What kind of test code do you think?  It'll surely change but it really matters at the beginning of function?

Now I'm thinking again, functions like `perror()` depends on the previous `errno` value and should be supported.  So ok, let's save and restore both of entry and exit paths.

ParkHanbum:
@namhyung I think it can be done by adding code to keep its value if we change errno value while execute uftrace entry/exit. 

however if we not need to change errno, then it would nice to adding test to clearly that we do not change errno while uftrace entry/exit.

so, do you think it is need to change errno while executing uftrace entry/exit?

namhyung:
uftrace/libmcount itself doesn't change errno, but it's changed indirectly by calling library functions and we don't know which one is actually changing it.

ParkHanbum:
@namhyung how about this? is this acceptable?
```
+ /*
+  * Uftrace could be change the errno while executing uftrace's entry/exiti
+  * function. this will affect target program's executing. for example,
+  * if errno has been changed while executing perror(), it could be print
+  * wrong error message to user. so, we must guarantee that errno do not
+  * changed while executing uftrace's entry/exit functions.
+  *
+  * following macros do stuff for this.
+  *
+  * _store_errno - it must be called at top of function.
+  * _restore_errno - it must be called at end of function.
+  */
+ #define _store_errno()                          \
+         int __stored_errno;                     \
+         __stored_errno = errno;
+
+ #define _restore_errno()                        \
+         errno = (1 ? __stored_errno : 0);
```

namhyung:
I think it's better to have a wrapper function to handle all return paths but without a macro.  Like this?

```c

static int __mcount_entry(...)
{
    ...
}

int mcount_entry(...)
{
   int saved_errno = errno;
   int ret = __mcount_entry();

   errno = saved_errno;
   return ret;
}
```

libmcount/mcount.c
```diff
@@ -1408,8 +1436,16 @@ static void cygprof_exit(unsigned long parent, unsigned long child)
...
+static void cygprof_exit(unsigned long parent, unsigned long child)
+{
+	int saved_errno = errno;
+
+	__cygprof_exit(parent, child);
+	errno = saved_errno;
+}
+
+static void _xray_entry(unsigned long parent, unsigned long child,
+			struct mcount_regs *regs)
```
honggyukim: Did you use a single underline to avoid the conflict with `__xray_entry` in `arch/x86_64/xray.S`?

libmcount/plthook.c
```diff
@@ -817,9 +819,21 @@ unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
 	return real_addr;
 }
 
+unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
+			    unsigned long module_id, struct mcount_regs *regs)
+{
+	int saved_errno = errno;
+	unsigned long ret;
+	ret = __plthook_entry(ret_addr, child_idx, module_id, regs);
```
honggyukim: It doesn't look consistent with other code. What about this if you want to avoid 80 chars in a line?
```c
unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
                            unsigned long module_id, struct mcount_regs *regs)
{
        int saved_errno = errno;
        unsigned long ret =
                __plthook_entry(ret_addr, child_idx, module_id, regs);

        errno = saved_errno;
        return ret;
}
```
Otherwise, there should be a empty line after declarations.
```c
unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
                            unsigned long module_id, struct mcount_regs *regs)
{
        int saved_errno = errno;
        unsigned long ret;

        __plthook_entry(ret_addr, child_idx, module_id, regs);
        errno = saved_errno;
        return ret;
}
```

honggyukim:
And please rebase to master whenever you update the commit. It doesn't have a conflict but it'd be better to do it. Thanks.

ParkHanbum:
this PR has been refreshed..

libmcount/plthook.c
```diff
@@ -850,9 +852,21 @@ unsigned long plthook_entry(unsigned long *ret_addr, unsigned long child_idx,
...
```
honggyukim: Please give a single empty line between declarations and statements.

ParkHanbum: applied!

honggyukim:
Please rebase it to master :)

ParkHanbum:
updated!

PR #704, opened 2019-03-18
mcount: change TLS model to 'initial-exec'

ParkHanbum:
Currently uftrace utilizes TLS variable mcount thread_data type mtd.
Each entry function calls the mcount_prepare function, which prepares
mcount_thread_data. The mcount_prepare function has the following code:

```
struct mcount_thread_data * mtdp = & mtd;
```

The compiler translates this code as follows:
```
data16 lea 0x2d6d7 (% rip),% rdi # 35f98 <.got + 0x8>
data16 data16 callq 6400 <__ tls_get_addr @ plt>
```

The __tls_get_addr@plt function is called to reference the TLS variable
mtd. which in some cases calls malloc@plt or the free@plt function
internally. And in this case it will re-enter plthook_entry. In other
words, infinite recursion occurs.

This is because every action that attempts to access mtd with a
mechanism caused by the TLS model is made by a call to
__tls_get_addr@plt.

For this reason, changed the model of the TLS variable mtd to
initial-exec. The initial-exec model solves the problem of infinite
recursion because it is a way to refer to variables by adding or
subtracting offsets from static TLS blocks.

```
mov 0x2d6a6 (% rip),% rax # 35fc8 <.got + 0x38>
mov% fs: (% rax),% rbx
```

For more information on this, please see the following link:
https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt

If it change to the initial-exec model, it can expect some improvement
in performance compared to the existing one. Please refer to the
following link for more details.
https://die4taoam.tistory.com/37

### DIFF
--- a/libmcount/internal.h
+++ b/libmcount/internal.h
@@ -119,7 +119,6 @@ struct mcount_thread_data {
 	int				tid;
 	int				idx;
 	int				record_idx;
-	bool				recursion_marker;
 	bool				in_exception;
 	bool				dead;
 	unsigned long			cygprof_dummy;
@@ -145,20 +144,24 @@ static inline void mcount_restore_arch_context(struct mcount_arch_context *ctx) 
 
 #ifdef SINGLE_THREAD
 # define TLS
-# define get_thread_data()  &mtd
-# define check_thread_data(mtdp)  (mtdp->rstack == NULL)
+# define attr_tls_model
 #else
 # define TLS  __thread
-# define get_thread_data()  pthread_getspecific(mtd_key)
-# define check_thread_data(mtdp)  (mtdp == NULL)
+# define attr_tls_model __attribute__((tls_model("initial-exec")))
 #endif
 
-extern TLS struct mcount_thread_data mtd;
+extern TLS struct mcount_thread_data *_mtdp attr_tls_model;
+extern TLS bool mcount_recursion_marker attr_tls_model;
 
-void __mcount_guard_recursion(struct mcount_thread_data *mtdp);
-void __mcount_unguard_recursion(struct mcount_thread_data *mtdp);
-bool mcount_guard_recursion(struct mcount_thread_data *mtdp);
-void mcount_unguard_recursion(struct mcount_thread_data *mtdp);
+# define get_thread_data() _mtdp
+# define check_thread_data(mtdp)  (mtdp == NULL)
+
+bool __mcount_check_dead(struct mcount_thread_data *mtdp);
+void __mcount_should_dead(struct mcount_thread_data *mtdp);
+void __mcount_guard_recursion();
+void __mcount_unguard_recursion();
+bool mcount_guard_recursion();
+void mcount_unguard_recursion();
 
 extern uint64_t mcount_threshold;  /* nsec */
 extern pthread_key_t mtd_key;
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -51,7 +51,8 @@ unsigned long mcount_global_flags = MCOUNT_GFL_SETUP;
 pthread_key_t mtd_key = (pthread_key_t)-1;
 
 /* thread local data to trace function execution */
-TLS struct mcount_thread_data mtd;
+TLS struct mcount_thread_data *_mtdp attr_tls_model;
+TLS bool mcount_recursion_marker attr_tls_model;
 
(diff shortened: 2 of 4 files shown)


libmcount/internal.h
```diff
@@ -145,20 +144,24 @@ static inline void mcount_restore_arch_context(struct mcount_arch_context *ctx)
...
 #else
 # define TLS  __thread
-# define get_thread_data()  pthread_getspecific(mtd_key)
-# define check_thread_data(mtdp)  (mtdp == NULL)
+# define attr_tls_model __attribute__((tls_model("initial-exec")))
 #endif
 
-extern TLS struct mcount_thread_data mtd;
+extern TLS struct mcount_thread_data *_mtdp attr_tls_model;
+extern TLS bool mcount_recursion_marker attr_tls_model;
```
namhyung: You need to add explanation why you use a pointer to mtd rather than the mtd itself and why the recursion marker is separated from the mtd.

libmcount/internal.h
```diff
@@ -145,20 +144,24 @@ static inline void mcount_restore_arch_context(struct mcount_arch_context *ctx)
...
+
+# define get_thread_data() _mtdp
+# define check_thread_data(mtdp)  (mtdp == NULL)
```
namhyung: But it requires all the callers to have a local variable named `mtdp` (though it does for now).

libmcount/internal.h
```diff
@@ -145,20 +144,24 @@ static inline void mcount_restore_arch_context(struct mcount_arch_context *ctx)
...
-void __mcount_guard_recursion(struct mcount_thread_data *mtdp);
-void __mcount_unguard_recursion(struct mcount_thread_data *mtdp);
-bool mcount_guard_recursion(struct mcount_thread_data *mtdp);
-void mcount_unguard_recursion(struct mcount_thread_data *mtdp);
+bool __mcount_check_dead(struct mcount_thread_data *mtdp);
+void __mcount_should_dead(struct mcount_thread_data *mtdp);
+void __mcount_guard_recursion();
+void __mcount_unguard_recursion();
+bool mcount_guard_recursion();
+void mcount_unguard_recursion();
```
namhyung: For C, you'd better adding `void` type for function prototype.

libmcount/internal.h
```diff
@@ -145,20 +144,24 @@ static inline void mcount_restore_arch_context(struct mcount_arch_context *ctx)
...
 #ifdef SINGLE_THREAD
 # define TLS
-# define get_thread_data()  &mtd
-# define check_thread_data(mtdp)  (mtdp->rstack == NULL)
+# define attr_tls_model
 #else
 # define TLS  __thread
-# define get_thread_data()  pthread_getspecific(mtd_key)
-# define check_thread_data(mtdp)  (mtdp == NULL)
+# define attr_tls_model __attribute__((tls_model("initial-exec")))
```
namhyung: Do you think we need to have other TLS model as well?  If not why not changing the CFLAGS instead of source code?

honggyukim:
I've just tested this and found that it works fine for node programs both for arm and thumb modes. But it doesn't fix the stack overflow problem on chrome execution and the symptom is same as before. It may be a different issue.

honggyukim:
The stack overflow problem can be fixed by #725. It's not related to TLS issue.

honggyukim:
There is an option to control `tls-model` as @namhyung mentioned.
```
$ man gcc

-ftls-model=model

  Alter the thread-local storage model to be used.  The model argument
  should be one of global-dynamic, local-dynamic, initial-exec or local-
  exec.  Note that the choice is subject to optimization: the compiler
  may use a more efficient model for symbols not visible outside of the
  translation unit, or if -fpic is not given on the command line.

  The default without -fpic is initial-exec; with -fpic the default is
  global-dynamic.
```

namhyung:
@ParkHanbum could you please update the PR to contain just built flag change?  Other changes can be included to the inject work.

ParkHanbum:
@namhyung this PR have purpose to fix infinity recursive calling in attaching. but change the TLS model bring to us only advantage for performance slightly. I think this is not critical and not necessary for now. it has better to close this and push it again with attaching feature for later. how do you think?

namhyung:
OK, I'm fine with it.

PR #707, opened 2019-03-20
record: Send logfile for -H/--host option

honggyukim:
It'd be helpful to send logfile as well as recorded trace data.

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1390,6 +1390,14 @@ static void send_event_file(int sock, const char *dirname)
 	send_trace_metadata(sock, dirname, "events.txt");
 }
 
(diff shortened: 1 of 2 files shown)


honggyukim:
In addition to this, I'm wondering why logfile is opened in "append" mode. I always have to remove the previous logfile to see the correct log, so it seems to be changed to just "write" mode.

namhyung:
It's natural to use 'append' mode for logfiles since it guarantees atomicity for concurrent writes.  You'd better use a different name for different session or remove the old file like you did.

honggyukim:
Ah.. It's to collect logs from multiple processes and threads. Thanks for the explanation :)

PR #708, opened 2019-03-23
uftrace: Fix coding style from whitespaces to tabs

honggyukim:
This patch only changes whitespaces to tabs and no functional change is
applied.


honggyukim:
No problem. I also often switch expandtab on and off :)

PR #713, opened 2019-03-26
utils: Fix segfault while busybox checking

honggyukim:
In some environments, path resolution is failed so it gets crashed while
checking if it's busybox or not.

This patch adds a fail safe routine for such cases.


bavernet:
Without the patch, I found that uftrace had a segfault with any commands except `tui` on my machine with ubuntu 16.04.6 LTS and zsh.
After applying it, the segfault error is disappeared and it works well.

honggyukim:
@bavernet Thanks a lot for testing it :)

PR #716, opened 2019-03-31
Field/refine with module addr

honggyukim:
This PR includes some refactoring on field printing and adds 'module_addr'
field to show relative address from the base address of module.

The change logs are as follows:
- field: Add 'addr' to struct field_data
- field: Add 'print' member to struct field_data
- field: Combine address print functions
- field: Fix the output of -f addr to fit in the width
- replay: Add -f module_addr field
- replay: live: Update man page for 'module_addr` field

Especially it adds `-f module_addr` field in replay command. The existing 'addr'
field is useful information, but it's difficult to recalculate the relative address
after subtracting by the base address of module.

The address of 'module_addr' is useful when reading the disassembled output
because the address is exactly matched with it.

The example output looks as follows:
```
  $ uftrace replay -f module,addr,module_addr
  #     MODULE NAME    ADDRESS    MODULE ADDR   FUNCTION
           t-lib.pg       400888       400888 | main() {
           t-lib.pg       40086f       40086f |   foo() {
           t-lib.pg       4006c0       4006c0 |     lib_a() {
   libabc_test_lib. 7fcdd46446de          6de |       lib_a() {
   libabc_test_lib. 7fcdd4644701          701 |         lib_b() {
   libabc_test_lib. 7fcdd4644724          724 |           lib_c() {
           [kernel] ffff8106dba0 ffff8106dba0 |             __do_page_fault();
   libabc_test_lib. 7fcdd46445b0          5b0 |             getpid();
   libabc_test_lib. 7fcdd4644724          724 |           } /* lib_c */
   libabc_test_lib. 7fcdd4644701          701 |         } /* lib_b */
   libabc_test_lib. 7fcdd46446de          6de |       } /* lib_a */
           t-lib.pg       4006c0       4006c0 |     } /* lib_a */
           t-lib.pg       40086f       40086f |   } /* foo */
           t-lib.pg       400888       400888 | } /* main */
```
For example, the module address of 'getpid()' is `5b0` and it can be easily found
in the disassembled output of the module.
```
  $ objdump -d libabc_test_lib.so
      ...
  00000000000005b0 <getpid@plt>:
   5b0:   ff 25 62 0a 20 00       jmpq   *0x200a62(%rip)
   5b6:   68 00 00 00 00          pushq  $0x0
   5bb:   e9 e0 ff ff ff          jmpq   5a0 <_init+0x20>
```
It can be useful when debugging the trace.


utils/field.h
```diff
@@ -9,6 +9,7 @@ struct field_data {
 	struct uftrace_task_reader *task;
 	struct fstack *fstack;
 	void *arg;
+	unsigned long addr;
```
namhyung: It's unnatural to add 'addr' field as it can be get from others.  Why do you make it special?  Also `unsigned long` is not sufficient since 32 bit system should run 64 bit data.

utils/field.h
```diff
@@ -10,6 +10,7 @@ struct field_data {
 	struct fstack *fstack;
 	void *arg;
 	unsigned long addr;
+	void (*print)(const char *fmt, ...);
 };
```
namhyung: This also looks strange since it's only used for a single field.

utils/field.c
```diff
@@ -49,6 +49,17 @@ int print_empty_field(struct list_head *output_fields, int space)
...
+void print_addr(struct field_data *fd)
+{
+	/* uftrace records (truncated) 48-bit addresses */
+	int width = sizeof(long) == 4 ? 8 : 12;
+
+	if (fd->addr == 0)  /* for EVENT or LOST record */
+		fd->print("%*s", width, "");
+	else
+		fd->print("%*lx", width, fd->addr);
+}
```
namhyung: I understand your intention to have a common function to deal with different address sizes.  But I'm not sure it's worth specializing the address handling.

Maybe we could add a function to get a portion of the effective address from a given address just before printing.  Also using `sizeof(long)` is wrong since it didn't consider cross-platform data analysis.  I know it's not your fault, but I'd like to change it to use size of address in the data file - something like `data_is_lp64()`.

utils/field.h
```diff
@@ -4,6 +4,8 @@
 #include "utils/fstack.h"
 #include "utils/list.h"
 
+#define ADDR_IN_48_BITS(addr) ((addr) & 0xffffffffffffUL)
```
namhyung: I guess the `UL` suffix is not enough for 32 bit systems.  Please make it `uint64_t` for sure.  Also I prefer having it as a (static inline) function over a macro (like `uint64_t effective_addr(uint64_t addr)`?)

cmds/replay.c
```diff
@@ -108,6 +108,41 @@ static void print_module(struct field_data *fd)
...
+
+	/* for EVENT or LOST record */
+	if (fstack == NULL) {
+		pr_out("%*s", width, "");
+		return;
+	}
+
+	s = find_task_session(&task->h->sessions, task->t, timestamp);
+	if (s == NULL) {  /* LOST */
+		pr_out("%*s", width, "");
```
namhyung: If you add `return` here, the else block can be unindented.

cmds/replay.c
```diff
@@ -185,6 +220,20 @@ static struct display_field field_module = {
...
+static struct display_field field_module_addr = {
+	.id      = REPLAY_F_MODULE_ADDR,
+	.name    = "module_addr",
+#if __SIZEOF_LONG__ == 4
+	.header  = "MOD ADDR",
+	.length  = 8,
+#else
+	.header  = " MODULE ADDR",
+	.length  = 12,
+#endif
```
namhyung: This should be changed at runtime as well depending on the size of the data.

cmds/replay.c
```diff
@@ -185,6 +220,20 @@ static struct display_field field_module = {
 	.list    = LIST_HEAD_INIT(field_module.list),
 };
 
+static struct display_field field_module_addr = {
+	.id      = REPLAY_F_MODULE_ADDR,
+	.name    = "module_addr",
```
namhyung: How about adding an alias like `modaddr` for short?

honggyukim:
@namhyung I appreciate your feedback. I will update this when I have some time.

honggyukim:
I just fixed some conflicts without any change yet. Never mind yet.

PR #717, opened 2019-04-01
uftrace: Revert 'shell-support' branch

honggyukim:
It seems to be tricky to handle shell script tracing.  Especially, the
problem looks clear when tracing a big program such as chrome.

It shows too many small ignorable short trace records and it increases
the number of tasks too much.

To avoid such problems, this patch reverts the following commits:

  0eab415d Merge branch 'shell-support'
   85c3202f test: Add a test case for script tracing
   8f910e8b record: Support tracing script files
   ae3faff7 symbol: Add check_script_file()


honggyukim:
Tracing shell script is useful but I have some problems when tracing chrome because it creates too many small tasks and it makes the trace difficult to understand.  I think it'd be better to revert shell script tracing until we find the better solution.

honggyukim:
I misundertood something. It's not the problem of shell script support. After using named pipe, chrome has been able to trace renderer processes and it seems that the renderer processes internally calls some bash scripts. I would like to filter them out but don't have a good idea yet.

PR #724, opened 2019-04-12
mcount: Fix mtdp assertion failure in plthook_exit

honggyukim:
The mtd_key deletion in mcount_cleanup() sometimes makes other thread
get crashed because they may try to get mtdp based on this mtd_key after
being deleted.  The error message is as follows:

    plthook_exit: Assertion `!check_thread_data(mtdp)' failed.

The expected behavior is to call mcount_cleanup() after all the other
threads are destoryed, but it doesn't work as expected in some cases.

Since this key deletion is not that important, it'd be better not to do
it until we find a better solution.

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -1751,8 +1751,16 @@ static void mcount_cleanup(void)
 	mcount_finish();
 	destroy_dynsym_indexes();
 
+#if 0
+	/*
+	 * This mtd_key deletion sometimes makes other thread get crashed
+	 * because they may try to get mtdp based on this mtd_key after being
+	 * deleted.  Since this key deletion is not mandatory, it'd be better
+	 * not to delete it until we find a better solution.
+	 */
 	pthread_key_delete(mtd_key);
 	mtd_key = -1;
+#endif
 
 	mcount_filter_finish();
 


honggyukim:
Hi @namhyung, I've seen this crash quite often when testing it in chrome and other modules. Please let me know if you have better solution than this.

namhyung:
I have no idea why it's called after the destructor but it's ok to leave the key not deleted for this problem.  But I don't want to change messages unless it's really needed.

honggyukim:
Do you mean the debug messages?

namhyung:
Ah, yes

honggyukim:
I simply commented out the `mtd_key` deletion part without changing the debug messages. Thanks.

namhyung:
I guess that when `mcount_cleanup()` is called after all threads was dead and then the program is going to exit.  So I'm not sure this patch fixes the problem - yes it won't trigger the assertion failure but underlying code in the plthook might have some problem still...

honggyukim:
I can only say it from my experience as of now. But if so, would it be possible to call `pthread_join` here to wait for all the other threads? I don't know how we can get the thread IDs here though.

namhyung:
I guess it could be racy and dangerous since it could cause some problems like lockup...

honggyukim:
Hmm.. if so, I don't have any better idea than this.

namhyung:
So you didn't see any problem after applying this patch?

honggyukim:
I saw the assertion fail quite often before in chrome, but I need to test it again. Current my target is qml engine and it clearly shows the difference with this change.

namhyung:
So I worry about that it's not a real fix.  But as it's the exit path in the process, I'd like to know whether it's just ok for most (or many?) cases...

honggyukim:
Hmm.. It's very strange that chrome in x86_64 never crashed in the current master and cannot see this assertion fail unlike before. I often saw it's crashed when the tab is closed before.

chrome survived with 3 tabs - youtube, naver, and cnn. I played youtube video more than 5 mins and switched many times from tab to tab, but never crashed. :)(

namhyung:
Then it's ok but I also care about the QML on ARM.

honggyukim:
QML on ARM gets crashed 100% without this.

namhyung:
how often does it crash with this?

honggyukim:
It always crashed but never crashed with this change.

namhyung:
Did it finish normally (without a problem)?

honggyukim:
Yes, actually the return value was a bit different but normally finished.

PR #725, opened 2019-04-12
plthook: Skip dlsym to avoid infinite loop

honggyukim:
In some cases, it's tricky to avoid infinite loop if dlsym is hooked
by libmcount.so.  Here is the problematic condition.

1. dlsym is invoked in a library. Let's say "libxxx.so".
2. dlsym is called to get the address of "foo". - dlsym(RTLD_NEXT, "foo")
3. The actual implementation of "foo" is inside "libfoo.so".
4. Inside the library "libxxx.so", "foo" is defined to override it.
5. The execution sequence is "foo@libxxx.so", then "foo@libfoo.so".

If "libmcount.so" is preloaded in this case, it hooks "foo@libxxx.so"
and executes "dlsym" instead of "libxxx.so".

The problem is that the return address of dlsym(RTLD_NEXT, "foo") is
expected the address of "foo@libfoo.so", but it returns "foo@libxxx.so".

If the subsequent call is made based on this return address, then
"foo@libxxx.so" recursively calls itself and never finished.

Since I don't have any better solution to make dlsym to find second next
"foo" in the order of dynamic library list, this patch simply skips
dlsym to avoid the such problem.

This problem is happened when --nest-libcall is used.

Fixed: #714

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -323,6 +323,7 @@ static const char *skip_syms[] = {
 	"__cxa_end_catch",
 	"__cxa_finalize",
 	"_Unwind_Resume",
+	"dlsym",
 };
 
 static const char *setjmp_syms[] = {


honggyukim:
To simply see the problem, I will post the same reproducible example as I wrote in #714.
```cpp
$ cat localtime_r.cc
#include <stdio.h>
#include <pthread.h>
#include <time.h>
#include <dlfcn.h>

typedef struct tm* (*LocaltimeRFunction)(const time_t* timep, struct tm* result);

static pthread_once_t g_libc_localtime_funcs_guard = PTHREAD_ONCE_INIT;
static LocaltimeRFunction g_libc_localtime_r;

static void InitLibcLocaltimeFunctions() {
  g_libc_localtime_r = (LocaltimeRFunction)dlsym(RTLD_NEXT, "localtime_r");
}


__attribute__ ((__visibility__("default")))
struct tm* localtime_r_override(const time_t* timep,
                                struct tm* result) __asm__ ("localtime_r");

__attribute__ ((__visibility__("default")))
struct tm* localtime_r_override(const time_t* timep, struct tm* result) {
    pthread_once(&g_libc_localtime_funcs_guard, InitLibcLocaltimeFunctions);
    struct tm* res = g_libc_localtime_r(timep, result);
    return res;
}
```
```cpp
$ cat main.cc
#include <stdio.h>
#include <pthread.h>
#include <time.h>
#include <dlfcn.h>

int main(int argc, char **argv)
{
        time_t t;
        struct tm lt;

        t = time(NULL);
        localtime_r(&t, &lt);

        printf("current time: %04d-%02d-%02d %02d:%02d:%02d\n",
                lt.tm_year + 1900, lt.tm_mon + 1, lt.tm_mday,
                lt.tm_hour, lt.tm_min, lt.tm_sec);

        return 0;
}
```
```
$ g++ -pg -g -shared -fPIC -o libxxx.so localtime_r.cc -ldl -pthread

$ g++ -pg -g main.cc -L. -Wl,-rpath=. -lxxx -ldl -pthread

$ ldd a.out
        linux-vdso.so.1 =>  (0x00007ffe54d6f000)
        libxxx.so => ./libxxx.so (0x00007f77c19b1000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f77c15e7000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f77c13e3000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f77c11c6000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f77c1bb3000)
```
It calls `localtime_r` twice - one from `a.out` and the other from `libxxx.so`.
```
$ uftrace record -a a.out
current time: 2019-04-07 23:53:12

$ uftrace replay -f module,addr
#     MODULE NAME    ADDRESS     FUNCTION
            a.out       400923 | main(1, 0x7ffc00e92308) {
            a.out       400790 |   time();
            a.out       4007c0 |   localtime_r() {
        libxxx.so 7f2920b877b5 |     localtime_r(0x7ffc00e921d8, 0x7ffc00e921e0) {
        libxxx.so 7f2920b8778a |       InitLibcLocaltimeFunctions();
        libxxx.so 7f2920b877b5 |     } = 0x7ffc00e921e0; /* localtime_r */
            a.out       4007c0 |   } /* localtime_r */
            a.out       400760 |   printf("current time: %04d-%02d-%02d %02d:%02d:%02d\n") = 34;
            a.out       400923 | } = 0; /* main */
```
However, it falls into infinite recursion when `--nest-libcall` is given.
```
$ uftrace record -a --nest-libcall a.out
WARN: too deeply nested calls: 1024

$ uftrace replay -f module,addr
#     MODULE NAME    ADDRESS     FUNCTION
            a.out       400923 | main(1, 0x7ffd1b69b488) {
            a.out       400790 |   time();
            a.out       4007c0 |   localtime_r() {
        libxxx.so 7f0e1f3a77b5 |     localtime_r(0x7ffd1b69b358, 0x7ffd1b69b360) {
        libxxx.so 7f0e1f3a7650 |       pthread_once(&g_libc_localtime_funcs_guard, &InitLibcLocaltimeFunctions) {
        libxxx.so 7f0e1f3a778a |         InitLibcLocaltimeFunctions() {
        libxxx.so 7f0e1f3a7660 |           dlsym(0xffffffffffffffff, "localtime_r") = &localtime_r;
        libxxx.so 7f0e1f3a778a |         } /* InitLibcLocaltimeFunctions */
        libxxx.so 7f0e1f3a7650 |       } = 0; /* pthread_once */
        libxxx.so 7f0e1f3a77b5 |       localtime_r(0x7ffd1b69b358, 0x7ffd1b69b360) {
        libxxx.so 7f0e1f3a7650 |         pthread_once(&g_libc_localtime_funcs_guard, &InitLibcLocaltimeFunctions) = 0;
        libxxx.so 7f0e1f3a77b5 |         localtime_r(0x7ffd1b69b358, 0x7ffd1b69b360) {
        libxxx.so 7f0e1f3a7650 |           pthread_once(&g_libc_localtime_funcs_guard, &InitLibcLocaltimeFunctions) = 0;
        libxxx.so 7f0e1f3a77b5 |           localtime_r(0x7ffd1b69b358, 0x7ffd1b69b360) {
        libxxx.so 7f0e1f3a7650 |             pthread_once(&g_libc_localtime_funcs_guard, &InitLibcLocaltimeFunctions) = 0;
        libxxx.so 7f0e1f3a77b5 |             localtime_r(0x7ffd1b69b358, 0x7ffd1b69b360) {
    ...
```

namhyung:
Hmm.. actually the expected return value of `dlsym(RTLD_DEFAULT, "foo")` in `libmcount` is `libxxx`.  But the `dlsym(RTLD_NEXT, "foo")` in `libxxx` should find a symbol in `libfoo`.  I don't know why it cannot find the symbol in `libfoo` since uftrace doesn't have the symbol ("foo").

honggyukim:
> Hmm.. actually the expected return value of dlsym(RTLD_DEFAULT, "foo") in libmcount is libxxx.

Right.

> But the dlsym(RTLD_NEXT, "foo") in libxxx should find a symbol in libfoo.

The problem is that `dlsym(RTLD_NEXT, "foo")` in libxxx actually executed inside libmcount instead, so dynamic linker thinks that the next library that has "foo" is libxxx again.

The tricky part is that "foo" is defined in both libxxx and libfoo.

honggyukim:
The above simple example can reproduce the problem in x86_64.

ParkHanbum:
nice job. it seems not easy.

namhyung:
> The problem is that dlsym(RTLD_NEXT, "foo") in libxxx actually executed inside libmcount instead

Do you mean `resolve_pltgot()`?  I still don't understand why it's a problem since libmcount doesn't have "foo" symbol.

honggyukim:
I think dynamic linker thinks that the current library is `libmcount.so` instead of `libxxx.so`, so it tries to find symbol "foo" in the next library, which is `libxxx.so` again.

honggyukim:
>> The problem is that dlsym(RTLD_NEXT, "foo") in libxxx actually executed inside libmcount instead

> Do you mean resolve_pltgot()? I still don't understand why it's a problem since libmcount doesn't have "foo" symbol.

From the perspective of dynamic linker, the library search order is as follows:
1. libxxx.so
2. libfoo.so

So `dlsym(RTLD_NEXT, "foo")` in `libxxx.so` returns "foo" of the next library even though it also has "foo" inside. However, `dlsym` of `libxxx.so` is hooked by `libmcount.so` so the library search order becomes as follows:
1. libmcount.so
2. libxxx.so
3. libfoo.so

If `dlsym(RTLD_NEXT, "foo")` is executed inside `libmcount.so` instead, it searches "foo" from the next library, but the next library of current `libmcount.so` is `libxxx.so` because `libxxx.so` comes before `libfoo.so`.  So `dlsym` returns the address of "foo" in `libxxx.so` again and it recursively calls itself and never finished.

namhyung:
> dlsym of libxxx.so is hooked by libmcount.so

I don't know where in the code you mentioned above.  I don't think we hook `dlsym` itself, only does for `dlopen` (see libmcount/wrap.c).

honggyukim:
The following code is compiled as `libxxx.so`.
```cpp
$ cat localtime_r.cc
#include <stdio.h>
#include <pthread.h>
#include <time.h>
#include <dlfcn.h>

typedef struct tm* (*LocaltimeRFunction)(const time_t* timep, struct tm* result);

static pthread_once_t g_libc_localtime_funcs_guard = PTHREAD_ONCE_INIT;
static LocaltimeRFunction g_libc_localtime_r;

static void InitLibcLocaltimeFunctions() {
  g_libc_localtime_r = (LocaltimeRFunction)dlsym(RTLD_NEXT, "localtime_r");
}


__attribute__ ((__visibility__("default")))
struct tm* localtime_r_override(const time_t* timep,
                                struct tm* result) __asm__ ("localtime_r");

__attribute__ ((__visibility__("default")))
struct tm* localtime_r_override(const time_t* timep, struct tm* result) {
    pthread_once(&g_libc_localtime_funcs_guard, InitLibcLocaltimeFunctions);
    struct tm* res = g_libc_localtime_r(timep, result);
    return res;
}
```

honggyukim:
> I don't know where in the code you mentioned above. I don't think we hook dlsym itself, only does for dlopen (see libmcount/wrap.c).

I'm not sure why `dlsym` returns the address of the function in the same library even with `RTLD_NEXT`, but you can try the above reproducible example. I clearly observed this problem.

namhyung:
I'm just trying to reason about the cause...

honggyukim:
It's not directly related to this issue, but this blog post explains well enough about how `dlsym` finds the function.
- http://optumsoft.com/dangers-of-using-dlsym-with-rtld_next/

honggyukim:
@namhyung closed this in f4ff91a.

Issue #735, opened 2019-05-01
chrome dump output bug


honggyukim:
There is a bug when dumping it in `--chrome` format so cannot be opened in a chrome browser.
```
            [ 63030]  |           fill_taskinfo(0x7ffd66261820) {
            [ 63030]  |             read_task_txt_file(0x7ffd66261700, "aaa", 0, 0) {
   0.992 us [ 63030]  |               __asprintf_chk();
   2.249 us [ 63030]  |               fopen("aaa/task.txt", "r") = 0x2651d60;
   3.774 us [ 63030]  |               __getdelim();
   3.188 us [ 63030]  |               sscanf();
   1.580 us [ 63030]  |               strstr("SESS timestamp=7121477.564831948 pid=63044 sid=12cbd8d09e0683ec exename="/home/honggyu/work/uft...", "exename=") = "exename="/home/honggyu/work/uftrace/t-abc"\n";
   1.266 us [ 63030]  |               strrchr("/home/honggyu/work/uftrace/t-abc"\n", '"') = ""\n";
```
In the above replay output, the last line is as follows:
```
$ uftrace replay -F strrchr -f none
strrchr("/home/honggyu/work/uftrace/t-abc"\n", '"') = ""\n";
```
Here is the dump output of `strrchr` above.
```js
$ uftrace dump --chrome -F strrchr
    ...
{...,"args":{"arguments":"(\"/home/honggyu/work/uftrace/t-abc\"\n\", '"')"}},
    ...
```
The problem is because of the second argument as `'"')"`.

PR #736, opened 2019-05-01
chrome: Fix a bug when displaying character type

honggyukim:
It didn't fully handle the char type dump when --chrome option is used.
The following dump output gets failed when loading it in chrome browser.
```
  {...,"args":{"arguments":"('"')"}},
```
It's because the second argument '"' is misinterpreted as " is the
content, but it looks like a closing quotation mark.  The " character
should be escaped from '"' to '\"' as follows:
```
  {...,"args":{"arguments":"('\"')"}},
```
This also handles \ character to be escaped because the following json
cannot be interpreted in chrome browser.
```
  {...,"args":{"arguments":"(\"Next\:\")"}},
```
It becomes as follows:
```
  {...,"args":{"arguments":"(\"Next\\:\")"}},
```
Fixed: #735

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -447,6 +447,22 @@ static void print_task_newline(int current_tid)
 #define print_args(fmt, ...)						\
 ({ int _x = snprintf(args + n, len, fmt, ##__VA_ARGS__); n += _x; len -= _x; })
 
+#define print_escaped_char(c)						\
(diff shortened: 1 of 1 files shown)


honggyukim:
I found this problem when I dump in chrome format recorded for `uftrace record` and `uftrace replay` commands.

PR #737, opened 2019-05-02
record: Set writer thread name to "WriterThread"

honggyukim:
It'd be better to have a task name for writer threads.


honggyukim:
I tried both "WriterThread" and "writer-thread" but I personally think that the first one looks better.

Issue #474, opened 2018-09-04
Finding target binary in $PATH


honggyukim:
uftrace can also be used for pre-built binaries with `--force` or `--nest-libcall` so I think it can just find target binary in $PATH instead of writing its full path, which doesn't seem to be intuitive.

I think It makes easier to use for new users. Please let me know how you think about it.

namhyung:
Maybe we can enable it when those options are given, or (full) dynamic tracing is available.

honggyukim:
Okay. I'm fine with that.

honggyukim:
We now have dynamic tracing support so I implemented it in #741.

PR #741, opened 2019-05-03
record: Find the execution binary in PATH

honggyukim:
Currently uftrace explicitly requires to have a correct pathname, so it
has to be used with full pathname as follows:

    $ uftrace --force /usr/bin/python -c ''

This patch allows the user can provide the command name, which is placed
one of directories in PATH environment.
So uftrace can be simply used as follows:

    $ uftrace --force python -c ''

It'd be helpful especially when tracing with -P option and it will make
easier to use full dynamic tracing feature as follows:

    $ uftrace -P. -a python -c 'print("Hi, Python")'

Closes: #474

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1467,10 +1467,7 @@ static void print_child_usage(struct rusage *ru)
 	       ru->ru_utime.tv_sec, ru->ru_utime.tv_usec);
 }
 
-#define UFTRACE_MSG  "Cannot trace '%s': No such executable file\n"	\
-"\tNote that uftrace doesn't search $PATH for you.\n"			\
-"\tIf you really want to trace executables in the $PATH,\n"		\
-"\tplease give it the absolute pathname (like /usr/bin/%s).\n"
+#define UFTRACE_MSG  "Cannot trace '%s': No such executable file.\n"
 
 #define MCOUNT_MSG  "Can't find '%s' symbol in the '%s'.\n"		\
 "\tIt seems not to be compiled with -pg or -finstrument-functions flag\n" 	\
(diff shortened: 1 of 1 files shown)


honggyukim:
Hi @namhyung, I think we better find the executable file in `PATH` environment for easier usage.
```
$ uftrace -P. -a echo "Hello"
Hello
 321.932 us [177318] | getenv("POSIXLY_CORRECT") = "NULL";
   1.644 us [177318] | strrchr("echo", '/') = "NULL";
  62.873 us [177318] | setlocale(LC_ALL, "") = "en_US.UTF-8";
   7.377 us [177318] | bindtextdomain("coreutils", "/usr/share/locale");
   1.876 us [177318] | textdomain("coreutils");
   1.037 us [177318] | __cxa_atexit();
   1.454 us [177318] | strcmp("Hello", "--help") = 27;
   0.523 us [177318] | strcmp("Hello", "--version") = 27;
   3.773 us [177318] | fputs_unlocked();
  12.184 us [177318] | __overflow();
   1.457 us [177318] | __fpending();
   1.280 us [177318] | fileno(&_IO_2_1_stdout_) = 1;
   0.803 us [177318] | __freading();
   0.190 us [177318] | __freading();
   1.104 us [177318] | fflush(&_IO_2_1_stdout_) = 0;
   2.580 us [177318] | fclose(&_IO_2_1_stdout_) = 0;
   0.184 us [177318] | __fpending();
   0.307 us [177318] | fileno(&_IO_2_1_stderr_) = 2;
   0.163 us [177318] | __freading();
   0.146 us [177318] | __freading();
   0.376 us [177318] | fflush(&_IO_2_1_stderr_) = 0;
   1.230 us [177318] | fclose(&_IO_2_1_stderr_) = 0;
```
It can be used just like any other tracing tools such as `strace`, `ltrace`, etc. Please let me know how you think about it.

cmds/record.c
```diff
@@ -1497,6 +1494,52 @@ static void print_child_usage(struct rusage *ru)
 # define EM_AARCH64  183
 #endif
 
+static bool is_regular_executable(const char *pathname)
+{
+	struct stat sb;
+
+	if (!stat(pathname, &sb))
+		if (S_ISREG(sb.st_mode) && (sb.st_mode & S_IXUSR))
+			return true;
```
namhyung: I prefer adding curly brackets to the outer if statement when body is more than one line (even when not strictly needed).

cmds/record.c
```diff
@@ -1497,6 +1494,52 @@ static void print_child_usage(struct rusage *ru)
...
+	bool found = false;
+
+	if (!env || *exename[0] == '/')
+		pr_err_ns(UFTRACE_MSG, *exename);
+
+	path = fullpath = xstrdup(env);
+
+	/* search opts->exename in PATH one by one */
+	p = strchr(path, ':');
+	while (p) {
```
namhyung: You'd better using `strv` for this kind of string handling.

cmds/record.c
```diff
@@ -1510,14 +1553,11 @@ static void check_binary(struct opts *opts)
 	};
 
 again:
-	pr_dbg3("checking binary %s\n", opts->exename);
+	pr_dbg("checking binary %s\n", opts->exename);
```
namhyung: What about moving this under the `find_in_path()`?

cmds/record.c
honggyukim: Okay.

cmds/record.c
honggyukim: Applied.

cmds/record.c
honggyukim: If I move this message inside `find_in_path`, then it will not print anything when the user gives the full pathname. The first `is_regular_executable` check will be successful so doesn't call `find_in_path` in this case.

honggyukim:
@namhyung I just updated it again based on your feedback. Thanks!

cmds/record.c
namhyung: Oh, I meant the callsite - as `find_in_path()` can change the exename, it'd be nice to print the final name IMO.

honggyukim: I see. I think it's enough to place after the searching the binary as I updated. Thanks.

honggyukim:
Thanks for merging this. Let's modify documents that includes the absolute path and `which` command. I mean let's keep it mind before making a release, not right now. :)

namhyung:
You can file an issue for that. :)

Issue #745, opened 2019-05-05
map file open failure due to internal chdir()


honggyukim:
Here is the sequence to reproduce the problem.
```                                                                                                                                                                                                                                                               
$ uftrace record --force ./configure
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
uftrace detected system features:
...         prefix: /home/honggyu/usr
...         libelf: [ on  ] - more flexible ELF data handling
...          libdw: [ on  ] - DWARF debug info support
...   libpython2.7: [ on  ] - python scripting support
...    libncursesw: [ on  ] - TUI support
...   cxa_demangle: [ on  ] - full demangler support with libstdc++
...     perf_event: [ on  ] - perf (PMU) event support
...       schedule: [ on  ] - scheduler event support
...       capstone: [ on  ] - full dynamic tracing support
```

honggyukim:
Here are the sessions of the above record.
```
$ uftrace tui
Key uftrace command
 G  call Graph for session #1: env
    call Graph for session #2: bash
    call Graph for session #3: readlink
    call Graph for session #4: dirname
    call Graph for session #5: readlink
    call Graph for session #6: uname
    call Graph for session #7: sed
    call Graph for session #8: grep
    call Graph for session #9: readlink
    call Graph for session #10: make
    call Graph for session #11: make
    call Graph for session #12: cat
    call Graph for session #13: cat
    call Graph for session #14: id
 R  Report functions
 I  uftrace Info
 h  Help message
 q  quit
```

Taeung:
I think this error `ENOENT 2 No such file or directory` of fopen() for sid-xxxxx.map can be caused by changed current path. For example, the first current path is `/home/taeung/uftrace/`
but it can be changed during shell-script execution.

[...]

PR #748, opened 2019-05-09
record: Use absolute path of uftrace data directory when tracing (shell) scripts

Taeung:
Current working directory can be changed during (shell) scripts execution
due to internal chdir() inside scripts.
If it is changed, the problem that uftrace.data can't be found can occur:

```
  $ uftrace record --force ./configure
  mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
   ERROR: cannot open for writing maps file: No such file or directory
```

Because of trying to create "uftrace.data/sid-xxxxx.map" of which the prefix
is a relative path. So convert uftrace data directory path that can be
relative to an absolute path in tracing (shell) scripts case.

Fixed: #745

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -2012,6 +2012,7 @@ int do_child_exec(int ready, struct opts *opts,
 {
 	uint64_t dummy;
 	char *shebang;
+	char fullpath[PATH_MAX];
 	struct strv new_args = STRV_INIT;
 
 	if (opts->no_randomize_addr) {
@@ -2019,6 +2020,13 @@ int do_child_exec(int ready, struct opts *opts,
 		if (personality(ADDR_NO_RANDOMIZE) < 0)
 			pr_dbg("disabling ASLR failed\n");
 	}
+
+	/*
+	 * The current working directory can be changed by calling chdir.
+	 * So dirname has to be converted to an absolute path to avoid unexpected problems.
+	 */
+	if (realpath(opts->dirname, fullpath) != NULL)
+		opts->dirname = fullpath;
 
 	shebang = check_script_file(argv[0]);
 	if (shebang) {


honggyukim:
Thanks for doing this. I think it's better to make the title of the commit shorter.
What about "record: Use absolute path for uftrace.data"?

Taeung:
It is to use absolute path in only tracing script case.
So how about "record: Use absolute path for uftrace.data in tracing script case" ?

honggyukim:
It's not only for script files. Here is another program that makes the problem.
```c
$ cat chdir-test.c 
#include <stdio.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/wait.h>

int main()
{
        if (fork()) {
                printf("parent!\n");
                wait(0);
        }
        else {
                printf("child!\n");
                chdir("cmds");
                execl("/bin/pwd", "/bin/pwd", NULL);
        }
}
```
```
$ gcc -pg chdir-test.c

$ ./a.out
parent!
child!
/home/honggyu/work/uftrace/cmds
```
But it makes the same problem when tracing with uftrace.
```
$ uftrace record ./a.out 
parent!
child!
mcount: /home/honggyu/work/uftrace/libmcount/record.c:1093:record_proc_maps
 ERROR: cannot open for writing maps file: No such file or directory
```
So it's not only for shell script problem, but related to `chdir` usage.

honggyukim:
The following change covers both cases.
```diff
diff --git a/cmds/record.c b/cmds/record.c
index c909c255..b7891341 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -2010,6 +2010,7 @@ int do_child_exec(int ready, struct opts *opts,
 {
        uint64_t dummy;
        char *shebang;
+       char fullpath[PATH_MAX];
        struct strv new_args = STRV_INIT;
 
        if (opts->no_randomize_addr) {
@@ -2018,6 +2019,9 @@ int do_child_exec(int ready, struct opts *opts,
                        pr_dbg("disabling ASLR failed\n");
        }
 
+       if (realpath(opts->dirname, fullpath) != NULL)
+               opts->dirname = fullpath;
+
        shebang = check_script_file(argv[0]);
        if (shebang) {
                char *s, *p;
```

cmds/record.c
```diff
@@ -2023,6 +2024,13 @@ int do_child_exec(int ready, struct opts *opts,
 		char *s, *p;
 		int i;
 
+		/* current working directory can be changed during shell-script execution
+		 * so convert uftrace data directory path that can be relative
+		 * to an absolute path.
+		 */
```
honggyukim: What about changing the comment as follows?
```c
		/*
                 * The current working directory can be changed by calling chdir.
		 * So dirname has to be converted to an absolute path to avoid unexpected problems.
		 */
```

Taeung:
> The following change covers both cases.
> 
> ```diff
> diff --git a/cmds/record.c b/cmds/record.c
> index c909c255..b7891341 100644
> --- a/cmds/record.c
> +++ b/cmds/record.c
> @@ -2010,6 +2010,7 @@ int do_child_exec(int ready, struct opts *opts,
>  {
>         uint64_t dummy;
>         char *shebang;
> +       char fullpath[PATH_MAX];
>         struct strv new_args = STRV_INIT;
>  
>         if (opts->no_randomize_addr) {
> @@ -2018,6 +2019,9 @@ int do_child_exec(int ready, struct opts *opts,
>                         pr_dbg("disabling ASLR failed\n");
>         }
>  
> +       if (realpath(opts->dirname, fullpath) != NULL)
> +               opts->dirname = fullpath;
> +
>         shebang = check_script_file(argv[0]);
>         if (shebang) {
>                 char *s, *p;
> ```

@namhyung what's your opinion about his change ?
I also agonized about converting it to absolute path in the only script case or in all case..

cmds/record.c
Taeung: Looks better !

honggyukim:
> I also agonized about converting it to absolute path in the only script case or in all case..

As @namhyung mentioned in https://github.com/namhyung/uftrace/issues/745#issuecomment-490390358, it's not only related to shell script issue.

namhyung:
Yes, let's cover all cases.

Taeung:
I got it ! thanks @honggyukim @namhyung 
will resend this PR !

Issue #764, opened 2019-05-14
uftrace compilation warnings when building using clang


honggyukim:
```
$ CC=clang ./configure --prefix=$HOME/usr
uftrace detected system features:
...         prefix: /home/honggyu/usr
...         libelf: [ on  ] - more flexible ELF data handling
...          libdw: [ on  ] - DWARF debug info support
...   libpython2.7: [ on  ] - python scripting support
...    libncursesw: [ on  ] - TUI support
...   cxa_demangle: [ on  ] - full demangler support with libstdc++
...     perf_event: [ on  ] - perf (PMU) event support
...       schedule: [ on  ] - scheduler event support
...       capstone: [ on  ] - full dynamic tracing support

$ cat .config | grep CC
override CC     := clang
```
It shows following warning messages.
```
$ make
    ...
CC FPIC  libmcount/mcount-single.op
/home/honggyu/work/uftrace/libmcount/mcount.c:294:30: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int]
        if (!strncmp(spec, "SIGRTM" + off, 6 - off)) {
                           ~~~~~~~~~^~~~~
/usr/include/x86_64-linux-gnu/bits/string2.h:900:35: note: expanded from macro 'strncmp'
                      || (__builtin_constant_p (s2)                           \
                                                ^~
/home/honggyu/work/uftrace/libmcount/mcount.c:294:30: note: use array indexing to silence this warning
        if (!strncmp(spec, "SIGRTM" + off, 6 - off)) {
                                    ^
                           &        [
/usr/include/x86_64-linux-gnu/bits/string2.h:900:35: note: expanded from macro 'strncmp'
                      || (__builtin_constant_p (s2)                           \
                                                ^
    ...
```

honggyukim:
The following change removes the warning messages.
```diff
diff --git a/libmcount/mcount.c b/libmcount/mcount.c
index 2fd70d1..a5eed37 100644
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -259,6 +259,9 @@ static int parse_sigspec(char *spec, struct uftrace_filter_setting *setting)
                .sa_handler  = mcount_signal_trigger,
                .sa_flags    = SA_RESTART,
        };
+       const char *sigrtm = "SIGRTM";
+       const char *sigrtmin = "SIGRTMIN";
+       const char *sigrtmax = "SIGRTMAX";

        pos = strchr(spec, '@');
[...]

PR #766, opened 2019-05-16
build: Fix a few warnings in clang build

honggyukim:
clang compiler doesn't allow the following syntax:
```
  warning: adding 'int' to a string does not append to the string
  strncmp(spec, "SIGRTM" + off, 6 - off)
                ~~~~~~~~~^~~~~
```
There are a few more warnings as follows:
```
  warning: passing 'unsigned char *' to parameter of type 'char *'
           converts between pointers to integer types with different sign
          __builtin___clear_cache(origin_code_addr,
                                  ^~~~~~~~~~~~~~~~
  warning: passing 'unsigned char *' to parameter of type 'char *'
           converts between pointers to integer types with different sign
                                  origin_code_addr + origin_code_size);
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This patch is to avoid such warnings.

Fixed: #764

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -369,7 +369,7 @@ static int patch_xray_func(struct mcount_dynamic_info *mdi, struct sym *sym)
 static void patch_code(struct mcount_dynamic_info *mdi,
 		       uintptr_t addr, uint32_t origin_code_size)
 {
-	unsigned char *origin_code_addr;
+	void *origin_code_addr;
 	unsigned char call_insn[] = { 0xe8, 0x00, 0x00, 0x00, 0x00 };
 	uint32_t target_addr = get_target_addr(mdi, addr);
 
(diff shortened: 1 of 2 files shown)


arch/x86_64/mcount-dynamic.c
```diff
@@ -369,8 +369,8 @@ static int patch_xray_func(struct mcount_dynamic_info *mdi, struct sym *sym)
 static void patch_code(struct mcount_dynamic_info *mdi,
 		       uintptr_t addr, uint32_t origin_code_size)
 {
-	unsigned char *origin_code_addr;
-	unsigned char call_insn[] = { 0xe8, 0x00, 0x00, 0x00, 0x00 };
+	char *origin_code_addr;
+	char call_insn[] = { 0xe8, 0x00, 0x00, 0x00, 0x00 };
```
namhyung: I don't know why clang declared arguments in `__builtin_clear_cache()` as `char *`.  To me, `char *` looks like a string.  So I use unsigned one when dealing with a binary data.

As the problem comes from the `origin_code_addr` what about making it `void *` instead and leave `call_insn` as is?

honggyukim: I just changed it from `char*` to `void*`. Thanks.

PR #770, opened 2019-05-18
misc: Add clock_gettime family to misc/prototypes.h

honggyukim:
This patch adds the prototypes of the following functions.
```
  int clock_getres(clockid_t clk_id, struct timespec *res);
  int clock_gettime(clockid_t clk_id, struct timespec *tp);
  int clock_settime(clockid_t clk_id, const struct timespec *tp);
```
Let's say there is an example related to clock in C++ chrono as follows:
```
  $ cat chrono_clock.cc
  #include <chrono>
  int main()
  {
    std::chrono::system_clock::now();
    std::chrono::steady_clock::now();
    std::chrono::high_resolution_clock::now();
  }
```
It's very easy to see how each C++ chrono class internally calls
clock_gettime() as follows:
```
  $ g++ -std=c++11 -pg chrono_clock.cc
  $ uftrace -la -f none --no-comment a.out
  main() {
    std::chrono::_V2::system_clock::now() {
      clock_gettime(CLOCK_REALTIME, 0x7ffd800ee4c0) = 0;
    }
    std::chrono::_V2::steady_clock::now() {
      clock_gettime(CLOCK_MONOTONIC, 0x7ffd800ee4c0) = 0;
    }
    std::chrono::_V2::system_clock::now() {
      clock_gettime(CLOCK_REALTIME, 0x7ffd800ee4c0) = 0;
    }
  }
```

### DIFF
--- a/misc/prototypes.h
+++ b/misc/prototypes.h
@@ -522,4 +522,22 @@ int setresgid(uid_t rgid, uid_t egid, uid_t sgid);
 int chown(const char *path, uid_t uid, uid_t gid);
 int lchown(const char *path, uid_t uid, uid_t gid);
 int fchown(int fd, uid_t uid, uid_t gid);
+
+#include <time.h>
+enum uft_clockid_t {
+	CLOCK_REALTIME = 0,
+	CLOCK_MONOTONIC,
+	CLOCK_PROCESS_CPUTIME_ID,
+	CLOCK_THREAD_CPUTIME_ID,
+	CLOCK_MONOTONIC_RAW,
+	CLOCK_REALTIME_COARSE,
+	CLOCK_MONOTONIC_COARSE,
+	CLOCK_BOOTTIME,
+	CLOCK_REALTIME_ALARM,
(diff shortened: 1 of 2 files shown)


misc/prototypes.h
```diff
@@ -522,4 +522,22 @@ int setresgid(uid_t rgid, uid_t egid, uid_t sgid);
...
+	CLOCK_MONOTONIC,
+	CLOCK_PROCESS_CPUTIME_ID,
+	CLOCK_THREAD_CPUTIME_ID,
+	CLOCK_MONOTONIC_RAW,
+	CLOCK_REALTIME_COARSE,
+	CLOCK_MONOTONIC_COARSE,
+	CLOCK_BOOTTIME,
+	CLOCK_REALTIME_ALARM,
+	CLOCK_BOOTTIME_ALARM,
+	CLOCK_TAI,
```
namhyung: Looks like there's a hole before TAI.

honggyukim: Right. I found the hole and fixed it. Thanks!

PR #782, opened 2019-06-05
arm: aarch64: Add description for arch registers

honggyukim:
It'd be a lot easier to understand the assembly code by having
description table for architecture registers.

This patch adds the register table to mcount.S for ARM[1] and AArch64[2].

References:
[1] https://docs.microsoft.com/ko-kr/cpp/build/overview-of-arm-abi-conventions?view=vs-2019
[2] https://docs.microsoft.com/ko-kr/cpp/build/arm64-windows-abi-conventions?view=vs-2019

### DIFF
--- a/arch/aarch64/mcount.S
+++ b/arch/aarch64/mcount.S
@@ -1,5 +1,32 @@
 #include "utils/asm.h"
 
+/*
+Integer registers
+  The AArch64 architecture supports 32 integer registers:
+
+    Regster  |   Volatile?  | Role
+  -----------+--------------+---------------------------------------------------
+    x0       |     Volatile | Parameter/scratch register 1, result register
+    x1 - x7  |     Volatile | Parameter/scratch register 2-8
+    x8 - x15 |     Volatile | Scratch registers
+   x16 - x17 |     Volatile | Intra-procedure-call scratch registers
+   x18       | Non-Volatile | Platform register: in kernel mode, points to KPCR
+             |              | for the current processor; in user mode, points to TEB
+   x19 - x28 | Non-Volatile | Scratch registers
+   x29 / fp  | Non-Volatile | Frame pointer
+   x30 / lr  | Non-Volatile | Link register
+
+Floating-point/SIMD registers
+  The AArch64 architecture also supports 32 floating-point/SIMD registers:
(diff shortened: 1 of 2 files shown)


honggyukim:
It seems that we don't support aarch64 hard float as of now so didn't handle it in `arch/aarch64/mcount.S` yet.

namhyung:
In libmcount we avoid accessing floating-point arithmetic so those registers will be intact.  And it's little bit different Linux ABI and Windows ABI so I don't think it's a good idea to copy from MS.

honggyukim:
I have to compare once again with Linux ABI, but it seems that most of the info looks same so I will modify only the different parts later on.

namhyung:
Also I'm not sure about the copyright problems.  Isn't it better copy from ARM documents?

honggyukim:
I can update it based on ARM documents but need to find the info from the doc.

yskelg:
I found the reference in the Arm documentation.

*32bit*
- https://github.com/ARM-software/abi-aa/blob/2023Q3/aapcs32/aapcs32.rst#611core-registers

*64bit*
- https://github.com/ARM-software/abi-aa/blob/2023Q3/aapcs64/aapcs64.rst#general-purpose-registers

yskelg:
I think other architectures would also look good with these register references added.

yskelg:
RISC-V Register Convention tables.
- https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#integer-register-convention

Name | ABI Mnemonic | Meaning | Preserved across calls?
-- | -- | -- | --
x0 | zero | Zero | — (Immutable)
x1 | ra | Return address | No
x2 | sp | Stack pointer | Yes
x3 | gp | Global pointer | — (Unallocatable)
x4 | tp | Thread pointer | — (Unallocatable)
x5 - x7 | t0 - t2 | Temporary registers | No
x8 - x9 | s0 - s1 | Callee-saved registers | Yes
x10 - x17 | a0 - a7 | Argument registers | No
x18 - x27 | s2 - s11 | Callee-saved registers | Yes
x28 - x31 | t3 - t6 | Temporary registers | No

PR #787, opened 2019-06-08
build: Fix a build error due to missing errno.h

honggyukim:
This patch is to fix the following build error for ARM arch.

    libmcount/plthook.c:899:20: error: 'errno' undeclared

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -7,6 +7,7 @@
 #include <assert.h>
 #include <dlfcn.h>
 #include <fnmatch.h>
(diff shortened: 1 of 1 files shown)


honggyukim:
The same problem is also shown in `i386` build.
```
$ ARCH=i386 CFLAGS=-m32 LDFLAGS=-m32 ./configure
$ make
    ...
/home/honggyu/work/uftrace/libmcount/plthook.c:899:20: error: ‘errno’ undeclared
    ...
```

Issue #658, opened 2019-01-21
schedule event doesn't work with time-filter


honggyukim:
Please see below. The time-filter is set to 10ms, but there are events which are under the time threshold.
```
$ uftrace record -t 10ms ./clang hello.c

$ uftrace replay
# DURATION     TID     FUNCTION
            [178116] | _GLOBAL__sub_I_ARMFrameLowering.cpp() {
   5.945 ms [178116] |   /* linux:schedule */
            [178116] |   llvm::cl::Option::addArgument() {
 271.693 us [178116] |     /* linux:schedule */
            [178116] |     llvm::ManagedStaticBase::RegisterManagedStatic() {
   1.760 ms [178116] |       /* linux:schedule */
   9.018 ms [178116] |       /* linux:schedule */
 197.215 us [178116] |       /* linux:schedule */
  11.100 ms [178116] |     } /* llvm::ManagedStaticBase::RegisterManagedStatic */
 191.312 us [178116] |     /* linux:schedule */
  11.631 ms [178116] |   } /* llvm::cl::Option::addArgument */
  17.620 ms [178116] | } /* _GLOBAL__sub_I_ARMFrameLowering.cpp */
            [178116] | _GLOBAL__sub_I_ARMAsmParser.cpp() {
            [178116] |   __static_initialization_and_destruction_0() {
  19.550 ms [178116] |     /* linux:schedule */
   2.124 ms [178116] |     /* linux:schedule */
  21.767 ms [178116] |   } /* __static_initialization_and_destruction_0 */
  21.768 ms [178116] | } /* _GLOBAL__sub_I_ARMAsmParser.cpp */
            [178116] | _GLOBAL__sub_I_MachineScheduler.cpp() {
            [178116] |   __static_initialization_and_destruction_0() {
   2.450 ms [178116] |     /* linux:schedule */
   7.352 ms [178116] |     /* linux:schedule */
 344.297 us [178116] |     /* linux:schedule */
  10.342 ms [178116] |   } /* __static_initialization_and_destruction_0 */
  10.343 ms [178116] | } /* _GLOBAL__sub_I_MachineScheduler.cpp */
            [178116] | _GLOBAL__sub_I_DwarfDebug.cpp() {
            [178116] |   __static_initialization_and_destruction_0() {
  19.927 ms [178116] |     /* linux:schedule */
  20.027 ms [178116] |   } /* __static_initialization_and_destruction_0 */
[...]

Issue #739, opened 2019-05-02
schedule event needs to be fixed


honggyukim:
I think current `linux:schedule` event is not mature and it has the following problems as of now.
- It doesn't work well with record time time-filter
  - So when distributing the recorded data, always have to mention to use time-filter for analysis.
  - otherwise, the output screen is covered by too many noisy scheduling events as mentioned in  https://github.com/namhyung/uftrace/issues/699#issuecomment-474568996.
- `--no-event` simply hides this scheduling events, but it also disables task-name event as well.
  - The only way to work around this is to use `--no-event -E linux:task-name`, but it doesn't look nice.

In most cases, I'm only interested in task-name and not interested in scheduling event.
I would like to suggest the following choices.
1. `--no-sched` option to disable only schedule event.
2. disable schedule event by default and add `--sched` option to turn it on.
3. https://github.com/namhyung/uftrace/issues/658#issuecomment-458010226 can be a solution

honggyukim:
As we discussed today, I think we better provide a way to apply default options from a pre-defined opt file that is inside of `uftrace.data`.

We can generate the opt file inside `uftrace.data` to apply the time-filter used by `record` command. This can be the simplest fix as wrote in "3" above.

honggyukim:
Otherwise, It would be also good to always use `-E linux:task-name` even if `--no-event` is used.

honggyukim:
> Otherwise, It would be also good to always use -E linux:task-name even if --no-event is used.

@namhyung What about applying this as well? Is there any benefit of not recording `linux:task-name`?

namhyung:
Well, it doesn't honor the meaning of **NO** events and maybe sometimes users want to remove overhead of perf events at all.

PR #788, opened 2019-06-08
uftrace: Introduce uftrace.data/default.opts file

honggyukim:
This introduces a new file "default.opts" in uftrace.data directory.
The file "uftrace.data/default.opts" will contain the default options to
be applied during running analysis commands.

In some cases, the record time filters are not properly applied for
analysis commands especially for kernel and event records.

It can be fixed by applying the record time filter again for analysis
commands.

Fixed: #658 #739

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -47,6 +47,8 @@ const char *argp_program_bug_address = "https://github.com/namhyung/uftrace/issu
 
 static bool dbg_domain_set = false;
 
+static bool parsing_default_opts = false;
+
 enum options {
 	OPT_flat	= 301,
 	OPT_no_libcall,
@@ -460,6 +462,10 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
 		break;
 
 	case 't':
+		/* do not override time-filter if it's already set */
+		if (parsing_default_opts && opts->threshold)
+			break;
+
 		/* add time-filter to uftrace.data/default.opts */
 		strv_append(&default_opts, "-t");
 		strv_append(&default_opts, arg);
@@ -1003,6 +1009,36 @@ static void free_opts(struct opts *opts)
 }
 
 #ifndef UNIT_TEST
+static void apply_default_opts(int *argc, char ***argv, struct opts *opts)
+{
+	char *basename = "default.opts";
+	char opts_file[PATH_MAX];
+	struct stat stbuf;
+
+	/* default.opts is only for analysis commands */
+	if (opts->mode == UFTRACE_MODE_RECORD ||
+	    opts->mode == UFTRACE_MODE_LIVE ||
+	    opts->mode == UFTRACE_MODE_RECV)
+		return;
+
+	/* this is not to override user given time-filter by default opts */
+	parsing_default_opts = true;
+
+	snprintf(opts_file, PATH_MAX, "%s/%s", opts->dirname, basename);
+	if (!stat(opts_file, &stbuf) && stbuf.st_size > 0) {
+		pr_dbg("apply '%s' option file\n", opts_file);
+		parse_opt_file(argc, argv, opts_file, opts);
+	}
+	else if (!strcmp(opts->dirname, UFTRACE_DIR_NAME) &&
+		 !access("./info", F_OK)) {
+		/* try again applying default.opts in the current dir */
+		if (!stat(basename, &stbuf) && stbuf.st_size > 0) {
+			pr_dbg("apply './%s' option file\n", basename);
+			parse_opt_file(argc, argv, basename, opts);
+		}
+	}
+}
+
 int main(int argc, char *argv[])
 {
 	struct opts opts = {
@@ -1089,6 +1125,9 @@ int main(int argc, char *argv[])
 	/* the srcline info is used for TUI status line by default */
 	if (opts.mode == UFTRACE_MODE_TUI)
 		opts.srcline = true;
+
+	/* apply 'default.opts' options for analysis commands */
+	apply_default_opts(&argc, &argv, &opts);
 
 	if (opts.idx == 0)
 		opts.idx = argc;
--- a/uftrace.c
+++ b/uftrace.c
@@ -460,6 +460,10 @@ static error_t parse_option(int key, char *arg, struct argp_state *state)
 		break;
 
 	case 't':
+		/* add time-filter to uftrace.data/default.opts */
+		strv_append(&default_opts, "-t");
+		strv_append(&default_opts, arg);
+
 		opts->threshold = parse_time(arg, 3);
 		if (opts->range.start || opts->range.stop) {
 			pr_use("--time-range cannot be used with --time-filter\n");
--- a/uftrace.h
+++ b/uftrace.h
@@ -272,6 +272,8 @@ struct opts {
 	enum uftrace_pattern_type patt_type;
 };
 
+extern struct strv default_opts;
+
 static inline bool opts_has_filter(struct opts *opts)
 {
 	return opts->filter || opts->trigger || opts->threshold ||
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -184,6 +184,7 @@ failed:
 static bool is_uftrace_directory(const char *path)
 {
 	int fd;
+	bool ret = false;
 	char *info_path = NULL;
 	char sig[UFTRACE_MAGIC_LEN] = {0,};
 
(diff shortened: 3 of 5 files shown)


utils/utils.c
```diff
@@ -232,6 +235,29 @@ static bool can_remove_directory(const char *path)
 	return is_uftrace_directory(path) || is_empty_directory(path);
 }
 
+static bool create_uftrace_opts(const char *dirname)
+{
+	const char *opts_str = strv_join(&uftrace_opts, "");
```
namhyung: I think the delimiter should be a space.  Also it'd be better to bail out if `uftrace_opts` is already empty.

utils/utils.c
```diff
@@ -192,18 +192,23 @@ static bool is_uftrace_directory(const char *path)
...
+		close(fd);
+		return !memcmp(sig, UFTRACE_MAGIC_STR, UFTRACE_MAGIC_LEN);
 	}
 
-	close(fd);
-	return !memcmp(sig, UFTRACE_MAGIC_STR, UFTRACE_MAGIC_LEN);
+	/* if "info" file is missing, also check that there is "uftrace.opts" */
+	xasprintf(&info_path, "%s/uftrace.opts", path);
+	if (!access(info_path, F_OK))
+		return true;
```
namhyung: So what's the point of checking this?  I think it cannot work if info file is missing.

uftrace.c
```diff
@@ -1072,6 +1092,9 @@ int main(int argc, char *argv[])
 	if (opts.use_pager && opts.mode != UFTRACE_MODE_LIVE)
 		start_pager(pager);
 
+	/* apply 'uftrace.opts' options for analysis commands */
+	apply_uftrace_opts(&argc, &argv, &opts);
```
namhyung: Are you sure about the priority of option?  I mean command line option should override one in the `uftrace.opts`.  It'd be nice to have a test case to verify it works correctly.

uftrace.c
```diff
@@ -860,6 +861,8 @@ static void parse_opt_file(int *argc, char ***argv, char *filename, struct opts
 	int orig_idx = 0;
 
 	if (stat(filename, &stbuf) < 0) {
+		if (keep_going)
+			return false;
```
namhyung: I think we can do the check in the call-site and get rid of this change.

uftrace.h
```diff
@@ -270,6 +270,8 @@ struct opts {
 	enum uftrace_pattern_type patt_type;
 };
 
+extern struct strv uftrace_opts;
```
namhyung: I think `replay_opt_str` or `analysis_opt_str` is better.

utils/utils.c
honggyukim: Okay. I will change it.

utils/utils.c
honggyukim: I've seen many cases when running it on embedded devices. As of now, uftrace can be broken many times and it generates broken data, which doesn't have `info` file. If there is `uftrace.data` without `info` file. It will show some kind of unclear message as follows:
```
$ uftrace record t-abc
WARN: creating directory failed: File exists
```
Even if the user tries to record again, it just shows the same messages.
```
$ uftrace record t-abc
WARN: creating directory failed: File exists

$ uftrace record t-abc
WARN: creating directory failed: File exists

$ uftrace record t-abc
WARN: creating directory failed: File exists
```
It will make novice users got stuck and cannot do anything after that. So we better make another way to identify if it's uftrace directory or not. So I think `uftrace.opts` can be a good candidate because it's created right after the directory is created.

uftrace.c
honggyukim: You're right. I have to move this in a different place.

uftrace.c
honggyukim: Okay. I will.

uftrace.h
honggyukim: I've used `uftrace_opts` in many places including the related functions. If it's changed, maybe we better change all the other places. And I personally think the current name is better.

I can rather change the name `uftrace.opts` itself to something else if there's better one.

namhyung: `default.opts`?

honggyukim: It looks better. I will change all the namings to that. Thanks.

utils/utils.c
honggyukim: done.

uftrace.c
honggyukim: I added a test if it doesn't override user setting time-filter.

uftrace.h
honggyukim: I changed it to `default.opts`.

honggyukim:
I just rebased it to current master and changed the test name number from t224 to t226.

utils/utils.c
```diff
@@ -232,6 +235,30 @@ static bool can_remove_directory(const char *path)
...
+	bool ret = false;
+
+	snprintf(opts_filename, PATH_MAX, "%s/default.opts", dirname);
+	fp = fopen(opts_filename, "w");
+	if (fp == NULL) {
+		pr_dbg("Open failed: %s\n", opts_filename);
+		goto out;
+	}
+
+	fprintf(fp, "%s\n", opts_str ? opts_str : "");
```
namhyung: So you want to make this file even if no option is needed, right?  Then please call `fprintf()` only when the default_opts is set so that it can have 0 size if not used.

uftrace.c
```diff
@@ -1003,6 +1009,34 @@ static void free_opts(struct opts *opts)
...
+	char *basename = "default.opts";
+	char opts_file[PATH_MAX];
+	struct stat stbuf;
+
+	/* this is not to override user given time-filter by default opts */
+	parsing_default_opts = true;
+
+	if (opts->mode != UFTRACE_MODE_RECORD &&
+	    opts->mode != UFTRACE_MODE_LIVE &&
+	    opts->mode != UFTRACE_MODE_RECV) {
```
namhyung: You can just return if the mode is one of the case.

uftrace.c
```diff
@@ -1003,6 +1009,34 @@ static void free_opts(struct opts *opts)
...
+		snprintf(opts_file, PATH_MAX, "%s/%s", opts->dirname, basename);
+		if (!stat(opts_file, &stbuf)) {
```
namhyung: And if its size is not zero.

utils/utils.c
honggyukim: done.

uftrace.c
honggyukim: done.

uftrace.c
honggyukim: done.

honggyukim:
I just updated it following your comments. Thanks for the review.

uftrace.c
```diff
@@ -1003,6 +1009,36 @@ static void free_opts(struct opts *opts)
...
+	else if (!strcmp(opts->dirname, UFTRACE_DIR_NAME) &&
+		 !access("./info", F_OK)) {
+		/* try again applying default.opts in the current dir */
+		if (!stat(basename, &stbuf) && stbuf.st_size > 0) {
+			pr_dbg("apply './%s' option file\n", basename);
+			parse_opt_file(argc, argv, basename, opts);
+		}
+	}
+}
+
```
namhyung: I think it's more useful if we can support any option in the `default.opts` file.  But let's think about it later..

utils/utils.c
```diff
@@ -192,18 +192,23 @@ static bool is_uftrace_directory(const char *path)
...
 	}
 
-	close(fd);
-	return !memcmp(sig, UFTRACE_MAGIC_STR, UFTRACE_MAGIC_LEN);
+	/* if "info" file is missing, also check that there is "default.opts" */
+	xasprintf(&info_path, "%s/default.opts", path);
+	if (!access(info_path, F_OK))
+		return true;
+
+	return false;
```
namhyung: You need to free `info_path` before the return.

honggyukim: I missed it. Thanks.

uftrace.c
honggyukim: The `default.opts` can manually be modified to add more options later. I plan to use it when distributing some pre-recorded data to apply other default filters.

honggyukim:
Thanks for the review. I just updated it.

uftrace.c
namhyung: But I'm afraid of the priority.  It's not acceptable to apply all the options like you did for the `-t` option.

namhyung:
Hmm.. it seems that github only closes the first issue and #739 is not closed automatically.  I'll close it manually.

uftrace.c
honggyukim: I also think that it might not be a perfect solution, but I think at least normal users will not modify `default.opts` manually. I will think about the better solution later.

honggyukim:
It may be because I didn't add `Fixed` line in the commit message, but only in the PR message.

namhyung:
But it closed the first one (#658).

PR #790, opened 2019-06-09
utils: Add pr_dbg4 and use it for trivial messages

honggyukim:
It'd be better to distinguish between mcount and plthook related debug
messages because they are directly related to program execution.

It's better not to print too many symbol, dwarf, and demangle messages
when we want to see where the program is segfaulted.

### DIFF
--- a/utils/demangle.c
+++ b/utils/demangle.c
@@ -175,9 +175,9 @@ static void dd_debug_print(struct demangle_data *dd)
 		dd_eof(dd) ? " (EOF)" : "", dd->level ? " (not finished)" : "",
 		dd->old, dd->pos + 1, '^', dd->func, dd->line, expected);
 
-	pr_dbg2("current: %s (pos: %d/%d)\n", dd->new, dd->pos, dd->len);
+	pr_dbg3("current: %s (pos: %d/%d)\n", dd->new, dd->pos, dd->len);
 	for (i = 0; i < dd->nr_dbg; i++)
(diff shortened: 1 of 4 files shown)


namhyung:
You can use `--debug-domain` with `mcount:1` or `plthook:2` for that.

honggyukim:
> You can use --debug-domain with mcount:1 or plthook:2 for that.

Sorry for later answer. It absolutely fine to provide a long list of `--debug-domain`.

However, I've been struggling to find segfault on various environments but most of the problems are plthook and mcount parts. I hardly see that symbol, dwarf and demangle are the issues, but those messages are too verbose in most cases.

Especially for external users, It's a lot simpler to make a request to run uftrace with `-vvv` rather than `--debug-domain plthook:3 --debug-domain mcount:3 --debug-domain uftrace:3`.

So I would like to make more debug domain levels. I personally want to make it in 5 levels.

namhyung:
Yep, I also think current level 2 or 3 is too verbose sometimes.  I'm ok to having more fine-grained debug domain and level

honggyukim:
Thanks. I think the current change is enough for my purpose. If you want to add more levels, please fix it later. I've exchanged some debug reports with internal engineers so I would like to ask them to send me with `-vvv`.

honggyukim:
@namhyung Can it be merged without update?

honggyukim:
I just rebased it to master and added more symbol debug logs to `pr_dbg4`.

Issue #791, opened 2019-06-09
node recording gets crashed in AArch64 


honggyukim:
I tested tracing node in AArch64, but it gets segfault as follows:
```
$ file node
node: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, BuildID[sha1]=10b4c6f29ba4f89875e8ccd239c825fadcc0a273, with debug_info, not stripped

$ uftrace record --no-libcall ./node -e ''
child terminated by signal: 11: Segmentation fault
```
The node was cross-compiled as described in [uftrace for node](https://github.com/namhyung/uftrace/wiki/uftrace-for-node).

honggyukim:
```
$ uftrace replay
# DURATION     TID     FUNCTION
            [  2093] | _GLOBAL__sub_I__register_cares_wrap() {
   2.135 us [  2093] |   uv_mutex_init();
  48.073 us [  2093] | } /* _GLOBAL__sub_I__register_cares_wrap */
            [  2093] | _GLOBAL__sub_I_node::reverted() {
            [  2093] |   node::MutexBase::MutexBase() {
  20.416 us [  2093] |     uv_mutex_init();
  22.708 us [  2093] |   } /* node::MutexBase::MutexBase */
  66.979 us [  2093] |   node::PerProcessOptions::PerProcessOptions();
            [  2093] |   node::MutexBase::MutexBase() {
   0.677 us [  2093] |     uv_mutex_init();
   2.292 us [  2093] |   } /* node::MutexBase::MutexBase */
 126.302 us [  2093] | } /* _GLOBAL__sub_I_node::reverted */
            [  2093] | _GLOBAL__sub_I_node::http2::Http2Session::callback_struct_saved() {
            [  2093] |   node::http2::Http2Session::Callbacks::Callbacks() {
   5.677 us [  2093] |     nghttp2_session_callbacks_new();
   0.468 us [  2093] |     nghttp2_session_callbacks_set_on_begin_headers_callback();
   0.365 us [  2093] |     nghttp2_session_callbacks_set_on_header_callback2();
   0.521 us [  2093] |     nghttp2_session_callbacks_set_on_frame_recv_callback();
...
   0.469 us [  2093] |                                     v8::internal::WeakFixedArray::Set();
   6.979 us [  2093] |                                   } /* v8::internal::DescriptorArray::CopyFrom */
[...]

PR #792, opened 2019-06-09
aarch64: Handle struct return type by keeping more regs on stack

honggyukim:
Current implementation for mcount_return and plthook_return only reserve
x0 register assuming the size of return value is enough with 8 bytes.

But some functions may have a return type in struct and its size is more
than 8 bytes as following example:
```c
  struct { int64_t quot, int64_t rem}
         __aeabi_ldivmod(int64_t numerator, int64_t denominator) {
    int64_t rem, quot;
    quot = __divmoddi4(numerator, denominator, &rem);
    return {quot, rem};
  }
```
In this case, we might partially lose return values because the return
value is stored in x0 and x1 registers to keep 16 bytes in the return
type of struct { int64_t quot, int64_t rem}.

So this patch fixes to keep both x0 and x1 registers to prevent such
problems.

It's also required to save d0 and d1 VFP registers in return hook.

This patch is inspired by the following commit for arm.

  ffb69ce arm: Handle struct return type by keeping more regs on stack

It also saves `x8` register for library calls same as user functions.

Fixed: #791

### DIFF
--- a/arch/aarch64/plthook.S
+++ b/arch/aarch64/plthook.S
@@ -81,9 +81,19 @@ ENTRY(plthook_return)
 	stp	x0, x1, [sp, #-16]!
 	stp	d0, d1, [sp, #-16]!
 
-	add	x0, sp, #16
+	/*
+	 * save indirect result location register
+	 * used in C++ for returning non-trivial objects
+	 */
+	str	x8, [sp, #-16]!
+
+	add	x0, sp, #32
+
 	bl	plthook_exit
 	mov	x16, x0
+
+	/* restore indirect result location register */
+	ldr	x8, [sp], #16
 
 	/* restore return values */
 	ldp	d0, d1, [sp], #16
--- a/arch/aarch64/plthook.S
+++ b/arch/aarch64/plthook.S
@@ -76,16 +76,19 @@ END(plt_hooker)
 ENTRY(plthook_return)
 	/* setup frame pointer */
 	stp	x29, x30, [sp, #-16]!
-	stp	x0, x8, [sp, #-16]!
-	str	d0, [sp, #-16]!
+
+	/* save return values */
+	stp	x0, x1, [sp, #-16]!
+	stp	d0, d1, [sp, #-16]!
 
 	add	x0, sp, #16
 	bl	plthook_exit
 	mov	x16, x0
 
 	/* restore return values */
-	ldr	d0, [sp], #16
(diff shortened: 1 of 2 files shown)


arch/aarch64/plthook.S
```diff
@@ -76,16 +76,29 @@ END(plt_hooker)
...
+	stp	x0, x1, [sp, #-16]!
+	stp	d0, d1, [sp, #-16]!
+
+	/*
+	 * save indirect result location register
+	 * used in C++ for returning non-trivial objects
+	 */
+	str	x8, [sp, #-16]!
+
+	add	x0, sp, #32
```
honggyukim: @namhyung Actually, I don't clearly understand this calculation. Could you please confirm this?

namhyung: `x0` is the first argument of `plthook_exit` (`long *retval`) so it should point the location where original `x0` is saved which is `sp + 16` in this case.  But I'm not sure it should save `x8` register as well as it's the "indirect result location register".

namhyung: Oh, I missed this commit.  It look ok now.

honggyukim:
> It'd be better making it 16 in the first commit and change it to 32 in the second commit.

Thanks. I just updated it.

arch/aarch64/plthook.S
honggyukim: > But I'm not sure it should save x8 register as well as it's the "indirect result location register".

I'm not 100% sure as well, but I think the same problem can be happened because when library calls are done indirectly, but their return is made directly. So the library function will immediately return its return value back to the caller.

honggyukim: The node test works fine even without saving `x8` in `plthook.S` but just added to prevent any kind of future problem.

honggyukim: If you're not sure about this, I can drop this as of now because it's more important to fix the current problem first.

namhyung: I think we should take the safer way and the original commit already did it.

honggyukim: So do you mean to keep the second commit for `plthook.S` as is?

namhyung: Yep

honggyukim: Thanks. Have a good night!

honggyukim:
I just rebased it to master again without changing anything.

PR #793, opened 2019-06-10
info: Disable pager for info command

honggyukim:
uftrace info output doesn't have to use pager so disable it.
It makes the output urgly especially in busybox environment.

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -1049,7 +1049,8 @@ int main(int argc, char *argv[])
 
 	if (opts.mode == UFTRACE_MODE_RECORD ||
 	    opts.mode == UFTRACE_MODE_RECV ||
-	    opts.mode == UFTRACE_MODE_TUI)
+	    opts.mode == UFTRACE_MODE_TUI ||
+	    opts.mode == UFTRACE_MODE_INFO)
 		opts.use_pager = false;
 	if (opts.nop)
 		opts.use_pager = false;


honggyukim:
We better not use pager if possible for embedded environment.

namhyung:
You can set `PAGER` env to "cat".

honggyukim:
I think general users have a problem using it if we make another step to use.

honggyukim:
Hmm.. what's the benefit of using pager for `info` command?

namhyung:
What's problem of using pager?

honggyukim:
It makes the output ugly on serial output on embedded environment and users have to explicitly type "q" to quit.

honggyukim:
I know that there is a way to avoid the problem, but I wonder if there is a benefit of using pager for `info` command?

namhyung:
> It makes the output ugly on serial output on embedded environment

Could you please be more specific?

honggyukim:
I don't have such environment right now, but I felt many times when I tried. Could you explain about the benefit of using pager?

honggyukim:
Never mind. It's not an important issue so let's focus on v0.9.3 release instead for now.

namhyung:
The benefit is same as other analysis commands.  It could have some options to list long outputs (like symbols or tasks).

honggyukim:
Sorry for the late answer. I didn't think that it's needed for horizontally long output, but it's okay. It may also be needed for `info --symbols` option. Thanks for the explanation.

PR #808, opened 2019-06-15
dynamic: Allow NOP as a patchable instruction

honggyukim:
Current implementation doesn't allow NOP as a patchable target
instruction.  This patch makes it as a patchable instruction.


namhyung:
I have similar patch but it's unlikely to have the NOP instruction in a function.  I'm thinking about allowing instructions don't have operands instead.

honggyukim:
Okay. I see. Please close this when you add the commit.

namhyung:
Merged: 99c7622

PR #809, opened 2019-06-15
slide: Add full dynamic tracing section

honggyukim:

### DIFF
--- a/doc/uftrace.html
+++ b/doc/uftrace.html
@@ -96,6 +96,17 @@
         width: 49%;
       }
 
+      /* Two-column layout */
+      .left30-column {
+        float: left;
+        width: 29%;
+      }
+      .right70-column {
+        float: right;
+        width: 69%;
+      }
+
+
       /* Code Font Size */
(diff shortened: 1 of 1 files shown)


honggyukim:
Please take a look at https://uftrace.github.io/slide/#full-dynamic-tracing.

namhyung:
Actually it's not adding calls to `mcount` it replaces.  Hmm.. but it'd be more difficult to display and explain in the slide...

honggyukim:
Do you think it'd better to change `mcount` to `__dentry__`?

namhyung:
No, I think it'd be easier to understand to use `mcount`.

honggyukim:
It actually goes to `__dentry__` but it calls `mcount_entry` inside so it seems reasonable.

namhyung:
I'm ok with that part.  What I concerned was it overwrites the existing instructions instead of inserting to some (empty?) space.

namhyung:
But now I'm ok with the current state - don't need to update..

PR #828, opened 2019-07-09
utils: Print pr_color to logfp as well

honggyukim:
In some cases, it's difficult to analyze a bug report based on logfile
output because it doesn't contain some critical messages, which are
printed by pr_color.  In most cases, such colored messages are crash
points, but it's difficult to match where the program gets segfault
among the long list of debug messages.

This patch makes pr_color print message to logfp as well as outfp only
if logfp is not stderr so that it doesn't print duplicate message.

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1878,12 +1878,12 @@ static int stop_tracing(struct writer_data *wd, struct opts *opts)
 				ret = UFTRACE_EXIT_SUCCESS;
 		}
 		else if (WIFSIGNALED(status)) {
-			pr_yellow("child terminated by signal: %d: %s\n",
+			pr_warn("child terminated by signal: %d: %s\n",
 				  WTERMSIG(status), strsignal(WTERMSIG(status)));
 			ret = UFTRACE_EXIT_SIGNALED;
 		}
 		else {
-			pr_yellow("child terminated with unknown reason: %d\n",
+			pr_warn("child terminated with unknown reason: %d\n",
 				  status);
 			memset(&wd->usage, 0, sizeof(wd->usage));
 			ret = UFTRACE_EXIT_UNKNOWN;
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -709,18 +709,18 @@ static void segv_handler(int sig, siginfo_t *si, void *ctx)
 				break;
 
 			if (si->si_code == sigsegv_codes[i].code) {
-				pr_red("Segmentation fault: %s (addr: %p)\n",
+				pr_warn("Segmentation fault: %s (addr: %p)\n",
 				       sigsegv_codes[i].msg, si->si_addr);
 				break;
 			}
 		}
 		if (sig != SIGSEGV || i == (int)ARRAY_SIZE(sigsegv_codes)) {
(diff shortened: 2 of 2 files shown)


honggyukim:
I added this PR because it's difficult to analyze the segfault problem based on the current logfile. It's important to fine where the segfault was happened but there's no such message because it's only printed to `outfp`.
**Before:**
```
$ uftrace record -v --logfile=log.before a.out
retval = 0x111
handler() called!
retval = (nil)
Segmentation fault: address not mapped (addr: (nil))
Backtrace from uftrace:
=====================================
[0] (main[400928] <= __libc_start_main[7ffbf6dbb830])

$ cat log.before
uftrace: checking binary a.out
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
uftrace: creating 10 thread(s) for recording
plthook: setup PLT hooking "/home/honggyu/work/uftrace/git/uftrace/a.out"
mcount: mcount setup done
mcount: new session started: 948990508a3639bc: a.out
mcount: shmem_finish: tid: 39341 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-948990508a3639bc-39330-000
```
The logfile of after version contains where the segfault was happend.
**After:**
```
$ uftrace record -v --logfile=log.after a.out
retval = 0x111
handler() called!
retval = (nil)
Segmentation fault: address not mapped (addr: (nil))
Backtrace from uftrace:
=====================================
[0] (main[400928] <= __libc_start_main[7f2554b75830])

$ cat log.after
uftrace: checking binary a.out
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
uftrace: creating 10 thread(s) for recording
plthook: setup PLT hooking "/home/honggyu/work/uftrace/git/uftrace/a.out"
mcount: mcount setup done
mcount: new session started: ec8b474edf63ad23: a.out
mcount: shmem_finish: tid: 39056 seqnum = 0 curr = 0, nr_buf = 2 max_buf = 2
Segmentation fault: address not mapped (addr: (nil))
Backtrace from uftrace:
=====================================
[0] (main[400928] <= __libc_start_main[7f2554b75830])
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-ec8b474edf63ad23-39045-000
```

utils/debug.c
```diff
@@ -230,6 +230,10 @@ void __pr_color(char code, const char *fmt, ...)
 
 	va_start(ap, fmt);
 	vfprintf(outfp, fmt, ap);
+	if (logfp != stderr) {
+		va_start(ap, fmt);
+		vfprintf(logfp, fmt, ap);
+	}
```
namhyung: But this will add other (normal) messages to the logfile.  I think you'd better changing the segfault code to use `pr_warn()` instead.

honggyukim:
@namhyung You're right. I just changed the messages using `pr_warn` as you suggested. I also made one more change. I think it's a lot better to show backtrace info if the target program gets crashed regardless of `-v` options.

honggyukim:
@namhyung Please take a look this if possible. Thanks.

namhyung:
Could you please rebase?

honggyukim:
I just rebased it. Thanks!

PR #833, opened 2019-07-13
misc: Add capstone package separately

honggyukim:
In some old distros such as Ubuntu 14.04, capstone is not provided.
In this case, the entire package installation gets failed.
```
  $ sudo ./misc/install-deps.sh
  Reading package lists... Done
  Building dependency tree
  Reading state information... Done
  E: Unable to locate package libcapstone-dev
```
To prevent such problems, it'd be better to add capstone installtion
separetely so that it doesn't affect other pacakge installation.

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -13,12 +13,15 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 
 case $distro in
     "ubuntu" | "debian")
-        apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev libcapstone-dev pkg-config ;;
+        apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config
+        apt-get $OPT install libcapstone-dev ;;
     "fedora")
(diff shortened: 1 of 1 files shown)


namhyung:
It seems `apt` has `-m`/`--ignore-missing` but I couldn't find something similar for other package managers.

namhyung:
But is Ubuntu 14.04 still used?  I guess it's already in EOL (End-of-Life).

honggyukim:
> It seems apt has -m/--ignore-missing but I couldn't find something similar for other package managers.

If so, we can use the option.

> But is Ubuntu 14.04 still used? I guess it's already in EOL (End-of-Life).

Yes, I was together with someone who use Ubuntu 14.04. It's not mandatory but better to consider some old environments if possible.

honggyukim:
> It seems apt has -m/--ignore-missing but I couldn't find something similar for other package managers.

I tested again with `-m` option, but it doesn't work as expected.
```
$ sudo apt-get install -m libdw-dev libcapstone-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package libcapstone-dev
```
It also fails to install `libdw-dev` as well so I think we better keep the current change.

namhyung:
Hmm.. ok then

honggyukim:
Is there any other issue on this?

namhyung:
Nop, will merge.

Issue #807, opened 2019-06-13
segfault in ARM binary


honggyukim:
There is a segfault when tracing a specific ARM program which is compiled with `-pg` flag.
```
# uftrace record --no-libcall -D1 ./prog.pg
child terminated by signal: 11: Segmentation fault
```
Here is verbose messages.
```
# uftrace record --debug-domain mcount:3 --debug-domain uftrace:3 --no-libcall -D1 ./prog.pg
uftrace: checking binary ./prog.pg
uftrace: using /usr/lib/libmcount.so library for tracing
uftrace: creating 1 thread(s) for recording
uftrace: start writer thread 0
mcount: initializing mcount library
mcount: mcount setup done
uftrace: MSG SESSION: 14955: /usr/bin/prog.pg (9249968c421815e2)
mcount: new session started: 9249968c421815e2: prog.pg
mcount: preparing shmem buffers: tid = 14955
mcount: <0> enter 7e7ae3
mcount:  tr->flags: 0, filter mode, count: [0] 0/0
uftrace: copying instructions of _GLOBAL__sub_I_Item::Item
uftrace: MSG START: /uftrace-9249968c421815e2-14955-000
uftrace: MSG TASK_START : 14955/14955
uftrace: all process/thread exited
child terminated by signal: 11: Segmentation fault
uftrace: stop writer thread 0
uftrace: fill header (metadata) info in uftrace.data/info
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-9249968c421815e2-14955-000
uftrace: unlink for session: 9249968c421815e2
uftrace: unlink /uftrace-9249968c421815e2-14955-000
uftrace: unlink /uftrace-9249968c421815e2-14955-001
uftrace: reading symbols for session 9249968c421815e2
```
It gets crashed at the first function entry point. The most eye-catching lines are as follows:
```
mcount: <0> enter 7e7ae3
mcount:  tr->flags: 0, filter mode, count: [0] 0/0
uftrace: copying instructions of _GLOBAL__sub_I_Item::Item
```
Let's calculate the address.
```
# cat uftrace.data/sid-9249968c421815e2.map | grep prog.pg
762000-963000 r-xp 00000000 00:00 0                          /usr/bin/prog.pg

# python -c "print(hex(0x7e7ae3 - 0x762000))"
0x85ae3
```
Here is disassembly output.
```asm
00085ad8 <_GLOBAL__sub_I__ZN4ItemC2Ev>:
[...]

PR #838, opened 2019-07-29
arm: Fix segfault by adjusting map start address

honggyukim:
As of now, record command always gets segfault in arm environment.
It's a regression that was introduced by the following commit.

  d8eb70b4 symbol: Use module symbol table in find_symtabs()

This patch fixes the problem by adding map start address.  In addition
to this, it removes the code for cache lookup because it's not
thread-safe code.

Fixed: #807

### DIFF
--- a/arch/arm/mcount-support.c
+++ b/arch/arm/mcount-support.c
@@ -20,8 +20,6 @@ struct lr_offset {
 
 #define REG_SP  13
 
-static struct rb_root offset_cache = RB_ROOT;
-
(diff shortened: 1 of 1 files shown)


honggyukim:
@namhyung I know that `find_map` is previously done inside `find_symtabs`, but just added once again to minimize the code change. Please let me know if you have any better idea.

namhyung:
I'm ok with it.

honggyukim:
Thanks a lot! I can use it for TV now :)

PR #843, opened 2019-07-31
arch: Fix wrong comment of x86_64/mcount.S

JinDDeok:
If you look at the bottom line, it says 0xfffffffffffffff0.
The stack pointer in the comment is incorrectly written as 16 bytes.

This commit replaces 16 bytes with 16 bits.

Signed-off-by: JinDDeok wlstjr112@naver.com

### DIFF
--- a/arch/x86_64/mcount.S
+++ b/arch/x86_64/mcount.S
@@ -51,7 +51,7 @@ GLOBAL(mcount)
 	movq %rsp, %rdx
 	.cfi_def_cfa_register rdx
 
-	/* align stack pointer to 16-byte */
+	/* align stack pointer to 16-bit */
 	andq $0xfffffffffffffff0, %rsp
 	push %rdx
 


namhyung:
Thanks for contributing to uftrace.

You need to write a proper commit message because it's recorded in the log history.  We often look at the logs if something goes bad so having a good message is an important thing.  Also it requires a "Signed-off-by" line for your name - it indicates it's ok for used in the open source project.

Other comments will be added inline.

arch/x86_64/mcount.S
```diff
@@ -51,7 +51,7 @@ GLOBAL(mcount)
 	movq %rsp, %rdx
 	.cfi_def_cfa_register rdx
 
-	/* align stack pointer to 16-byte */
+	/* align stack pointer to 16-bit */
 	andq $0xfffffffffffffff0, %rsp
```
namhyung: This is actually to clear last 4 bits in the `rsp` register (stack pointer).  Clearing 4 bits makes the content a multiple of 16 (= 2^4).  So after the instruction the stack pointer will point somewhere in the stack memory aligned by 16 **byte**.

JinDDeok: Thank you for your comments!

namhyung:
@JinDDeok hi, I'm not accepting this patch since the comment is correct.  But still thank you for taking your time for the contribution.

PR #847, opened 2019-08-05
dynamic: Disallow patching functions with an indirect jump

AnsBal:
Since it's hard to predict the target of an indirect jump, we just disallow patching it for now.
Allowing it may cause a thread to jump to the prologue and corrupt the execution. 

Signed-off-by: Anas Balboul <anasbalbo@gmail.com>

### DIFF
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -140,8 +140,11 @@ static bool check_unsupported(struct mcount_disasm_engine *disasm, cs_insn *insn
 			}
 			break;
 		case X86_OP_MEM:
-			/* TODO */
+			/* indirect jumps are not allowed */
+			return false;
 			break;
+		case X86_OP_REG:
+			return false;
 		default:
 			break;
 		}


AnsBal:
Thank you for the comment !
The program didn't crash with/without the patch. but it may crash in some cases. I got the idea from kprobes. They don't instrument a function with an indirect jump for a safety reason :


> Kprobes analyzes the entire function and verifies that there is no
>   jump into the optimized region.  Specifically: - the function contains no indirect jump;
[Source: (Safety Check section)](https://github.com/torvalds/linux/blob/master/Documentation/kprobes.txt)

Here is a -v run without the patch:

	$ uftrace -v -P. -t 1ms -D3 uftrace replay --no-pager 
	...
	dynamic:    total:      752
	dynamic:  patched:      582 (77.39%)
	dynamic:   failed:      161 (21.40%)
	dynamic:  skipped:        9 ( 1.19%)
	dynamic: no match:        0
	...

And another run with the patch:

	$ uftrace -v -P. -t 1ms -D3 uftrace replay --no-pager
	...
	dynamic:    total:      752
	dynamic:  patched:      559 (74.33%)
	dynamic:   failed:      184 (24.46%)
	dynamic:  skipped:        9 ( 1.19%)
	dynamic: no match:        0
	...

namhyung:
Great, could you please update the commit message with the link to the kprobes doc and it affects ~3% of (uftrace) functions?

AnsBal:
Sure! Please let me know if the commit message is too long.

namhyung:
Thanks, the message itself is fine but it'd be nice if you wrap the lines around 70 column.

namhyung:
Sorry, one more trivia.  I usually see the commit log in the console.  So please write it like a footnote.
```
... safety check section [1] ...

[1] http://safety.check.url/here
```

AnsBal:
It makes the message clear now, Thanks!

Issue #853, opened 2019-09-01
segfault when given logfile name is already taken by a directory


honggyukim:
Here is an example recording a simple program.
```
$ gcc -pg tests/s-abc.c -o t-abc

$ uftrace record t-abc

$ uftrace replay
# DURATION     TID     FUNCTION
   1.700 us [ 16855] | __monstartup();
   1.300 us [ 16855] | __cxa_atexit();
            [ 16855] | main() {
            [ 16855] |   a() {
            [ 16855] |     b() {
            [ 16855] |       c() {
   1.400 us [ 16855] |         getpid();
   3.600 us [ 16855] |       } /* c */
   5.600 us [ 16855] |     } /* b */
   7.300 us [ 16855] |   } /* a */
   9.200 us [ 16855] | } /* main */
```
The above example works fine and the debug message can be stored with `--logfile` option as follows:
```
$ uftrace record -v --logfile=dir t-abc
```
However, it gets crashed if the given file name is already used by another directory as follows:
```
$ mkdir dir

$ uftrace record -v --logfile=dir t-abc
Segmentation fault (core dumped)
```
It needs to check if the file is already exists or not.

honggyukim:
The crash point is in `uftrace.c` file as follows:
```c
1042│ int main(int argc, char *argv[])
1043│ {
        ...
1080│         if (dbg_domain_set && !debug)
1081│                 debug = 1;
1082│
1083│         if (opts.logfile) {
1084│                 logfp = fopen(opts.logfile, "a");
1085│                 if (logfp == NULL)
1086├───────────────────────> pr_err("cannot open log file");
1087│
1088│                 setvbuf(logfp, NULL, _IOLBF, 1024);
1089│         }
1090│         else if (debug) {
1091│                 /* ensure normal output is not mixed by debug message */
1092│                 setvbuf(outfp, NULL, _IOLBF, 1024);
1093│         }
        ...
```

honggyukim:
This is a simple bug so anyone can fix this if interested.

rls1004:
Can i take this issue?

honggyukim:
Sure, please take it. I would like to provide more info regarding this.

Please refer to the following discussion.
- https://github.com/namhyung/uftrace/pull/707#issuecomment-475830493
- https://github.com/namhyung/uftrace/issues/831

I think we better have some protection code for this.

rls1004:
The problem seems to be caused by writing an error message to logfd when logfd is NULL.
[...]

PR #861, opened 2019-09-09
uftrace: Fix segfault when given logfile name is already taken by a directory

rls1004:
This patch is to fix segfault in --logfile.
If logfd is NULL, set logfd to stderr before calling pr_err().

Fixed: #853

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -1082,8 +1082,10 @@ int main(int argc, char *argv[])
 
 	if (opts.logfile) {
 		logfp = fopen(opts.logfile, "a");
-		if (logfp == NULL)
+		if (logfp == NULL) {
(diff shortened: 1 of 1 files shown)


honggyukim:
Thanks for the fix. Could you please add `Signed-off-by` statement at the end of the commit message? You can simply run `git amend -s` to add it. And please also make sure your real name is updated in `~/.gitignore`.

honggyukim:
The first line of commit message is better to be less than about 50 characters.

rls1004:
Did you mean `~/.gitconfig`, not `~/.gitignore`?

honggyukim:
Right. I mean `~/.gitconfig`.

Issue #858, opened 2019-09-07
segfault when function argument to display exceeds memory range


rls1004:
I found some problem when using -A option.
Here is the steps to reproduce it.

Code to be traced (simple.c) :
```C
#include <stdio.h>

int main() {
	puts("123");
	return 0;
}
```

Display the first argument :
```
$ uftrace -A puts@arg1/s ./simple
123
# DURATION     TID     FUNCTION
   0.957 us [ 13412] | __monstartup();
   0.640 us [ 13412] | __cxa_atexit();
            [ 13412] | main() {
            [ 13412] |   puts("123") {
 113.447 us [ 13412] |     /* linux:schedule */
  95.143 us [ 13412] |     /* linux:schedule */
 402.826 us [ 13412] |   } /* puts */
 404.100 us [ 13412] | } /* main */
```

Display the 99999999999th argument :
```
$ uftrace -A puts@arg99999999999/s ./simple
WARN: Segmentation fault: address not mapped (addr: 0x7fff5b85aa80)
WARN: Backtrace from uftrace:
WARN: =====================================
WARN: [1] (puts[4004f0] <= main[400699])
WARN: [0] (main[40068f] <= __libc_start_main[7efc46702830])
WARN: child terminated by signal: 11: Segmentation fault
# DURATION     TID     FUNCTION
   0.838 us [ 13404] | __monstartup();
   0.549 us [ 13404] | __cxa_atexit();
            [ 13404] | main() {
            [ 13404] |   puts() {
 112.727 us [ 13404] |     /* linux:schedule */
 101.142 us [ 13404] |     /* linux:schedule */
  50.856 us [ 13404] |     /* linux:schedule */
  76.728 us [ 13404] |     /* linux:schedule */
  72.285 us [ 13404] |     /* linux:schedule */
  73.302 us [ 13404] |     /* linux:schedule */
  48.374 us [ 13404] |     /* linux:schedule */
  79.110 us [ 13404] |     /* linux:schedule */
  48.258 us [ 13404] |     /* linux:schedule */
  76.708 us [ 13404] |     /* linux:schedule */
  70.086 us [ 13404] |     /* linux:schedule */
  86.564 us [ 13404] |     /* linux:schedule */
  72.046 us [ 13404] |     /* linux:schedule */
  72.247 us [ 13404] |     /* linux:schedule */
  71.403 us [ 13404] |     /* linux:schedule */
  49.131 us [ 13404] |     /* linux:schedule */
[...]

PR #865, opened 2019-09-10
mcount: fix segfault in mcount_get_stack_arg()

rls1004:
If the given argument index is out of memory or inaccessible, a segfault will occur.
I added code to check the validity of the memory address before accessing the address.

Before :
```
$ uftrace -A puts@arg123123123 ./simple
WARN: Segmentation fault: address not mapped (addr: 0x7ffc65604290)
WARN: Backtrace from uftrace:
WARN: [1] (puts[5597527fe660] <= main[5597527fe86b])
WARN: [0] (main[5597527fe838] <= __libc_start_main[7feaa464bb97])
   0.840 us [ 13506] | __monstartup();
   0.305 us [ 13506] | __cxa_atexit();
            [ 13506] | main() {
            [ 13506] |   puts() {

task: 13506
[1] puts
[0] main
```

After :  
```
$ uftrace -A puts@arg123123123 ./simple
mcount: /home/rls1004/contributhon/uftrace/arch/x86_64/mcount-support.c:116:mcount_get_stack_arg
 ERROR: invalid argument: 123123117
   0.567 us [ 18317] | __monstartup();
   0.257 us [ 18317] | __cxa_atexit();
```

Fixed: #858

### DIFF
--- a/arch/aarch64/mcount-support.c
+++ b/arch/aarch64/mcount-support.c
@@ -114,6 +114,7 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 			  struct uftrace_arg_spec *spec)
 {
 	int offset = 1;
+	unsigned long *addr = ctx->stack_base;
 
 	switch (spec->type) {
 	case ARG_TYPE_STACK:
@@ -132,10 +133,20 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 		break;
 	}
 
-	if (offset < 1 || offset > 100)
+	if (offset < 1 || offset > 100) {
 		pr_dbg("invalid stack offset: %d\n", offset);
+		memset(ctx->val.v, 0, sizeof(ctx->val));
+		return;
+	}
 
-	memcpy(ctx->val.v, ctx->stack_base + offset, spec->size);
+	addr += offset;
+
+	if (check_mem_region(ctx, (unsigned long)addr))
+		memcpy(ctx->val.v, addr, spec->size);
+	else {
+		pr_dbg("stack address is not allowed: %p\n", addr);
+		memset(ctx->val.v, 0, sizeof(ctx->val));
+	}
 }
 
 void mcount_arch_get_arg(struct mcount_arg_context *ctx,
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -363,8 +363,8 @@ static bool find_mem_region(struct rb_root *root, unsigned long addr)
 	return false;
 }
 
-static bool check_mem_region(struct mcount_arg_context *ctx,
-			     unsigned long addr)
+bool check_mem_region(struct mcount_arg_context *ctx,
+		      unsigned long addr)
 {
 	bool update = true;
 	struct mcount_mem_regions *regions = ctx->regions;
@@ -796,6 +796,12 @@ void save_watchpoint(struct mcount_thread_data *mtdp,
 		     struct mcount_ret_stack *rstack,
 		     unsigned long watchpoints)
 {
+}
+
+bool check_mem_region(struct mcount_arg_context *ctx,
+		      unsigned long addr)
+{
+	return true;
 }
 #endif
 
--- a/arch/arm/mcount-support.c
+++ b/arch/arm/mcount-support.c
@@ -412,6 +412,7 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 			  struct uftrace_arg_spec *spec)
 {
 	int offset = 1;
+	unsigned long *addr = ctx->stack_base;
 
 	switch (spec->type) {
 	case ARG_TYPE_STACK:
@@ -438,10 +439,20 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 		break;
 	}
 
-	if (offset < 1 || offset > 100)
+	if (offset < 1 || offset > 100) {
 		pr_dbg("invalid stack offset: %d\n", offset);
+		memset(ctx->val.v, 0, sizeof(ctx->val));
+		return;
+	}
 
-	memcpy(ctx->val.v, ctx->stack_base + offset, spec->size);
+	addr += offset;
+
+	if (check_mem_region(ctx, (unsigned long)addr))
+		memcpy(ctx->val.v, addr, spec->size);
+	else {
+		pr_dbg("stack address is not allowed: %p\n", addr);
+		memset(ctx->val.v, 0, sizeof(ctx->val));
+	}
 }
 
 void mcount_arch_get_arg(struct mcount_arg_context *ctx,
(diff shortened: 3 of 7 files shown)


arch/aarch64/mcount-support.c
```diff
@@ -110,10 +110,44 @@ int mcount_get_register_arg(struct mcount_arg_context *ctx,
...
+			return true;
+
+		if (iter->start > addr)
+			p = &parent->rb_left;
+		else
+			p = &parent->rb_right;
+	}
+
+	return false;
+}
```
namhyung: Please just use existing `check_mem_region()` instead of duplicating the code for each arch.  You'll need to make it a global function (by removing `static` keyword) first.

arch/aarch64/mcount-support.c
```diff
@@ -135,7 +169,12 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 	if (offset < 1 || offset > 100)
 		pr_dbg("invalid stack offset: %d\n", offset);
 
-	memcpy(ctx->val.v, ctx->stack_base + offset, spec->size);
+	addr += offset;
+
+	if( check_arg_mem_region(ctx->regions, (unsigned long)addr) )
```
namhyung: Please follow the coding style with regard to spaces.

arch/aarch64/mcount-support.c
```diff
@@ -135,7 +169,12 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
...
+		memcpy(ctx->val.v, addr, spec->size);
+	else
+		pr_err_ns("invalid argument: %d\n", offset);
```
namhyung: I don't think we want to terminate the program (`pr_err_ns()` will do) because of an invalid argument.  Probably `pr_dbg()` would be enough.  Or maybe we could zero-fill for such invalid arguments or offsets.

libmcount/record.c
```diff
@@ -416,6 +416,7 @@ static unsigned save_to_argbuf(void *argbuf, struct list_head *args_spec,
 		if (is_retval != (spec->idx == RETVAL_IDX))
 			continue;
 
+		update_mem_regions(ctx->regions);
```
namhyung: No need to do it here since `mcount_arch_get_arg/retval()` will call it if needed.  And I don't want to add the overhead of doing that unconditionally.

arch/aarch64/mcount-support.c
rls1004: The `check_mem_region` function is located between `#ifdef DISABLE_MCOUNT_FILTER` and `#endif` so that it will be compiled only when the `DISABLE_MCOUNT_FILTER` flag is set.

To use this function as a global function, it must be changed from conditional compilation to unconditional compilation, as well as removing the static keyword.

In addition, some of the functions and structures associated with this function are set to conditional compilation, which must be changed to unconditional compilation.

Is it okay to change like this?

namhyung:
Could you please squash two commits into one?

namhyung:
Please note that I didn't receive any notification if you don't leave a comment after you update it.

arch/aarch64/mcount-support.c
```diff
@@ -114,6 +114,7 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 			  struct uftrace_arg_spec *spec)
 {
 	int offset = 1;
+	unsigned long *addr = ctx->stack_base;
```
namhyung: Why not adding offset here?

arch/aarch64/mcount-support.c
```diff
@@ -135,7 +136,12 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 	if (offset < 1 || offset > 100)
 		pr_dbg("invalid stack offset: %d\n", offset);
 
-	memcpy(ctx->val.v, ctx->stack_base + offset, spec->size);
+	addr += offset;
+
+	if (check_mem_region(ctx, (unsigned long)addr))
+		memcpy(ctx->val.v, addr, spec->size);
+	else
+		pr_dbg("invalid argument offset: %d\n", offset);
```
namhyung: I think following message looks better:

```c
pr_dbg("stack address is not allowed: %p\n", addr);
```

namhyung:
Also there's a typo in the commit message header.

arch/aarch64/mcount-support.c
namhyung: Oh, I forgot the code.. sorry for the noise. ;-p

Looking at the code, it looks the initial value of 1 is just for documentation purpose.  It should be greater than or equal to 1.  And the actual value will be calculated from the argspec type anyway.

rls1004:
Thank you for your review.
I fixed a typo in the commit header and changed the message in pr_dbg.

arch/aarch64/mcount-support.c
```diff
@@ -135,7 +136,12 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 	if (offset < 1 || offset > 100)
 		pr_dbg("invalid stack offset: %d\n", offset);
 
-	memcpy(ctx->val.v, ctx->stack_base + offset, spec->size);
+	addr += offset;
+
+	if (check_mem_region(ctx, (unsigned long)addr))
+		memcpy(ctx->val.v, addr, spec->size);
+	else
+		pr_dbg("stack address is not allowed: %p\n", addr);
```
namhyung: Could you please set `ctx->val.v` to zero for the error cases?  Maybe it'd be better doing it at the beginning of the function unconditionally for possible changes...

rls1004:
ctx->val.v is now set to 0 for the error case.
and I added new test case.

arch/aarch64/mcount-support.c
```diff
@@ -135,7 +136,14 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 	if (offset < 1 || offset > 100)
 		pr_dbg("invalid stack offset: %d\n", offset);
```
namhyung: Oops, sorry, you should handle this case as well.  It'd be nice call `memset()` with 0 and return to prevent invalid values.

rls1004: If a return is made in this code, it is not necessary to check the memory range.
If the offset is between 1 and 100, it does not exceed the memory range.
Why is the code limit the offset range between 1 and 100?

namhyung: Because it's very unlikely to have arguments more than 100.

rls1004: I added code to handle this case.

namhyung:
Could you please add your "Signed-off-by" line for the second commit too?

honggyukim:
In addition, please rebase it before force push the branch.

rls1004:
I added "Signed-off-by" and rebase it.

namhyung:
Thanks for your work and endurance!

PR #866, opened 2019-09-11
script: Add luajit scripting support

gonapps:
Add luajit scripting support


honggyukim:
Thanks for doing this great work! I tested this in my environment but found a few problems.

**1. Good parts**
- The output of `count.py` and `count.lua` are same.
- The output of `simple.py` and `simple.lua` are same.

**2. Bad parts**
- The output of `info.lua` doesn't provide the info of `info.py`.
```

$ uftrace record -S scripts/info.py t-abc 
True
v0.9.3-15-g5d53 ( dwarf python luajit tui perf sched dynamic )
('t-abc',)

$ uftrace record -S scripts/info.lua t-abc 
(... nothing printed ...)
```
- It seems that there is something wrong reading arguments and return values.
```
$ gcc -pg -g -o t-arg tests/s-arg.c

$ uftrace record -a -S scripts/strings.py t-arg
"a"
---
"c"
---
"b"
---

$ uftrace record -a -S scripts/strings.lua t-arg 
WARN: 0
WARN: 1
WARN: 2
WARN: 0
WARN: 1
WARN: 0
WARN: 1
...
WARN: 0
WARN: 0
WARN: 0
b
---
c
---
a
---
```
Please fix this problem first, then we can discuss further. I appreciate your hard work!

honggyukim:
Please rebase it before updating this PR. Thanks!

namhyung:
Could you please squash some commits so that we can have only meaningful ones?  Also the python scripting engine is not linked directly to reduce unnecessary overhead when it's not used.  Please use `dlopen` + `dlsym` to call luajit APIs for the same reason.

gonapps:
> Could you please squash some commits so that we can have only meaningful ones? Also the python scripting engine is not linked directly to reduce unnecessary overhead when it's not used. Please use `dlopen` + `dlsym` to call luajit APIs for the same reason.

Sure as soon as possible I'll reflect what you said.

honggyukim:
@gonapps You also have to think how to split the current single commit into multiple separate commits. It has to be divided based on related changes.

You will better understand by reading Google Review Developer Guide.
- https://google.github.io/eng-practices/review/

Among the subsections, I recommend you to read why we have to write small commits in other words CL in Google terms.
- https://google.github.io/eng-practices/review/developer/small-cls.html

gonapps:
@honggyukim Thanks for the tip!

gonapps:
**Current status: segfault in some situations**
```console
uftrace record -S scripts/strings.lua ./t-arg
```

In scripts/strings.lua
```lua
print('"' .. strval .. '"') -- this causes segfault
print(strval) -- this is ok
```
In utils/script-luajit.c
lua_pcall() in luajit_uftrace_end() is the direct cause of the segfault

Also wrong arg 'uftrace_end' is passed to strings.lua and it is caused by lua_getglobal() call in luajit_uftrace_end()

Trying to find the fundamental reason for these behaviors..
If someone has idea or clue about it, please reply to me.

honggyukim:
@gonapps Thanks for the update, but it works fine on my environment.

I just made some simple modification to remove debug message as follows:
```diff
diff --git a/scripts/strings.lua b/scripts/strings.lua
index 1bc4a23..037ffad 100644
--- a/scripts/strings.lua
+++ b/scripts/strings.lua
@@ -13,7 +13,6 @@ function uftrace_entry(ctx)
                 arg = string.gsub(arg, '^%s+', '')
                 arg = string.gsub(arg, '%s+$', '')
                 if arg ~= '' then
-                    print('arg ' .. arg)
                     strset[arg] = true
                 end
             end
@@ -27,7 +26,6 @@ function uftrace_exit(ctx)
             ret = string.gsub(ret, '^%s+', '')
             ret = string.gsub(ret, '%s+$', '')
             if ret ~= '' then
-                print('ret ' .. ret)
                 strset[ret] = true
             end
         end
@@ -39,4 +37,5 @@ function uftrace_end()
         print('"' .. strval .. '"') -- this causes segfault
         print("---")
     end
+    print("uftrace_end")
 end

```
Then, it's output looks good as follows:
```
$ uftrace record -S scripts/strings.lua ./t-arg
"b"
---
"c"
---
"a"
---
uftrace_end
```

gonapps:
@honggyukim What is your environment? mine is Manjaro Linux
```console
uname -r
4.19.69-1-MANJARO
```
The reason I added the code printing arg is to check arg values.
In my environment wrong argument 'uftrace_end' is passed to strings.lua
```console
gcc tests/s-arg.c -pg -o t-arg
uftrace record -S scripts/strings.lua ./t-arg
arg c
arg c
arg b
arg b
arg a
arg a
arg uftrace_end

```

honggyukim:
Without modifying your patches, I get the following output without being crashed.
```
$ uftrace record -S scripts/strings.lua ./t-arg
arg c
arg c
arg c
arg b
arg b
arg b
arg a
arg a
arg a
"b"
---
"c"
---
"a"
---
```
My system is `Ubuntu 16.04.5 LTS`.

honggyukim:
@gonapps Can it be used as a way to do some kind of fault injection? For example, it may always replace the return value of `fopen` to NULL for security tests or fuzzing. Please read more about it in the following link.
- https://en.wikipedia.org/wiki/Fault_injection#Application_of_fault_injection

I'm just wondering if it's possible to write an arbitrary value to some specific memory location.

honggyukim:
Thanking it again, it can be simply used in the command line as follows:
```
$ uftrace record -R fopen@retval=0 ./a.out
```
`fopen@retval=0` indicates to overwrite any return values of fopen to `NULL`. By using scripting feature, it can be done conditionally by checking the value first.

gonapps:
@honggyukim Thanks I'll try that soon.

honggyukim:
> @honggyukim Thanks I'll try that soon.

I was just wondering if it's possible to do fault injection using lua script without the above syntax.

namhyung:
@gonapps do you have any update?  I think you need to prevent lua functions from being called after `mcount_finish()`.

gonapps:
Sorry.. currently I am kind of busy due to team projects in my university.
I think I can do more work in next week.

2019년 10월 2일 (수) 오후 2:47, Namhyung Kim <notifications@github.com>님이 작성:

namhyung:
Nevermind, I really don't want to push you.  You don't need to say "sorry" and take your time. :)

gonapps:
I investigated about this yesterday and definitely calling lua api functions in luajit_uftrace_end() causes problem. Somehow lua_getglobal() function is get hooked by plthook.
However, what I don't understand is __PyObject_CallObject(pFuncEnd, NULL); in python_uftrace_end() doesn't cause any problems.

honggyukim:
Hmm.. I've been trying to reproduce the problem you mentioned, but it works fine on my environment. Could you please show us the backtrace when it gets segfault in `luajit_uftrace_end`?

gonapps:
@honggyukim 
```console
Linux gonapps-laptop 4.19.69-1-MANJARO #1 SMP PREEMPT Thu Aug 29 08:51:46 UTC 2019 x86_64 GNU/Linux
```
My .gdbinit file is as follows:
```console
set follow-fork-mode child
set follow-exec-mode new
```
As you can see, lua api functions are traced by plthook.
```console
gdb --args ./uftrace record -S scripts/strings.lua ./t-arg 
GNU gdb (GDB) 8.3
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./uftrace...
(gdb) b luajit_uftrace_entry
Breakpoint 1 at 0x38a50: file /home/gonapps/workspace/uftrace/utils/script-luajit.c, line 220.
(gdb) b luajit_uftrace_end
Breakpoint 2 at 0x38b20: luajit_uftrace_end. (2 locations)
(gdb) r
Starting program: /home/gonapps/workspace/uftrace/uftrace record -S scripts/strings.lua ./t-arg
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Attaching after Thread 0x7ffff6fbf780 (LWP 3615) fork to child process 3619]
[New inferior 2 (process 3619)]
[Detaching after fork from parent process 3615]
[Inferior 1 (process 3615) detached]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
process 3619 is executing new program: /home/gonapps/workspace/uftrace/t-arg
[New inferior 3]
[New process 3619]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Switching to Thread 0x7ffff6ff1880 (LWP 3619)]

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9c0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9d0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffda10)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9e0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9c0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9a0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.
arg c
arg c

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9c0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9a0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.
arg b
arg b

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9c0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd9a0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.
arg a
arg a

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd8d0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd990)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd920)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 2, luajit_uftrace_end ()
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:267
267		pr_dbg("%s()\n", __func__);
(gdb) bt
#0  luajit_uftrace_end () at /home/gonapps/workspace/uftrace/utils/script-luajit.c:267
#1  0x00007ffff7f95de6 in mcount_trace_finish (send_msg=<optimized out>)
    at /home/gonapps/workspace/uftrace/libmcount/mcount.c:554
#2  0x00007ffff7f95ca2 in mcount_finish () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:782
#3  mcount_cleanup () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1797
#4  mcount_fini () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1881
#5  0x00007ffff7fe2b2b in _dl_fini () from /lib64/ld-linux-x86-64.so.2
#6  0x00007ffff7db9e70 in __run_exit_handlers () from /usr/lib/libc.so.6
#7  0x00007ffff7db9fae in exit () from /usr/lib/libc.so.6
#8  0x00007ffff7da3eea in __libc_start_main () from /usr/lib/libc.so.6
#9  0x00005555555550ae in _start ()
(gdb) n
268		lua_getglobal(L, "uftrace_end"); // uftrace_end
(gdb) n

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd850)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) bt
#0  luajit_uftrace_entry (sc_ctx=0x7fffffffd850) at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
#1  0x00007ffff7f96dc2 in script_hook_entry (tr=0x7fffffffd920, tr=0x7fffffffd920, rstack=0x555555634920, 
    mtdp=0x7ffff6ff0798) at /home/gonapps/workspace/uftrace/libmcount/mcount.c:918
#2  mcount_entry_filter_record (mtdp=mtdp@entry=0x7ffff6ff0798, rstack=rstack@entry=0x555555634920, 
    tr=0x7fffffffd920, regs=regs@entry=0x7fffffffd9a8) at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1036
#3  0x00007ffff7f9b36e in __plthook_entry (regs=0x7fffffffd9a8, module_id=<optimized out>, child_idx=6, 
    ret_addr=0x7fffffffd9e8) at /home/gonapps/workspace/uftrace/libmcount/plthook.c:828
#4  plthook_entry (ret_addr=0x7fffffffd9e8, child_idx=<optimized out>, module_id=<optimized out>, 
    regs=0x7fffffffd9a8) at /home/gonapps/workspace/uftrace/libmcount/plthook.c:903
#5  0x00007ffff7fb4e08 in plt_hooker () at /home/gonapps/workspace/uftrace/arch/x86_64/plthook.S:40
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) n
222		lua_getglobal(L, "uftrace_entry"); // uftrace_entry
(gdb) bt
#0  luajit_uftrace_entry (sc_ctx=0x7fffffffd850) at /home/gonapps/workspace/uftrace/utils/script-luajit.c:222
#1  0x00007ffff7f96dc2 in script_hook_entry (tr=0x7fffffffd920, tr=0x7fffffffd920, rstack=0x555555634920, 
    mtdp=0x7ffff6ff0798) at /home/gonapps/workspace/uftrace/libmcount/mcount.c:918
#2  mcount_entry_filter_record (mtdp=mtdp@entry=0x7ffff6ff0798, rstack=rstack@entry=0x555555634920, 
    tr=0x7fffffffd920, regs=regs@entry=0x7fffffffd9a8) at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1036
#3  0x00007ffff7f9b36e in __plthook_entry (regs=0x7fffffffd9a8, module_id=<optimized out>, child_idx=6, 
    ret_addr=0x7fffffffd9e8) at /home/gonapps/workspace/uftrace/libmcount/plthook.c:828
#4  plthook_entry (ret_addr=0x7fffffffd9e8, child_idx=<optimized out>, module_id=<optimized out>, 
    regs=0x7fffffffd9a8) at /home/gonapps/workspace/uftrace/libmcount/plthook.c:903
#5  0x00007ffff7fb4e08 in plt_hooker () at /home/gonapps/workspace/uftrace/arch/x86_64/plthook.S:40
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) n
223		if (lua_isnil(L, -1)) { // nil
(gdb) n
228		setup_common_context(sc_ctx); // uftrace_entry ctx
(gdb) n
229		if (sc_ctx->arglen)
(gdb) n
230			setup_argument_context(false, sc_ctx); // uftrace_entry ctx
(gdb) n
232		if (lua_pcall(L, 1, 0, 0) != 0) { // err
(gdb) n
arg uftrace_end
script_hook_entry (tr=0x7fffffffd920, tr=0x7fffffffd920, rstack=0x555555634920, mtdp=0x7ffff6ff0798)
    at /home/gonapps/workspace/uftrace/libmcount/mcount.c:919
919		mcount_restore_arch_context(&mtdp->arch);
(gdb) n
922		symbol_putname(sym, symname);
(gdb) n
mcount_entry_filter_record (mtdp=mtdp@entry=0x7ffff6ff0798, rstack=rstack@entry=0x555555634920, tr=0x7fffffffd920, 
    regs=regs@entry=0x7fffffffd9a8) at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1040
1040			if (tr->flags & FLAGS_TO_CHECK) {
(gdb) n
__plthook_entry (regs=0x7fffffffd9a8, module_id=<optimized out>, child_idx=6, ret_addr=0x7fffffffd9e8)
    at /home/gonapps/workspace/uftrace/libmcount/plthook.c:830
830		if (unlikely(special_flag)) {
(gdb) n
892			mcount_unguard_recursion(mtdp);
(gdb) n
plthook_entry (ret_addr=<optimized out>, child_idx=<optimized out>, module_id=<optimized out>, 
    regs=<optimized out>) at /home/gonapps/workspace/uftrace/libmcount/plthook.c:904
904		errno = saved_errno;
(gdb) n
905		return ret;
(gdb) n
plt_hooker () at /home/gonapps/workspace/uftrace/arch/x86_64/plthook.S:41
41		movq %rax, %r11
(gdb) n
43		pop  %rax
(gdb) n
46		pop  %rcx
(gdb) c
Continuing.

Thread 3.1 "t-arg" hit Breakpoint 1, luajit_uftrace_entry (sc_ctx=0x7fffffffd7b0)
    at /home/gonapps/workspace/uftrace/utils/script-luajit.c:220
220		pr_dbg("%s()\n", __func__);
(gdb) n
222		lua_getglobal(L, "uftrace_entry"); // uftrace_entry
(gdb) n
223		if (lua_isnil(L, -1)) { // nil
(gdb) n
228		setup_common_context(sc_ctx); // uftrace_entry ctx
(gdb) n
229		if (sc_ctx->arglen)
(gdb) n
230			setup_argument_context(false, sc_ctx); // uftrace_entry ctx
(gdb) n
232		if (lua_pcall(L, 1, 0, 0) != 0) { // err
(gdb) n
script_hook_entry (tr=0x7fffffffd880, tr=0x7fffffffd880, rstack=0x555555634920, mtdp=0x7ffff6ff0798)
    at /home/gonapps/workspace/uftrace/libmcount/mcount.c:919
919		mcount_restore_arch_context(&mtdp->arch);
(gdb) n
922		symbol_putname(sym, symname);
(gdb) c
Continuing.

Thread 3.1 "t-arg" received signal SIGSEGV, Segmentation fault.
0x00007ffff7b4d4af in ?? () from /usr/lib/libluajit-5.1.so.2
(gdb) bt
#0  0x00007ffff7b4d4af in ?? () from /usr/lib/libluajit-5.1.so.2
#1  0x00007ffff7b1b9e5 in ?? () from /usr/lib/libluajit-5.1.so.2
#2  0x00007ffff7b6297c in lua_pcall () from /usr/lib/libluajit-5.1.so.2
#3  0x00007ffff7fac611 in luajit_uftrace_end () at /home/gonapps/workspace/uftrace/utils/script-luajit.c:273
#4  0x00007ffff7f95de6 in mcount_trace_finish (send_msg=<optimized out>)
    at /home/gonapps/workspace/uftrace/libmcount/mcount.c:554
#5  0x00007ffff7f95ca2 in mcount_finish () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:782
#6  mcount_cleanup () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1797
#7  mcount_fini () at /home/gonapps/workspace/uftrace/libmcount/mcount.c:1881
#8  0x00007ffff7fe2b2b in _dl_fini () from /lib64/ld-linux-x86-64.so.2
#9  0x00007ffff7db9e70 in __run_exit_handlers () from /usr/lib/libc.so.6
#10 0x00007ffff7db9fae in exit () from /usr/lib/libc.so.6
#11 0x00007ffff7da3eea in __libc_start_main () from /usr/lib/libc.so.6
#12 0x00005555555550ae in _start ()
(gdb)
```

honggyukim:
I'd like to ask you to test with the following 3 different examples.

**1. lua runtime execution with `--no-libcall`**
```
$ uftrace record -S scripts/strings.lua --no-libcall ./t-arg
```

**2. lua analysis execution after lua record**
```
$ uftrace record -S scripts/strings.lua ./t-arg
$ uftrace script -S scripts/strings.lua
```

**3. lua analysis execution after normal record**

I found another problem that doesn't print anything.
```
$ uftrace record ./t-arg
$ uftrace script -S scripts/strings.lua
```

honggyukim:
I found a different between script-python and script-lua implementation. script-python loads its shared library with `dlopen`, but script-lua doesn't in the current implementation.
```
$ ldd libmcount/libmcount.so | grep python

$ ldd libmcount/libmcount.so | grep lua
        libluajit-5.1.so.2 => /usr/lib/x86_64-linux-gnu/libluajit-5.1.so.2 (0x00007fbf1e973000)

$ ldd uftrace | grep python

$ ldd uftrace | grep lua
        libluajit-5.1.so.2 => /usr/lib/x86_64-linux-gnu/libluajit-5.1.so.2 (0x00007fd3a76cf000)
```
The related code can be found as follows:
```c
$ cat utils/script-python.c
    ...
#define INIT_PY_API_FUNC(func) \
        do { \
                __##func = dlsym(python_handle, #func); \
                if (!__##func) { \
                        pr_err("dlsym for \"" #func "\" is failed!\n"); \
                        return -1; \
                } \
        } while (0)

static int load_python_api_funcs(void)
{
        python_handle = dlopen(libpython, RTLD_LAZY | RTLD_GLOBAL);
        if (!python_handle) {
                pr_warn("%s cannot be loaded!\n", libpython);
                return -1;
        }

        INIT_PY_API_FUNC(Py_Initialize);
        INIT_PY_API_FUNC(Py_Finalize);
                ...
        INIT_PY_API_FUNC(PyObject_CallObject);
                ...
}
        ...
int python_uftrace_end(void)
{
                ...
        /* Call python function "uftrace_end". */
        __PyObject_CallObject(pFuncEnd, NULL);
                ...
}
    ...
```

gonapps:
@honggyukim 

> I found a different between script-python and script-lua implementation. script-python loads its shared library with dlopen, but script-lua doesn't in the current implementation.

Yes I know I didn't use dlopen to use luajit. But is it possible that it leads to that behavior of luajit functions being traced?

gonapps:
@honggyukim 

> 1. lua runtime execution with --no-libcall

both python and luajit prints nothing.

> 2. lua analysis execution after lua record



```
arg c
arg c
arg b
arg b
arg a
arg a
arg uftrace_end
arg c
arg c
arg b
arg b
arg a
arg a
arg uftrace_end
"b"
---
"uftrace_end"
---
"c"
---
"a"
---

```

> 3. lua analysis execution after normal record

python script doesn't print anything either

honggyukim:
It you don't use `dlopen` and `dlsym`, it their library functions might be hooked by plthook routines. Here are the code for avoiding not to hook `mcount` related functions in the target binary.

https://github.com/namhyung/uftrace/blob/322185a09c85e280733949db661ce975253aef7f/libmcount/plthook.c#L90-L100

https://github.com/namhyung/uftrace/blob/322185a09c85e280733949db661ce975253aef7f/libmcount/plthook.c#L128-L141

The following commit describes about the problem.
- b7f4c69 plthook: Add plt_skip_syms[] and plt_skip_nr

honggyukim:
>> **1. lua runtime execution with --no-libcall**

> both python and luajit prints nothing.
```
$ uftrace record -S scripts/strings.py --no-libcall ./t-arg
"a"
---
"c"
---
"b"
---

$ uftrace record -S scripts/strings.lua --no-libcall ./t-arg
"b"
---
"c"
---
"a"
---
```

>> **3. lua analysis execution after normal record**

> python script doesn't print anything either

I missed to add `-a` option during `record`. Here is the fixed execution output.
```
$ uftrace record -a ./t-arg

$ uftrace script -S scripts/strings.py
"a"
---
"c"
---
"b"
---

$ uftrace script -S scripts/strings.lua
"b"
---
"c"
---
"a"
---
```
It seems working fine in my environment.

gonapps:
Yes, '3. lua analysis execution after normal record' works fine after -a option
And I'll try to modify my code to use dlopen, dlsym

honggyukim:
>> **1. lua runtime execution with --no-libcall**

> both python and luajit prints nothing.

Hmm.. strange thing is that my test result shows that both python and lua print the same output. But I don't know why it doesn't print anything in your environment.

> Yes, '3. lua analysis execution after normal record' works fine after -a option
And I'll try to modify my code to use dlopen, dlsym

Thanks.

gonapps:
@honggyukim 
Now I make code use dlopen() to load the lua library and looks like the bug is fixed!

honggyukim:
Thanks for the update. But please rebase it onto the current master whenever you force-push the commit.

In addition, please fix the following segfault.
```
$ cd scripts

$ uftrace record -S ./trace-memcpy.py -la ls
count.lua  dump.py   info.lua  replay.py          simple.lua  strings.lua  trace-memcpy.lua  trace-memcpy.pyc
count.py   dump.pyc  info.py   report-libcall.py  simple.py   strings.py   trace-memcpy.py   uftrace.data

$ uftrace record -S ./trace-memcpy.lua -la ls
```
Here is the backtrace for the crash.
```
Thread 2.1 "ls" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7fc4900 (LWP 128113)]
0x0000000000000000 in ?? ()
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007ffff7bb7b03 in script_init_for_luajit (info=info@entry=0x7fffffffdf50, ptype=ptype@entry=PATT_REGEX) at /home/honggyu/work/uftrace/util
s/script-luajit.c:406
#2  0x00007ffff7bb5303 in script_init (info=info@entry=0x7fffffffdf50, ptype=ptype@entry=PATT_REGEX) at /home/honggyu/work/uftrace/utils/script.c:
117
#3  0x00007ffff7ba26c2 in mcount_script_init (patt_type=PATT_REGEX) at /home/honggyu/work/uftrace/libmcount/mcount.c:1635
#4  mcount_startup () at /home/honggyu/work/uftrace/libmcount/mcount.c:1786
#5  0x00007ffff7de76ca in call_init (l=<optimized out>, argc=argc@entry=1, argv=argv@entry=0x7fffffffe228, env=env@entry=0x7fffffffe238) at dl-init.c:72
#6  0x00007ffff7de77db in call_init (env=0x7fffffffe238, argv=0x7fffffffe228, argc=1, l=<optimized out>) at dl-init.c:30
#7  _dl_init (main_map=0x7ffff7ffe168, argc=1, argv=0x7fffffffe228, env=0x7fffffffe238) at dl-init.c:120
#8  0x00007ffff7dd7c6a in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
#9  0x0000000000000001 in ?? ()
#10 0x00007fffffffe521 in ?? ()
#11 0x0000000000000000 in ?? ()
```

gonapps:
>In addition, please fix the following segfault.

Thanks I fixed it.

>Thanks for the update. But please rebase it onto the current master whenever you force-push the commit.

You mean syncing forked master branch with upstream master branch then rebase master from current branch?

honggyukim:
>> In addition, please fix the following segfault.

> Thanks I fixed it.

I can see it's fixed. Thanks for the quick fix.

>> Thanks for the update. But please rebase it onto the current master whenever you force-push the commit.

> You mean syncing forked master branch with upstream master branch then rebase master from current branch?

Let's say `origin` is namhyung's repo, then run the following commands.
```
$ git fetch origin
$ git rebase origin/master
```
Then force-push it again. The next step can be dividing your large single commit into multiple related commits as I mentioned in https://github.com/namhyung/uftrace/pull/866#issuecomment-532031615.

However, if you don't feel comfortable to fix and update multiple commits with `rebase` command, then you can wait for @namhyung's review first.

honggyukim:
There is another problem, which is different from python script output.
```
$ uftrace record -S ./strings.py pwd
/home/honggyu/work/uftrace/scripts

$ uftrace record -S ./strings.lua pwd
WARN: uftrace_exit failed: ./strings.lua:27: bad argument #1 to 'gsub' (string expected, got nil)
WARN: uftrace_exit failed: ./strings.lua:27: bad argument #1 to 'gsub' (string expected, got nil)
WARN: uftrace_exit failed: ./strings.lua:27: bad argument #1 to 'gsub' (string expected, got nil)
WARN: uftrace_exit failed: ./strings.lua:27: bad argument #1 to 'gsub' (string expected, got nil)
/home/honggyu/work/uftrace/scripts
```

gonapps:
@honggyukim Thanks for the help!
I fixed strings.lua and rebase master branch from the current branch.
However I didn't split this commit into multiple commits yet.
Maybe we can wait until @namhyung says it needs splitting?

honggyukim:
Thanks for the update. I just tested it with `strings.lua` again and it works fine.

I'm okay to wait for @namhyung's review before you split the commit.

gonapps:
@namhyung Okay I split the commit into multiples.

>Plus, it'd be much nice to add a test case for luajit script.

Can you elaborate more about this?
Did you mean to test lua files in scripts directory or test utils/script-luajit.c ?

honggyukim:
>> Plus, it'd be much nice to add a test case for luajit script.

> Can you elaborate more about this?
> Did you mean to test lua files in scripts directory or test utils/script-luajit.c ?

It means the test inside `tests` directory. The test files are like `tests/t001_basic.py`.

The entire unittest can be executed with `make test`, but I prefer to fine tune it as follows:
```
$ cd tests
$ ./runtest.py -j8
```

gonapps:
@honggyukim Oh thanks, I understand what to do.

honggyukim:
Thanks for splitting the commits. They look mostly fine, but have the following problem.

Even if this PR contains multiple commits, each step must be compilable. It's important especially when we try to fine some bug with `git bisect` command. If some commit point cannot be compiled, then it's difficult to move the commits back and forward for testing. 

So please fix the following problem. It cannot be compiled after having only your first commit. The compilation error is expected to be fixed with the later commit, but it's not desirable.
```
$ git checkout 36dc261

$ git log --oneline -1
36dc261 (HEAD) utils: Make script engine generic

$ make
  CC       uftrace.o
In file included from /home/honggyu/work/uftrace/uftrace.c:41:0:
/home/honggyu/work/uftrace/utils/script.h:13:33: fatal error: utils/script-luajit.h: No such file or directory
compilation terminated.
Makefile:256: recipe for target '/home/honggyu/work/uftrace/uftrace.o' failed
make: *** [/home/honggyu/work/uftrace/uftrace.o] Error 1
```
For more info about `git bisect` please see below:
- https://flaviocopes.com/git-bisect

Thanks for your work!

honggyukim:
In order to make your work easier with multiple commit modification, you will need to know more about the followings:
- interactive rebase and edit mode
- `git rebase --continue`

honggyukim:
Thanks for the update. But it seems that it cannot apply in-script options by `uftrace-option`.
```
$ uftrace record -S scripts/strings.lua pwd
uftrace: /home/honggyu/work/uftrace/cmds/record.c:1622:check_binary
 ERROR: Can't find 'mcount' symbol in the '/bin/pwd'.
        It seems not to be compiled with -pg or -finstrument-functions flag
        which generates traceable code.  Please check your binary file.
```
The following output is also different.
```
$ uftrace record -S scripts/strings.lua --no-libcall ./t-arg
$ uftrace replay
            [ 56637] | main() {
            [ 56637] |   foo() {
   1.407 us [ 56637] |     bar();
   0.227 us [ 56637] |     bar();
   0.140 us [ 56637] |     bar();
   4.660 us [ 56637] |   } /* foo */
   0.427 us [ 56637] |   many();
            [ 56637] |   pass() {
   0.296 us [ 56637] |     check();
   0.820 us [ 56637] |   } /* pass */
   8.313 us [ 56637] | } /* main */
```
I can see that each step can be compiled without a problem. Thanks.

gonapps:
@honggyukim Oops, somehow I missed committing modified uftrace.c.. I fixed it

honggyukim:
Thanks for the update. It's fixed now, but I found something else.

Python output shows the strings at the end, but lua prints nothing.
```
$ uftrace script -S scripts/strings.lua --record pwd
/home/honggyu/work/uftrace

$ uftrace script -S scripts/strings.py --record pwd
/home/honggyu/work/uftrace
"/home/honggyu/work/uftrace"
---
"POSIXLY_CORRECT"
---
"en_US.UTF-8"
---
"/"
---
"coreutils"
---
"pwd"
---
"LP"
---
"NULL"
---
"/usr/share/locale"
---
```

scripts/strings.lua
```diff
@@ -0,0 +1,43 @@
...
+local strset = {}
+
+function uftrace_entry(ctx)
+    if ctx['args'] ~= nil then
+        for i, arg in ipairs(ctx['args']) do
+            if type(arg) == 'string' then
+                arg = string.gsub(arg, '^%s+', '')
+                arg = string.gsub(arg, '%s+$', '')
+                if arg ~= '' then
+                    -- print('arg ' .. arg)
```
honggyukim: Please remove this comment.

scripts/strings.lua
```diff
@@ -0,0 +1,43 @@
...
+end
+
+function uftrace_exit(ctx)
+    if ctx['retval'] ~= nil then
+        local ret = ctx['retval'] 
+        if type(ctx['retval']) == 'string' then
+            ret = string.gsub(ret, '^%s+', '')
+            ret = string.gsub(ret, '%s+$', '')
+            if ret ~= '' then
+                -- print('ret ' .. ret)
```
honggyukim: Ditto.

scripts/strings.lua
```diff
@@ -0,0 +1,43 @@
...
+                strset[ret] = true
+            end
+        end
+    end
+end
+
+function uftrace_end()
+    for strval, _ in pairs(strset) do
+        print('"' .. strval .. '"') -- this causes segfault
```
honggyukim: Please remove the comment at the end too.

gonapps:
@honggyukim Would you check the result again? Because in my environment results were the same.
```console
[gonapps@gonapps-laptop uftrace]$ uftrace script -S scripts/strings.lua --record pwd
/home/gonapps/workspace/uftrace
"pwd"
---
"/home/gonapps/workspace/uftrace"
---
"/usr/share/locale"
---
"POSIXLY_CORRECT"
---
"/"
---
"NULL"
---
"LC_CTYPE=en_US.UTF-8;LC_NUMERIC=ko_KR.UTF-8;LC_TIME=ko_KR.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONET..."
---
"coreutils"
---
"LP"
---
[gonapps@gonapps-laptop uftrace]$ uftrace script -S scripts/strings.py --record pwd
/home/gonapps/workspace/uftrace
"POSIXLY_CORRECT"
---
"/"
---
"coreutils"
---
"pwd"
---
"LP"
---
"LC_CTYPE=en_US.UTF-8;LC_NUMERIC=ko_KR.UTF-8;LC_TIME=ko_KR.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONET..."
---
"NULL"
---
"/usr/share/locale"
---
"/home/gonapps/workspace/uftrace"
---

```

honggyukim:
> @honggyukim Would you check the result again?

It works fine now and it's also very good to have a test file. Thanks!

In addition, could you please add `dump.lua` if possible?

gonapps:
@honggyukim Added replay.lua and dump.lua.
Except report-libcall.lua which needs external package to get pid of a process, now there are all equivalent lua scripts to python scripts.

honggyukim:
Great! `replay.lua` and `dump.lua` all work fine. Thanks very much for your effort!

honggyukim:
Here is a bit of performance comparison between python and lua scripting during record time.
```
$ uftrace record -S scripts/trace-memcpy.py --time uptime
 17:35:25 up 250 days,  6:46, 50 users,  load average: 0.10, 0.17, 0.09
elapsed time:      0.053731300 sec
 system time:      0.012000000 sec
   user time:      0.040000000 sec

$ uftrace record -S scripts/trace-memcpy.lua --time uptime
 17:35:26 up 250 days,  6:46, 50 users,  load average: 0.10, 0.17, 0.09
elapsed time:      0.032286664 sec
 system time:      0.012000000 sec
   user time:      0.016000000 sec

$ uftrace record -S scripts/trace-memcpy.py --time pwd
/home/honggyu/work/uftrace
elapsed time:      0.044874418 sec
 system time:      0.004000000 sec
   user time:      0.040000000 sec

$ uftrace record -S scripts/trace-memcpy.lua --time pwd
/home/honggyu/work/uftrace
elapsed time:      0.023431618 sec
 system time:      0.000000000 sec
   user time:      0.020000000 sec
```
They are very simple programs, but show that lua engine is more lightweight than python engine. We can reduce slowdown during scripting execution.

gonapps:
@honggyukim, @namhyung Thank you for the support!

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+
+		switch (spec->fmt) {
+		case ARG_FMT_AUTO:
+		case ARG_FMT_SINT:
+		case ARG_FMT_UINT:
+		case ARG_FMT_HEX:
+		case ARG_FMT_PTR:
+		case ARG_FMT_ENUM:
+			memcpy(val.v, data, spec->size);
+			switch (spec->size) {
```
namhyung: I think you just can use val.v without the switch statement.

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+	dllua_settable(L, -3); // * ctx
+	dllua_pushstring(L, "duration"); // * ctx 'duration'
+	dllua_pushinteger(L, sc_ctx->duration); // * ctx 'duration' duration
+	dllua_settable(L, -3); // * ctx
+	dllua_pushstring(L, "address"); // * ctx 'address'
+	dllua_pushinteger(L, sc_ctx->address); // * ctx 'address' address
+	dllua_settable(L, -3); // * ctx
+	dllua_pushstring(L, "name"); // * ctx 'name'
+	dllua_pushstring(L, sc_ctx->name); // * ctx 'name' name
+	dllua_settable(L, -3); // * ctx
```
namhyung: All these comments look not so meaningful.  Also please use C-style comments if possible.

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+
+	lua_newtable(L); // uftrace_begin ctx
+	dllua_pushstring(L, "record"); // uftrace_begin ctx 'record'
+	dllua_pushboolean(L, info->record); // uftrace_begin ctx 'record' record
+	dllua_settable(L, -3); // uftrace_begin cx
+	dllua_pushstring(L, "version"); // uftrace_begin ctx 'version'
+	dllua_pushstring(L, info->version); // uftrace_begin ctx 'version' version
+	dllua_settable(L, -3); // uftrace_begin ctx
+	dllua_pushstring(L, "cmds"); // uftrace_begin ctx 'cmds'
+	lua_newtable(L); // uftrace_begin ctx 'cmds' cmds
```
namhyung: Do you really need these comments?

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+	int i;
+	char *s;
```
namhyung: I know it's old-fashioned, but I prefer putting declarations at the beginning of function.  Or at least you can use a blank line to separate them from others.

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+}
+
+static int luajit_uftrace_begin(struct script_info *info)
+{
+	pr_dbg("%s()\n", __func__);
+
+	lua_getglobal(L, "uftrace_begin"); // uftrace_begin
+	if (lua_isnil(L, -1)) // nil
+	{
+		lua_pop(L, 1); //
```
namhyung: Can you get rid of all these empty comments?

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+}
+
+static int luajit_uftrace_entry(struct script_context *sc_ctx)
+{
+	pr_dbg("%s()\n", __func__);
+
+	lua_getglobal(L, "uftrace_entry"); // uftrace_entry
+	if (lua_isnil(L, -1)) { // nil
+		lua_pop(L, 1); //
+		return -1;
```
namhyung: Use dllua functions?

misc/install-deps.sh
```diff
@@ -14,20 +14,20 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 case $distro in
     "ubuntu" | "debian")
         apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config
-        apt-get $OPT install libcapstone-dev ;;
+        apt-get $OPT install libcapstone-dev libluajit-5.1-dev ;;
```
namhyung: ok, please squash this into an earlier commit.

gonapps: Do you mean squashing into c5b83a367b18eda50c4c48250ffd97cb82b835ba and change the commit message ?

uftrace.c
```diff
@@ -968,20 +969,39 @@ void parse_script_opt(struct opts *opts)
...
 
+	script_type = get_script_type(opts->script_file);
+
+	if (script_type == SCRIPT_UNKNOWN) {
+		fclose(fp);
+		pr_err("unknown script type");
+	}
+
+        comment_string = comment_strings[script_type];
+        comment_string_length = strlen(comment_string);
```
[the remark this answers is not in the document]
gonapps: Fixed

utils/script-luajit.c
gonapps: Fixed

utils/script-luajit.c
gonapps: Fixed

utils/script-luajit.c
gonapps: Fixed

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+		count++;
+	}
+	if (count == 0)
+		return;
+
+	if (is_retval)
+		dllua_pushstring(L, "retval"); // * ctx 'retval'
+	else
+		dllua_pushstring(L, "args"); // * ctx 'args'
+	lua_newtable(L); // * ctx '<retval-or-args>' <retval-or-args-table>
```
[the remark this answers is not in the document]
gonapps: Fixed

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+}
+
+void __attribute__((weak)) lua_settop(lua_State *L, int index)
+{
+	return dllua_settop(L, index);
+}
+
+static void setup_common_context(struct script_context *sc_ctx)
+{
+	lua_newtable(L); // * ctx
```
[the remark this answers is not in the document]
gonapps: lua_newtable is a macro defined in lua.h. it internally uses lua_createtable()

utils/script-luajit.c
gonapps: lua_global(), lua_isnil(), lua_pop() are macros.

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+	lua_pop(L, 1); //
+
+	luajit_uftrace_begin(info);
+	return 0;
+}
+
+void script_finish_for_luajit(void)
+{
+	pr_dbg("%s()\n", __func__);
+	dllua_close(L);
```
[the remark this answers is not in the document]
gonapps: Fixed. I set script_str to NULL in mcount_cleanup() instead.

configure
```diff
@@ -178,6 +180,7 @@ printf "...%15s: %s\n" "prefix" "${prefix}"
 print_feature "libelf" "have_libelf" "more flexible ELF data handling"
 print_feature "libdw" "have_libdw" "DWARF debug info support"
 print_feature "libpython2.7" "have_libpython2.7" "python scripting support"
+print_feature "libluajit" "have_libluajit" "luajit scripting support"
```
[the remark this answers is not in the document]
gonapps: Ok I'll squash later commit into this commit.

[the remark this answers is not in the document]
gonapps: Fixed

misc/install-deps.sh
gonapps: Fixed

utils/script.h
```diff
@@ -10,12 +10,17 @@
 
 #include "libmcount/mcount.h"
 #include "utils/script-python.h"
+#include "utils/script-luajit.h"
```
gonapps: This file has to be moved if utils/script-luaijt.h is moved.

utils/script-python.c
```diff
@@ -340,17 +340,7 @@ static void setup_argument_context(PyObject **pDict, bool is_retval,
...
-		short         s;
-		int           i;
-		long          l;
-		long long     L;
-		float         f;
-		double        d;
-		long double   D;
-		unsigned char v[16];
-	} val;
+	union script_arg_val val;
```
gonapps: This file also has to be moved if utils/script.h is moved

utils/script.c
```diff
@@ -46,6 +46,8 @@ static enum script_type_t get_script_type(const char *str)
 	 */
 	if (!strcmp(ext, ".py"))
 		return SCRIPT_PYTHON;
+	else if (!strcmp(ext, ".lua"))
+		return SCRIPT_LUAJIT;
```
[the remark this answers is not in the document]
gonapps: If I move these files you mentioned, all the files also have to be moved.

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+#include <lualib.h>
+#include <lauxlib.h>
+#include "utils/utils.h"
+#include "utils/symbol.h"
+#include "utils/fstack.h"
+#include "utils/filter.h"
+#include "utils/script.h"
+#include "utils/script-luajit.h"
+
+static const char *libluajit = "libluajit-5.1.so";
```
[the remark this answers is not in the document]
gonapps: It is like python2.
For backward compatibility libluajit is not going to support full 5.2, 5.3 and actual version is 2.x.x
So every system libluajit will be named like libluajit-5.1.so
```console
pacman -Ql luajit
...
luajit /usr/lib/libluajit-5.1.so
luajit /usr/lib/libluajit-5.1.so.2
luajit /usr/lib/libluajit-5.1.so.2.0.5
...
```

utils/script.c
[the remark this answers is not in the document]
namhyung: Hmm.. I meant that we should not put any luajit changes here since it's not added yet.

utils/script-luajit.c
namhyung: ok, thanks!

utils/script-luajit.c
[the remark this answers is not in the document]
namhyung: good!

utils/script-luajit.c
namhyung: Thoughts?

gonapps:
@namhyung Fixed

utils/script-luajit.c
```diff
@@ -0,0 +1,425 @@
...
+
+const __attribute__((weak)) char *lua_tolstring(lua_State *L, int index, size_t *len)
+{
+	return dllua_tolstring(L, index, len);
+}
+
+void __attribute__((weak)) lua_settop(lua_State *L, int index)
+{
+	return dllua_settop(L, index);
+}
```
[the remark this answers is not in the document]
namhyung: Ok, you remove the weak specifier.  Then why do you need such functions?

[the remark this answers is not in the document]
namhyung: It could be called from outside of libmcount...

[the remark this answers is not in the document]
gonapps: e159ffd953faf2be8978e297d9af62a6f2393aaa

utils/script-python.c
gonapps: Fixed

gonapps:
@namhyung Fixed the commit message. And I already removed weak specifiers.

utils/script-luajit.c
gonapps: memcpy to val.v only occurs in the cases of ARG_FMT_FLOAT and ARG_FMT_ENUM

utils/script-luajit.c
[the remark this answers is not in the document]
namhyung: What do you mean?  What if we remove these functions?

utils/script-luajit.c
namhyung: As val.v is zero initialized and copied only meaningful size, you can just use `dllua_pushinteger(L, val.v)` regardless of spec->size IMO.

namhyung:
I'd be nice if you could add some description of luajit and its merit in the third commit message.  Since some people (including me) are not familiar with Lua and Lua-JIT.

utils/script-luajit.c
gonapps: Hmm I'm kind of worried about endian-safeness. Would it be safe in both big-endian and little-endian systems?

utils/script-luajit.c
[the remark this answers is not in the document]
gonapps: Can't remove these functions. macros used in the code internally calls those functions.

gonapps:
@namhyung Okay I'll add the description

utils/script-luajit.c
namhyung: I'm not sure.  OK let's just leave it.

utils/script-luajit.c
```diff
@@ -0,0 +1,405 @@
...
+			memcpy(&slen, data, 2);
+
+			str = xmalloc(slen + 1);
+
+			/* copy real string contents */
+			memcpy(str, data + 2, slen);
+			str[slen] = '\0';
+
+			/* NULL string is encoded as '0xffffffff' */
+			if (!memcmp(str, &null_str, sizeof(null_str)))
```
namhyung: The str might be less than 4 bytes, you need to check `slen == 4` before this.

gonapps: Fixed.

PR #875, opened 2019-09-17
build: Add docdir to make for later usage

honggyukim:
This patch adds docdir variable to make it changable.  It could be used
to store translated versions of man pages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ bindir = $(prefix)/bin
 libdir = $(prefix)/lib
 etcdir = $(prefix)/etc
 mandir = $(prefix)/share/man
+docdir = $(srcdir)/doc
 
 srcdir = $(CURDIR)
 # set objdir to $(O) by default (if any)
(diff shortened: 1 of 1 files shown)


honggyukim:
With this change, we can use to support Korean man page after applying the following change.
```diff
diff --git a/Makefile b/Makefile
index 38b68ef..c93dc00 100644
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,10 @@ etcdir = $(prefix)/etc
 mandir = $(prefix)/share/man
 docdir = $(srcdir)/doc

+ifeq ($(DOC), ko)
+  docdir = $(srcdir)/doc/ko
+endif
+
 srcdir = $(CURDIR)
 # set objdir to $(O) by default (if any)
 ifeq ($(objdir),)
```
The build command can be as follows:
```
$ make DOC=ko install
```

honggyukim:
> LGTM. I prefer DOCLANG rather than DOC btw.

I'm okay, but it's not included in this PR anyway.

Issue #901, opened 2019-09-24
Differences between option -K and -K 1


yeominstall:
I found an ambiguity about kernel option.
I expected to get the same result from commands below, but I didn’t.

```
parallels@parallels-Parallels-Virtual-Platform:~$ sudo uftrace -k record a.out 
3628800
```
```
parallels@parallels-Parallels-Virtual-Platform:~$ sudo uftrace -K 1 record a.out 
3628800
```
```
parallels@parallels-Parallels-Virtual-Platform:~$ sudo uftrace -K record a.out 
Usage: invalid kernel depth: record (ignoring...)
3628800
# DURATION     TID     FUNCTION
            [  6075] | _GLOBAL__sub_I_fac() {
            [  6075] |   __static_initialization_and_destruction_0() {
            [  6075] |     std::ios_base::Init::Init() {
  11.238 us [  6075] |       __do_page_fault();
...
```

In the help page, we can show about kernel option like:
```
  -k, --kernel               Trace kernel functions also (if supported)
      --keep-pid             Keep same pid during execution of traced program
      --kernel-buffer=SIZE   Size of kernel tracing buffer (default: 1408K)
      --kernel-full          Show kernel functions outside of user
      --kernel-only          Dump kernel data only
      --kernel-skip-out      Skip kernel functions outside of user (deprecated)
                            
  -K, --kernel-depth=DEPTH   Trace kernel functions within DEPTH (default: 1)
      --libmcount-single     Use single thread version of libmcount
      --list-event           List avaiable events
      --logfile=FILE         Save log messages to this file
```

It seems like option `-K` and `-K 1` are same because it described `-K` has default value `1`. 

Could we improve this?
If I misunderstood something about this, please let me know.

honggyukim:
Hmm.. there’s no default value for `-K`. It always has to accept a value for depth setting. The option description looks incorrect. Can you fix the description?

yeominstall:
Sure, I will do.

namhyung:
The number is required. Please simply remove the default value from the help string.

PR #902, opened 2019-09-24
record: Fix description for option -K

yeominstall:
remove description about default value of option -K
because it doesn't actually have a default value.

Fixed: #901

Signed-off-by: Yeomin Nam <y.minn13@gmail.com>


honggyukim:
Could you change "fixed" to "Fixed"? And please write a simple description in the commit message why this change is made.

yeominstall:
@honggyukim Thank you for your advice :) I applied it.

PR #910, opened 2019-09-27
support Korean in configure

junhoryu:
currently it supports only English, when you execute configure.
After apply this patch, if locale has ko_KR it will show Korean instead of English

signed-off-by: JunhoRyu <ruujoon93@gmail.com>


namhyung:
Thanks for the contribution, but I don't think we need translations for this script.

Issue #760, opened 2019-05-13
Document update is needed for dynamic tracing and absolute path usage


honggyukim:
**1. dynamic tracing**
uftrace is now able to trace prebuilt binaries with full dynamic tracing support. So we have to update related documents to explain this promising feature.

**2. absolute pathname**
Since we don't have to give full pathname to uftrace with #741 is merged, let's modify documents that includes the absolute path and which command.

honggyukim:
Can it be added to v0.9.3 milestone?

namhyung:
For item 2, where do you want to change?

namhyung:
For item 1, what about this?

```diff
diff --git a/doc/uftrace-record.md b/doc/uftrace-record.md
index 428e75d..5e00962 100644
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -32,9 +32,11 @@ RECORD OPTIONS
     Implies \--srcline.  See *ARGUMENTS*.
 
 -P *FUNC*, \--patch=*FUNC*
-:   Patch FUNC dynamically.  This is only applicable binaries built by
-    gcc with `-pg -mfentry -mnop-mcount` or clang with `-fxray-instrument`.
-    This option can be used more than once.  See *DYNAMIC TRACING*.
+:   Patch FUNC dynamically.  This option can be used more than once.
+    See *DYNAMIC TRACING*.
+
+-Z *SIZE*, \--size-filter=*SIZE*
+:   Patch functions bigger than SIZE bytes dynamically.  See *DYNAMIC TRACING*.
 
 -E *EVENT*, \--event=*EVENT*
 :   Enable event tracing.  The event should be available on the system.
@@ -541,19 +543,25 @@ normally you need to build the target program with `-pg` (or
 funtions call `mcount()`.
 
 With dynamic tracing, you can trace specific functions only given by the
-`-P`/`--patch` option.  However you need to add some more compiler (gcc) options
-when building the target program.  The gcc 5.1 or more recent versions provide
-`-mfentry` and `-mnop-mcount` options which add instrumentation code (i.e.
-calling `mcount()` function) at the very beginning of a function and convert
-the instruction to a NOP.  Then it has almost zero performance overhead when
-running in a normal condition.  The uftrace can convert it back to call
+`-P`/`--patch` option.  With capstone disassembly engine you even don't need to
+(re)compile the target with the option above.  Now uftrace can analyze the
+instructions and (if possible) it can copy them to a different place and rewrite
[...]

PR #911, opened 2019-09-28
modified MCOUNT_MSG in record.c

junhoryu:
Using full dynamic tracing, uftrace could be worked with binary files which are not compiled with -pg or -finstrument-functions flag
but, MCOUNT_MSG does not have explaination for it.

fixed:#760

signed-off-by: JoonhoRyu <ruuojoon93@gmail.com>

### DIFF
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -433,9 +433,11 @@ static unsigned save_to_argbuf(void *argbuf, struct list_head *args_spec,
 				 */
 				long *base = ctx->val.p;
 				long *_M_string_length = base + 1;
-				char *_M_dataplus = (char*)(*base);
-				len = *_M_string_length;
-				str = _M_dataplus;
+				if (check_mem_region(ctx, (unsigned long)base)) {
+					char *_M_dataplus = (char*)(*base);
+					len = *_M_string_length;
+					str = _M_dataplus;
+				}
 			}
 
 			if (str) {
@@ -443,7 +445,7 @@ static unsigned save_to_argbuf(void *argbuf, struct list_head *args_spec,
 				char *dst = ptr + 2;
 				char buf[32];
 
-				if (!check_mem_region(ctx, ctx->val.i)) {
+				if (!check_mem_region(ctx, (unsigned long)str)) {
 					len = snprintf(buf, sizeof(buf), "<%p>", str);
 					str = buf;
 				}


gy741:
Please include the subject word in the commit message header.
You can solve it by using the git rebase.

Ref:
https://github.com/namhyung/uftrace/blob/master/CONTRIBUTING.md

junhoryu:
I modified commit message header and pushed to my git hub
I was wondering if do I have to request pull request again or not.
thank you.

DanielTimLee:
Just force-pushing the commit to your branch will be fine. 
No need to re-send pull request.

namhyung:
Yes, you can just push the commit to your branch.  But you need add a comment in the pull request conversation so that I can notice there's an update.

namhyung:
Both of the tags in the commit message should start with a capital letter like "Fixed: " and "Signed-off-by: ".

namhyung:
Also please don't use a long line in the commit message.  Instead you need to break it at around 72 characters.

namhyung:
And you pushed an unrelated commit to the master branch.  That's not a good habit and it makes impossible to merge your change.  I recommend that you have a separate branch for each of your work.

cmds/record.c
```diff
@@ -1474,8 +1474,10 @@ static void print_child_usage(struct rusage *ru)
 
 #define UFTRACE_MSG  "Cannot trace '%s': No such executable file.\n"
 
-#define MCOUNT_MSG  "Can't find '%s' symbol in the '%s'.\n"		\
-"\tIt seems not to be compiled with -pg or -finstrument-functions flag\n" 	\
+#define MCOUNT_MSG  "Cannot trace '%s' symbol in the '%s'.\n" \
+"\tusing full dynamic tracing which is -P option works without" \
+ " -pg or -finstrument-functions.\n" \
+"\totherwise, It seems need to be compiled with -pg or -finstrument-functions flag\n" 	\
```
namhyung: I still see that the first line should use 'find' instead of 'trace' since we don't trace "mcount" itself.  How about this?

Cannot find '%s' symbol in the '%s'.\n
It seems not to be compiled with -pg or -finstrument-functions flag.\n
You can rebuild your program with it or use -P option for dynamic tracing.

junhoryu:
I modified the comment as you said.
I will work my next jobs in separated branches as you recommended
thank you for coaching me.

cmds/record.c
junhoryu: sorry, there was something wrong.
after fixing I didn't commit again.
I fixed it.
hoping all of proccesses has done correctly in this time. :)
Thank you.

honggyukim:
@junhoryu Please rebase it before force pushing your commit. This PR contains @yeominstall's previous commit, which looks weird.

You can easily see that this PR contains "two" commits although you wrote only a single commit.

honggyukim:
You should follow the steps below:
- remove @yeominstall's commit first maybe with `git reset --hard HEAD^^^`
- update the repo to the current master with `git pull`
- apply your commit
- force push the commit again

honggyukim:
@junhoryu The reason why you see 2 commits in your PR is as follows:
```
old master --- yeomin's updated commit (uftrace master is here)
           \
            \--- yeomin's modified commit --- your commit
```
It's because you didn't push your commit based on the current master, which includes Yeomin's recent commit. But somehow you have different Yeomin's commit whose hash ID is different, then pushed your commit in the wrong place. Please check this again.

honggyukim:
```
$ git format-patch a68315e^..a68315e
```

junhoryu:
$ git format-patch a68315e^..a68315e
fatal: ambiguous argument 'a68315e^..a68315e': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

namhyung:
I think you can just do `git rebase`.

PR #915, opened 2019-09-30
Makefile: Fix conversion error for ARCH variable already set to x86

sparkleholic:
In open-embedded, the toolchain for  qemux86 machine sets ARCH as x86,
and "-m32" option are appended in CC env variable, not in CFLAGS.
Previous script checked the host arch if ARCH absent, and
checked whether -m32 option is set from CFLAGS only.
This patch fixs above mentioned.

Signed-off-by: Junil Kim <logyourself@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,9 @@ $(call allow-override,LD,$(CROSS_COMPILE)ld)
 
 uname_M := $(shell uname -m 2>/dev/null || echo not)
 
-ARCH ?= $(shell echo $(uname_M) | sed -e s/i.86/i386/ -e s/arm.*/arm/ )
+ARCH := $(shell echo $(uname_M) | sed -e s/i.86/i386/ -e s/arm.*/arm/ )
 ifeq ($(ARCH),x86_64)
-  ifneq ($(findstring m32,$(CFLAGS)),)
+  ifneq ($(findstring -m32,$(CC) $(CFLAGS)),)
     override ARCH := i386
   endif
 endif
--- a/configure
+++ b/configure
@@ -89,12 +89,10 @@ for arg; do
     eval "$opt='$val'"
 done
 
-if [ -z "$ARCH" ]; then
-    uname_M=$(uname -m 2>/dev/null || echo not)
-    ARCH=$(echo $uname_M | sed -e s/i.86/i386/ -e s/arm.*/arm/ )
-    if [ "$ARCH" = "x86_64" ] && echo "$CFLAGS" | grep -w m32 ; then
-        ARCH=i386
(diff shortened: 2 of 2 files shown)


sparkleholic:
The cross toolchain for webOS qemux86 machine which generated by `populate_sdk`
```
$ env
...
ARCH=x86
CC=i586-webos-linux-gcc -m32 -march=i586 --sysroot=/opt/webos-sdk-x86_64/8.0~s6/sysroots/i586-webos-linux
CFLAGS=-O2 -pipe -g -feliminate-unused-debug-types -Werror=return-type
```

honggyukim:
Hi @sparkleholic, It looks a bit strange that `-m32` is provided inside `CC` variable. 

I think you don't have to modify script, but build it as follows:
```
$ ARCH=i386 CFLAGS=-m32 LDFLAGS=-m32 ./configure
```
It seems that we might be better to detect `x86` and change it to `i386` inside the script.

sparkleholic:
Thanks! Yes, adding the variables to calling configure could solve this.

And there is a mention why the open-embedded(yocto)’s CC contains the compiler arguments.

https://lists.yoctoproject.org/pipermail/poky/2018-September/011497.html

honggyukim:
I see. I understand why `CC` is used. Thanks for sharing the link.

@namhyung what do you think about having the following change?
```diff
diff --git a/configure b/configure
index 213f4091..c01a9fe0 100755
--- a/configure
+++ b/configure
@@ -91,10 +91,12 @@ done
 
 if [ -z "$ARCH" ]; then
     uname_M=$(uname -m 2>/dev/null || echo not)
-    ARCH=$(echo $uname_M | sed -e s/i.86/i386/ -e s/arm.*/arm/ )
+    ARCH=$(echo $uname_M | sed -e s/i.86/i386/ -e s/^x86$/i386/ -e s/arm.*/arm/ )
     if [ "$ARCH" = "x86_64" ] && echo "$CFLAGS" | grep -w m32 ; then
         ARCH=i386
     fi
+elif [ "$ARCH" == "x86" ]; then
+    ARCH=i386
 fi
 
 bindir=${bindir:-${prefix}/bin}
```

namhyung:
I'm ok with checking both CC and CFLAGS variables.  But the reason why it checks ARCH before is to override it for cross-build.

namhyung:
@sparkleholic are you ok with the change from @honggyukim ?  Then please update your code (with rebase) and force-push it to this branch.

namhyung:
@sparkleholic I slightly modified your code and merged it.  Please let me know if you still have the problem.

PR #922, opened 2019-10-05
record: Modify MCOUNT_MSG in record.c

junhoryu:
Using full dynamic tracing, uftrace could be worked with binary files
which are not compiled with -pg or -finstrument-functions flag
but, MCOUNT_MSG does not have explaination for it.

Fixed:#760

Signed-off-by: JoonhoRyu <ruujoon93@gmail.com>

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1475,8 +1475,8 @@ static void print_child_usage(struct rusage *ru)
 #define UFTRACE_MSG  "Cannot trace '%s': No such executable file.\n"
 
 #define MCOUNT_MSG  "Can't find '%s' symbol in the '%s'.\n"		\
-"\tIt seems not to be compiled with -pg or -finstrument-functions flag\n" 	\
-"\twhich generates traceable code.  Please check your binary file.\n"
+"\tIt seems not to be compiled with -pg or -finstrument-functions flag.\n" 	\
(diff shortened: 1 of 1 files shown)


cmds/record.c
```diff
@@ -1474,9 +1474,9 @@ static void print_child_usage(struct rusage *ru)
 
 #define UFTRACE_MSG  "Cannot trace '%s': No such executable file.\n"
 
-#define MCOUNT_MSG  "Can't find '%s' symbol in the '%s'.\n"		\
+#define MCOUNT_MSG  "Cannot find '%s' symbol in the '%s'.\n"		\
```
namhyung: Please don't change this.

cmds/record.c
```diff
@@ -1474,9 +1474,9 @@ static void print_child_usage(struct rusage *ru)
...
 "\tIt seems not to be compiled with -pg or -finstrument-functions flag\n" 	\
```
namhyung: I think it's better to add a period ('.') since it became the end of the statement.

junhoryu: I fixed them.

cmds/record.c
junhoryu: I fixed it

junhoryu:
모든 의견 반영하였습니다!

Issue #914, opened 2019-09-30
heap-buffer overflow in setargs()


gy741:
Hello,

I found heap-buffer overflow bug.

Due to an infinite loop, a crash occurs.

https://github.com/namhyung/uftrace/blob/b735366116ce47cd26e65d3488fb6ad77d63d708/utils/utils.c#L727-L734

Crash info:
```
$ cat uftrace.data/default.opts
'

(gdb) p * args
Cannot access memory at address 0x55da7b3da000
(gdb) bt
#0  0x000055da7a5d2f64 in setargs (args=0x55da7b3da000 <error: Cannot access memory at address 0x55da7b3da000>, argv=0x55da7b3cb0e8) at /workspace/karas/uftrace/utils/utils.c:730
#1  0x000055da7a5d31a0 in parse_cmdline (cmd=0x55da7b3cae70 "'\n", argc=0x7ffe8b342ec0) at /workspace/karas/uftrace/utils/utils.c:802
#2  0x000055da7a583389 in parse_opt_file (argc=0x7ffe8b3440cc, argv=0x7ffe8b3440c0, filename=0x7ffe8b3430a0 "uftrace.data/default.opts", opts=0x7ffe8b344120)
    at /workspace/karas/uftrace/uftrace.c:904
#3  0x000055da7a583920 in apply_default_opts (argc=0x7ffe8b3440cc, argv=0x7ffe8b3440c0, opts=0x7ffe8b344120) at /workspace/karas/uftrace/uftrace.c:1038
#4  0x000055da7a583dad in main (argc=2, argv=0x7ffe8b344368) at /workspace/karas/uftrace/uftrace.c:1140
```

Thanks.

gy741:
If someone is interested,  can contribute to this problem.

gkarlos:
How can this be reproduced?

gy741:
@gkarlos Thank you for your interest.

Steps to Reproduce:
```
$ gcc -pg -o t-abc tests/s-abc.c
$ uftrace record t-abc
$ echo "'" > uftrace.data/default.opts
$ uftrace replay
segmentation fault ( (core dumped)
```

gkarlos:
@gy741 Thanks. I will have a look

honggyukim:
@gkarlos Thanks for your effort. The simpler way to reproduce is as follows:
```
$ echo "'" > opt

$ uftrace record --opt-file opt t-abc
Segmentation fault (core dumped)
```

[...]

PR #932, opened 2019-10-07
utils: Fix segfaults caused by setargs()

gkarlos:
This PR fixes a number of segfaults caused when parsing options from a file ( `--opt-file`). Relevant issue #914 .

Note that there are still some leaks related to option parsing but they are not particularly tied to `setargs()`. They are caused by `exit()`-ing when an executable cannot be opened. This probably needs a new issue opened.

Signed-off-by: George Karlos gkarloscodes@gmail.com

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -725,24 +725,30 @@ static int setargs(char *args, char **argv)
 
 		/* consider quotes and update argv */
 		if (*args == QUOTE) {
+			++args;
+			if (*args == QUOTE )
+				continue;
 			if (argv)
-				argv[count] = ++args;
-			while (*args != QUOTE)
+				argv[count] = args;
+			while (*args && *args != QUOTE)
 				++args;
-			if (argv)
+			if (argv && *args)
 				*args = ' ';
 		}
 		else if (*args == DQUOTE) {
+			++args;
+			if (*args == DQUOTE)
+				continue;
 			if (argv)
-				argv[count] = ++args;
-			while (*args != DQUOTE)
+				argv[count] = args;
+			while (*args && *args != DQUOTE)
 				++args;
-			if (argv)
+			if (argv && *args)
 				*args = ' ';
 		}
 		else if (*args == '#') {
 			/* ignore comment line */
-			while (*args != '\n' || *args == '\0')
+			while (*args != '\n' && *args != '\0')
 				++args;
 			continue;
 		}


cmds/record.c
```diff
@@ -1554,6 +1554,9 @@ static void check_binary(struct opts *opts)
 		EM_X86_64, EM_ARM, EM_AARCH64, EM_386
 	};
 
+if ( !(opts->exename))
+	pr_err_ns("Cannot open '(null)' : Invalid filename.\n");
+
```
namhyung: Do you really need this?  It already checks the number of arguments and will print an error message if it misses exename.

uftrace.c
```diff
@@ -912,7 +912,8 @@ static void parse_opt_file(int *argc, char ***argv, char *filename, struct opts
 		opts->idx = 0;
 	}
 
-	argp_parse(&file_argp, file_argc, file_argv,
+	if ( file_argv) /* workaround for when the file is empty */
+		argp_parse(&file_argp, file_argc, file_argv,
```
namhyung: We don't add a space after open parenthesis.  Also you'd better use a pair of curly braces when the body of if statement is multi-line.  Please indent below lines as well.

uftrace.c
```diff
@@ -1188,7 +1189,6 @@ int main(int argc, char *argv[])
 
 	if (opts.opt_file)
 		free_parsed_cmdline(argv - opts.idx);
-
```
namhyung: Unnecessary change.

utils/utils.c
```diff
@@ -712,56 +712,66 @@ char * json_quote(char *str, int *len)
...
+
 static int setargs(char *args, char **argv)
 {
 	int count = 0;
 
 	while (*args) {
-		/* ignore spaces */
-		if (isspace(*args)) {
+		/* ignore spaces and newlines */
+		if (isspace(*args) || isnewline(*args)) {
```
namhyung: AFAIK `isspace()` covers newlines too.

gkarlos: I believe you are right actually. I will revert that change

uftrace.c
gkarlos: This is also relevant when input is empty, i.e `""` or `"  "` or any number of spaces. So i guess i shall leave it out.

cmds/record.c
namhyung: OK, then please write the exact command line to reproduce in the commit message.

utils/utils.c
```diff
@@ -725,24 +728,28 @@ static int setargs(char *args, char **argv)
 
 		/* consider quotes and update argv */
 		if (*args == QUOTE) {
+			if (argv && *(++args) == QUOTE ) /* skip '' */
+				continue;
```
namhyung: Does it need to check `argv` here?  And I don't want to make the if statement complex.  Also I think the comment can be removed.  What about this?

```c
if (*args == QUOTE) {
   args++;
   if (*args == QUOTE)
      continue;
   ...
```

utils/utils.c
```diff
@@ -725,24 +728,28 @@ static int setargs(char *args, char **argv)
...
+			while (*args && *args != DQUOTE) /* advance until " or end */
 				++args;
-			if (argv)
+			if (argv && *args)
 				*args = ' ';
 		}
 		else if (*args == '#') {
 			/* ignore comment line */
-			while (*args != '\n' || *args == '\0')
+			while (*args && !isnewline(*args) && (*args != '\0'))
```
namhyung: This can be `while (*args != '\n' && *args != '\0')`

gkarlos:
@namhyung I think so. This fixes invalid filenames involving quotes that were causing segfaults before. e.g `echo "'" > opt` , `echo -n "'" > opt`, `echo "'''" > opt` etc.

PR #942, opened 2019-10-13
plthook: Remove dead code in find_got

gy741:
Hello.

I think, the if conditional statement in libmcount/plthook.c on line 246 is useless.

If `plt_found` is not true, it is processed on line 202.

So, The if conditional statement is always true.

what do you think?

https://github.com/namhyung/uftrace/blob/519ecef22613cca6260e8a2b52b5cf8a2903b087/libmcount/plthook.c#L202-L249

Thanks.

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -243,27 +243,25 @@ static int find_got(struct uftrace_elf_data *elf,
 	mcount_arch_plthook_setup(pd, elf);
 	list_add_tail(&pd->list, &plthook_modules);
 
-	if (plt_found) {
-		if (plthook_resolver_addr == 0)
-			plthook_resolver_addr = pd->pltgot_ptr[2];
+	if (plthook_resolver_addr == 0)
+		plthook_resolver_addr = pd->pltgot_ptr[2];
 
-		/*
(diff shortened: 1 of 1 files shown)


honggyukim:
It becomes true after the following commit was applied.
- 5d81130 plthook: Remove mcount_arch_undo_bindnow()

It partially contains the following diff.
```diff
diff --git a/libmcount/plthook.c b/libmcount/plthook.c
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -200,9 +186,9 @@
-       if (!pltgot_addr && !plt_found) {
+       if (!plt_found) {
                pd = mcount_arch_hook_no_plt(elf, modname, offset);
                if (pd == NULL)
-                       pr_dbg2("no PLTGOT nor BIND-NOW.. ignoring...\n");
+                       pr_dbg2("no PLTGOT found.. ignoring...\n");
                else
                        list_add_tail(&pd->list, &plthook_modules);

                return 0;
        }

```
It looks that the if statement is not needed at the moment. I'm not sure yet in the long term to maintain the code.

namhyung:
Yeah, it looks not needed anymore.  Please send a PR.

Issue #939, opened 2019-10-13
stack buffer overflow in json_quote()


rls1004:
`fread()` reads the contents of the `/proc/self/cmdline` file by the size of `fha->buf` and stores it in `buf`.

https://github.com/namhyung/uftrace/blob/322185a09c85e280733949db661ce975253aef7f/cmds/info.c#L215-L232

However, `fread()` does not append null to the end of the string read, so the string in `buf` can be recognized as a string longer than 4096, which is the size of `fha->buf`.

`strchr()` used in `json_quote()` recognizes `p`(which points to buf) as a string longer than 4096 and reads unintended strings beyond the range of `buf`.

https://github.com/namhyung/uftrace/blob/322185a09c85e280733949db661ce975253aef7f/utils/utils.c#L693-L696

Here is the crash info : 
```
$ uftrace ./test aaaaaaaaaa...<4096 or more>
...
=================================================================
==49856==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc3eadca30 at pc 0x7f5a9447f30e bp 0x7ffc3eadb790 sp 0x7ffc3eadaf38
READ of size 4103 at 0x7ffc3eadca30 thread T0
    #0 0x7f5a9447f30d  (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x5b30d)
    #1 0x55d44404691a in json_quote /home/rls1004/uftrace/utils/utils.c:693
    #2 0x55d443f96bc5 in fill_cmdline /home/rls1004/uftrace/cmds/info.c:231
    #3 0x55d443f9c6bc in fill_uftrace_info /home/rls1004/uftrace/cmds/info.c:906
    #4 0x55d443fccbd6 in fill_file_header /home/rls1004/uftrace/cmds/record.c:439
    #5 0x55d443fd7e7c in finish_writers /home/rls1004/uftrace/cmds/record.c:1922
    #6 0x55d443fd91d3 in do_main_loop /home/rls1004/uftrace/cmds/record.c:2054
    #7 0x55d443fd9e16 in command_record /home/rls1004/uftrace/cmds/record.c:2171
    #8 0x55d443fbb144 in command_live /home/rls1004/uftrace/cmds/live.c:138
    #9 0x55d443f8752c in main /home/rls1004/uftrace/uftrace.c:1165
    #10 0x7f5a92de1b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
[...]

PR #946, opened 2019-10-13
info: Fix stack buffer overflow when reading cmdline

rls1004:
Fixed the stack overflow by inserting null byte at the end of the string.

Fixed: #939


rls1004:
I modified the pull request.

PR #961, opened 2019-10-16
misc: Add .gdbinit for debugging

gonapps:
This .gdbinit contains:
set follow-fork-mode child
set follow-exec-mode new

You don't have to set fork-mode and exec-mode everytime when using gdb
if your ~/.gdbinit is like following:
set auto-load safe-path /

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,0 +1,2 @@
+set follow-fork-mode child
+set follow-exec-mode new


gonapps:
I thought it would be convenient to people who want to load .gdbinit in a current directory.
If you don't agree with this, you can also close this pull request.

Issue #952, opened 2019-10-15
incorrectly checks fread() retval


gy741:
Hello.

The return value of the free() function is "size_t", which cannot be a negative value.

Therefore, the following conditions always fail.

https://github.com/namhyung/uftrace/blob/2d6c907dedaecb37fb73b9a42f97020842c5b6d4/cmds/info.c#L219-L224

https://github.com/namhyung/uftrace/blob/322185a09c85e280733949db661ce975253aef7f/utils/symbol.c#L1237-L1241

fix code:

``` diff
diff --git a/cmds/info.c b/cmds/info.c
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -219,7 +219,7 @@ static int fill_cmdline(void *arg)
	ret = fread(buf, 1, sizeof(fha->buf), fp);
	fclose(fp);

-	if (ret < 0)
+	if (ret != sizeof(fha->buf))
		return ret;

	/* cmdline separated by NUL character - convert to space */

diff --git a/cmds/info.c b/cmds/info.c
--- a/utils/symbol.c
+++ b/ utils/symbol.c
@@ -1237,7 +1237,7 @@ int save_kernel_symbol(char *dirname)
	while ((len = fread(buf, 1, sizeof(buf), ifp)) > 0)
		fwrite(buf, 1, len, ofp);

-	if (len < 0)
+	if (len != sizeof(buf))
		ret = len;

	fclose(ifp);

```

Ref:    https://github.com/torvalds/linux/commit/f7679dabfaf69840b000d238a020cee7157aca17
         https://hg.mozilla.org/mozilla-central/rev/ed19305f89e6

namhyung:
While the return type is unsigned, we save it into a signed type so it can be negative.

gy741:
@namhyung  Thanks for the good code review.

There is a way to change the data type. `int` -> `size_t`
Maybe you don't want it.

I'll close the problem.

namhyung:
Hmm.. it was confusing, you said `free()` function but it was actually `fread()`, right?  I checked again that `read()` (not started by "f") returns ssize_t while `fread()` returns size_t.  So I think your change is correct.

gy741:
Oops, sorry. 
Yes. It's about `fread()`.
I will send PR.

PR #966, opened 2019-10-17
uftrace: Fix fread() error handling

gy741:
This isn't really accurate right. fread() doesn't always
return 0 in error. It could return < number of elements
and set errno.

Fixed: #952

### DIFF
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -209,7 +209,8 @@ static int fill_cmdline(void *arg)
 	struct fill_handler_arg *fha = arg;
 	char *buf = fha->buf;
 	FILE *fp;
-	int ret, i;
+	size_t ret;
+	int i;
 	char *p;
 
 	fp = fopen("/proc/self/cmdline", "r");
@@ -219,7 +220,7 @@ static int fill_cmdline(void *arg)
 	ret = fread(buf, 1, sizeof(fha->buf), fp);
 	fclose(fp);
 
-	if (ret < 0)
+	if (ret != sizeof(fha->buf))
 		return ret;
 
 	/* cmdline separated by NUL character - convert to space */
--- a/utils/symbol.c
+++ b/utils/symbol.c
@@ -1237,7 +1237,7 @@ int save_kernel_symbol(char *dirname)
 	while ((len = fread(buf, 1, sizeof(buf), ifp)) > 0)
 		fwrite(buf, 1, len, ofp);
 
-	if (len < 0)
+	if (len != sizeof(buf))
 		ret = len;
 
 	fclose(ifp);


cmds/info.c
```diff
@@ -219,7 +220,7 @@ static int fill_cmdline(void *arg)
 	ret = fread(buf, 1, sizeof(fha->buf), fp);
 	fclose(fp);
 
-	if (ret < 0)
+	if (ret != sizeof(fha->buf))
```
honggyukim: `ret` is the string length of `buf`, but `sizeof(fha->buf)` is always `4096` so this condition is always true.

utils/symbol.c
```diff
@@ -1237,7 +1237,7 @@ int save_kernel_symbol(char *dirname)
 	while ((len = fread(buf, 1, sizeof(buf), ifp)) > 0)
 		fwrite(buf, 1, len, ofp);
 
-	if (len < 0)
+	if (len != sizeof(buf))
```
honggyukim: Ditto.

namhyung:
Ooops, I should've checked it!  @gy741 please run the tests after change something.

honggyukim:
In addition, why don't you write the commit message as follows?
```
The following commit is a regression that breaks every normal recording.

  5cffda2 uftrace: Fix fread() error handling

Unfortunately, this change breaks normal recording as follows:

  $ uftrace t-abc
  uftrace: /home/honggyu/work/uftrace/utils/data-file.c:493:open_info_file

In uftrace.data/info file, cmdline line is not found at all.

The man page of fread describes as follows:

  fread() does not distinguish between end-of-file and error, and callers
  must use feof(3) and ferror(3) to determine which occurred.

In cmds/info.c checking ret != sizeof(fha->buf) doesn't always mean that
an error occured.  If the buffer is large enough, return value of fread()
can be smaller than the size of the buffer.

In utils/symbol.c when while loop ends, len will be always zero.
ferror() can be used to check if an error actually occured in this case.

There are compile warnings as well.

  cmds/info.c: In function ‘fill_cmdline’:
  cmds/info.c:230:25: warning: comparison of integer expressions of different signedness:
                      ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
    230 |  for (i = 0, p = buf; i < ret; i++, p++) {
        |                         ^
  cmds/info.c:237:22: warning: passing argument 2 of ‘json_quote’ from
                      incompatible pointer type [-Wincompatible-pointer-types]
    237 |  p = json_quote(buf, &ret);
        |                      ^~~~
        |                      |
        |                      size_t * {aka long unsigned int *}
  In file included from /home/gonapps/workspace/uftrace/utils/symbol.h:15,
                   from /home/gonapps/workspace/uftrace/uftrace.h:11,
                   from /home/gonapps/workspace/uftrace/cmds/info.c:20:
  utils/utils.h:337:8: note: expected ‘int *’ but argument is of type ‘size_t *’
                       {aka ‘long unsigned int *’}
    337 | char * json_quote(char *str, int *len);
        |        ^~~~~~~~~~
  cmds/info.c:241:30: warning: comparison of integer expressions of different signedness:
                      ‘ssize_t’ {aka ‘long int’} and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]

The warnings can simply be suppressed by changing ret variable to int type.

Fixed: #952

```

honggyukim:
I should have checked it as well.

gy741:
@namhyung My mistake.
I'm very sorry. Since we don't test in a clean environment, 
I seem to have missed the problem.
I'll be more careful next time.

namhyung:
That's ok.  I should run test before accepting patches.

PR #973, opened 2019-10-17
utils: Add posix.h for systems which don't support posix shm

binkoni:

### DIFF
--- a/check-deps/__have_posix_shm.c
+++ b/check-deps/__have_posix_shm.c
@@ -1,0 +1,10 @@
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+int main(void)
+{
+	shm_open("/uftrace_test", O_RDWR, 0600);
+	shm_unlink("/uftrace_test");
+	return 0;
+}
--- a/utils/posix.h
+++ b/utils/posix.h
@@ -1,0 +1,52 @@
+#ifndef POSIX_H
+#define POSIX_H
+
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <errno.h>
+
+#ifdef HAVE_POSIX_SHM
+# define UFTRACE_SHMDIR_NAME "/dev/shm"
+#else
+# define UFTRACE_SHMDIR_NAME "uftrace.shm"
+#endif
+
+#ifndef HAVE_POSIX_SHM
+static inline int shm_open(const char *file, int oflag, mode_t mode)
+{
+	int ret;
+	char* path;
+
+	if (mkdir(UFTRACE_SHMDIR_NAME, 700) && errno != EEXIST)
+		return -1;
+
+	xasprintf(&path, "%s%s", UFTRACE_SHMDIR_NAME, file);
+	ret = open(path, oflag, mode);
+	free(path);
+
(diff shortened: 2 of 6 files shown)


utils/posix.h
```diff
@@ -0,0 +1,47 @@
...
+static inline int shm_open(const char *file, int oflag, mode_t mode)
+{
+	int ret;
+	char* path;
+	if(mkdir(UFTRACE_SHMDIR_NAME, S_IRWXU | S_IRWXG | S_IRWXO) && errno != EEXIST)
+		return -1;
+	const size_t path_size = sizeof(UFTRACE_SHMDIR_NAME) - 1 + strlen(file) + 1;
+	path = calloc(path_size, sizeof(char));
+	strncat(path, UFTRACE_SHMDIR_NAME, path_size - strlen(path) - 1);
+	strncat(path, file, path_size - strlen(path) - 1);
```
namhyung: I prefer `xasprintf()` for this kind of thing.

utils/posix.h
```diff
@@ -0,0 +1,47 @@
...
+	ret = open(path, oflag, mode);
```
namhyung: And I think we should fallback to other shared memory (if exists) instead of a regular file.  Maybe Sys-V shmem or ashmem (on Android) or just a regular file in a tmpfs.

gonapps: There is a problem.
Sys-V shmem and ashmem accepts size of the shared memory as a argument.

check-deps/__have_posix_shm.c
```diff
@@ -0,0 +1,10 @@
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+int main(void)
+{
+	shm_open("/uftrace_test", O_RDWR, 0600);
+	shm_unlink("/uftrace_test");
+	return 0;
+}
```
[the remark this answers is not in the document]
gonapps: Should I remove this file and just check when using posix_open() and posix_unlink() with ifdef ANDROID?

utils/posix.h
```diff
@@ -0,0 +1,47 @@
...
+#endif
+
+#ifndef HAVE_POSIX_SHM
+static inline int shm_open(const char *file, int oflag, mode_t mode)
+{
+	int ret;
+	char* path;
+	if(mkdir(UFTRACE_SHMDIR_NAME, S_IRWXU | S_IRWXG | S_IRWXO) && errno != EEXIST)
+		return -1;
+	const size_t path_size = sizeof(UFTRACE_SHMDIR_NAME) - 1 + strlen(file) + 1;
```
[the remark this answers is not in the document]
gonapps: Fixed

utils/posix.h
gonapps: Yeah you're right

gonapps:
@namhyung 
from https://github.com/SHallimani/Android-NDK-Examples
>>
There is no way to 'open' an existing region by name. An ashmem i-node, and corresponding memory, is automatically reclaimed when the last file descriptor to it is closed. This is useful because it means that if your process dies due to a crash, the memory will be reclaimed by the kernel automatically. This is not the case with regular SysV shared memory (a crashing process just leaks the memory! Something unacceptable on an embedded system like Android).

PR #974, opened 2019-10-17
live: Use current directory when /tmp is failed

gonapps:
In some systems like a non-rooted android environment (termux)
You don't have a write access to /tmp directory

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -104,22 +104,30 @@ int command_live(int argc, char *argv[], struct opts *opts)
 	};
 	int ret;
 
+	tmp_dirname = template;
 	umask(022);
 	fd = mkstemp(template);
-	if (fd < 0)
-		pr_err("cannot create temp name");
+	if (fd < 0) {
+		if (errno != EPERM)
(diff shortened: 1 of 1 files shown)


cmds/live.c
```diff
@@ -104,15 +104,23 @@ int command_live(int argc, char *argv[], struct opts *opts)
...
+
+		fd = mkstemp(template + sizeof("/tmp/") - 1);
+
+		if (fd < 0)
+			pr_err("cannot create temp name");
+		tmp_dirname = template + sizeof("/tmp/") - 1;
+	}
 
 	close(fd);
 	unlink(template);
```
namhyung: You could either overwrite template or use tmp_dirname here (and below).

gonapps: Modified to tmp_dirname += sizeof("/tmp/") - 1;

gonapps:
@namhyung Updated the commit.

cmds/live.c
gonapps: Thanks. I've fixed now

PR #983, opened 2019-10-18
doc/ko: Minimize difference between report and dump

honggyukim:
Since there are a lot of common options between report and dump
commands, it'd be better to write the exactly the same description so
that one can easily see the difference.

This patch is to minimize the different lines between report and dump
man pages.  For same options, diff tool shows nothing with this change.

Before:
```
  $ diff -u uftrace-report.md uftrace-dump.md | wc -l
  361
```
After:
```
  $ diff -u uftrace-report.md uftrace-dump.md | wc -l
  277
```


honggyukim:
@namhyung I will merge this first so that it can be used as a reference.

Issue #993, opened 2019-10-19
heap buffer overflow in get_argspec_string()


rls1004:
https://github.com/namhyung/uftrace/blob/7817a96eb33f28e0fbb931301a493c633994af65/cmds/replay.c#L592-L603

The `memcmp()` function makes a 4-byte string comparison of str.
However, it can happen that the memory size requested to `xmalloc()` is less than 4 bytes.

```
$ uftrace -A printf@arg1/s ./fmt a
a
=================================================================
==109346==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000db5f2 at pc 0x7f177ba29f54 bp 0x7f
READ of size 4 at 0x6020000db5f2 thread T0
    #0 0x7f177ba29f53  (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xaff53)
    #1 0x55f7d7643561 in get_argspec_string /home/rls1004/uftrace/cmds/replay.c:602
    #2 0x55f7d7646917 in print_graph_rstack /home/rls1004/uftrace/cmds/replay.c:862
    #3 0x55f7d7648d32 in command_replay /home/rls1004/uftrace/cmds/replay.c:1187
    #4 0x55f7d763a3e3 in command_live /home/rls1004/uftrace/cmds/live.c:158
    #5 0x55f7d7606561 in main /home/rls1004/uftrace/uftrace.c:1166
    #6 0x7f177a337b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
    #7 0x55f7d75ffab9 in _start (/usr/local/bin/uftrace+0x46ab9)

0x6020000db5f2 is located 0 bytes to the right of 2-byte region [0x6020000db5f0,0x6020000db5f2)
allocated by thread T0 here:
    #0 0x7f177ba58b50 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb50)
    #1 0x55f7d764335c in get_argspec_string /home/rls1004/uftrace/cmds/replay.c:592
    #2 0x55f7d7646917 in print_graph_rstack /home/rls1004/uftrace/cmds/replay.c:862
    #3 0x55f7d7648d32 in command_replay /home/rls1004/uftrace/cmds/replay.c:1187
    #4 0x55f7d763a3e3 in command_live /home/rls1004/uftrace/cmds/live.c:158
[...]

PR #994, opened 2019-10-19
replay: Fix heap buffer overflow in get_argspec_string()

rls1004:
Fixed to check if string is 4 bytes long before comparing it to null_str.
Removed unnecessary code for newline.

Fixed: #993

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -580,26 +580,14 @@ void get_argspec_string(struct uftrace_task_reader *task,
 		if (spec->fmt == ARG_FMT_STR ||
 		    spec->fmt == ARG_FMT_STD_STRING) {
 			unsigned short slen;
-			unsigned short newline = 0;
-			char last_ch;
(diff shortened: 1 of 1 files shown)


cmds/replay.c
```diff
@@ -589,7 +589,7 @@ void get_argspec_string(struct uftrace_task_reader *task,
 			if (last_ch == '\n')
 				newline = 1;
 
-			str = xmalloc(slen + newline + 1);
+			str = xmalloc(slen + newline + sizeof(null_str));
```
namhyung: Hmm.. I don't like it.  Actually I found two other concerns.  First, checking newline character is not needed anymore since we use `print_escape_char()`.  And the `null_str` check is just for older data formats and I think you'd better checking `slen == 4` before calling `memcmp` instead.

rls1004: much better, I modified it.

Issue #996, opened 2019-10-19
error when using multiple command lines


rls1004:
steps to reproduce this error : 
```
$ uftrace ./fmt "
> "


uftrace: /home/rls1004/uftrace/utils/data-file.c:493:open_info_file
 ERROR: cannot read uftrace header info!
```

honggyukim:
Could you please explain more about it?

rls1004:
For example, in this case
```
$ uftrace record ./test "a
b"
```
The recorded cmdline is shown below.
```
...
cmdline:uftrace record ./fmt a
b
cpuinfo:lines=2
...
```

The `read_cmdline()` read it using the `fgets()`, so reading stops after a newline.

https://github.com/namhyung/uftrace/blob/7817a96eb33f28e0fbb931301a493c633994af65/cmds/info.c#L249-L257

The `read_cpuinfo()` reads the next line as well.
An error occurs because the string "cpuinfo:" cannot be read.

PR #998, opened 2019-10-19
info: Fix error when reading multiple cmdline

rls1004:
Fixed `\n` to be replaced with `space`.

Fixed: #996

### DIFF
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -227,7 +227,7 @@ static int fill_cmdline(void *arg)
 
 	/* cmdline separated by NUL character - convert to space */
 	for (i = 0, p = buf; i < ret; i++, p++) {
-		if (*p == '\0')
+		if (*p == '\0' || *p == '\n')
 			*p = ' ';
 	}
 


cmds/info.c
```diff
@@ -229,6 +229,10 @@ static int fill_cmdline(void *arg)
 	for (i = 0, p = buf; i < ret; i++, p++) {
 		if (*p == '\0')
 			*p = ' ';
+		if (*p == '\n') {
+			memcpy(p, p+1, ret-i);
```
honggyukim: I didn't take a look for this change yet, but put a single white space on the left and right of "+" and "-" signs.

rls1004: I modified it.

cmds/info.c
```diff
@@ -229,6 +229,10 @@ static int fill_cmdline(void *arg)
 	for (i = 0, p = buf; i < ret; i++, p++) {
 		if (*p == '\0')
 			*p = ' ';
+		if (*p == '\n') {
+			memcpy(p, p + 1, ret - i);
+			p--;
+		}
```
namhyung: Why not simply changing it as a normal whitespace?

rls1004: It is my mistake.
```
$ uftrace ./test "a\
b"
```
I wrongly thought it should be handled the same way as above.
So I tried to remove the '\n' and join the string.
```
$ uftrace ./test "a
b"
```
However, in this case, the strings are separated.
I fixed it to be handled the same way as white space.

PR #1001, opened 2019-10-19
dynamic: support relative conditional branches and relative jumps 

AnsBal:
Conditional branches and relative jumps in the prologue appears in 8% of uftrace functions. Thus there is a need to support them. 
The idea is to use a jcc8 to jump (when condition is met) to a trampoline which jumps to the target of the original instruction. The offset of the new jcc is patched later when the code is saved in the out of line buffer.

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -643,3 +643,71 @@ void mcount_arch_dynamic_recover(struct mcount_dynamic_info *mdi,
 		free(badsym);
 	}
 }
+
+int mcount_arch_branch_table_size(struct mcount_disasm_info *info)
+{
+	struct cond_branch_info *jcc_info;
+	unsigned long jcc_target;
+	int count = 0;
+	int i;
+
+	for (i = 0; i < info->nr_branch; i++) {
+		jcc_info = &info->branch_info[i];
+		jcc_target = jcc_info->branch_target;
+
+		/* no need to allocate entry for jcc that jump directly to prologue */
+		if (info->addr <= jcc_info->branch_target &&
+			jcc_target < info->addr + info->orig_size) {
+			continue;
+		}
+		count++;
+	}
+	return count * ARCH_BRANCH_ENTRY_SIZE;
+}
+
+void mcount_arch_patch_branch(struct mcount_disasm_info *info,
+					struct mcount_orig_insn *orig)
+{
+	/*
+	 * The first entry in the table starts right after the out-of-line 
+	 * execution buffer.
+	 */
+	uint64_t entry_offset = orig->insn_size;
+	uint8_t trampoline[ARCH_TRAMPOLINE_SIZE] = { 0xff, 0x25, };
+	unsigned long jcc_target;
+	unsigned long jcc_index;
+	struct cond_branch_info *jcc_info;
+	uint32_t disp;
+	int i;
+
+	for (i = 0; i < info->nr_branch; i++) {
+		jcc_info = &info->branch_info[i];
+		jcc_target = jcc_info->branch_target;
+		jcc_index = jcc_info->insn_index;
+
+		/* leave the original disp of jcc that target the prologue as it is */
+		if (info->addr <= jcc_info->branch_target &&
+			jcc_target < info->addr + info->orig_size) {
+
+			info->insns[jcc_index + 1] = jcc_target - (jcc_info->insn_addr + jcc_info->insn_size);
+			continue;
+		}
+
+		/* setup the branch entry trampoline */
+		memcpy(trampoline + JMP_INSN_SIZE, &jcc_target, sizeof(jcc_target));
+
+		/* write the entry to the branch table */
+		memcpy(orig->insn + entry_offset, trampoline, sizeof(trampoline));
+
+		/* previously, all jcc32 are downgraded to jcc8 */
+		disp = entry_offset - (jcc_index + ARCH_JCC8_SIZE);
+		if (disp > SCHAR_MAX) { /* should not happen */
+			pr_err("target is not in reach"); 
+		}
+
+		/* patch jcc displacement to target correspending entry in the table */
+		info->insns[jcc_index + 1] = disp;
+
+		entry_offset += ARCH_BRANCH_ENTRY_SIZE;
+	}
+}
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -79,6 +79,94 @@ static int opnd_reg(int capstone_reg)
 }
 
 /*
+ * Handle relative conditional jumps and relative unconditional jumps.
+ *
+ * This function relocates jcc8 and jcc32 instructions by replacing them with a jcc8 
+ * that has a null offset. The offset will be patched later when the code is saved
+ * in the of line execution buffer. The new jcc8 will bounce (if condition is met)
+ * on a trampoline that jumps to the target of the orginal instruction.
+ *
+ * The relocation of jmp8 and jmp32 is achieved by replacing them with an absolute 
+ * indirect jump to the target.
+ *
+ */
+static int handle_rel_jmp(cs_insn *insn, uint8_t insns[], struct mcount_disasm_info *info)
+{
+	cs_x86 *x86 = &insn->detail->x86;
+	uint8_t relocated_insn[ARCH_TRAMPOLINE_SIZE] = { 0xff, 0x25,};
+	uint8_t opcode = insn->bytes[0];
+	uint64_t target;
+	struct cond_branch_info *cbi;
+
+#define JMP8_OPCODE 0xEB
+#define JMP32_OPCODE 0xE9
+#define IND_JMP_SIZE 6
+#define OP   0
+#define OFS	  1
+	cs_x86_op *opnd = &x86->operands[0];
+
+	if (x86->op_count != 1 || opnd->type != X86_OP_IMM)
+		goto out;
+
+	if ((opcode == JMP8_OPCODE || opcode == JMP32_OPCODE)) {
+		if (strcmp(insn->mnemonic, "jmp") != 0)
+			goto out;
+
+		target = opnd->imm;
+
+		memcpy(insns, relocated_insn, IND_JMP_SIZE);
+		memcpy(insns + IND_JMP_SIZE, &target, sizeof(target));
+
+		info->modified = true;
+		/* 
+		 * If this jump is the last insn in the prologue, we can ignore
+		 * the one in patch_normal_func()
+		 */
+		if (info->orig_size + insn->size >= ARCH_JMP32_SIZE)
+			info->has_jump = true;
+		return IND_JMP_SIZE + sizeof(target);
+
+	}/* Jump relative 8 if condition is met (except JCXZ, JECXZ and JRCXZ) */
+	else if((opcode & 0xF0) == 0x70) {
+
+		cbi = &info->branch_info[info->nr_branch++];
+		cbi->insn_index = info->copy_size;
+		cbi->branch_target = opnd->imm;
+		cbi->insn_addr = insn->address;
+		cbi->insn_size = insn->size;
+
+		relocated_insn[OP] = opcode;
+		relocated_insn[OFS] = 0x00;
+
+		memcpy(insns, (void *)relocated_insn, ARCH_JCC8_SIZE);
+
+		info->modified = true;
+		return ARCH_JCC8_SIZE;
+
+	}/* Jump relative 32 if condition is met */
+	else if(opcode == 0x0F && (insn->bytes[1] & 0xF0) == 0x80) {
+
+		cbi = &info->branch_info[info->nr_branch++];
+		cbi->insn_index = info->copy_size;
+		cbi->branch_target = opnd->imm;
+		cbi->insn_addr = insn->address;
+		cbi->insn_size = insn->size;
+
+		/* We use the equivalent jcc8 of the original jcc32 */
+		relocated_insn[OP] = insn->bytes[1] - 0x10;
+		relocated_insn[OFS] = 0x00;
+
+		memcpy(insns, (void *)relocated_insn, ARCH_JCC8_SIZE);
+		
+		info->modified = true;
+		return ARCH_JCC8_SIZE;
+	}
+
+out:
+	return -1;
+}
+
+/*
  *  handle PIC code.
  *  for currently, this function targeted specific type of instruction.
  *
@@ -202,6 +290,11 @@ static int manipulate_insns(cs_insn *insn, uint8_t insns[], int* fail_reason,
 	pr_dbg3("Manipulate instructions having PC-relative addressing.\n");
 
 	switch (*fail_reason) {
+	case INSTRUMENT_FAIL_RELJMP:
+		res = handle_rel_jmp(insn, insns, info);
+		if (res > 0)
+			*fail_reason = 0;
+		break;
 	case INSTRUMENT_FAIL_PICCODE:
 		res = handle_pic(insn, insns, info);
 		if (res > 0)
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -78,7 +78,9 @@ struct mcount_orig_insn *mcount_save_code(struct mcount_disasm_info *info,
 		/* it needs to save original instructions as well */
 		int orig_size = ALIGN(info->orig_size, 16);
 		int copy_size = ALIGN(info->copy_size + jmp_size, 16);
-		patch_size = ALIGN(copy_size + orig_size, 32);
+		int table_size = mcount_arch_branch_table_size(info);
+
+		patch_size = ALIGN(copy_size + orig_size + table_size, 32);
 	}
 	else {
 		patch_size = ALIGN(info->copy_size + jmp_size, 32);
@@ -108,6 +110,8 @@ struct mcount_orig_insn *mcount_save_code(struct mcount_disasm_info *info,
 		/* save original instructions before modification */
 		orig->orig = orig->insn + patch_size - ALIGN(info->orig_size, 16);
 		memcpy(orig->orig, (void *)info->addr, info->orig_size);
+
+		mcount_arch_patch_branch(info, orig);
 	}
 
 	memcpy(orig->insn, info->insns, info->copy_size);
@@ -180,6 +184,15 @@ __weak void mcount_disasm_init(struct mcount_disasm_engine *disasm)
 }
 
 __weak void mcount_disasm_finish(struct mcount_disasm_engine *disasm)
+{
+}
+
+__weak int mcount_arch_branch_table_size(struct mcount_disasm_info *info)
+{
+	return 0;
+}
+
+__weak void mcount_arch_patch_branch(struct mcount_disasm_info *info, struct mcount_orig_insn *orig)
 {
 }
 
--- a/libmcount/internal.h
+++ b/libmcount/internal.h
@@ -415,6 +415,13 @@ struct mcount_disasm_engine {
 #define INSTRUMENT_FAILED                       -1
 #define INSTRUMENT_SKIPPED                      -2
 
+/* 
+ * Supposing the size of smallest conditional branch is 2 byte.
+ * We can replace, at most, 3 of them by the instrumentation
+ * instruction.
+ */
+#define MAX_COND_BRANCH 3
+
 int mcount_dynamic_update(struct symtabs *symtabs, char *patch_funcs,
 			  enum uftrace_pattern_type ptype,
 			  struct mcount_disasm_engine *disasm);
@@ -426,6 +433,15 @@ struct mcount_orig_insn {
 	void			*insn;
 	int			orig_size;
 	int			insn_size;
+};
+
+struct cond_branch_info {
+	/* where the insn starts in the out-of-line exec buffer*/
+	unsigned long insn_index;
+	/* the original target address of the branch */
+	unsigned long branch_target;
+	unsigned long insn_addr;
+	unsigned long insn_size;
 };
 
 /*
@@ -446,6 +462,8 @@ struct mcount_disasm_info {
 	int			copy_size;
 	bool			modified;
 	bool			has_jump;
+	uint8_t			nr_branch;
+	struct cond_branch_info branch_info[MAX_COND_BRANCH];
 };
 
 struct mcount_orig_insn *mcount_save_code(struct mcount_disasm_info *info,
@@ -463,6 +481,9 @@ int mcount_patch_func(struct mcount_dynamic_info *mdi, struct sym *sym,
 
 void mcount_disasm_init(struct mcount_disasm_engine *disasm);
 void mcount_disasm_finish(struct mcount_disasm_engine *disasm);
+
+int mcount_arch_branch_table_size(struct mcount_disasm_info *info);
+void mcount_arch_patch_branch(struct mcount_disasm_info *info, struct mcount_orig_insn *orig);
 
 struct dynamic_bad_symbol {
 	struct list_head	list;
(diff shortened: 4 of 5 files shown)


arch/x86_64/dynamic.S
```diff
@@ -42,6 +42,7 @@
 
 GLOBAL(__dentry__)
 	.cfi_startproc
+	pushf
```
namhyung: I'm afraid of 16-byte stack alignment

AnsBal:
Correct me if i'm wrong, but I don't think jcc can be handled the same way as a call. Conditional jumps don't always branch, which means that the next instruction after them may be executed. That's why the trampoline was inserted at the end of the out of line code. For jumps, I can see the light way of handling them.

arch/x86_64/dynamic.S
AnsBal: Can you please share more details about your worry.

namhyung: x86 ABI requires 16 byte alignment for the function entry.  So `mcount` and `__dentry__` in libmcount take care of the alignment when saving registers.

AnsBal: I thought that the line dynamic.S:67 (andq $0xfffffffffffffff0, %rsp) is taking care of that.

namhyung: Oh, you're right.  This part was changed several times so I forgot the history, sorry! :)

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,78 @@ static int opnd_reg(int capstone_reg)
...
+		memcpy(insns, (void *)relocated_insn, JMP_REG_SIZE);
+		memcpy(insns + JMP_REG_SIZE, &target, sizeof(target));
+		return JMP_REG_SIZE + sizeof(target);
+
+	}/* Jump relative 8 if condition is met (except JCXZ, JECXZ and JRCXZ) */
+	else if((opcode & 0xF0) == 0x70){
+
+		info->cond_info[info->nbr_cond_branch].insn_index = info->copy_size;
+		info->cond_info[info->nbr_cond_branch].jmp_addr =  opnd->imm;
+		info->nbr_cond_branch++;
```
namhyung: I prefer having a local variable like below:

```c
   struct cond_branch_info *cbi = &info->cond_info[info->nbr_cond_branch++];
```

arch/x86_64/mcount-insn.c
```diff
@@ -367,6 +446,11 @@ int disasm_check_insns(struct mcount_disasm_engine *disasm,
 			break;
 	}
 
+	/* 
+	 * instructions in prologue could jump back to it, the func should be 
+	 * unsupported in this case.
+	 */
+	i = 0;
```
namhyung: This can be a separate fix.  Hmm.. is it really a problem?

libmcount/dynamic.c
```diff
@@ -26,6 +26,8 @@ static struct mcount_dynamic_stats {
 
 #define PAGE_SIZE   4096
 #define CODE_CHUNK  (PAGE_SIZE * 8)
+#define TRAMPOLINE_SIZE 14
+#define JMP_INSN_SIZE   6
```
namhyung: The word trampoline is too generic and we use it for other places too.  Please be more specific.  Also this file is arch-agnostic please move arch-specific code under arch directory.  You might want to add some weak functions with `mcount_arch_` prefix.

libmcount/dynamic.c
```diff
@@ -110,6 +115,18 @@ struct mcount_orig_insn *mcount_save_code(struct mcount_disasm_info *info,
...
 	}
 
+	index = info->copy_size + jmp_size;
+	/* patch conditional jumps instructions */
+	for(i = 0; i < info->nbr_cond_branch; i++){
+		memcpy(cond_branch_trampoline + JMP_INSN_SIZE, &info->cond_info[i].jmp_addr, sizeof(info->cond_info[i].jmp_addr));
+		memcpy(orig->insn + index, cond_branch_trampoline, sizeof(cond_branch_trampoline));
+
+		uint8_t target = (unsigned long)info->insns + index - ((unsigned long)info->insns + info->cond_info[i].insn_index + 2);
+		info->insns[info->cond_info[i].insn_index + 1] = target;
```
namhyung: Please try to follow the coding style.  The lines are too long, please consider using a local variable.  Also it seems not possible that target is great than 128 but it would be better to add a check to verify that.

libmcount/internal.h
```diff
@@ -415,6 +415,8 @@ struct mcount_disasm_engine {
 #define INSTRUMENT_FAILED                       -1
 #define INSTRUMENT_SKIPPED                      -2
 
+#define MAX_COND_BRANCH 3
```
namhyung: Please add a comment why it cannot be more than 3.

libmcount/internal.h
```diff
@@ -428,6 +430,11 @@ struct mcount_orig_insn {
 	int			insn_size;
 };
 
+struct cond_branch_info{
```
namhyung: Please add a space before "{".

libmcount/internal.h
```diff
@@ -440,9 +447,11 @@ struct mcount_disasm_info {
 	struct sym		*sym;
 	unsigned long		addr;
 	unsigned char   	insns[64];
+	struct cond_branch_info cond_info[MAX_COND_BRANCH];
```
namhyung: I think "branch_info" is slightly better.

libmcount/internal.h
```diff
@@ -440,9 +447,11 @@ struct mcount_disasm_info {
...
 	int			orig_size;
 	int			copy_size;
 	bool			modified;
+	uint8_t		nbr_cond_branch;
```
namhyung: Please fix the indentation.  And we use "nr" for numbers like in Linux kernel, or "num" looks fine.  Also how about calling it "nr_branch" according to the suggestion above.

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,78 @@ static int opnd_reg(int capstone_reg)
...
+
+	if ((opcode == JMP8_OPCODE || opcode == JMP32_OPCODE)) {
+		if (strcmp(insn->mnemonic, "jmp") != 0)
+			goto out;
+
+		target = opnd->imm;
+
+		memcpy(insns, (void *)relocated_insn, JMP_REG_SIZE);
+		memcpy(insns + JMP_REG_SIZE, &target, sizeof(target));
+		return JMP_REG_SIZE + sizeof(target);
```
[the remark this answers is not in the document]
AnsBal: Exactly. That's why i was set to 0 in mcount-insn.c:453. To disallow jumps that are in the prologue and that jumps to it.

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,78 @@ static int opnd_reg(int capstone_reg)
...
+ *
+ * The relocation of jmp8 and jmp32 is similair to jcc.
+ *
+ * TODO: relocate relative jumps that branch to the prologue by jumping to the 
+ * 		 relocated instruction instead.
+ */
+static int handle_rel_jmp(cs_insn *insn, uint8_t insns[], struct mcount_disasm_info *info)
+{
+	cs_x86 *x86 = &insn->detail->x86;
+	uint8_t relocated_insn[15] = { 0xff, 0x25,};
```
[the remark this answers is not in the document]
AnsBal: Because it's the size of the longest insn in x86. It could be replaced with 14 (the max size used by the function to patch jumps).

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,78 @@ static int opnd_reg(int capstone_reg)
...
+	uint8_t opcode = insn->bytes[0];
+	uint64_t target;
+
+#define JMP8_OPCODE 0xEB
+#define JMP32_OPCODE 0xE9
+#define JCC8_SIZE   2
+#define JMP_REG_SIZE 6
```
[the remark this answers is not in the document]
AnsBal: Ohh it's not a jump near and absolute indirect. IND_JMP_SIZE is a right name.

arch/x86_64/mcount-insn.c
AnsBal: It's only a problem if branches in the prologue are relocated. I tried it with a function that has a 2 byte jmp 0xfc (2 bytes backward) in the prologue and the program crashed because the target insn of the jump was replaced by the instrumentation code.

libmcount/internal.h
AnsBal: Since the size of the smallest jcc is 2 bytes, at most we can replace 3 of them by the call in the trampoline. This may be true for x86, maybe not for other archs. Knowing that this file is also arch-agnostic, what is the recommended value to use ?

libmcount/internal.h
AnsBal: For the struct tag or the variable ?

libmcount/internal.h
namhyung: I think 3 is ok.

libmcount/internal.h
namhyung: The variable.

arch/x86_64/mcount-dynamic.c
```diff
@@ -633,3 +634,37 @@ void mcount_arch_dynamic_recover(struct mcount_dynamic_info *mdi,
...
+{
+	uint8_t trampoline[14] = { 0xff, 0x25, };
+	uint64_t target = orig->insn_size;
+	unsigned long jmp_target;
+	unsigned long jcc_index;
+	uint32_t disp;
+	int i;
+
+	/* patch conditional jumps instructions */
+	for(i = 0; i < info->nr_branch; i++){
```
namhyung: Please add a whitespace after `for` and opening brace.

arch/x86_64/mcount-dynamic.c
```diff
@@ -633,3 +634,37 @@ void mcount_arch_dynamic_recover(struct mcount_dynamic_info *mdi,
...
+		jmp_target = info->branch_info[i].branch_target;
+		jcc_index = info->branch_info[i].insn_index;
+
+		memcpy(trampoline + JMP_INSN_SIZE, &jmp_target, sizeof(jmp_target));
+		memcpy(orig->insn + target, trampoline, sizeof(trampoline));
+
+		disp = target - (jcc_index + 2);
+		if(disp > SCHAR_MAX) { /* should not happen */
```
namhyung: Ditto, after `if`

arch/x86_64/mcount-dynamic.c
```diff
@@ -633,3 +634,37 @@ void mcount_arch_dynamic_recover(struct mcount_dynamic_info *mdi,
...
+			pr_err("target is not in reach"); 
+		}
+		info->insns[jcc_index + 1] = disp;
+
+		target += sizeof(trampoline);
+	}
+}
```
namhyung: Please add a newline.

arch/x86_64/mcount-dynamic.c
```diff
@@ -633,3 +634,37 @@ void mcount_arch_dynamic_recover(struct mcount_dynamic_info *mdi,
 		free(badsym);
 	}
 }
+
+int mcount_arch_branch_table_size(struct mcount_disasm_info *info)
+{
+#define TRAMPOLINE_SIZE 14
```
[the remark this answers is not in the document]
namhyung: Maybe rename it to `BRANCH_TABLE_SIZE` or `JMPTABLE_SIZE`?

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,79 @@ static int opnd_reg(int capstone_reg)
...
+	}/* Jump relative 32 if condition is met */
+	else if(opcode == 0x0F && (insn->bytes[1] & 0xF0) == 0x80) {
+
+		cbi = &info->branch_info[info->nr_branch++];
+		cbi->insn_index = info->copy_size;
+		cbi->branch_target = opnd->imm;
+
+		/* We use the equivalent jcc8 of the original jcc32 */
+		relocated_insn[OP] = insn->bytes[1] - 0x10;
+		relocated_insn[OFS] = 0x00;
```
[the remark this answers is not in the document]
AnsBal: > Hmm.. can we keep jcc32 as is? I think it'd be a problem if we change it to jcc8 and the condition is not met. As we relocate jump with a branch table in the trampoline, the upper bytes of offset should be zero. So we can change LSB even in jcc32 in `mcount_arch_patch_branch()` IMO.

Can you please give more details about the problem when the condition is not met.
I think if the condition is not met, the next instruction will be executed. Since handle_rel_jmp() returns the size of the jcc8 insn (2 bytes in this case), the remaining bytes of the buffer won't be used.

arch/x86_64/mcount-insn.c
namhyung: > The relocated jcc jumps to the address of the original instruction. If that target address falls in the prologue (that was overwritten), we may have a problem.

I don't think it's a problem, and we can simply ignore that instruction as it's PC-relative so the target address is still correct IMO.

arch/x86_64/mcount-insn.c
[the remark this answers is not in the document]
namhyung: Hmm.. you changed jcc32 (5 bytes) to jcc8 (2 bytes). So when the condition is not met, the instruction should move to +5 instead of +2, right?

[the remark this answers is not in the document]
namhyung: Maybe it doesn't matter since it won't add more instruction after the jcc32..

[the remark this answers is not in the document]
namhyung: Nevermind, I was confused that we don't change it in-place..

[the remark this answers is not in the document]
AnsBal: Speaking about changing jcc32 and jcc8; to be sure that our jcc will always reach the trampoline, we can always use jcc32. If it's a jcc32 we keep it. else (If it's a jcc8) we upgrade it to jcc32. Do you think it's worth it ?

arch/x86_64/mcount-insn.c
AnsBal: Oh it's PC-relative, indeed. But we patched the jcc to jump to a trampoline when condition is met. The trampoline jumps to an absolute address (the address of the code in the text segment).

arch/x86_64/mcount-dynamic.c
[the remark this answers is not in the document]
AnsBal: Technically, it's the size of an entry in the table. Maybe BRANCH_ENTRY_SIZE ?

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,87 @@ static int opnd_reg(int capstone_reg)
...
+
+		target = opnd->imm;
+
+		memcpy(insns, relocated_insn, IND_JMP_SIZE);
+		memcpy(insns + IND_JMP_SIZE, &target, sizeof(target));
+
+		info->modified = true;
+		/* Since we added JMP here, we can ignore the one in patch_normal_func() */
+		info->has_jump = true;
+		return IND_JMP_SIZE + sizeof(target);
```
namhyung: This assumes that current jump is the last instruction in the prologue we copy.  But there's no guarantee for jcc8 case, right?  I think we can use this only if `info->orig_size + insn->size >= CALL_INSN_SIZE`.

arch/x86_64/mcount-insn.c
namhyung: I thought you only patched offset in the jcc leaving the condition as is, no?

libmcount/internal.h
```diff
@@ -415,6 +415,12 @@ struct mcount_disasm_engine {
 #define INSTRUMENT_FAILED                       -1
 #define INSTRUMENT_SKIPPED                      -2
 
+/* Supposing the size of smallest conditional branch is 1 byte.
+ * We can replace, at most, 5 of them by the instrumentation 
+ * instruction.
+ */
+#define MAX_COND_BRANCH 5
```
namhyung: No, I think 3 is enough for now.  The prologue size needed for other arch is different and they usually require only 2 instructions.

libmcount/internal.h
```diff
@@ -446,6 +457,8 @@ struct mcount_disasm_info {
 	int			copy_size;
 	bool			modified;
 	bool			has_jump;
+	uint8_t			nr_branch;
+	struct			cond_branch_info branch_info[MAX_COND_BRANCH];
```
namhyung: No struct name should go left, please adjust the indentation if needed.

arch/x86_64/mcount-dynamic.c
```diff
@@ -461,6 +461,7 @@ static int patch_normal_func(struct mcount_dynamic_info *mdi, struct sym *sym,
 	struct mcount_disasm_info info = {
 		.sym  = sym,
 		.addr = mdi->map->start + sym->addr,
+		.nr_branch = 0,
```
namhyung: This is not needed since it defaults to 0.

arch/x86_64/mcount-insn.c
AnsBal: Yes, the offset in the jcc was patched leaving the condition as is. But the new offset is the displacement between the jcc and the trampoline, not the old displacement between the jcc and it's original target.

arch/x86_64/mcount-insn.c
AnsBal: > This assumes that current jump is the last instruction in the prologue we copy. But there's no guarantee for **jcc8** case, right?

Do you mean no guarantee for jmp8 ?



> I think we can use this only if `info->orig_size + insn->size >= CALL_INSN_SIZE`.

I agree. This works because, during the call to this function, info->orig_size contains the accumulated size starting from the first instruction of the instrumented symbol to the currently handled instruction.

libmcount/internal.h
AnsBal: I usually use spaces for structs. I'm not sure about the indentation with tabs. I hope I got it right this time =)

libmcount/internal.h
AnsBal: I was taking about safety for other arch. I'll change it back to 3.

arch/x86_64/mcount-insn.c
namhyung: Yep, jmp8 :)

arch/x86_64/mcount-insn.c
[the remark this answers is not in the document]
namhyung: I think jcc8 is fine.

arch/x86_64/mcount-dynamic.c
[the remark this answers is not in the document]
namhyung: I'm ok with it

arch/x86_64/dynamic.S
```diff
@@ -106,7 +107,7 @@ GLOBAL(__dentry__)
 
 	add $48, %rsp
 	.cfi_adjust_cfa_offset -48
-
+	popf
```
namhyung: Do we really need to save and restore the flags?  Then we should do it for normal mcount as well, no?

arch/x86_64/mcount-insn.c
namhyung: Right, then I think it's ok as rel-jump in prologue should go an insn in the prologue and see the same condition and then go to the trampoline (or not).  So we don't need to worry about jumps to prologue in prologue, right?

If it's not, I still think it should be a different issue so a different commit.

arch/x86_64/mcount-insn.c
```diff
@@ -78,6 +78,90 @@ static int opnd_reg(int capstone_reg)
...
+
+		target = opnd->imm;
+
+		memcpy(insns, relocated_insn, IND_JMP_SIZE);
+		memcpy(insns + IND_JMP_SIZE, &target, sizeof(target));
+
+		info->modified = true;
+		/* If this jump is the last insn in the prologue, we can ignore
+		 * the one in patch_normal_func()
+		 */
```
namhyung: It's just a style issue, but it'd be better being consistent.  Please start from the second line for multi-line comments.

libmcount/internal.h
```diff
@@ -415,6 +415,12 @@ struct mcount_disasm_engine {
 #define INSTRUMENT_FAILED                       -1
 #define INSTRUMENT_SKIPPED                      -2
 
+/* Supposing the size of smallest conditional branch is 2 byte.
```
namhyung: Here too.

arch/x86_64/dynamic.S
AnsBal: It should work without it if we assume that the flag register is always set in the function before using it. If it's set by the caller, mcount_entry may change its state and undefined behavior may occur when the flag is used in the callee.
IMO we are better safe than sorry.

namhyung: Yeah, but then it's not a problem of dynamic tracing.  The normal mcount call should consider that also IMO.  But I've never seen the compiler generating that code for mcount (caller) and mcount itself (in [glibc](https://code.woboq.org/userspace/glibc/gmon/mcount.c.html)).  So I assumed the flag is not propagated to a different function.

With that being said, I agree with your safety concern.  But please seperate it from this change and cover the other cases (mcount, fentry) as well.  Also I think it'd be easier to handle if you save it after the normal registers so that you can leave the offsets unchanged.

arch/x86_64/mcount-insn.c
namhyung: What I'm concerning is when the **target** instruction (of the rel-jump in prologue) is modified.  We don't need to worry about the unmodified instructions IMO.  The modified case is only either LEA, CALL or JMP/Jcc (with this change).  And LEA, CALL and Jcc32 are fine since they can come at the last in the prologue so won't change start offsets of other instructions.  And I think Jcc8 is also fine since it keeps the instruction size.

AnsBal: The disp of Jcc is not the only part of the problem. In this patch (and the patch that handles CALLs as well) The target of the trampoline is the target of the Jcc. If the target of the Jcc is the middle of the prologue we may have a problem. Here is an example with Jcc (Same can happen with JMPs and CALLs):

Let's say we want to instrument a function with 3 instructions in the prologue. This function has a Jcc8 that branches to the 2nd instruction.
 ```
Sym:
0x0: Insn1
0x2: Insn2
0x4: Jcc -4
 ```
The result is something similar to this:
 ```
Sym:
0x0: CALL dentry_trampoline
0x5: NOP
 ```
And the out of line execution buffer should be similar to this:
 ```
0x200: Insn1 
0x202: Insn2 
0x204: Jcc +10
0x206: Some padding
0x210: Indirect JMP 0x2   <-- Notice the 2nd instruction (Insn2) is not in 0x2 anymore 
  ```

namhyung: Thanks a lot for providing an example.  So my idea is just to keep `Jcc -4` unchanged.  We don't need the jump table in this case, right?

AnsBal: Yes, in this case we don't need it. But we can't keep the same offset all the time. We need to patch it because if we relocate an instruction between the jcc and the target, its size may change. And the displacement of the jcc may be invalid.

AnsBal:
> Sorry for the late review. I'm still in a business trip.

Sorry for late updates as well. I'm preparing for exams and interviews

> Would be please consider line wrap in the commit logs?

Should I wrap the line to reach smaller window width ?

arch/x86_64/mcount-insn.c
namhyung: But as I mentioned, (currently) all the instructions that change the size is at least 5 byte so it cannot have Jcc after itself.

namhyung:
> Sorry for late updates as well. I'm preparing for exams and interviews

That's totally fine.  Take the time for your real work :)

> Should I wrap the line to reach smaller window width ?

We usually assume 80 for the column width, but git also indent log messages
so general guideline is wrap lines around 72.  And the subject line is supposed
to be shorter than that and people said it would be around 50..

arch/x86_64/mcount-insn.c
AnsBal: Right, but we need to be careful later and take this into consideration.

namhyung:
I'm really sorry it's way too late..  The change looks ok overall, thanks a lot!  I'll add some cleanup later.

PR #1002, opened 2019-10-20
uftrace: Fix .travis.yml to use install-dep.sh

gonapps:
With this change you will immediately know
when change to install-deps.sh introduces some errors especially in Ubuntu.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,15 @@
 language: c
 
-# Ubuntu 14.04 LTS
-dist: trusty
+# Ubuntu 16.04 LTS
+dist: xenial
 
 sudo: required
 
 install:
 - sudo apt-get -qq update
-- sudo apt-get -y  install libelf-dev libstdc++-4.8-dev pandoc
+- sudo misc/install-deps.sh -y
 
-script: ./configure && make
+script: ./configure && make && make unittest
 
 env:
   global:
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -14,20 +14,25 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 case $distro in
     "ubuntu" | "debian")
         apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config
-        apt-get $OPT install libcapstone-dev libluajit-5.1-dev ;;
+        apt-get $OPT install libluajit-5.1-dev || true
+        apt-get $OPT install libcapstone-dev || true ;;
     "fedora")
         dnf install $OPT pandoc elfutils-devel python2-devel ncurses-devel pkgconf-pkg-config
-        dnf install $OPT capstone-devel luajit-devel ;;
+        dnf install $OPT luajit-devel || true
+        dnf install $OPT capstone-devel || true ;;
     "rhel" | "centos")
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
         yum install $OPT pandoc elfutils-devel python2-devel ncurses-devel pkgconfig
-        yum install $OPT capstone-devel luajit-devel ;;
+        yum install $OPT luajit-devel || true
+        yum install $OPT capstone-devel || true ;;
     "arch" | "manjaro")
         pacman $OPT -S pandoc libelf python2 ncurses pkgconf
-        pacman $OPT -S capstone luajit ;;
+        pacman $OPT -S luajit || true
+        pacman $OPT -S capstone || true ;;
     "alpine")
         apk $OPT add elfutils-dev python2-dev ncurses-dev pkgconf
-        apk $OPT add capstone-dev luajit-dev ;;
+        apk $OPT add luajit-dev || true
+        apk $OPT add capstone-dev || true ;;
     *) # we can add more install command for each distros.
         echo "\"$distro\" is not supported distro, so please install packages manually." ;;
 esac


honggyukim:
The current setting looks Ubuntu 14.04 LTS, but it doesn't have libcapstone-dev so might be a problem. In addition, you better add `-y` option to `install-deps.sh`.

gonapps:
@honggyukim Thanks, I fixed.

misc/install-deps.sh
```diff
@@ -14,20 +14,20 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 case $distro in
     "ubuntu" | "debian")
         apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config
-        apt-get $OPT install libcapstone-dev libluajit-5.1-dev ;;
+        apt-get $OPT install libcapstone-dev libluajit-5.1-dev || true ;;
```
DanielTimLee: Seems you're trying to suppress the error from the absence of the package `libcapstone-dev` which doesn't exist in Ubuntu 14.04 LTS. 

But with this approach, this will skip the installation of `libcapstone-dev` as well as `libluajit-5.1-dev`.

https://pkgs.org/download/libluajit-5.1-dev

Since luajit package exist in Ubuntu 14.04 LTS, how about moving `libluajit-5.1-dev` to upper line?

gonapps:
@DanielTimLee Okay I fixed it.

DanielTimLee:
@gonapps  Thanks for the applying the feedback!
But actually what I meant was, `libluajit-5.1-dev` won't be necessary with the true option,
since the it is already there in the Ubuntu 14.04 package repository.

(I'm not so sure about the other disto)

namhyung:
Actually I'd like to update to Ubuntu 16.04 if possible since 14.04 is in EOL (end-of-life) now.

gonapps:
@namhyung @DanielTimLee I modified travis to use ubuntu 16.04, so.. how do you think about '|| true'? should I remove '|| true' as before modification?

.travis.yml
```diff
@@ -1,15 +1,15 @@
 language: c
 
 # Ubuntu 14.04 LTS
-dist: trusty
+dist: xenial
```
namhyung: Please update the comment too,

.travis.yml
```diff
@@ -1,15 +1,15 @@
...
 
 sudo: required
 
 install:
 - sudo apt-get -qq update
-- sudo apt-get -y  install libelf-dev libstdc++-4.8-dev pandoc
+- sudo misc/install-deps.sh -y
 
-script: ./configure && make
+script: ./configure && make && tests/unittest
```
namhyung: It won't build the unittest binary so will fail.  I think you can call `make test TESTARG=unittest` instead.

misc/install-deps.sh
```diff
@@ -14,20 +14,25 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 case $distro in
     "ubuntu" | "debian")
         apt-get $OPT install pandoc libdw-dev libpython2.7-dev libncursesw5-dev pkg-config
-        apt-get $OPT install libcapstone-dev libluajit-5.1-dev ;;
+        apt-get $OPT install libluajit-5.1-dev || true
+        apt-get $OPT install libcapstone-dev || true ;;
```
namhyung: Is this needed for travis?  Anyway I think this change should be in a separate commit.

misc/install-deps.sh
```diff
@@ -14,20 +14,25 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
...
         yum install $OPT pandoc elfutils-devel python2-devel ncurses-devel pkgconfig
-        yum install $OPT capstone-devel luajit-devel ;;
+        yum install $OPT luajit-devel || true
+        yum install $OPT capstone-devel || true ;;
     "arch" | "manjaro")
-        pacman $OPT -S pandoc libelf python2 ncurses pkgconf
-        pacman $OPT -S capstone luajit ;;
+	pacman $OPT -S pandoc libelf python2 ncurses pkgconf
+	pacman $OPT -S luajit || true
+	pacman $OPT -S capstone || true ;;
```
namhyung: Please fix the indentation, we used all spaces here.

.travis.yml
gonapps: Thanks for the review. I fixed this

.travis.yml
gonapps: Oh, my mistake. I changed this to 'make uniitest'

misc/install-deps.sh
gonapps: Because .travis.yml is now using ubuntu 16.04, this change is not much necessary. Should I revert this change as before?

misc/install-deps.sh
gonapps: Thanks. I fixed.

misc/install-deps.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+
```
namhyung: ???

misc/install-deps.sh
namhyung: Well, it's not important.  Just leave it.

misc/install-deps.sh
gonapps: My mistake.. I fixed

namhyung:
Well, I think the most important change here is upgrading Ubuntu version and running unit test.  Could you please update the commit message?

gonapps:
@namhyung Okay I updated the commit message!

namhyung:
The subject line should be shorter than that.  How about "build: Update Travis CI to use Ubuntu 16.04 and unit test"?

gonapps:
@namhyung Updated

.travis.yml
namhyung: Oh, you're right

PR #1006, opened 2019-10-20
misc: Fix tab size in install-deps.sh

sangyun-han:
./misc/install-deps.sh: There is wrong tab size in the recently added line.

Signed-off-by: sangyun-han <sangyun628@gmail.com>


namhyung:
Actually the arch and manjaro used a TAB but others used SPACEs.  But anyway I'm ok with the change too.

PR #1009, opened 2019-10-20
dynamic: Fix a bug disallowing some functions to be patched

AnsBal:
I think that orig_size should be used here instead of copy_size.

Signed-off-by: Anas Balboul <anasbalbo@gmail.com>


namhyung:
Could you please reduce the length of the commit subject line?  And move the detailed description to the body if needed.

AnsBal:
> Right, thanks for the fix!

Thanks for the review !

> Could you please reduce the length of the commit subject line? And move the detailed description to the body if needed.

Done. Let me know if the commit message is clear enough.

PR #953, opened 2019-10-15
doc: Fix typo and grammar in doc files

henrychoi7:
Hi, I just found some typo and grammar issues.
Also, some lines of the document should be ignored markdown formatting, so I fixed it by using `\`.

Signed-off-by: Handong Choi <wschd770@gmail.com>

PR #1010, opened 2019-10-21
doc: Fixed a symbol that was mistaken for MD syntax

rls1004:
Tilde is incorrectly recognized as a markdown syntax.
Added a backslash to display this correctly.


rls1004:
**before :**
\<start\>~\<stop\> (separated by "~")
**after :**
\<start\>\~\<stop\> (separated by "\~")

A tilde is recognized as a strikethrough in markdown syntax.

honggyukim:
It's duplicated to #953, but I can understand why it's need because of your explanation. Thanks!

namhyung:
But the problem is that it removes "~" in the man page according to the https://github.com/namhyung/uftrace/pull/953#discussion_r335794021 .

PR #1012, opened 2019-10-21
misc: Modify dockerfiles to use -y option with install-deps.sh

gonapps:


gonapps:
@namhyung No currently there aren't. I modified this because 
@honggyukim recommended to use -y at https://github.com/namhyung/uftrace/pull/1002#issuecomment-544225383

Because it is not bug fix, I changed commit message.

PR #1014, opened 2019-10-23
tests: Make unittest stricter

gonapps:
So far tests/unittest has always returned EXIT_SUCCESS even if there is a failed test.
I modified tests/unittest.c to return EXIT_FAILURE when there is at least
one of FAILED, SIGNALED, UNKNOWN test.
I also modified .travis.yml to build unittest with ASAN=1 and this reveals some new memory bugs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
 - sudo apt-get -qq update
 - sudo misc/install-deps.sh -y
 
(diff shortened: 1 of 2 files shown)


honggyukim:
@gonapps Thanks for the commit, but please do not write more than 72 characters in a single line for the commit message.

PR #1028, opened 2019-11-10
tests: Fix some tests for ARM 32-bit raspbian

honggyukim:
Since raspbian uses the same package install commands, it just simply
adds raspbian to ubuntu case.

It also applies some partial test fixes.


honggyukim:
I added two more commits to fix some test failures in #1029.

PR #1032, opened 2019-11-11
mcount: Add arch postfix for libmcount*.so

honggyukim:
Since "libmcount*.so" have the same names for each architecture, there
is no way to have multiple libraries for different arch binaries.

This patch allows to have multiple libmcount*.so by adding architecture
postfix.

Here are the execution output for both x86_64 and i386 binaries, and it
shows the single "uftrace" binary can trace for both x86_64 and i386.

For x86_64 arch:
```
  $ uftrace -v -F main -f +addr t-abc
  uftrace: checking binary t-abc
  uftrace: using /home/honggyu/usr/lib/libmcount-x86_64.so library for tracing
    ...
  # DURATION     TID      ADDRESS     FUNCTION
              [179327]       40072a | main() {
              [179327]       4006cf |   a() {
              [179327]       4006e2 |     b() {
              [179327]       4006f5 |       c() {
     0.830 us [179327]       400520 |         getpid();
     2.797 us [179327]       4006f5 |       } /* c */
     3.454 us [179327]       4006e2 |     } /* b */
     4.031 us [179327]       4006cf |   } /* a */
     4.907 us [179327]       40072a | } /* main */
```
For i386 arch:
```
  $ gcc -pg -o t-abc-32 -m32 tests/s-abc.c
  $ uftrace -v -F main -f +addr t-abc-32
  uftrace: checking binary t-abc-32
  uftrace: using /home/honggyu/usr/lib/libmcount-i386.so library for tracing
    ...
  # DURATION     TID      ADDRESS     FUNCTION
              [179347]      804860e | main() {
              [179347]      80485a6 |   a() {
              [179347]      80485bb |     b() {
              [179347]      80485d0 |       c() {
     0.727 us [179347]      8048400 |         getpid();
     2.746 us [179347]      80485d0 |       } /* c */
     3.347 us [179347]      80485bb |     } /* b */
     3.810 us [179347]      80485a6 |   } /* a */
     4.703 us [179347]      804860e | } /* main */
```


honggyukim:
I think we can have a discussion about adding postfix for each libmcount*.so libraries so that it can be used for multiple architectures.

honggyukim:
@namhyung Please let me know how you think. I made this PR based on an internal request and it's also needed for me sometimes.

honggyukim:
@namhyung If you think that changing all the library name is not acceptable, then why don't we keep the original `libmcount.so`, but search `libmcount-$(ARCH).so` when the target binary arch is different from the arch currently running `uftrace` binary?

namhyung:
Actually we cannot run multiple architecture.  It only can run 32bit compat layer.

honggyukim:
Thinking it again, what about making it work as follows:
- provide a way to install `libmcount*.so` into the path with arch sub-directory
  - e.g. `make SUBDIR=i386-linux-gnu install` installs `/usr/lib/i386-linux-gnu/libmcount*.so`
- preload `libmcount*.so` from the default path, but if the `ELFCLASS` is different, then search it inside arch triple sub-directory as a fall-back routine.

It doesn't require `libmcount*.so` name change, but just provide a fall-back routine.

@namhyung If you agree with it, I will change this PR as mentioned.

namhyung:
I'm not sure how long we need to support those 32-bit compatibility in the long run.  So I think we can make it optional like a `configure --enable-compat` switch.  Then it can build 32 bit object files into 'libmcount-compat.so'.

honggyukim:
What about building with `configure --libdir=/usr/local/lib/i386-linux-gnu`?

Then we can make a fall back routine when ELFCLASS of `libmcount.so` is different from the target binary to search `libmcount.so` under `/usr/local/lib/i386-linux-gnu`.

It doesn't require any library name change, but have to decide the target tripple such as `i386-linux-gnu`.

PR #1041, opened 2019-11-19
build: Reduce common code, add SAN flag

gonapps:
This commit tie flags that appear in common into {TRACE,COVERAGE,ASAN}_CFLAGS
Additionally, it adds an UndefinedBehaviorSanitizer flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,28 +91,44 @@ else
 endif
 
 ifeq ($(TRACE), 1)
-  UFTRACE_CFLAGS    += -pg -fno-omit-frame-pointer
-  DEMANGLER_CFLAGS  += -pg -fno-omit-frame-pointer
-  SYMBOLS_CFLAGS    += -pg -fno-omit-frame-pointer
-  TRACEEVENT_CFLAGS += -pg -fno-omit-frame-pointer
-  TEST_CFLAGS       += -pg -fno-omit-frame-pointer
+  TRACE_CFLAGS      := -pg -fno-omit-frame-pointer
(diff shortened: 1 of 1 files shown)


gonapps:
@namhyung 
According to this [article](https://developers.redhat.com/blog/2014/10/16/gcc-undefined-behavior-sanitizer-ubsan/), UBSan seems to be introduced in gcc 4.7

> Also it'd be better to split the UB-san change in a different commit so we can handle it more easily.

Okay then I'll make it as a separate commit and a pull request.

honggyukim:
It seems that it has to be changed more general way using `CHECK` macro.
The possible usage can be one of the followings:
- `make CHECK=all`
- `make CHECK=asan,lsan,ubsan`
- `make CHECK=address,leak,undefined`

We can still keep `ASAN=1` usage for simplicity and old users.

gonapps:
@honggyukim That sounds good too! if @namhyung agrees, I'll reflect what you said

namhyung:
ACK

gonapps:
@namhyung @honggyukim  I added Makefile code as you said, but I named the new flag 'SAN' instead of 'CHECK'. How do you two think?

honggyukim:
Thanks for your work, but I personally think that `CHECK` is easier than `SAN`. It's just a matter of preference though.

gonapps:
@honggyukim Okay, let's wait for @namhyung 's opinion

honggyukim:
@gonapps Sure.

Issue #1037, opened 2019-11-16
change the behaviors of --report option


honggyukim:
The `--report` option is used for showing report output before replay output in `live` command only. But I think it'd be used in a different way.

**1. reuse it for `tui` command.**

It can be used for displaying report mode when `tui` command is launched.

**2. displaying only report output in live mode.**

I think it's not needed to show both report and replay at the same time. It can be used to show only report output without useless commented message.

**Before**:
```
$ uftrace --report -F main t-abc
#
# uftrace report
#
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
   10.033 us    5.660 us           1  main
    4.373 us    0.683 us           1  a
    3.690 us    0.680 us           1  b
    3.010 us    2.147 us           1  c
    0.863 us    0.863 us           1  getpid

#
# uftrace replay
#
# DURATION     TID     FUNCTION
            [ 28340] | main() {
            [ 28340] |   a() {
            [ 28340] |     b() {
            [ 28340] |       c() {
   0.863 us [ 28340] |         getpid();
   3.010 us [ 28340] |       } /* c */
   3.690 us [ 28340] |     } /* b */
   4.373 us [ 28340] |   } /* a */
  10.033 us [ 28340] | } /* main */
```
**After**:
```
$ uftrace --report -F main t-abc
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
   10.033 us    5.660 us           1  main
    4.373 us    0.683 us           1  a
    3.690 us    0.680 us           1  b
    3.010 us    2.147 us           1  c
    0.863 us    0.863 us           1  getpid
```

namhyung:
Well, it was sometimes useful for me to see the both of report and replay output.  But I'm ok with the first change.

honggyukim:
I'm okay if you don't want to change it for live command. I just made a PR for tui command in #1042. Thanks!

PR #1042, opened 2019-11-20
tui: Use --report option to start with report mode

honggyukim:
This patch reuses --report option in tui command so that it can be used
to start the tui window in report mode.

Closes: #1037


honggyukim:
I didn't include because it's a partial fix, but I will add it because item 2 is not going to be applied.

honggyukim:
I just updated. Thanks!

Issue #920, opened 2019-10-03
memory leak report in tui


honggyukim:
Here is the steps to reproduce the problem.
```
$ make ASAN=1 -j8

$ gcc -pg -g -o t-abc tests/s-abc.c

$ ./uftrace record -L. -a t-abc

$ ./uftrace tui

=================================================================
==766==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 849 byte(s) in 63 object(s) allocated from:
    #0 0x7ffff6ecc30f in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x6230f)
    #1 0x48e5fb in add_arg_spec /home/honggyu/work/uftrace/utils/filter.c:217
    #2 0x48ea08 in add_trigger /home/honggyu/work/uftrace/utils/filter.c:242
    #3 0x48f497 in add_filter /home/honggyu/work/uftrace/utils/filter.c:330
    #4 0x49266c in add_trigger_entry /home/honggyu/work/uftrace/utils/filter.c:939
    #5 0x492dd6 in setup_trigger /home/honggyu/work/uftrace/utils/filter.c:1028
    #6 0x4932ad in uftrace_setup_argument /home/honggyu/work/uftrace/utils/filter.c:1104
    #7 0x496598 in build_arg_spec /home/honggyu/work/uftrace/utils/fstack.c:362
    #8 0x45569a in walk_sessions /home/honggyu/work/uftrace/utils/session.c:234
    #9 0x496727 in setup_fstack_args /home/honggyu/work/uftrace/utils/fstack.c:400
    #10 0x472cd4 in open_data_file /home/honggyu/work/uftrace/utils/data-file.c:589
    #11 0x4249b0 in command_tui /home/honggyu/work/uftrace/cmds/tui.c:2584
    #12 0x40b77b in main /home/honggyu/work/uftrace/uftrace.c:1177
    #13 0x7ffff585c82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

Direct leak of 595 byte(s) in 44 object(s) allocated from:
    #0 0x7ffff6ecc30f in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x6230f)
    #1 0x48ffc0 in parse_spec /home/honggyu/work/uftrace/utils/filter.c:512
    #2 0x49076c in parse_argument_spec /home/honggyu/work/uftrace/utils/filter.c:597
[...]

PR #1044, opened 2019-11-22
utils: Fix memory leaks in tui

lntuition:
Fix memory leaks in argument enum string.

Fixed: #920


honggyukim:
I just confirmed that it makes the leak report in #920 disappeared. Thanks!

lntuition:
@namhyung Thank you for advice! I will reflect your opnion.

Issue #1035, opened 2019-11-15
full dynamic tracing failure for V8


honggyukim:
I tested to trace current V8 master with full dynamic tracing, but failed as follows:
```
$ uftrace record -P. ./d8 --expose-gc --trace-gc -e 'gc()'
dynamic: /home/honggyu/work/uftrace/arch/x86_64/mcount-dynamic.c:126:mcount_cleanup_trampoline
 ERROR: cannot restore trampoline due to protection: Invalid argument
```
I bisected the commits and found the problem from the following one.
- 1147486 dynamic: support libraries for dynamic tracing

It cannot simply be reverted right now, so need to take care of the problem later.

honggyukim:
@ParkHanbum Please take a look if possible.

honggyukim:
You can download the following binaries to reproduce the problem.
- [d8.txt](https://github.com/namhyung/uftrace/files/3853618/d8.txt)
- [snapshot_blob.bin.txt](https://github.com/namhyung/uftrace/files/3853619/snapshot_blob.bin.txt)

Then, try to run it if it works fine.
```
$ mv d8.txt d8
$ mv snapshot_blob.bin.txt snapshot_blob.bin

$ ./d8 -e 'console.log("Hello")'
Hello
```

AnsBal:
Hello @honggyukim, I think the problem is the that the address is not aligned to the page size.

This patch should solve the problem:

```diff
diff --git a/arch/x86_64/mcount-dynamic.c b/arch/x86_64/mcount-dynamic.c
index 96912682..03737d2b 100644
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -122,7 +122,9 @@ int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 
 void mcount_cleanup_trampoline(struct mcount_dynamic_info *mdi)
 {
-       if (mprotect((void *)mdi->text_addr, mdi->text_size, PROT_EXEC))
+       void* text_addr = (void *)(mdi->text_addr & ~(getpagesize() - 1));
+
+       if (mprotect(text_addr, mdi->text_size, PROT_EXEC))
                pr_err("cannot restore trampoline due to protection");
 }
 
```

honggyukim:
Hello @AnsBal,

Thanks very much for your effort, but it doesn't actually fix the problem. Here is the execution log.
```
[...]

Issue #827, opened 2019-07-07
another problem of fully dynamic tracing for chrome


honggyukim:
I tested chrome with the current master again and it also gets segfault. Here is gdb test output.
```
(gdb) r
Starting program: /home/honggyu/usr/bin/uftrace record -P. --no-libcall --no-event ./chrome --no-sandbox --headless https://www.google.com
    ...
Thread 2.1 "chrome" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7f1abc0 (LWP 5922)]
0x00007ffff19cc090 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
(gdb) bt
#0  0x00007ffff19cc090 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#1  0x00007ffff19cc4a1 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#2  0x00007ffff19cd08d in _Unwind_Backtrace () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#3  0x00007ffff1707b4f in __GI___backtrace (array=<optimized out>, size=<optimized out>) at ../sysdeps/x86_64/backtrace.c:110
#4  0x00007ffff7ba6eb7 in backtrace (buffer=0x7fffffffc9b0, sz=250) at /home/honggyu/work/uftrace/git/uftrace/libmcount/wrap.c:271
#5  0x000055555a1f614f in base::debug::StackTrace::StackTrace (this=0x7fffffffc9b0, count=<optimized out>) at ./../../base/debug/stack_trace_posix.cc:820
#6  0x00007ffff7bc0ede in __dentry__ () at /home/honggyu/work/uftrace/git/uftrace/arch/x86_64/dynamic.S:110
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```
We can see some code by moving to the upper frame.
```c
 94│ int
 95│ __backtrace (void **array, int size)
 96│ {
 97│   struct trace_arg arg = { .array = array, .cfa = 0, .size = size, .cnt = -1 };
 98│
 99│   if (size <= 0)
100│     return 0;
101│
102│ #ifdef SHARED
103│   __libc_once_define (static, once);
104│
105│   __libc_once (once, init);
106│   if (unwind_backtrace == NULL)
107│     return 0;
108│ #endif
109│
110├─> unwind_backtrace (backtrace_helper, &arg);
111│
[...]

PR #1047, opened 2019-11-22
support PIC code which has 'mov' as opcode.

ParkHanbum:
support PIC code which has 'mov' as opcode.

# statistics

### before
```
dynamic:    total:    17538
dynamic:  patched:    15702 (89.53%)
dynamic:   failed:     1836 (10.46%)
dynamic:  skipped:        0 ( 0.00%)
dynamic: no match:        0
mcount: mcount setup done
```
### after
```
dynamic:    total:    17538
dynamic:  patched:    15717 (89.61%)
dynamic:   failed:     1821 (10.38%)
dynamic:  skipped:        0 ( 0.00%)
dynamic: no match:        0
mcount: mcount setup done
```

### before
```
dynamic:    total:     1109
dynamic:  patched:      336 (30.29%)
dynamic:   failed:      756 (68.16%)
dynamic:  skipped:       17 ( 1.53%)
dynamic: no match:        0
```
### after
```
dynamic:    total:     1109
dynamic:  patched:      368 (33.18%)
dynamic:   failed:      724 (65.28%)
dynamic:  skipped:       17 ( 1.53%)
dynamic: no match:        0
mcount: mcount setup done
```

### DIFF
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -194,6 +194,209 @@ out:
 }
 
 /*
+ *  handle PIC code for MOV instruction.
+ *  this function manipulate the instruction.
+ *
+ *  for examples :
+ *
+ *  case LOAD:
+ *	mov rcx, qword ptr [rip + 0x8f3f85]
+ *  to manipulate like below:
+ *	push rax
+ *	mov rax, PC_BASE + 0x8f3f85
+ *	mov rcx, [rax]
+ *	pop rax
+ *
+ *  case STORE:
+ *  	mov qword ptr [rip+0x8f3f85], rcx
+ *  to manipulate like below:
+ *	push rax
+ *	mov rax, PC_BASE + 0x8f3f85
+ *	mov [rax], rcx
+ *	pop rax
+ *
+ *  exceptional case in LOAD :
+ *  	mov rax, qword ptr [rip+0x8f3f85]
+ *  this will change like below:
+ * 	xchg rax, rcx
+ *  	push rax
+ *  	mov rax, PC_BASE + 0x8f3f85
+ *  	mov rcx, [rax]
+ *  	pop rax
+ *  	xchg rax, rcx
+ *
+ *  exceptional case in STORE :
+ *  	mov qword ptr [rip+0x8f3f85], rax
+ *  to manipulate like below:
+ *  	xchg rax, rcx
+ *	push rax
+ *	mov rax, PC_BASE + 0x8f3f85
+ *	mov [rax], rcx
+ *	pop rax
+ *	xchg rax, rcx
+ *
+ */
+static int handle_pic_mov(cs_insn *insn, uint8_t insns[],
+		      struct mcount_disasm_info *info)
+{
+	struct MODRM modrm;
+	int modrm_rows, modrm_cols;
+	effect_t effect_opnd;
+	uint64_t PC_base;
+
+	cs_x86 *x86 = &insn->detail->x86;
+	cs_x86_op *opnd1 = &x86->operands[0];
+	cs_x86_op *opnd2 = &x86->operands[1];
+
+	static const int NEED_XCHG = 0;
+	static const int NON_XCHG = 2;
+	static const int LOAD = 0;
+	static const int STORE = 1;
+
+	int adjust = NON_XCHG;
+	int insns_type = LOAD;
+
+	// typically instruction elements.
+	uint8_t mov_insns[20] = {0xCC};
+	int insns_size;
+	enum mov_insn_map {
+		XCHG_1 = 0,
+		PUSH = 2,
+		REX_P,
+		OP,
+		IMM,
+		IMM_END = 12,
+		REX_P2 = 13,
+		PREFIX_P2 = 13,
+		OP2,
+		MODRM,
+		POP = 16,
+		XCHG_2,
+	};
+
+	/* support MOV instruction only */
+	if (strcmp(insn->mnemonic, "mov") != 0)
+		goto out;
+
+	/* check PC-relative addressing mode */
+	if (opnd1->type == X86_OP_MEM && opnd1->mem.base == X86_REG_RIP) {
+		insns_type = STORE;
+	}
+	else if (opnd2->type == X86_OP_MEM && opnd2->mem.base == X86_REG_RIP) {
+		insns_type = LOAD;
+	}
+	else
+		goto out;
+
+	/* the SIB addressing is not supported yet */
+	if (opnd1->mem.scale > 1 || opnd2->mem.scale > 1)
+		goto out;
+
+	/* for handle exceptional case, */
+	if (insns_type == LOAD) {
+		/* for now, support only 64bit registere */
+		if (!(X86_REG_RAX <= opnd1->reg && opnd1->reg <= X86_REG_R15))
+			goto out;
+
+		if (opnd1->reg == X86_REG_RAX)
+			adjust = NEED_XCHG;
+
+		PC_base = insn->address + insn->size + opnd2->mem.disp;
+
+		if (X86_REG_RAX <= opnd1->reg && opnd1->reg <= X86_REG_RSP)
+			mov_insns[REX_P2 - adjust] = 0x48;
+		else if (X86_REG_R8 <= opnd1->reg && opnd1->reg <= X86_REG_R15)
+			mov_insns[REX_P2 - adjust] = 0x4C;
+		else
+			goto out;
+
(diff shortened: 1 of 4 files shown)


honggyukim:
@ParkHanbum Thanks for the PR, but I'm afraid of adding more and more techniques to increase success rates because we still see some bugs in full dynamic tracing.

It might add more issues when applying full dynamic tracing to large scale programs so I think we have to fix the existing issues as of now.
- #827 another problem of full dynamic tracing for chrome
- #1035 full dynamic tracing failure for V8

namhyung:
Thanks for the work!  I'll take a look at it.

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+		PREFIX_P2 = 13,
+		OP2,
+		MODRM,
+		POP = 16,
+		XCHG_2,
+	};
+
+	/* support MOV instruction only */
+	if (strcmp(insn->mnemonic, "mov") != 0)
+		goto out;
```
namhyung: This is not necessary since you already checked it in `handle_pic()` below.

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+	uint64_t PC_base;
+
+	cs_x86 *x86 = &insn->detail->x86;
+	cs_x86_op *opnd1 = &x86->operands[0];
+	cs_x86_op *opnd2 = &x86->operands[1];
+
+	static const int NEED_XCHG = 0;
+	static const int NON_XCHG = 2;
+	static const int LOAD = 0;
+	static const int STORE = 1;
```
namhyung: Please make them macro constants.

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+ *
+ *  for examples :
+ *
+ *  case LOAD:
+ *	mov rcx, qword ptr [rip + 0x8f3f85]
+ *  to manipulate like below:
+ *	push rax
+ *	mov rax, PC_BASE + 0x8f3f85
+ *	mov rcx, [rax]
+ *	pop rax
```
namhyung: Can't we do just like below?

```
mov rcx, PC_BASE + 0x8f3f85
mov rcx, [rcx]
```

At least my simple program has following ...
```
    122b:	48 8b 40 08          	mov    rax,QWORD PTR [rax+0x8]
```

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+	else
+		goto out;
+
+	/* the SIB addressing is not supported yet */
+	if (opnd1->mem.scale > 1 || opnd2->mem.scale > 1)
+		goto out;
+
+	/* for handle exceptional case, */
+	if (insns_type == LOAD) {
+		/* for now, support only 64bit registere */
```
namhyung: typo,  anyway as far as I know RIP-relative addressing is only available in 64-bit mode.  Do you really need this?  Also it might not be a contiguous range from RAX to R15.

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+		OP,
+		IMM,
+		IMM_END = 12,
+		REX_P2 = 13,
+		PREFIX_P2 = 13,
+		OP2,
+		MODRM,
+		POP = 16,
+		XCHG_2,
+	};
```
namhyung: What about adding a struct to construct new instructions instead?  For example

```c
// should be packed
struct {
   char push[2];
   struct { char rex, op, imm; uint64_t imm; } mov1;
   ...
};
```

arch/x86_64/mcount-insn.c
```diff
@@ -193,6 +193,209 @@ static int handle_pic_lea(cs_insn *insn, uint8_t insns[],
...
+ *
+ *  exceptional case in LOAD :
+ *  	mov rax, qword ptr [rip+0x8f3f85]
+ *  this will change like below:
+ * 	xchg rax, rcx
+ *  	push rax
+ *  	mov rax, PC_BASE + 0x8f3f85
+ *  	mov rcx, [rax]
+ *  	pop rax
+ *  	xchg rax, rcx
```
namhyung: Why is this exceptional?  Why not doing below?  (But I still think we can do this just using single register like above)

```
push rcx
mov rcx, PC_BASE + 0xf8f385
mov rax, [rcx]
pop rcx
```

arch/x86_64/x64_modrm.c
```diff
@@ -0,0 +1,145 @@
+#include <stdio.h>
+#include <stdint.h>
+#include "x64_modrm.h"
+
+uint8_t modrm_to_byte(struct MODRM modrm)
```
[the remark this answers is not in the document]
ParkHanbum: @namhyung but it is abbreviations. I'll change it if you don't want to use uppercase for abbreviations.

arch/x86_64/x64_modrm.c
```diff
@@ -0,0 +1,145 @@
...
+{
+	uint8_t res = 0;
+
+	res |= (modrm.mod << 6);
+	res |= (modrm.reg << 3);
+	res |= modrm.rm;
```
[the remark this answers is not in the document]
ParkHanbum: @namhyung I'll follow your advise, thanks.

arch/x86_64/mcount-insn.c
ParkHanbum: @namhyung because registers except RAX only can take 4byte for displacement.

arch/x86_64/mcount-insn.c
ParkHanbum: @namhyung I'm pretty sure about this. because operand is not taken 64 bit register only, it can be "EAX or AX".

arch/x86_64/mcount-insn.c
ParkHanbum: @namhyung because only allowed 8byte displacement to RAX register. so, we use RAX to store address of PIC. therefore, need to keep the value of rax register if instruction which we want to manipulate use rax.

arch/x86_64/mcount-insn.c
AnsBal: Hey @ParkHanbum I think with prefix REX.w we can have an eight-byte disp :
0:  48 b9 88 77 66 55 44    movabs rcx,0x1122334455667788
7:  33 22 11

ParkHanbum: @AnsBal you right, maybe it is better choice that using calculated address(imm64) than register base instruction. @namhyung how do you think?

arch/x86_64/x64_modrm.c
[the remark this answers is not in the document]
namhyung: please change.

arch/x86_64/mcount-insn.c
namhyung: I even was not aware of the RAX limitation. :)  I'm fine with anything works!

namhyung:
@ParkHanbum hi, sorry for the looooooong delay.

It seems we can have a much simpler approach for 64-bit loads using the existing LEA handling logic.  I think it can be divided to LEA + MOV (using only single register) then we don't need to use another register with PUSH/POP.  It'd be beneficial since I guess we have more loads and stores.  For stores, I think we can use @AnsBal 's general approach in #1100.

namhyung:
I've added the above mentioned approach (only for loads).  Please take a look if you have time.

honggyukim:
@ParkHanbum @namhyung Thanks for the update. I've seen the difference for `gcc` binary.

**before**
```
$ uftrace record -P. --force --no-libcall -v gcc version.h
    ...
dynamic: dynamic patch stats for 'cc1'
dynamic:    total:    15420
dynamic:  patched:    10541 (68.35%)
dynamic:   failed:     4281 (27.76%)
dynamic:  skipped:      598 ( 3.87%)
dynamic: no match:        0
```
**after**
```
$ uftrace record -P. --force --no-libcall -v gcc version.h
    ...
dynamic: dynamic patch stats for 'cc1'
dynamic:    total:    15420
dynamic:  patched:    11172 (72.45%)
dynamic:   failed:     3650 (23.67%)
dynamic:  skipped:      598 ( 3.87%)
dynamic: no match:        0
```

PR #1048, opened 2019-11-23
build: Fix make ctags rule not to include unintended files

honggyukim:
The current 'make ctags' rule contains unintended files because 'find'
command start from the current directory and matches the entire files
that contains '*.[chS]'.

If uftrace base directory contains other files and directories,
unintended files may be included in 'tags' file.  So this patch
explicitly list all uftrace files and directories only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -358,7 +358,7 @@ reset-coverage:
 	$(Q)$(RM) coverage.info
 
 ctags:
-	@find . -name "*\.[chS]" -o -path ./tests -prune -o -path ./check-deps -prune \
+	@find uftrace.c uftrace.h libmcount arch utils misc libtraceevent -name "*\.[chS]" \
 		| xargs ctags --regex-asm='/^(GLOBAL|ENTRY|END)\(([^)]*)\).*/\2/'
 
 .PHONY: all config clean test dist doc ctags PHONY


namhyung:
What are the unintended files actually?

honggyukim:
I tend to put more example codes in the uftrace base directory. In this case, the example files and its sub-directories are include in ctags.

namhyung:
Hmm.. that's not reasonable IMHO.  We should think all the source files belong to uftrace if it's in the source tree, otherwise you need to tweak it if the directory structure is changed.

honggyukim:
> Hmm.. that's not reasonable IMHO. We should think all the source files belong to uftrace if it's in the source tree, otherwise you need to tweak it if the directory structure is changed.

If so, I cannot put any other example code in the base directory. I don't think we will change the directory structure frequently, and `ctags` rule can also be changed very easily when it's needed.

namhyung:
What's the problem if `ctags` includes your test functions?

honggyukim:
I've made a test scenario that download test programs inside `uftrace` base directory. It's annoying to see some error messages when running `make ctags` after that.
```
$ make ctags
ctags: Warning: cannot open input file "./ksc2019/projects/clang/llvm-8.0.1.src/tools/clang/test/ARCMT/Inputs/with" : No such file or directory
ctags: Warning: cannot open input file "space/test.h" : No such file or directory
^Cmake: *** [ctags] Interrupt
```

yskelg:
This looks like a PR that `ctag` users will cheer for. I hope it gets rebased and applied soon.

PR #1050, opened 2019-11-23
graph: Enhance task graph with field, time-filter, color

honggyukim:
The current graph --task output show tid with task name, but it looks
confusing because it doesn't explain the number is tid info.

So it'd be better to show tid in a TID field and it looks more clear.
In addition, this patch also changes TASK to TASK-NAME in the header.

Before:
```
  $ uftrace graph --task
  ========== TASK GRAPH ==========
  # TOTAL-TIME   SELF-TIME : TASK
      9.002  m    1.041  m : [168086] chrome
                           :  +-[168130] chrome
                           :  |
      7.018  m    1.020  m :  +----[168131] chrome
                           :  |     |
      2.039  m    2.802  s :  |     +----[168200] chrome
                           :  |     |     +-[168201] chrome
      2.039  m   16.797 ms :  |     |     +-[168202] TaskSchedulerFo
      2.039  m  432.881 ms :  |     |     +-[168204] Chrome_ChildIOT
  ...
```
After:
```
  $ uftrace graph --task
  ========== TASK GRAPH ==========
  # TOTAL-TIME   SELF-TIME     TID   : TASK-NAME
      7.018  m   25.892  s  [168086] : chrome
                            [168130] : +- chrome
                                     : |
      5.059  m  682.897 ms  [168131] : +---- chrome
                                     : |     |
      2.039  m    2.972  s  [168200] : |     +---- chrome
                            [168201] : |     |     +- chrome
      2.039  m   16.415 ms  [168202] : |     |     +- TaskSchedulerFo
      2.039  m    1.352  s  [168204] : |     |     +- Chrome_ChildIOT
  ...
```

### DIFF
--- a/cmds/graph.c
+++ b/cmds/graph.c
@@ -14,6 +14,7 @@
 #include "utils/graph.h"
 
 static LIST_HEAD(output_fields);
+static LIST_HEAD(output_task_fields);
 
 struct graph_backtrace {
 	struct list_head list;
@@ -105,6 +106,46 @@ static struct display_field field_addr = {
 	.list    = LIST_HEAD_INIT(field_addr.list),
 };
 
+static void print_task_total_time(struct field_data *fd)
+{
+	struct uftrace_task *node = fd->arg;
+	uint64_t d;
+
+	d = node->time.run;
+
+	print_time_unit(d);
+}
+
+static void print_task_self_time(struct field_data *fd)
+{
+	struct uftrace_task *node = fd->arg;
+	uint64_t d;
+
+	d = node->time.run - node->time.idle;
+
+	print_time_unit(d);
+}
+
+static struct display_field field_task_total_time = {
+	.id      = GRAPH_F_TASK_TOTAL_TIME,
+	.name    = "total-time",
+	.alias   = "total",
+	.header  = "TOTAL TIME",
+	.length  = 10,
+	.print   = print_task_total_time,
+	.list    = LIST_HEAD_INIT(field_task_total_time.list),
+};
+
+static struct display_field field_task_self_time = {
+	.id      = GRAPH_F_TASK_SELF_TIME,
+	.name    = "self-time",
+	.alias   = "self",
+	.header  = " SELF TIME",
+	.length  = 10,
+	.print   = print_task_self_time,
+	.list    = LIST_HEAD_INIT(field_task_self_time.list),
+};
+
 /* index of this table should be matched to display_field_id */
 static struct display_field *field_table[] = {
 	&field_total_time,
@@ -112,9 +153,21 @@ static struct display_field *field_table[] = {
 	&field_addr,
 };
 
+/* index of this task table should be matched to display_field_id */
+static struct display_field *field_task_table[] = {
+	&field_task_total_time,
+	&field_task_self_time,
+};
+
 static void setup_default_field(struct list_head *fields, struct opts *opts)
 {
 	add_field(fields, field_table[GRAPH_F_TOTAL_TIME]);
+}
+
+static void setup_default_task_field(struct list_head *fields, struct opts *opts)
+{
+	add_field(fields, field_task_table[GRAPH_F_TASK_TOTAL_TIME]);
+	add_field(fields, field_task_table[GRAPH_F_TASK_SELF_TIME]);
 }
 
 static void print_field(struct uftrace_graph_node *node)
@@ -124,6 +177,16 @@ static void print_field(struct uftrace_graph_node *node)
 	};
 
 	if (print_field_data(&output_fields, &fd, 2))
+		pr_out(" : ");
+}
+
+static void print_task_field(struct uftrace_task *node)
+{
+	struct field_data fd = {
+		.arg = node,
+	};
+
+	if (print_field_data(&output_task_fields, &fd, 2))
 		pr_out(" : ");
 }
 
@@ -731,11 +794,7 @@ static bool print_task_node(struct uftrace_task *task,
 	if (uftrace_done)
 		return false;
 
-	pr_out("  ");
-	print_time_unit(task->time.run);
-	pr_out("  ");
-	print_time_unit(task->time.run - task->time.idle);
-	pr_out(" : ");
+	print_task_field(task);
 	pr_indent(indent_mask, indent, true);
 	pr_out("[%d] %s\n", task->tid, name);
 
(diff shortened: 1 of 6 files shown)


honggyukim:
By the way, what the difference between those two `chrome` tasks?
```
      9.002  m    1.041  m : [168086] chrome
                           :  +-[168130] chrome
                           :  |
      7.018  m    1.020  m :  +----[168131] chrome
```
The first `[168130] chrome` is followed by `+-`, but the second `[168131] chrome` is followed by `+----`, which is longer than the previous one.

namhyung:
> The first [168130] chrome is followed by +-, but the second [168131] chrome is followed by +----, which is longer than the previous one.

It's for thread and process.

honggyukim:
>> The first [168130] chrome is followed by +-, but the second [168131] chrome is followed by +----, which is longer than the previous one.

> It's for thread and process.

Hmm.. I still don't understand the difference. Could you explain more in details?

honggyukim:
> I think you'd better making it a field so that users can select it with -f option.

Sure, I will.

honggyukim:
> > The first [168130] chrome is followed by +-, but the second [168131] chrome is followed by +----, which is longer than the previous one.
> 
> It's for thread and process.

Ah. You mean `+-` is used for thread and `+---` is used for process? If so, why don't we just use only `+-` for both, but put some mark to processes? For example, `+- chrome*` for process.

namhyung:
> Ah. You mean +- is used for thread and +--- is used for process? If so, why don't we just use only +- for both, but put some mark to processes? For example, +- chrome* for process.

I want to distinguish them more easily with the different indents.  Maybe adding '*' helps too.  Anyway we need to document it..

Can you change the test case 230 as well?

honggyukim:
> I want to distinguish them more easily with the different indents. Maybe adding '*' helps too. Anyway we need to document it..

Do you prefer to keep the indentation and add '*' mark as well?

> Can you change the test case 230 as well?

Sure. I will.

honggyukim:
@namhyung I updated this PR. Please take a look when you have some time. Thanks!

ChangeLog:
- Enable graph --task output to use `-f`/`--field` option
- Move `TID` to a field
- Support time-filter option by comparing with task total time
- Distinguish process with an asterisk and thread in green color(just like htop tool)
- Fix test case 230 accordingly

cmds/graph.c
```diff
@@ -813,7 +813,14 @@ static bool print_task_node(struct uftrace_task *task,
 
 	print_task_field(task);
 	pr_indent(indent_mask, indent, true);
-	pr_out("%s\n", name);
+	if (parent && parent->pid == task->pid) {
+		/* print thread name in green color */
+		pr_green("%s\n", name);
```
namhyung: I'm not sure about this..  We already use different indent and now you added '*' too.

cmds/graph.c
```diff
@@ -871,21 +876,28 @@ static bool print_task_node(struct uftrace_task *task,
...
 	pr_out("========== TASK GRAPH ==========\n");
 	print_header(&output_task_fields, "# ", "TASK NAME", 2);
 
 	indent_mask = xcalloc(handle->nr_tasks, sizeof(*indent_mask));
-	print_task_node(handle->sessions.first_task, NULL, indent_mask, 0);
+
+	/* filter out if total time is less than time-filter */
+	if (opts->threshold <= task->time.run)
+		print_task_node(task, NULL, indent_mask, 0, opts);
+
```
namhyung: Not sure it'll work well for complex data.  I'm concerning the lines..  Did you test it?

honggyukim: Yes, I tested this with chrome data and other data as well.

cmds/graph.c
honggyukim: I actually prefer using color rather than indentation because it required more explanation for new users. Even I didn’t understand why their indentation is different and I thought it looked like a bug.

cmds/graph.c
```diff
@@ -796,7 +813,7 @@ static bool print_task_node(struct uftrace_task *task,
 
 	print_task_field(task);
 	pr_indent(indent_mask, indent, true);
-	pr_out("[%d] %s\n", task->tid, name);
+	pr_out("%s\n", name);
```
namhyung: I'm think of doing the same thing for function graph - moving call count to a field.  What do you think?  Obviously it's a separate change.

cmds/graph.c
namhyung: But color has no meaning what it is, so you still need an explanation.  From the indentation we can guess whether it's a closer descendant (thread) or not (process) IMO.

namhyung: Also you didn't fix(?) the indentation.

doc/uftrace-graph.md
```diff
@@ -28,6 +28,10 @@ GRAPH OPTIONS
     Multiple fields can be set by using comma.  Special field of 'none' can be
     used (solely) to hide all fields.  Default is 'total'.  See *FIELDS*.
 
+--task
+:   Print task graph instead of normal function graph.  The each node in the
+    output shows process(with an asterisk) or thread(printed in green color).
```
namhyung: It'd be nice if you add an example and descriptions for fields.

cmds/graph.c
```diff
@@ -105,18 +106,70 @@ static struct display_field field_addr = {
...
+	&field_task_total_time,
+	&field_task_self_time,
+};
+
 static void setup_default_field(struct list_head *fields, struct opts *opts)
 {
 	add_field(fields, field_table[GRAPH_F_TOTAL_TIME]);
 }
 
+static void setup_task_field(struct list_head *fields, struct opts *opts)
```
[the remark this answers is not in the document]
honggyukim: Okay. It's done.

cmds/graph.c
honggyukim: I makes sense. I will do that later.

doc/uftrace-graph.md
honggyukim: I added more examples.

cmds/graph.c
honggyukim: I think we can use both indentation and green color. I just removed asterisk for process, but didn't change for indentation and color.

honggyukim:
I just updated it again. Thanks.

ChangeLog:
- rename from `setup_task_field` to `setup_default_task_field`
- remove asterisk for process name
- improve documentation with more examples

honggyukim:
I fixed 230 test again by removing asterisk.

honggyukim:
I just removed '*' from the commit message. Thanks.

PR #1052, opened 2019-11-25
dynamic: Fix problem with mprotect's address that may be invalid

AnsBal:
This patch align mprotect's address to the page size. It solves issue (or part of it) #1035: full dynamic tracing failure for V8.

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -12,6 +12,8 @@
 #include "utils/symbol.h"
 
 #define PAGE_SIZE  4096
+#define PAGE_ADDR(a)    ((void *)((a) & ~(PAGE_SIZE - 1)))
+#define PAGE_LEN(a, l)    (a + l - (unsigned long)PAGE_ADDR(a))
 #define XRAY_SECT  "xray_instr_map"
 #define MCOUNTLOC_SECT  "__mcount_loc"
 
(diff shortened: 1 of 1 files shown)


honggyukim:
@AnsBal I recommend you to read other commit messages by running `git log` command. The first line is a commit title, which is no longer than 50 characters, then please give an empty line. Then we write message body, which is no longer than 72 characters in a single line.

Here is an example commit message.
```
arch/arm: Support binary has clang-specific mcount ABI

It was reported that clang generates different code pattern for
calling mcount().  Unlike gcc uses __gnu_mcount_nc() for compatibility
it just use the old mcount() name and saves fp and lr registers before
calling it.  It seems to be the case in Android NDK build (or termux?)

Fixed: #851

```

AnsBal:
@namhyung 
> And I think you should consider the size of the mapping covers the last page since you adjusted the start address.

I created a PAGE_LEN macro that's equivalent to PAGE_ADDR(a + PAGE_SIZE - 1). Do you think it's better to just reuse PAGE_ADDR ?

@honggyukim 
I was looking for an example like this one. Thank you !

namhyung:
I'm ok with having the `PAGE_LEN` macro.  But I wonder it can cover small region across a page boundary.  In that case it should access two pages.

AnsBal:
Since the macro align the size to next page size, PAGE_LEN of anything between 4097 and 8192 (both edges included) will give 8192.  And PAGE_LEN of any value between 0 and 4096 should always give 4096.

namhyung:
It depends on how you pass the value.  Say address = 4000, len = 512..

AnsBal:
Ohh, sorry I used the old len for this special case.
```
int end_addr = addr + len;
int new_len = end_addr - PAGE_ADDR(addr);
int page_len = PAGE_LEN(new_len );
```
Also I wonder if this needed because, quoting **mumap**'s description:  "The address addr must be a multiple of the page size (**but length need not be**)." it's clearly said that len don't need to be multiple of page size. **Mprotect** description says nothing about len but it says that addr must be multiple of page size.

namhyung:
Yep, that should work.  And right, you can just use `new_len` since kernel internally round it up to page size.

honggyukim:
@namhyung Does it need more change?

namhyung:
Hmm.. no. :)

honggyukim:
@AnsBal Thanks very much for fixing this!

Issue #803, opened 2019-06-11
heap-buffer overflow report from asan


honggyukim:
Please see the report from address sanitizer.
```
$ make ASAN=1
$ make install

$ uftrace -la pwd
/home/honggyu/work/uftrace
=================================================================
==34844==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000089ea1 at pc 0x7f98e5d61995 bp 0x7ffeb3d9ed00 sp 0x7ffeb3d9e4a8
READ of size 20 at 0x603000089ea1 thread T0
    #0 0x7f98e5d61994 in __asan_memcpy (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x8c994)
    #1 0x49085d in add_to_rstack_list /home/honggyu/work/uftrace/utils/fstack.c:926
    #2 0x4939d2 in get_task_ustack /home/honggyu/work/uftrace/utils/fstack.c:1478
    #3 0x494466 in read_user_stack /home/honggyu/work/uftrace/utils/fstack.c:1591
    #4 0x498133 in __read_rstack /home/honggyu/work/uftrace/utils/fstack.c:2060
    #5 0x498df4 in peek_rstack /home/honggyu/work/uftrace/utils/fstack.c:2227
    #6 0x442969 in command_replay /home/honggyu/work/uftrace/cmds/replay.c:1122
    #7 0x4369d5 in command_live /home/honggyu/work/uftrace/cmds/live.c:156
    #8 0x40ae94 in main /home/honggyu/work/uftrace/uftrace.c:1081
    #9 0x7f98e46bd82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #10 0x405bf8 in _start (/home/honggyu/usr/bin/uftrace+0x405bf8)

0x603000089ea1 is located 0 bytes to the right of 17-byte region [0x603000089e90,0x603000089ea1)
allocated by thread T0 here:
    #0 0x7f98e5d6d9c1 in realloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x989c1)
    #1 0x49151c in read_task_arg /home/honggyu/work/uftrace/utils/fstack.c:1058
```

honggyukim:
Here is one more.
```
$ uftrace -la uptime
 06:55:07 up 123 days, 20:06, 48 users,  load average: 0.25, 0.24, 0.19
=================================================================
[...]

Issue #1023, opened 2019-11-07
memory leak in report command


honggyukim:
```
$ uftrace report -t 1s --no-pager
    ...
=================================================================
==10673==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 149 byte(s) in 12 object(s) allocated from:
    #0 0x7f8208987538 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x77538)
    #1 0x55a1c717fb7c in process_perf_event /home/honggyu/work/uftrace/git/uftrace/utils/perf.c:576
    #2 0x55a1c71b291a in __read_rstack /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:2060
    #3 0x55a1c71b370f in read_rstack /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:2210
    #4 0x55a1c7120eec in build_function_tree /home/honggyu/work/uftrace/git/uftrace/cmds/report.c:117
    #5 0x55a1c712167b in report_functions /home/honggyu/work/uftrace/git/uftrace/cmds/report.c:227
    #6 0x55a1c71257e1 in command_report /home/honggyu/work/uftrace/git/uftrace/cmds/report.c:692
    #7 0x55a1c7117a16 in main /home/honggyu/work/uftrace/git/uftrace/uftrace.c:1190
    #8 0x7f82072cdb96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)

SUMMARY: AddressSanitizer: 149 byte(s) leaked in 12 allocation(s).
```

lntuition:
https://github.com/namhyung/uftrace/blob/ae49f162eb55b447e5085172eadea0fd7f0fa95d/utils/fstack.c#L925-L930

I think this problem occurs because of duplicate memory allocation. I think 927 line include behavior of 928-929 line. If we remove 928-929 line, memory leaking doesn't exist anymore. But i'm not sure that this approach does not hurt program...

So I need help. Is this correct approach? @honggyukim

honggyukim:
It doesn't work by removing 928-929 line. You will easily see the problem after recording a simple program with `-a` option, then replay it. Maybe `args->data` has to be free-ed somewhere else.

namhyung:
Agreed, we are overwriting `node->args` here and if it had data allocated, it'll be leaked.

honggyukim:
How about this?
```diff
[...]

PR #1055, opened 2019-11-27
Fix memory leaks and heap buffer overflows

honggyukim:
This PR fixes memory leaks and heap buffer overflow problems including the following commits
* fstack: Fix a heap buffer overflow
* replay: Fix a memory leak in event name
* script: Fix a memory leak in a struct strv variable

Fixes: #803 and other problems found in tests

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -314,66 +314,62 @@ static void print_event(struct uftrace_task_reader *task,
 			u.statm = task->args.data;
 			pr_color(color, "%s (size=%"PRIu64"KB, rss=%"PRIu64"KB, shared=%"PRIu64"KB)",
 				 evt_name, u.statm->vmsize, u.statm->vmrss, u.statm->shared);
-			return;
+			break;
 		case EVENT_ID_READ_PAGE_FAULT:
 			u.page_fault = task->args.data;
 			pr_color(color, "%s (major=%"PRIu64", minor=%"PRIu64")",
 				 evt_name, u.page_fault->major, u.page_fault->minor);
-			return;
+			break;
 		case EVENT_ID_READ_PMU_CYCLE:
 			u.cycle = task->args.data;
 			pr_color(color, "%s (cycle=%"PRIu64", instructions=%"PRIu64")",
 				 evt_name, u.cycle->cycles, u.cycle->instrs);
-			return;
+			break;
 		case EVENT_ID_READ_PMU_CACHE:
 			u.cache = task->args.data;
 			pr_color(color, "%s (refers=%"PRIu64", misses=%"PRIu64")",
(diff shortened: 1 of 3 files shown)


cmds/replay.c
```diff
@@ -314,66 +314,63 @@ static void print_event(struct uftrace_task_reader *task,
...
 		case EVENT_ID_WATCH_CPU:
 			u.cpu = task->args.data;
 			pr_color(color, "%s (cpu=%d)", evt_name, *u.cpu);
-			return;
-
+			break;
 		default:
 			pr_color(color, "%s", evt_name);
-			break;
+			pr_color(color, "user_event:%u", evt_id);
```
namhyung: OK, but it seems user events cannot get to here so let's remove the second print.

utils/fstack.c
```diff
@@ -677,6 +677,9 @@ static int fstack_check_skip(struct uftrace_task_reader *task,
 		return -1;
 
 	if (rstack->type == UFTRACE_EXIT) {
+		if (task->stack_count < 1)
+			return -1;
```
namhyung: Hmm.. it should not have 0 stack count here and be skipped.  Let's return 0 instead.

honggyukim:
I just updated it as you mentioned. Thanks!

cmds/replay.c
honggyukim: done.

utils/fstack.c
honggyukim: done.

namhyung:
Wow, you're so quick!  But I think it'd better removing the patch discussing in #1023 from here.

honggyukim:
Do you mean this change? Then I will remove it from this PR.
```diff
diff --git a/utils/perf.c b/utils/perf.c
index e891dd8..43b0c56 100644
--- a/utils/perf.c
+++ b/utils/perf.c
@@ -570,6 +570,8 @@ void process_perf_event(struct uftrace_data *handle)
                if (unlikely(task == NULL || task->fp == NULL))
                        continue;

+               args.data = NULL;
+
                if (perf->type == PERF_RECORD_COMM) {
                        rec->more = 1;
                        args.args = NULL;
@@ -605,6 +607,7 @@ void process_perf_event(struct uftrace_data *handle)
 add_it:
                add_to_rstack_list(&task->event_list, rec, &args);
                perf->valid = false;
+               free(args.data);
        }

        handle->perf_event_processed = true;
```

honggyukim:
I just rebased it onto master.

namhyung:
Yes, that one.

honggyukim:
I just removed the patch you mentioned and updated again. Thanks.

PR #1058, opened 2019-12-02
mcount: Flush data when call is deeply nested

honggyukim:
In some cases, the target program might be crashed while running in the
deeply nested call.  In this case, the current implementation cannot
record its internal calls and lose most of the calls.

For example, there is an incorrected written factorial example.
```
  $ cat factorial.c
  int factorial(int n)
  {
      return n * factorial(n - 1);
  }

  int main()
  {
      return factorial(8);
  }
```
The user expects the deeply nested calls, but it doesn't show anything
in the current implementation.
```
  $ gcc -pg -g factorial.c
  $ uftrace -a a.out
  WARN: too deeply nested calls: 1024
     2.010 us [ 36774] | __monstartup();
     0.847 us [ 36774] | __cxa_atexit();
```
This patch make it flush the recorded data when the call is deeply
nested and it shows the internal calls.
```
  $ gcc -pg -g factorial.c
  $ uftrace -a a.out
     2.223 us [ 36542] | __monstartup();
     1.154 us [ 36542] | __cxa_atexit();
              [ 36542] | main() {
              [ 36542] |   factorial(8) {
              [ 36542] |     factorial(7) {
              [ 36542] |       factorial(6) {
              [ 36542] |         factorial(5) {
              [ 36542] |           factorial(4) {
              [ 36542] |             factorial(3) {
              [ 36542] |               factorial(2) {
              [ 36542] |                 factorial(1) {
              [ 36542] |                   factorial(0) {
              [ 36542] |                     factorial(-1) {
              [ 36542] |                       factorial(-2) {
              [ 36542] |                         factorial(-3) {
...
```

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -788,14 +788,20 @@ static void mcount_finish(void)
 static bool mcount_check_rstack(struct mcount_thread_data *mtdp)
 {
 	if (mtdp->idx >= mcount_rstack_max) {
-		static bool warned = false;
+		if (!mtdp->warned) {
+			struct mcount_ret_stack *rstack;
 
-		if (!warned) {
-			pr_warn("too deeply nested calls: %d\n", mtdp->idx);
-			warned = true;
+			pr_warn("call depth beyond %d is not recorded.\n"
+			        "      (use --max-stack=DEPTH to record more)\n",
+				mtdp->idx);
+			/* flush current rstack */
+			rstack = &mtdp->rstack[mcount_rstack_max - 1];
+			record_trace_data(mtdp, rstack, NULL);
+			mtdp->warned = true;
 		}
 		return true;
 	}
+	mtdp->warned = false;
 	return false;
 }
 
--- a/libmcount/internal.h
+++ b/libmcount/internal.h
@@ -126,6 +126,7 @@ struct mcount_thread_data {
 	bool				recursion_marker;
 	bool				in_exception;
 	bool				dead;
+	bool				warned;
 	unsigned long			cygprof_dummy;
 	struct mcount_ret_stack		*rstack;
 	void				*argbuf;


honggyukim:
The problem is that it can record the data, but it gets crashed when running analysis commands which looks very weird.
```
$ uftrace graph --no-pager

$ uftrace graph --no-pager
=================================================================
==37943==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62e000052400 at pc 0x0000004851ee bp 0x7fffb50a53e0 sp 0x7fffb50a53d0
WRITE of size 8 at 0x62e000052400 thread T0
    #0 0x4851ed in fstack_account_time /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:1776
    #1 0x4871cb in __fstack_consume /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:1985
    #2 0x488324 in __read_rstack /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:2186
    #3 0x48842d in read_rstack /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:2210
    #4 0x419839 in build_graph /home/honggyu/work/uftrace/git/uftrace/cmds/graph.c:488
    #5 0x41b79f in command_graph /home/honggyu/work/uftrace/git/uftrace/cmds/graph.c:846
    #6 0x40b8e4 in main /home/honggyu/work/uftrace/git/uftrace/uftrace.c:1202
    #7 0x7f95fd07082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #8 0x405bf8 in _start (/home/honggyu/usr/bin/uftrace+0x405bf8)

0x62e000052400 is located 0 bytes to the right of 40960-byte region [0x62e000048400,0x62e000052400)
allocated by thread T0 here:
    #0 0x7f95fe7207fa in __interceptor_calloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x987fa)
    #1 0x479eb2 in setup_task_handle /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:61
    #2 0x47b0ca in fstack_setup_task /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:211
    #3 0x47c223 in fstack_setup_filters /home/honggyu/work/uftrace/git/uftrace/utils/fstack.c:443
    #4 0x41b703 in command_graph /home/honggyu/work/uftrace/git/uftrace/cmds/graph.c:838
    #5 0x40b8e4 in main /home/honggyu/work/uftrace/git/uftrace/uftrace.c:1202
    #6 0x7f95fd07082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
...
```

namhyung:
??? I can see the graph well...

```
$ uftrace graph | head -20
# Function Call Graph for 'bad-factorial' (session: e16b984f2f156fcb)
# TOTAL TIME   FUNCTION
   19.506 ms : (1) bad-factorial
    0.878 us :  +-(1) __monstartup
             :  | 
    0.464 us :  +-(1) __cxa_atexit
             :  | 
    9.752 ms :  +-(1) main
    9.751 ms :    (1) factorial
    9.751 ms :    (1) factorial
    9.751 ms :    (1) factorial
    9.749 ms :    (1) factorial
    9.749 ms :    (1) factorial
    9.749 ms :    (1) factorial
    9.748 ms :    (1) factorial
    9.747 ms :    (1) factorial
    9.746 ms :    (1) factorial
    9.746 ms :    (1) factorial
    9.746 ms :    (1) factorial
```

honggyukim:
It looks like an undefined behavior due to an incorrect memory usage somewhere.

honggyukim:
The recorded data can be downloaded here - [uftrace.data.issue.1058.tar.gz](https://github.com/namhyung/uftrace/files/4139420/uftrace.data.issue.1058.tar.gz).

The problem can easily be reproduced with `uftrace report --no-pager`.

honggyukim:
The crash problem in https://github.com/namhyung/uftrace/pull/1058#issuecomment-560386481 is fixed by [check/fstack-get](https://github.com/namhyung/uftrace/compare/check/fstack-get).

@namhyung Could you please review this change again?

libmcount/mcount.c
```diff
@@ -787,15 +787,21 @@ static void mcount_finish(void)
...
-		static bool warned = false;
+	static bool warned = false;
 
+	if (mtdp->idx >= mcount_rstack_max) {
 		if (!warned) {
+			struct mcount_ret_stack *rstack;
+
 			pr_warn("too deeply nested calls: %d\n", mtdp->idx);
+			/* flush current rstack */
+			rstack = &mtdp->rstack[mtdp->idx - 1];
```
namhyung: What if idx was increased more than the `mcount_rstack_max`?

libmcount/mcount.c
```diff
@@ -787,15 +787,21 @@ static void mcount_finish(void)
...
+			record_trace_data(mtdp, rstack, NULL);
```
namhyung: And I think we need to dump rstack for each task separately.  Why not having a warned variable in the mtd?

honggyukim: I just moved `warned` to `mcount_thread_data`. Thanks.

libmcount/mcount.c
honggyukim: I thought that `mtdp->idx` is only increased one by one, so never be beyond `mcount_rstack_max` in line 798 because of the if statement and `warned` variable. I just changed it to `mcount_rstack_max - 1` anyway.

libmcount/mcount.c
```diff
@@ -788,14 +788,20 @@ static void mcount_finish(void)
...
+			        "      (use --max-stack=DEPTH to record more)\n",
+				mtdp->idx);
+			/* flush current rstack */
+			rstack = &mtdp->rstack[mcount_rstack_max - 1];
+			record_trace_data(mtdp, rstack, NULL);
+			mtdp->warned = true;
 		}
 		return true;
 	}
+	mtdp->warned = false;
```
namhyung: Do we really need this?

honggyukim: Yes, it's needed to record next upcoming deeply nested call stack. It's not always happened before stack overflow, but can also be happened in a normal case, but deeply nested depending on `--max-stack` value.

namhyung: ok

PR #1062, opened 2019-12-12
tests: when the verbose option is set, show verbose messages from uftrace also

ParkHanbum:
for currently, when verbose option is set then it will show build and
test command only. so, if want see the verbose message, there is some
work to. to reduce it, this PR will show verbose message from uftrace
with test and build command.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -397,8 +397,13 @@ class TestBase:
         dif = ''
 
         test_cmd = self.runcmd()
-        self.pr_debug("test command: %s" % test_cmd)
+        if self.debug:
+            verbose_level = " -"
+            for i in range(self.debug):
+                verbose_level = verbose_level + "v"
+            test_cmd = test_cmd.replace("uftrace", "uftrace " + verbose_level)
 
+        self.pr_debug("test command: %s" % test_cmd)
         p = sp.Popen(test_cmd, shell=True, stdout=sp.PIPE, stderr=sp.PIPE)
 
         class Timeout(Exception):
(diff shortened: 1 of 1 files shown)


honggyukim:
Hmm.. Sorry but I disagree with it. I often use verbose option only to see its diff output.

honggyukim:
What about changing the current verbose option in `runtest.py` to `-d` instead and use `-v` for your purpose to show debug message of `uftrace`?

@ParkHanbum @namhyung Please let me know how you think.

tests/runtest.py
```diff
@@ -397,8 +397,13 @@ def run(self, name, cflags, diff, timeout):
...
 
         test_cmd = self.runcmd()
-        self.pr_debug("test command: %s" % test_cmd)
+        if self.debug:
+            verbose_level = " -"
+            for i in range(self.debug):
+                verbose_level = verbose_level + "v"
+            test_cmd = test_cmd.replace("uftrace", "uftrace " + verbose_level)
 
+        self.pr_debug("test command: %s" % test_cmd)
```
namhyung: Why is this line marked as changed?

tests/runtest.py
```diff
@@ -438,6 +444,7 @@ def timeout_handler(sig, frame):
                 dif = "%s: %s returns %d\n" % (name, cflags, ret)
             return TestBase.TEST_NONZERO_RETURN, dif
 
+        self.pr_debug("=========== %s ===========\n%s" % (" stderr ", err))
```
namhyung: I guess you also need to decode the `err` stream.

PR #1064, opened 2019-12-14
dump: Fix segfault in do_dump_replay on inverted timestamps

dramborleg:
This commit checks that the inverted_time function pointer is non-NULL
before calling it in do_dump_replay.

Signed-off-by: Colin Lord <clord@xcom-labs.com>

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1509,7 +1509,7 @@ static void do_dump_replay(struct uftrace_dump_ops *ops, struct opts *opts,
 			continue;
 
 		if (prev_time > frs->time)
-			ops->inverted_time(ops, task);
+			call_if_nonull(ops->inverted_time, ops, task);
 		prev_time = frs->time;
 
 


namhyung:
Thanks for your contribution!  The patch itself looks good but I'd like to know why it has such a record.  Is it possible to share your data file?

dramborleg:
Unfortunately I can't share the logs where this was happening. Thanks for taking a look at the pr though.

namhyung:
Ok, I understand.  So is it enough to handle your problem with this change?

dramborleg:
For now this is good enough, the inverted timestamps do cause a problem in the trace viewing tools, so I have some workarounds I've been using until I can dig in more to find out the root cause. At least this gets rid of the segfault for now, so I figured it is useful to contribute even if it isn't fully solving the problem.

namhyung:
OK, thanks again for your help! :)

dramborleg:
Sure, thanks also for all your work on uftrace!

Issue #989, opened 2019-10-19
broken replay output with 64-bit kernel address


honggyukim:
The current output with kernel address is broken as follows:
```
$ sudo ./uftrace -L. -k -f +addr ./hello
Hello World!
# DURATION     TID      ADDRESS     FUNCTION
   1.416 us [157644]       400510 | __monstartup();
   0.930 us [157644]       400530 | __cxa_atexit();
            [157644]       40068f | main() {
            [157644]       4004f0 |   puts() {
   6.371 us [157644] ffffffff812211d0 |     sys_newfstat();
  18.013 us [157644] ffffffff8121c910 |     sys_write();
  32.321 us [157644]       4004f0 |   } /* puts */
  33.640 us [157644]       40068f | } /* main */
```

honggyukim:
It has to be fixed like https://github.com/namhyung/uftrace/pull/716/commits/3267899c5e3e4d5a0d1cf5acbd31ec0210961eb7, but needs to update following @namhyung's comment in https://github.com/namhyung/uftrace/pull/716/files#r270887007.

PR #1066, opened 2019-12-15
field: Fix the output of -f addr to fit in the width

honggyukim:
Since kernel address is represented in 64-bits, it cannot fit in the
current 12 width field and it makes the output ugly.
```
  $ uftrace replay -f +addr
  # DURATION     TID      ADDRESS     FUNCTION
              [191538]       40068f | main() {
              [191538]       4004f0 |   puts() {
     6.337 us [191538] ffffffff812211d0 |     sys_newfstat();
     7.507 us [191538] ffffffff8106dba0 |     __do_page_fault();
    19.908 us [191538] ffffffff8121c910 |     sys_write();
    47.339 us [191538]       4004f0 |   } /* puts */
    48.488 us [191538]       40068f | } /* main */
```
This patch uses effective_addr function to cut the higher 16 bits for
better address display in replay, graph, and tui commands.

Fixed: #989

### DIFF
--- a/cmds/graph.c
+++ b/cmds/graph.c
@@ -67,7 +67,7 @@ static void print_addr(struct field_data *fd)
 	/* uftrace records (truncated) 48-bit addresses */
 	int width = sizeof(long) == 4 ? 8 : 12;
 
-	pr_out("%*lx", width, node->addr);
+	pr_out("%*"PRIx64, width, effective_addr(node->addr));
 }
 
 static struct display_field field_total_time= {
--- a/utils/field.h
+++ b/utils/field.h
@@ -39,6 +39,12 @@ struct display_field {
 	const char *alias;
 };
 
+static inline uint64_t effective_addr(uint64_t addr)
+{
+	/* return 48-bit truncated address info */
+	return addr & ((1ULL << 48) - 1);
+}
+
 void print_header(struct list_head *output_fields, const char *prefix,
 		  int space);
 int print_field_data(struct list_head *output_fields, struct field_data *fd,
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -53,7 +53,7 @@ static void print_addr(struct field_data *fd)
 	if (fstack == NULL)  /* LOST */
 		pr_out("%*s", width, "");
 	else
-		pr_out("%*lx", width, fstack->addr);
+		pr_out("%*"PRIx64, width, effective_addr(fstack->addr));
 }
 
 static void print_timestamp(struct field_data *fd)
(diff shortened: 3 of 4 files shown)


cmds/graph.c
```diff
@@ -67,7 +67,7 @@ static void print_addr(struct field_data *fd)
 	/* uftrace records (truncated) 48-bit addresses */
 	int width = sizeof(long) == 4 ? 8 : 12;
 
-	pr_out("%*lx", width, node->addr);
+	pr_out("%*llx", width, effective_addr(node->addr));
```
namhyung: I know it's annoying but you should use `PRIx64` for `uint64_t` type.

honggyukim: Thanks for pointing this out. I just updated it.

honggyukim:
I added one more commit to properly handle cross recorded data analysis with the following commit.
- d005b507 field: Fix field address length based on ELFCLASS in data

cmds/graph.c
```diff
@@ -835,6 +829,10 @@ int command_graph(int argc, char *argv[], struct opts *opts)
 		synthesize_depth_trigger(opts, &handle, func);
 	}
 
+	/* address length of recorded data might be different from the system */
+	is_lp64 = data_is_lp64(&handle);
+	setup_field_addr(field_table, GRAPH_F_ADDR, is_lp64);
+
```
namhyung: Thanks for doing this.  But I think it'd better passing a pointer to `display_field` to the `print` functions and use the length field.  I know it requires changes in all print functions but it seems the right thing in the long run.

utils/field.h
```diff
@@ -39,6 +39,12 @@ struct display_field {
 	const char *alias;
 };
 
+static inline uint64_t effective_addr(uint64_t addr)
+{
+	/* return 48-bit truncated address info */
+	return addr & 0xffffffffffffULL;
```
namhyung: How about `addr & ((1ULL << 48) - 1)`?

cmds/graph.c
honggyukim: Okay. I will think about it. If so, is it okay if I drop this change in this PR so that the first commit can be merged earlier? I think the second commit is not urgent at all so would like to do it maybe next week.

utils/field.h
honggyukim: Sure, I will.

honggyukim:
I just removed the second commit so that it can be handled in a different PR. Please take a look. Thanks.

honggyukim:
@namhyung I think it's ready to be merged. I will update the `lp64` related patch in a different PR later on. Thanks for your help.

PR #1081, opened 2020-01-01
script: Fix a segfault when module loading is failed

honggyukim:
This patch is to fix the following crash.
```
  $ uftrace script -S scripts/count.py --record t-abc
  WARN: libpython3.7m.so cannot be loaded!

```

### DIFF
--- a/cmds/script.c
+++ b/cmds/script.c
@@ -197,9 +197,9 @@ int command_script(int argc, char *argv[], struct opts *opts)
 
 	/* dtor for script support */
 	script_uftrace_end();
-out:
-	script_finish();
 
+	script_finish();
+out:
 	close_data_file(opts, &handle);
 
 	strv_free(&info.cmds);
--- a/utils/script.c
+++ b/utils/script.c
@@ -102,6 +102,7 @@ int script_init(struct script_info *info, enum uftrace_pattern_type ptype)
 	pr_dbg2("%s(\"%s\")\n", __func__, script_pathname);
 	if (access(script_pathname, F_OK) < 0) {
 		perror(script_pathname);
+		script_finish_filter();
 		return -1;
 	}
 
@@ -124,8 +125,10 @@ int script_init(struct script_info *info, enum uftrace_pattern_type ptype)
 		script_pathname = NULL;
 	}
 
-	if (script_pathname == NULL)
+	if (script_pathname == NULL) {
+		script_finish_filter();
 		return -1;
+	}
 
 	return 0;
 }


cmds/script.c
```diff
@@ -197,8 +197,10 @@ int command_script(int argc, char *argv[], struct opts *opts)
 
 	/* dtor for script support */
 	script_uftrace_end();
-out:
+
 	script_finish();
+out:
+	script_finish_filter();
```
namhyung: The `script_finish_filter()` is an API only for scripting engines (python, luafit) so I don't want to use it from outside.  I'd rather recommend skip calling `script_finish()` in case of failure.

honggyukim: `script_finish()` is called to avoid memory leak previously. Otherwise, we better pass a flag to `script_finish` then avoid calling `script_finish_for_python`.

namhyung: Then it'd be a responsibility of `script_init()` to release all the memory in case of failure.

honggyukim: Hmm.. may be right. I will see if it's okay.

honggyukim: I updated it as you suggested. Thanks.

utils/script.c
```diff
@@ -124,8 +125,10 @@ int script_init(struct script_info *info, enum uftrace_pattern_type ptype)
 		script_pathname = NULL;
 	}
 
-	if (script_pathname == NULL)
+	if (script_pathname == NULL) {
+		script_finish_filter();
```
namhyung: Hmm.. I still don't like this.  I guess it might leak some resources like `python_handle` and many memory objects in the cpython internal when module loading is failed..

namhyung:
What about this instead?

```diff
diff --git a/utils/script-python.c b/utils/script-python.c
index 4d144ba4..7dcd1dc8 100644
--- a/utils/script-python.c
+++ b/utils/script-python.c
@@ -38,6 +38,9 @@ static pthread_mutex_t python_interpreter_lock = PTHREAD_MUTEX_INITIALIZER;
 /* whether error in script was reported to user */
 static bool python_error_reported = false;
 
+/* whether script_init() was done successfully */
+static bool python_initialzed;
+
 static PyAPI_FUNC(void) (*__Py_Initialize)(void);
 static PyAPI_FUNC(void) (*__Py_Finalize)(void);
 static PyAPI_FUNC(void) (*__PySys_SetPath)(char *);
@@ -682,16 +685,20 @@ int script_init_for_python(struct script_info *info,
        if (load_python_api_funcs() < 0)
                return -1;
 
-       if (set_python_path(py_pathname) < 0)
+       if (set_python_path(py_pathname) < 0) {
+               dlclose(python_handle);
                return -1;
+       }
 
        pthread_mutex_lock(&python_interpreter_lock);
 
        __Py_Initialize();
+       python_initialized = true;
 
        /* Import python module that is passed by -p option. */
        if (import_python_module(py_pathname) < 0) {
                pthread_mutex_unlock(&python_interpreter_lock);
+               /* script_finish() will release resources */
                return -1;
        }
 
@@ -731,6 +738,9 @@ void script_finish_for_python(void)
 {
        pr_dbg("%s()\n", __func__);
 
+       if (!python_initialized)
+               return;
+
        pthread_mutex_lock(&python_interpreter_lock);
 
        __Py_Finalize();
```

honggyukim:
Thanks. It doesn't get crashed and show no report even with `ASAN=1` build.
```
$ uftrace script -S scripts/count.py --record t-abc
WARN: libpython3.7m.so cannot be loaded!
WARN: failed to init python scripting

```

PR #1082, opened 2020-01-01
build: Find and use a correct python3 version

honggyukim:
It might be failed to find a correct python3 version used for scripting.
It's because python3 version is found by the python3 binary, but not by
finding python3 library actually used.

The current implementation finds python3 version as follows:
```
  $ python3 -c 'import sysconfig; print(sysconfig.get_config_vars()["LDVERSION"])'
  3.7m
```
However, python3-config finds different version.

This patch is to find the correct version by parsing the output of
'python3-config --libs'.
```
  $ python3-config --libs | sed 's/-l\(python[0-9\.m]\+\) .*/\1/g'
  python3.5m

  $ python3.7-config --libs | sed 's/-l\(python[0-9\.m]\+\) .*/\1/g'
  python3.7m
```
This might be a corner case fix but to support many more different
environments.

### DIFF
--- a/check-deps/Makefile.check
+++ b/check-deps/Makefile.check
@@ -16,9 +16,10 @@ ifneq ($(wildcard $(srcdir)/check-deps/cc_has_mno_sse2),)
 endif
 
 ifneq ($(wildcard $(srcdir)/check-deps/have_libpython3),)
+  PYTHON3 := $(shell python3-config --libs | sed 's/-l\(python[0-9\.m]\+\) .*/\1/g')
   COMMON_CFLAGS += -DHAVE_LIBPYTHON3
-  COMMON_CFLAGS += -I $(shell python3 -c 'import sysconfig; print(sysconfig.get_config_vars()["INCLUDEPY"])')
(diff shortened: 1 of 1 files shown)


namhyung:
This is strange and I don't think it's a normal situation that having 2 different versions of python in a system.  I guess it's Ubuntu, right?  Did you confirm other Ubuntu machines are also in a same situation?  Why not simply changing `python3-config` to point to the `python-3.7-config`?

honggyukim:
> Did you confirm other Ubuntu machines are also in a same situation?

I know that my environment is a bit strange now, but I also think we better embrace such environments unless it affects the normal environments.

> Why not simply changing python3-config to point to the python-3.7-config?

I think we better not to use some specific version for this.

namhyung:
> I know that my environment is a bit strange now, but I also think we better embrace such environments unless it affects the normal environments.

We cannot consider every single strange configurations.  I think it's a reasonable expectation that `python3` and `python3-config` will use a same version.

> I think we better not to use some specific version for this.

Oh, it was a recommendation just for your system. :)

PR #981, opened 2019-10-18
utils: Support portable versionsort

gonapps:
In some non-libc envirnment such as bionic,
there is no versionsort and related functions.

Signed-off-by: Byeonggon Lee <gonny952@gmail.com>

PR #1085, opened 2020-01-03
utils: Replace versionsort() with custom comparison function.

gonapps:
Replace versionsort() for non-libc environment such as bionic libc.


gonapps:
This is continuation of this pull request https://github.com/namhyung/uftrace/pull/981
Added and test the code like you mentioned

PR #1087, opened 2020-01-09
arch/x86_64: Save RFLAGS register

AnsBal:
Hey, before talking about the commit details I want to wish you all the joy of a Happy New Year !

This commit is about saving _RFLAGS_ in stack prior to calling _mcount_entry_ and similar functions that could change it. _GDB_ and other tracing/debugging tools save it as well [1].

I only modified these 3 files for tonight:

_arch/x86_64/mcount.S
arch/x86_64/dynamic.S
arch/x86_64/fentry.S_

The ones left for _x86_64_ are _xray.S_, _cyg_profile_ and _plthook.S_ 
@namhyung  Thank you for your review ! I'll update the ones left in parallel with the review.

[1] https://github.com/bminor/binutils-gdb/blob/master/gdb/gdbserver/linux-x86-low.c
line 1084.

EDIT: I'm not really sure about the need to save the flag for PLT hooks.

### DIFF
--- a/arch/x86_64/dynamic.S
+++ b/arch/x86_64/dynamic.S
@@ -42,6 +42,10 @@
 
 GLOBAL(__dentry__)
 	.cfi_startproc
+
+	/* save RFLAGS register otherwise clobbered */
+	pushf
+
 	sub $48, %rsp
 	.cfi_adjust_cfa_offset 48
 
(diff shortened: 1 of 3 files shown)


namhyung:
@AnsBal happy new year!

I found this from the [x86_64 ABI document](https://github.com/hjl-tools/x86-psABI/wiki/x86-64-psABI-1.0.pdf) (3.2.1)

> The direction flag DF in the %rFLAGS register must be clear (set to “forward”
direction) on function entry and return. Other user flags have no specified role in
the standard calling sequence and are not preserved across calls.

This is somewhat vague but I interpret it as "it doesn't pass RFLAGS to a different function".

AnsBal:
In this case, as long as we only instrument functions entry, we are good.
What about functions written in assembly that does not follow the ABI, wont they be affected by this ?

namhyung:
I guess they'll have many other problems - not only RFLAGS.. ;-)

AnsBal:
I guess they are not supported for now ? Are they excluded when reading the symbols ?

PR #1100, opened 2020-01-28
dynamic: Relocate any rip-relative instruction

AnsBal:
I was inspired from uprobe emulating instruction in out-of-line to make this patch [1]. The core idea is to keep the instruction opcode but change modrm to force it to use a general purpose register instead of RIP. The chosen register to be used by the instruction will hold the computed [rip + dis] address.

Here is some examples: (The address are not exact)
```
original instruction:
0:  48 8b 15 fa 40 23 00    mov    rdx,QWORD PTR [rip+0x2340fa]  
relocated instruction:
0:  50                      push   rax
1:  48 b8 80 43 c7 56 5b    movabs rax,0x555b56c74380
8:  55 00 00
b:  48 8b 10                mov    rdx,QWORD PTR [rax]
e:  58                      pop    rax
```
```
original instruction:
0:  48 83 3d c4 ba 23 00    cmp    QWORD PTR [rip+0x23bac4],0x0 
7:  00 
relocated instruction:
0:  50                      push   rax
1:  48 b8 80 25 c7 56 5b    movabs rax,0x555b56c72580
8:  55 00 00
b:  48 83 38 00             cmp    QWORD PTR [rax],0x0
f:  58                      pop    rax
```
```
original instruction:
0:  f6 05 5f fe 29 00 20    test   BYTE PTR [rip+0x29fe5f],0x20
relocated instruction:
0:  50                      push   rax
1:  48 b8 f1 97 c8 56 5b    movabs rax,0x555b56c897f1
8:  55 00 00
b:  f6 00 20                test   BYTE PTR [rax],0x20
e:  58                      pop    rax
```
 [1] linux/arch/x86/kernel/uprobe.c:860 (arch_uprobe_analyze_insn)


namhyung:
@AnsBal Nice!  It looks like a clever way to handle this.  So didn't you see any problem with this?  I'll have a closer look after releasing the new version.

AnsBal:
Except the one with some instructions using implicitly some registers, I didn't run into problems when experimenting it.

> I'll have a closer look after releasing the new version.

Take the time. I'm looking forward to the release !

namhyung:
I'm sorry it's too late.

Generally looks good, but I'd like to keep the LEA and MOV (for loads) separate function as they can be optimized.  Instead of search available registers, can we just use RDI (or RSI if it's used) since it's not used for any instructions (based on your code)?

Issue #1112, opened 2020-02-08
CI support for multiple arch & compiler


DanielTimLee:
Currently, uftrace is using Travis CI for testing builds and unittests.

Recently, Travis CI released beta version of 
[Multi-cpu arch support](https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support) 

Since it only includes amd64, arm64, so it won't support all of our architectures,
But it will help to test briefly.

Also, travis support multiple compiler builds. (Gcc, clang)

For example, adding this derictives to `.travis.yml`
```yaml
language: c

arch:
  - amd64
  - arm64

compiler:
  - gcc
  - clang
```

Will create build matrix of 2x2 : each arch x compilers.

More detailed info can be found at [Here](https://docs.travis-ci.com/user/multi-cpu-architectures/)

honggyukim:
Thanks. It looks very useful.

PR #1113, opened 2020-02-08
build: Add multi-arch and multi-compiler build test

DanielTimLee:
Recently, Travis CI released support for multi-arch build. And also, it has multi-compiler support.

By this commit, travis will compile with amd64, arm64 architecture and with gcc, clang compiler. (Each arch x compilers)

Fixed: #1112 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: c
 
+arch:
+  - amd64
+  - arm64
+
+compiler:
+  - gcc
+
 # Ubuntu 16.04 LTS
 dist: xenial
 


DanielTimLee:
From the Travis CI build result, 
Clang with AMD64 and Clang with arm64 build failed at unittest.

AMD64 (Failed with ASAN)
ARM64 (Failed due to not found 'mcount' symbol)

DanielTimLee:
Not sure but on my AMD64 clang-10 environment, no error occurs.

namhyung:
Looks really nice!  I'll take a look at the failures..

DanielTimLee:
@namhyung @honggyukim 

Just came to mind, 
How about running test for each linker?

I mean, adding an option to `runtest.py` or `unittest.c` for alternative linker other than `GNU ld`.

There are different variants of  `ld` (GNU ld), such as  `lld` (LLVM linker), `gold` (binutils).
And by running various tests, the stability of the project is likely to increase.

I'm not sure how different they are implemented, but I don't think they'll get exactly the same result.

What do you think?

DanielTimLee:
Oops, sorry. Accidentally pressed the close button.

honggyukim:
I also think it's definitely needed. I've tested on many different environments myself, but it's a lot better to run such tests with Travis CI build. Thanks a lot!

honggyukim:
If we have compilation problem with clang, what about only using `aarch64`?

DanielTimLee:
It's an avail option. And I'm fine with that.

Not sure what is the exact problem with `armhf` compiler test failure.
I'm pretty sure that this issue is mixed with various problems.
(iirc, the compiler version `llvm-3.8` doesn't work with ASAN.)

Maybe, install higher version of clang only at armhf target might work with this test.
This will include the change of Travis test script with `travis.yml`.
(Not tested yet.)

But for now, just exclude armhf and merge will be fine.

namhyung:
If `armhf` has problems, let's enable `aarch64` first.

honggyukim:
@DanielTimLee Can you exclude `clang` compiler for now and update this PR again?

DanielTimLee:
Will do right away!

DanielTimLee:
Done. Thanks for the review.

namhyung:
Thanks, can you please update the commit message?

DanielTimLee:
Oops, my bad,
Just updated the commit message.

Issue #1110, opened 2020-02-05
memory leak in uftrace graph


honggyukim:
The test `181 graph_full` works fine with the current master.
```
$ ./runtest.py 181
Start 1 tests with 1 worker
Test case                 pg             finstrument-fu
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
181 graph_full          : OK OK OK OK OK OK OK OK OK OK
```
However, if `uftrace` is built with asan enabled, then it shows a heap buffer overflow error as follows:
```
$ make ASAN=1 install

$ gcc -pg -o t-abc tests/s-abc.c

$ gcc -pg -o t-forkexec tests/s-forkexec.c

$ uftrace record ./t-forkexec

$ uftrace graph --no-pager
=================================================================
==18294==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62e0000483d8 at pc 0x000000480e47 bp 0x7ffe39cb8f00 sp 0x7ffe39cb8ef0
READ of size 8 at 0x62e0000483d8 thread T0
    #0 0x480e46 in add_graph_entry /home/honggyu/uftrace/utils/graph.c:90
    #1 0x481df3 in graph_add_node /home/honggyu/uftrace/utils/graph.c:207
    #2 0x43b564 in build_graph_node /home/honggyu/uftrace/cmds/graph.c:535
    #3 0x43c02b in build_graph /home/honggyu/uftrace/cmds/graph.c:639
    #4 0x43e3ca in command_graph /home/honggyu/uftrace/cmds/graph.c:989
    #5 0x40b894 in main /home/honggyu/uftrace/uftrace.c:1202
    #6 0x7f6905c1082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #7 0x405ba8 in _start (/mnt/data1/home3/honggyu/usr/bin/uftrace+0x405ba8)

0x62e0000483d8 is located 40 bytes to the left of 40960-byte region [0x62e000048400,0x62e000052400)
allocated by thread T0 here:
    #0 0x7f69073177fa in __interceptor_calloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x987fa)
    #1 0x495bc6 in setup_task_handle /home/honggyu/uftrace/utils/fstack.c:61
    #2 0x496dde in fstack_setup_task /home/honggyu/uftrace/utils/fstack.c:211
[...]

PR #1119, opened 2020-02-11
utils: Fix memory leak in graph

lntuition:
Add missing free to add_graph_exit when recover from fork.
This patch will remove memory leak when use graph command with fork.

Fixed: #1110

### DIFF
--- a/utils/graph.c
+++ b/utils/graph.c
@@ -162,6 +162,7 @@ static int add_graph_exit(struct uftrace_task_graph *tg)
 			    snode->pid == tg->task->t->ppid) {
 				node = snode->node;
 				list_del(&snode->list);
+				free(snode);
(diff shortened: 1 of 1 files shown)


honggyukim:
@lntuition It's been a while, but thanks very much for doing this. I've tested it and the memory leak in #1110 is fixed with it.

Could you please rewrite the commit message without `-` at the beginning? We don't usually write the commit message that way. I personally write it like an email to someone who will read the message after a year. It's better to be a full sentence rather than a short note.

lntuition:
@honggyukim Thank you for your kind comment every time. I change commit message.

honggyukim:
@lntuition The commit message looks great now. Thanks a lot for keep contributing the project.

namhyung:
Thanks!

PR #1124, opened 2020-02-12
utils: Detect busybox and disable color

honggyukim:
In some cases, busybox is used with a script file.

The example case is as follows:
```
  $ realpath /usr/bin/less
  /usr/lib64/busybox/usr/bin/less
```
However, the file is not an ELF executable, but just a script.
```
  $ cat /usr/lib64/busybox/usr/bin/less
  #!/bin/busybox.nosuid
```
To detect such case, uftrace has to check whether the file is ELF
executable or not.  If it's just a script, then disable color.


utils/debug.c
```diff
@@ -104,7 +133,9 @@ static bool check_busybox(const char *pager)
 check:
 	path = realpath(pager, NULL);
 	if (path) {
-		ret = !strncmp("busybox", basename(path), 7);
+		if (!strncmp("busybox", basename(path), 7) ||
+		    !is_elf_file(path))
```
namhyung: I don't think this is a good guess.. if it's not an ELF binary, you should search it has some string like busybox.

honggyukim: Do you mean `is_elf_file` is called first, then check whether it has a substring "busybox" rather than `strncmp` with `basename`?

namhyung: I mean that if it's a script, then it should search "busybox" in the shebang.

Issue #1120, opened 2020-02-11
memory leak in tui


honggyukim:
It's a different memory leak problem in tui, but the example program is same as #1110.
```
$ make ASAN=1 install

$ gcc -pg -o t-abc tests/s-abc.c

$ gcc -pg -o t-forkexec tests/s-forkexec.c

$ uftrace record ./t-forkexec

$ uftrace tui
(press `q` key to quit)

=================================================================
==11994==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 239424 byte(s) in 14 object(s) allocated from:
    #0 0x7ff6438637fa in __interceptor_calloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x987fa)
    #1 0x48005c in load_debug_file /home/honggyu/uftrace/utils/dwarf.c:1333
    #2 0x480a09 in load_debug_info /home/honggyu/uftrace/utils/dwarf.c:1436
    #3 0x457b61 in create_session /home/honggyu/uftrace/utils/session.c:178
    #4 0x471745 in read_task_txt_file /home/honggyu/uftrace/utils/data-file.c:175
    #5 0x4743b3 in open_data_file /home/honggyu/uftrace/utils/data-file.c:530
    #6 0x424fc3 in command_tui /home/honggyu/uftrace/cmds/tui.c:2631
    #7 0x40b8e0 in main /home/honggyu/uftrace/uftrace.c:1208
    #8 0x7ff64215c82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

SUMMARY: AddressSanitizer: 239424 byte(s) leaked in 14 allocation(s).
```

ljw8161:
I've copied the same codes, but I can't get a memory leak like yours.

$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.6 LTS
Release:	16.04
Codename:	xenial

$ uftrace --version
uftrace v0.9.4 ( dwarf python luajit tui perf sched dynamic )

This is a specification for my computer.
[...]

PR #1125, opened 2020-02-12
utils: Fix memory leak in tui

lntuition:
When use tui with fork, dinfo->locs can be overwritten.
This patch will remove memory leak when use tui with fork

Fixed: #1120

### DIFF
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -1323,16 +1323,6 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
 	char *func = NULL;
 	uint64_t offset = 0;
 
-	dinfo->args = RB_ROOT;
-	dinfo->rets = RB_ROOT;
-	dinfo->enums = RB_ROOT;
-	dinfo->files = RB_ROOT;
-
-	if (needs_srcline) {
-		dinfo->nr_locs = symtab->nr_sym;
-		dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
-	}
-
 	xasprintf(&pathname, "%s/%s.dbg", dirname, basename(filename));
 
 	fp = fopen(pathname, "r");
@@ -1346,6 +1336,16 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
 	}
 
 	pr_dbg2("load debug info from %s\n", pathname);
+
+	dinfo->args = RB_ROOT;
+	dinfo->rets = RB_ROOT;
+	dinfo->enums = RB_ROOT;
+	dinfo->files = RB_ROOT;
+
+	if (needs_srcline && dinfo->locs == NULL) {
+		dinfo->nr_locs = symtab->nr_sym;
+		dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
+	}
 
 	while (getline(&line, &len, fp) >= 0) {
 		char *pos;
@@ -1432,7 +1432,7 @@ void load_debug_info(struct symtabs *symtabs, bool needs_srcline)
 		if (map->mod == NULL)
 			continue;
 
-		if (!debug_info_has_location(dinfo)) {
+		if (!debug_info_has_location(dinfo) && !debug_info_has_argspec(dinfo)) {
 			load_debug_file(dinfo, stab, symtabs->dirname,
 					map->libname, needs_srcline);
 		}


utils/dwarf.c
```diff
@@ -1330,6 +1330,9 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
 
 	if (needs_srcline) {
 		dinfo->nr_locs = symtab->nr_sym;
+		if (dinfo->locs) {
+			free(dinfo->locs);
+		}
```
honggyukim: Please remove curly braces as the body is just a single line.

honggyukim:
Hmm.. I've looked at this again. How about this?
```diff
diff --git a/utils/dwarf.c b/utils/dwarf.c
index b016b3f..81f2fbe 100644
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -1323,16 +1323,6 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
        char *func = NULL;
        uint64_t offset = 0;

-       dinfo->args = RB_ROOT;
-       dinfo->rets = RB_ROOT;
-       dinfo->enums = RB_ROOT;
-       dinfo->files = RB_ROOT;
-
-       if (needs_srcline) {
-               dinfo->nr_locs = symtab->nr_sym;
-               dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
-       }
-
        xasprintf(&pathname, "%s/%s.dbg", dirname, basename(filename));

        fp = fopen(pathname, "r");
@@ -1347,6 +1337,16 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,

        pr_dbg2("load debug info from %s\n", pathname);

+       dinfo->args = RB_ROOT;
+       dinfo->rets = RB_ROOT;
+       dinfo->enums = RB_ROOT;
+       dinfo->files = RB_ROOT;
+
+       if (needs_srcline && dinfo->locs == NULL) {
+               dinfo->nr_locs = symtab->nr_sym;
+               dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
+       }
+
        while (getline(&line, &len, fp) >= 0) {
                char *pos;
                struct rb_root *root = &dinfo->args;
```
If there is no '*.dbg` file, then no need to prepare for a buffer for source line info.

It also suppresses the memory leak report.

lntuition:
@honggyukim I agree. I will fix patch with your opinion.

namhyung:
Is this really related to `fork`?

lntuition:
@namhyung Yeah. Main purpose of this PR is adding NULL check. As @honggyukim commented, If there is no '*.dbg` file, then no need to prepare for a buffer for source line info. Maybe this PR can be divided into two commit. Or is it better to modify commit message?

With https://github.com/namhyung/uftrace/issues/1120#issuecomment-585556413, (Also i'm not sure) I think the essence of this problem is related to `fork` and `execl`

namhyung:
OK, this is a problem when `exec` is used so there're more than one sessions.  As symbol table and debug info will be loaded for each session, some files will try to be open more than once and memory will be leaked if it did.  That's why I added `debug_info_has_location()` check in the `load_debug_info()`.

Now I cannot remember exactly why I just checked the location info only.  Maybe it's mandated at the time, but now it's not so we should check it has argspec as well IMO.

lntuition:
@namhyung Thank you for comment! but just adding check argspec doesn't remove memory leak. Am i misunderstanding your points?

```diff
@@ -1432,7 +1432,7 @@ void load_debug_info(struct symtabs *symtabs, bool needs_srcline)
                if (map->mod == NULL)
                        continue;
 
-               if (!debug_info_has_location(dinfo)) {
+               if (!debug_info_has_location(dinfo) && !debug_info_has_argspec(dinfo)) {
                        load_debug_file(dinfo, stab, symtabs->dirname,
                                        map->libname, needs_srcline);
                }
```

Also, I don't know why just moving init section at `load_debug_file()` remove memory leak (without null checking) 

```diff
diff --git a/utils/dwarf.c b/utils/dwarf.c
index b016b3f..4d954f5 100644
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -1323,16 +1323,6 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
        char *func = NULL;
        uint64_t offset = 0;

-       dinfo->args = RB_ROOT;
-       dinfo->rets = RB_ROOT;
-       dinfo->enums = RB_ROOT;
-       dinfo->files = RB_ROOT;
-
-       if (needs_srcline) {
-               dinfo->nr_locs = symtab->nr_sym;
-               dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
-       }
-
        xasprintf(&pathname, "%s/%s.dbg", dirname, basename(filename));

        fp = fopen(pathname, "r");
@@ -1344,9 +1334,18 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,

                pr_err("failed to open: %s", pathname);
        }
-
        pr_dbg2("load debug info from %s\n", pathname);

+       dinfo->args = RB_ROOT;
+       dinfo->rets = RB_ROOT;
+       dinfo->enums = RB_ROOT;
+       dinfo->files = RB_ROOT;
+
+       if (needs_srcline) {
+               dinfo->nr_locs = symtab->nr_sym;
+               dinfo->locs = xcalloc(dinfo->nr_locs, sizeof(*dinfo->locs));
+       }
+
        while (getline(&line, &len, fp) >= 0) {
                char *pos;
                struct rb_root *root = &dinfo->args;
```

namhyung:
Ah, ok.  I think I can remember it now.

The TUI forces to turn on srcline option it tried to load debug info but actually the file was not found.  It may leak when location info is missing but arg info is there.  So I think it needs both changes from me and you. :)

lntuition:
@namhyung Ah, okay. I fix commit with including both changes. Could you review once more?

PR #1118, opened 2020-02-10
misc: Add kali to install-deps.sh

kimdo331:
Since kali uses the same package install commands, it just simply adds kali to ubuntu case.

PR #1134, opened 2020-03-02
misc: Add kali to install-deps.sh

kimdo331:
Since kali uses the same package install commands, it just simply adds kali to ubuntu case.

Signed-off-by: Dohyun Kim <kimdo331@naver.com>


kimdo331:
@honggyukim I send a new Pull Request. This is the same content as https://github.com/namhyung/uftrace/pull/1118

Issue #1088, opened 2020-01-12
arm floating point argument failures


honggyukim:
Here is a test program.
```c
$ cat float.c
#include <stdio.h>

float float_add(float a, float b)
{
        fprintf(stderr, "a = %f, b = %f\n", a, b);
        return a + b;
}

int main(int argc, char *argv[])
{
        double a;

        a = float_add(-0.1, 0.2);
        return a > 0;
}
```
It shows floating point numbers in normal execution.
```
$ gcc -pg -g float.c

$ file a.out
a.out: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), ...

$ ./a.out
a = -0.100000, b = 0.200000
```
It also shows correct numbers with simple uftrace record execution.
```
$ uftrace record --no-libcall --no-event a.out
a = -0.100000, b = 0.200000

$ uftrace record --no-libcall --no-event --debug-domain uftrace:1 a.out
uftrace: checking binary a.out
uftrace: removing uftrace.data.old directory
uftrace: creating 1 thread(s) for recording
uftrace: using /home/honggyu/usr/lib/libmcount.so library for tracing
a = -0.100000, b = 0.200000
uftrace: child terminated with exit code: 1
uftrace: reading uftrace.data/task.txt file
uftrace: flushing /uftrace-3a8039083d891c45-17880-000
```
However, the floating point numbers are somehow printed incorrect values if uftrace tries to record an argument.
```
$ uftrace record -A float_add@fparg1 --no-libcall --no-event a.out
a = 0.000000, b = 0.000000

$ uftrace replay
# DURATION     TID     FUNCTION
            [ 17914] | main() {
  47.760 us [ 17914] |   float_add(0.000000);
  54.740 us [ 17914] | } /* main */
```
The result is same with `--auto-args`.
```
$ uftrace record -a --no-libcall --no-event a.out
a = 0.000000, b = 0.000000
```

honggyukim:
Here is another problem in return value recording.
```
$ uftrace -R float_add@retval/f32 --no-libcall --no-event a.out
a = -0.100000, b = 0.200000
# DURATION     TID     FUNCTION
            [  1009] | main() {
   1.247 ms [  1009] |   float_add() = -nan;
   1.268 ms [  1009] | } /* main */
```

honggyukim:
It seems that `memset@plt` makes `s0` and `s1` cleared to zero.
```asm
 1│ Dump of assembler code for function save_argument:
[...]

PR #1135, opened 2020-03-05
arm: Fix floating point arguments in arm

honggyukim:
The designated initializer in C makes implicit memset internally in arm
architecture.  Due to the memset, VFP registers are clobbered
unexpectedly, so this patch fixes the problem by replacing the
designated initializer to explicit 'mcount_memset4'.

In addition, the offset of VFP registers were incorrect because the
following commit pushs d1 register on top of the original d0.

  ffb69ce arm: Handle struct return type by keeping more regs on stack

So the offset is adjusted from -2 to -4 to cope with the change

Source:
```c
  #include <stdio.h>

  float float_add(float a, float b)
  {
          fprintf(stderr, "a = %f, b = %f\n", a, b);
          return a + b;
  }

  {
          double c;

          c = float_add(-0.1, 0.2);
          fprintf(stderr, "c = %f\n", c);
          return c > 0;
  }
```
Before:
```
  $ uftrace -a -F main a.out
  a = 0.000000, b = 0.000000
  c = 0.000000
              [ 25362] | main(1, 0x7ea25344) {
              [ 25362] |   float_add(0.000000, 0.000000) {
   503.228 us [ 25362] |     fprintf(&_IO_2_1_stderr_, "a = %f, b = %f\n") = 27;
   511.197 us [ 25362] |   } = 0.000000; /* float_add */
     9.687 us [ 25362] |   fprintf(&_IO_2_1_stderr_, "c = %f\n") = 13;
   531.977 us [ 25362] | } = 0; /* main */
```
After:
```
  $ uftrace -a -F main a.out
  a = -0.100000, b = 0.200000
  c = 0.100000
              [ 25146] | main(1, 0x7edbb344) {
              [ 25146] |   float_add(-0.100000, 0.200000) {
   501.769 us [ 25146] |     fprintf(&_IO_2_1_stderr_, "a = %f, b = %f\n") = 28;
   509.321 us [ 25146] |   } = 0.100000; /* float_add */
    12.500 us [ 25146] |   fprintf(&_IO_2_1_stderr_, "c = %f\n") = 13;
   533.539 us [ 25146] | } = 1; /* main */
```
Fixed: #1088

### DIFF
--- a/arch/arm/mcount-support.c
+++ b/arch/arm/mcount-support.c
@@ -488,8 +488,10 @@ void mcount_arch_get_retval(struct mcount_arg_context *ctx,
 	/* type of return value cannot be FLOAT, so check format instead */
 #ifdef HAVE_ARM_HARDFP
 	if (spec->fmt == ARG_FMT_FLOAT && use_hard_float) {
-		/* d0 register (64 bit) was saved below the r0 */
-		memcpy(ctx->val.v, ctx->retval - 2, spec->size);
+		/* d0, d1 registers (64 bit) were saved below the r0 */
+		long *float_retval = ctx->retval - 4;
+
+		mcount_memcpy4(ctx->val.v, float_retval, spec->size);
 	}
 	else
 #endif /* HAVE_ARM_HARDFP */
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -732,9 +732,6 @@ static unsigned long __plthook_entry(unsigned long *ret_addr,
 	struct sym *sym;
 	struct mcount_thread_data *mtdp = NULL;
 	struct mcount_ret_stack *rstack;
-	struct uftrace_trigger tr = {
-		.flags = 0,
-	};
 	bool skip = false;
 	bool recursion = true;
 	enum filter_result filtered;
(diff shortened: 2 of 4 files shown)


arch/arm/mcount-support.c
```diff
@@ -488,8 +488,10 @@ void mcount_arch_get_retval(struct mcount_arg_context *ctx,
 	/* type of return value cannot be FLOAT, so check format instead */
 #ifdef HAVE_ARM_HARDFP
 	if (spec->fmt == ARG_FMT_FLOAT && use_hard_float) {
-		/* d0 register (64 bit) was saved below the r0 */
-		memcpy(ctx->val.v, ctx->retval - 2, spec->size);
+		/* d0, d1 registers (64 bit) were saved below the r0 */
+		long *float_retval = ctx->retval - 4;
+
+		memcpy(ctx->val.v, float_retval, spec->size);
```
namhyung: Why not using `mcount_memcpy4()` here?

honggyukim: I cannot test this now, but changed it to `mcount_memcpy4` as you suggested.

libmcount/plthook.c
```diff
@@ -732,16 +732,17 @@ static unsigned long __plthook_entry(unsigned long *ret_addr,
...
 	bool recursion = true;
 	enum filter_result filtered;
 	struct plthook_data *pd;
 	struct plthook_special_func *func;
 	unsigned long special_flag = 0;
 	unsigned long real_addr = 0;
+	struct uftrace_trigger tr;
+
+	mcount_memset4(&tr, 0, sizeof(tr));
+	tr.flags = 0;
```
namhyung: You don't need to reset flag anymore.. :)

honggyukim: I left it to keep the original code, but removed and updated it anyway.

PR #1140, opened 2020-03-18
doc: Fix deprecated --threads option to --task

ekyooo:


honggyukim:
Thanks for the fix. But it'd be better to make the commit title shorter.

How about this?
```
doc: Fix deprecated --threads option to --task
```
In addition, there is the same problem in `doc/ko/uftrace-report.md` so please fix it as well.

Issue #1071, opened 2019-12-24
always crashed at cs_close() when full dynamic tracing used


honggyukim:
Here is the execution environment.
```
$ uftrace --version
uftrace v0.9.3-189-gd9049 ( dwarf python luajit tui perf sched dynamic )

$ cat /etc/os-release | grep VERSION=
VERSION="18.04.3 LTS (Bionic Beaver)"
```
In this server, full-dynamic tracing always gets crashed as follows:
```
$ uftrace record -L. -P. --debug-domain dynamic:3 t-abc
dynamic: dynamic patch type: t-abc: 0 (none)
dynamic: patch normal func: a (patch size: 9)
dynamic: patch normal func: b (patch size: 9)
dynamic: patch normal func: c (patch size: 9)
dynamic: patch normal func: main (patch size: 8)
dynamic: patched all (4) functions in 't-abc'
WARN: child terminated by signal: 11: Segmentation fault
```
Here is the crash point.
```
$ cgdb -q --args uftrace record -L. -P. t-abc
    ...
 29│ void mcount_disasm_finish(struct mcount_disasm_engine *disasm)
 30│ {
 31├───────> cs_close(&disasm->engine);
 32│ }
 33│
 34│ enum fail_reason {
 35│         INSTRUMENT_FAIL_NODETAIL        = (1U << 0),
 36│         INSTRUMENT_FAIL_NOOPRND         = (1U << 1),
/home/honggyu/work/uftrace/arch/x86_64/mcount-insn.c
```
I deleted `cs_close` statement for debugging purpose.
```diff
diff --git a/arch/x86_64/mcount-insn.c b/arch/x86_64/mcount-insn.c
index 32c87c04..c1810489 100644
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -28,7 +28,6 @@ void mcount_disasm_init(struct mcount_disasm_engine *disasm)

 void mcount_disasm_finish(struct mcount_disasm_engine *disasm)
 {
-       cs_close(&disasm->engine);
 }

 enum fail_reason {
```
It still gets crashed in a different place.
```
1642│ static __used void mcount_startup(void)
1643│ {
1644│         char *pipefd_str;
1645│         char *logfd_str;
    ...
1769│         if (patch_str)
1770│                 mcount_dynamic_update(&symtabs, patch_str, patt_type, &disasm);
1771│
    ...
[...]

PR #1141, opened 2020-03-21
dynamic: Fix a crash in x86_64 full dynamic tracing

honggyukim:
In some environments, full dynamic tracing gets always crashed.
```
  $ uftrace record -P. --debug-domain dynamic:3 t-abc
  dynamic: dynamic patch type: t-abc: 0 (none)
  dynamic: patch normal func: a (patch size: 9)
  dynamic: patch normal func: b (patch size: 9)
  dynamic: patch normal func: c (patch size: 9)
  dynamic: patch normal func: main (patch size: 8)
  dynamic: patched all (4) functions in 't-abc'
```
It works fine on many other machines, but gets crashed on some specific
machines.

It might has some read-only data in trampoline code, but the cleanup
phase sets the permission only with PROT_EXEC.

The reason is not 100% clear now, but it's verified that adding
PROT_READ permission fixes the crash problem.

Fixed: #1071

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -125,8 +125,9 @@ int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 
 void mcount_cleanup_trampoline(struct mcount_dynamic_info *mdi)
 {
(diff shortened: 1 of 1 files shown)


honggyukim:
I found that this PR partially fixes some problems in #827 for chrome full dynamic tracing.

Issue #1152, opened 2020-04-10
display the file name of function


LoisHuang:
Hello, whether the output could contain the file name of the function?

honggyukim:
If you record with `--srcline` or `-a` option, then tui command shows the source location info at the bottom.

Please note that the target binary must have debug info, which is generated by `-g` option just like using `gdb` tool.

honggyukim:
#1133 is under review and it shows source location info in `report` command.

LoisHuang:
Thanks for your reply! But I seemingly encountered one problem  when I used `tui `command, that the foot line displayed `uftrace graph: source location is not available [at 0x40063b] `. How can I solve this problem?

honggyukim:
Please make sure the followings:
- if the binary is compiled with `-g` option.
- if the uftrace.data is recorded with one of `--srcline` or `-a` option. You can check if data directory has `.dbg` file inside.

Thanks for using uftrace.

LoisHuang:
Appreciate for your explanation! I used `gcc -finstrument-functions -g test.c` firstly, and it could not display the source line correctly. Then I changed the command to  `gcc -pg -g test.c` , and I can get the source line information now.

honggyukim:
Thanks for finding the problem. 

The problem is that it's not possible to read argument info if the given binary is built with `-finstrument-functions` due to the generated code patterns.

The `.dbg` files contain argument info along with source location info, but the current implementation do not generate `.dbg` file for `-finstrument-functions` built binaries.

But I think uftrace has to produce `.dbg` file only for `-finstrument-functions` built binaries if `--srcline` option is given regardless of argument info.

The following change will be working fine for `-finstrument-functions` built binaries.
```diff
diff --git a/utils/dwarf.c b/utils/dwarf.c
index ace3a63e..8c85d67e 100644
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -262,9 +262,6 @@ static int setup_dwarf_info(const char *filename, struct debug_info *dinfo,
 {
        int fd;

-       if (!force && check_trace_functions(filename) != TRACE_MCOUNT)
-               return 0;
-
        pr_dbg2("setup dwarf debug info for %s\n", filename);

        fd = open(filename, O_RDONLY);
```

honggyukim:
@LoisHuang I've made a PR to fix this problem. Please see #1155. Thanks very much for reporting this problem.

LoisHuang:
[...]

PR #1155, opened 2020-04-13
dwarf: Produce srcline in .dbg for -finstrument-functions

honggyukim:
LoisHuang reported that the current implementation does not generate
source location info when the target binary is compiled with
-finstrument-functions even though it has debug info.

Binary compilation:
```
  $ gcc -g -finstrument-functions -o t-abc.finst tests/s-abc.c
```
Before:
```
  $ uftrace record -a t-abc.finst
  $ cat uftrace.data/*.dbg
  cat: cannot access 'uftrace.data/*.dbg': No such file or directory
```
After:
```
  $ uftrace record -a t-abc.finst
  $ cat uftrace.data/t-abc.finst.dbg
  F: 73a a
  L: 11 tests/s-abc.c
  F: 77a b
  L: 16 tests/s-abc.c
  F: 7ba c
  L: 21 tests/s-abc.c
  F: 818 main
  L: 26 tests/s-abc.c
```
Please note that it doesn't generate arguments information unnecessarily
because TRACE_CYGPROF type only has source location information.

Fixed: #1152

### DIFF
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -262,8 +262,8 @@ static int setup_dwarf_info(const char *filename, struct debug_info *dinfo,
 {
 	int fd;
 
-	if (!force && check_trace_functions(filename) != TRACE_MCOUNT)
-		return 0;
+	if (force || check_trace_functions(filename) != TRACE_CYGPROF)
+		dinfo->needs_args = true;
 
 	pr_dbg2("setup dwarf debug info for %s\n", filename);
 
@@ -1043,6 +1043,12 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 
 		if (uftrace_done)
 			break;
+
+		/* do not read arguments when it's not needed */
+		if (!dinfo->needs_args) {
+			bd.nr_args = 0;
+			bd.nr_rets = 0;
+		}
 
 		dwarf_getsrcfiles(&cudie, &bd.files.files, &bd.files.num);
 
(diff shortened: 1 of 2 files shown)


utils/dwarf.c
```diff
@@ -261,9 +261,10 @@ static int setup_dwarf_info(const char *filename, struct debug_info *dinfo,
 			    unsigned long offset, bool force)
 {
 	int fd;
+	enum uftrace_trace_type trace_type = check_trace_functions(filename);
 
-	if (!force && check_trace_functions(filename) != TRACE_MCOUNT)
-		return 0;
+	if (force || trace_type == TRACE_MCOUNT || trace_type == TRACE_FENTRY)
+		dinfo->needs_args = true;
```
namhyung, line 267: Now I think that `force` is a really bad name.  Anyway we might add some more trace types and miss to update this.  How about this?

```c
if (force || trace_type != TRACE_CYGPROF)
    dinfo->needs_args = true;
```

honggyukim: I just updated it as you suggested. Thanks!

namhyung: Why not just moving the check call to the if statement as it was.

utils/dwarf.c
```diff
@@ -1044,6 +1045,12 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 		if (uftrace_done)
 			break;
 
+		/* do not read arguments when it's not needed */
+		if (dinfo->needs_args == false) {
```
namhyung, line 1049: For bool type, I prefer using `!` like

```c
if (!dinfo->needs_args)
   ...
}
```

honggyukim:
I just updated again. Thanks.

Issue #1151, opened 2020-04-10
failure on collecting correct kernel-cpuXX


honggyukim:
If some cpus are offlined especially from the initial numbers, then current implementation fails to collect correct `kernel-cpuXX.dat`. It's because it just simply count the number of cpus, then collect from `cpu0` by looping over the number of cpus.

Here is the problematic cpu settings.
```
$ cat /sys/devices/system/cpu/possible
0-7

$ cat /sys/devices/system/cpu/offline
0-3,6-7

$ cat /sys/devices/system/cpu/online
4-5
```
In this case, uftrace has to collect `kernel-cpu4.dat` and `kernel-cpu5.dat`, but it just collects `kernel-cpu0.dat` and `kernel-cpu1.dat`, which are empty because they are offlined.

We need to fix this by collecting the online cpus only.

honggyukim:
If this issue is fixed, then it can also be simpler to handle the problem as mentioned in https://github.com/namhyung/uftrace/pull/945#pullrequestreview-304215689.

namhyung:
I'm not sure but if it'd return a specific error code for offline cpus, then we could check it and move onto the next.

honggyukim:
I'm sorry but could you please explain what you mean?

namhyung:
Something like this?

```c
for (i = 0; i < num_cpu; i++) {
   if (check_cpu(i) < 0)  // check error code?
      continue;
   ...
   // creates "kernel-cpu${i}".dat
}
```

honggyukim:
Thank you. The following print statement shows only online cpus so it needs to be changes to find all the available cpus.
```diff
diff --git a/utils/kernel.c b/utils/kernel.c
index d7dc3571..23324563 100644
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -651,6 +652,7 @@ int setup_kernel_tracing(struct uftrace_kernel_writer *kernel, struct opts *opts
                return ret;

        kernel->nr_cpus = n = sysconf(_SC_NPROCESSORS_ONLN);
+       fprintf(stderr, "kernel->nr_cpus = %d\n", kernel->nr_cpus);

        kernel->traces  = xcalloc(n, sizeof(*kernel->traces));
        kernel->fds     = xcalloc(n, sizeof(*kernel->fds));
```

honggyukim:
We may have to use `sysconf(_SC_NPROCESSORS_CONF)` and loop over all the available cpus, then decide which cpu is online.

[...]

PR #945, opened 2019-10-13
util: Check/remove empty dat files

yeominstall:
remove automatically created dat files for each online CPUs like ‘perf-cpuXX.dat’ and 'kernel-cpuXX.dat’ if it doesn’t have no content in it.

Fixed: #893

PR #1156, opened 2020-04-18
kernel: Properly handle offlined CPUs for kernel tracing

honggyukim:
If some cpus are offlined especially from the initial numbers, then
the current implementation fails to collect correct kernel-cpuXX.dat.

It's because it just simply counts the number of cpus, then collects
from cpu0 by looping over the number of cpus.

Here is the problematic cpu settings.
```
  $ cat /sys/devices/system/cpu/possible
  0-7

  $ cat /sys/devices/system/cpu/offline
  0-3,6-7

  $ cat /sys/devices/system/cpu/online
  4-5
```
In this case, uftrace has to collect kernel-cpu4.dat and kernel-cpu5.dat,
but it just collects kernel-cpu0.dat and kernel-cpu1.dat, which are
empty because they are offlined.

This patch fixes the problem by looping over all the cpus rather than
considering only online cpus.  It changes from _SC_NPROCESSORS_ONLN to
_SC_NPROCESSORS_CONF.
```
  _SC_NPROCESSORS_ONLN
    The number of processors currently online

  _SC_NPROCESSORS_CONF
    The number of processors configured.
```
Fixed: #1151

### DIFF
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -650,7 +650,7 @@ int setup_kernel_tracing(struct uftrace_kernel_writer *kernel, struct opts *opts
 	if (ret < 0)
 		return ret;
 
-	kernel->nr_cpus = n = sysconf(_SC_NPROCESSORS_ONLN);
+	kernel->nr_cpus = n = sysconf(_SC_NPROCESSORS_CONF);
 
 	kernel->traces	= xcalloc(n, sizeof(*kernel->traces));
 	kernel->fds	= xcalloc(n, sizeof(*kernel->fds));


honggyukim:
No, the empty `kernel-cpuX.dat` files will be left unless we work on #945.

It will be a lot simpler to handle the issue with kernel files with this PR. Please read your previous comment at  https://github.com/namhyung/uftrace/pull/945#pullrequestreview-304215689.

namhyung:
Ah, I forgot it was not merged. :)

honggyukim:
I think the empty files are better to be removed. So we can work on it after this.

namhyung:
I'm ok as long as it makes any problem.

honggyukim:
Does it need any more change?

PR #1051, opened 2019-11-25
some misc refactoring and change code_tree to code_hmap

ParkHanbum:
this is rough benchmark test results.

* Benchmark Results of rbtree.......
```
OPTIMIZE 0
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    6.114 ms    6.114 ms           1  start_bmt
OPTIMIZE 1
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    3.799 ms    3.799 ms           1  start_bmt
OPTIMIZE 2
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    3.770 ms    3.770 ms           1  start_bmt
OPTIMIZE 3
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    3.796 ms    3.796 ms           1  start_bmt
OPTIMIZE s
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    3.983 ms    3.983 ms           1  start_bmt
```

* Benchmark Results of hashmap.......
```
OPTIMIZE 0
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
    1.274 ms    1.274 ms           1  start_bmt
OPTIMIZE 1
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
  734.949 us  734.949 us           1  start_bmt
OPTIMIZE 2
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
  637.115 us  637.115 us           1  start_bmt
OPTIMIZE 3
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
  651.478 us  651.478 us           1  start_bmt
OPTIMIZE s
  Total time   Self time       Calls  Function
  ==========  ==========  ==========  ====================
  886.333 us  886.333 us           1  start_bmt
```
this results mean how long time takes while searching against about 50,000 symbols.

You can check it from my git repository.
https://github.com/ParkHanbum/android_c_hashmap

clone and make. after finish benchmark, cat the report.log.

PR #1160, opened 2020-04-22
dwarf : Remove common compile directory prefix from file name in .dbg

ekyooo:
Remove the common directory prefix if file name is under
the compile unit's compilation directory (DW_AT_comp_dir).

### DIFF
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -6,6 +6,7 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <dlfcn.h>
+#include <stdlib.h>
 
 /* This should be defined before #include "utils.h" */
 #define PR_FMT     "dwarf"
@@ -910,6 +911,7 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
 	dinfo->locs[sym_idx].sym  = sym;
 	dinfo->locs[sym_idx].file = dfile;
 	dinfo->locs[sym_idx].line = dline;
+	dinfo->nr_locs_used++;
 }
 
 static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
@@ -996,6 +998,109 @@ out:
 	return DWARF_CB_OK;
 }
 
+struct comp_dir_entry {
+	struct rb_node	node;
+	char		*name;
+	int		nr_used;  /* number of times comp_dir is used in module */
+	int		nr_locs;  /* number of source locations built into comp_dir */
+};
+
+static int add_comp_dir(struct rb_root *root, char *name, int nr_locs)
+{
+	struct comp_dir_entry *entry, *iter;
+	struct rb_node *parent = NULL;
+	struct rb_node **p = &root->rb_node;
+	int cmp;
+
+	pr_dbg3("add dir entry: %s (%d)\n", name, nr_locs);
+
+	while (*p) {
+		parent = *p;
+		iter = rb_entry(parent, struct comp_dir_entry, node);
+
+		cmp = strcmp(iter->name, name);
+		if (cmp == 0) {
+			iter->nr_used++;
+			iter->nr_locs += nr_locs;
+			return 0;
+		}
+
+		if (cmp > 0)
+			p = &parent->rb_left;
+		else
+			p = &parent->rb_right;
+	}
+
+	entry = xmalloc(sizeof(*entry));
+	entry->name = xstrdup(name);
+	entry->nr_locs = nr_locs;
+	entry->nr_used = 1;
+
+	rb_link_node(&entry->node, parent, p);
+	rb_insert_color(&entry->node, root);
+
+	return 0;
+}
+
+static void free_comp_dir(struct rb_root *root)
+{
+	struct comp_dir_entry *entry;
+	struct rb_node *node;
+
+	while (!RB_EMPTY_ROOT(root)) {
+		node = rb_first(root);
+		entry = rb_entry(node, typeof(*entry), node);
+
+		rb_erase(node, root);
+		free(entry->name);
+		free(entry);
+	}
+}
+
+static struct comp_dir_entry * get_max_comp_dir(struct comp_dir_entry *a, struct comp_dir_entry *b)
+{
+	if (a->nr_used > b->nr_used ||
+	   (a->nr_used == b->nr_used && a->nr_locs > b->nr_locs))
+		return a;
+	else
+		return b;
+}
+
+static char* get_base_comp_dir(struct rb_root *dirs)
+{
+	struct rb_node *rbnode;
+	struct comp_dir_entry *e;
+	struct comp_dir_entry *prev = NULL;
+	struct comp_dir_entry *max;
+
+	if (RB_EMPTY_ROOT(dirs))
+		return NULL;
+
+	rbnode = rb_first(dirs);
+	max = rb_entry(rbnode, typeof(*e), node);
+	prev = max;
+	rbnode = rb_next(rbnode);
+
+	while (rbnode != NULL) {
+		e = rb_entry(rbnode, typeof(*e), node);
+
+		if (!strncmp(e->name, prev->name, strlen(prev->name))) {
+			prev->nr_used += e->nr_used;
+			prev->nr_locs += e->nr_locs;
+		}
+		else {
+			max = get_max_comp_dir(prev, max);
+			prev = e;
+		}
+
+		rbnode = rb_next(rbnode);
+	}
+
+	max = get_max_comp_dir(prev, max);
+
+	return max->name;
+}
+
 static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 			     enum uftrace_pattern_type ptype,
 			     struct strv *args, struct strv *rets)
@@ -1005,6 +1110,8 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 	size_t header_sz = 0;
 	struct uftrace_pattern *arg_patt;
 	struct uftrace_pattern *ret_patt;
+	struct rb_root comp_dirs = RB_ROOT;
+	char *dir;
 	char *s;
 	int i;
 
@@ -1054,7 +1161,23 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 
 		dwarf_getfuncs(&cudie, get_dwarfspecs_cb, &bd, 0);
 
+		if (dwarf_hasattr(&cudie, DW_AT_comp_dir)) {
+			dir = str_attr(&cudie, DW_AT_comp_dir, false);
+			add_comp_dir(&comp_dirs, dir, dinfo->nr_locs_used);
+			dinfo->nr_locs_used = 0;
+		}
+
 		curr = next;
+	}
+
+	dir = get_base_comp_dir(&comp_dirs);
+	if (dir) {
+		pr_dbg3("base dir: %s\n", dir);
+		dinfo->base_dir = xstrdup(dir);
+		free_comp_dir(&comp_dirs);
+	}
+	else {
+		dinfo->base_dir = NULL;
 	}
 
 	for (i = 0; i < args->nr; i++)
@@ -1106,6 +1229,8 @@ static void release_debug_info(struct debug_info *dinfo)
 
 	free(dinfo->locs);
 	dinfo->locs = NULL;
+
+	free(dinfo->base_dir);
 
 	release_dwarf_info(dinfo);
 }
@@ -1273,6 +1398,8 @@ static void save_debug_entries(struct debug_info *dinfo,
 {
 	int i;
 	FILE *fp;
+	int idx = 0;
+	int len;
 
 	fp = create_debug_file(dirname, basename(filename));
 	if (fp == NULL)
@@ -1288,7 +1415,15 @@ static void save_debug_entries(struct debug_info *dinfo,
 			continue;
 
 		save_debug_file(fp, 'F', loc->sym->name, loc->sym->addr);
-		save_debug_file(fp, 'L', loc->file->name, loc->line);
+
+		if (dinfo->base_dir) {
+			len = strlen(dinfo->base_dir);
+			if (!strncmp(loc->file->name, dinfo->base_dir, len))
+				idx = len + 1;
+		}
+
+		/* skip common parts with compile directory  */
+		save_debug_file(fp, 'L', loc->file->name + idx, loc->line);
 
 		entry = find_debug_entry(&dinfo->args, loc->sym->addr);
 		if (entry)
@@ -1380,7 +1515,7 @@ static int load_debug_file(struct debug_info *dinfo, struct symtab *symtab,
 		case 'R':
 			if (line[0] == 'R')
 				root = &dinfo->rets;
-			
+
 			if (func == NULL)
 				goto out;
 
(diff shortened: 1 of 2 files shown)


ekyooo:
This is test result.
```
$ readelf -wi t-sort |grep comp
 <0><b>: Abbrev Number: 1 (DW_TAG_compile_unit)
    <15>   DW_AT_comp_dir    : (indirect string, offset: 0x90): /home/bingbong/soft/uftrace
```

before
```
$ uftrace record --srcline t-sort
$ cat uftrace.data/t-sort.dbg
F: 77a loop
L: 3 /home/bingbong/soft/uftrace/tests/s-sort.c
F: 7a1 foo
L: 10 /home/bingbong/soft/uftrace/tests/s-sort.c
F: 7bd bar
L: 17 /home/bingbong/soft/uftrace/tests/s-sort.c
F: 7ee main
L: 25 /home/bingbong/soft/uftrace/tests/s-sort.c
```

after
```
$ ./uftrace record --srcline t-sort
$ cat uftrace.data/t-sort.dbg
F: 77a loop
L: 3 tests/s-sort.c
F: 7a1 foo
L: 10 tests/s-sort.c
F: 7bd bar
L: 17 tests/s-sort.c
F: 7ee main
L: 25 tests/s-sort.c
```

```
$ ./uftrace report --srcline
  Total time   Self time       Calls  Function [Source]
   13.146 ms    1.701 ms           1  main [tests/s-sort.c:25]
   10.714 ms  574.162 us           1  bar [tests/s-sort.c:17]
   10.140 ms   10.140 ms           1  usleep
  730.115 us   12.331 us           2  foo [tests/s-sort.c:10]
  717.784 us  717.784 us           6  loop [tests/s-sort.c:3]
    2.892 us    2.892 us           1  __monstartup
    1.048 us    1.048 us           1  __cxa_atexit
```

utils/dwarf.c
```diff
@@ -907,6 +907,25 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
 	if (dfile == NULL)
 		return;
 
+	Dwarf_Die cudie;
+	dwarf_diecu(die, &cudie, NULL, NULL);
```
ekyooo, line 911: I remember the coding style of uftrace about declaration you said before. 
but I followed style of previous code of this function. give me any opinion.

utils/dwarf.c
```diff
@@ -907,6 +907,25 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
...
+
+	if (dwarf_hasattr(&cudie, DW_AT_comp_dir)) {
+		char *comp_dir = str_attr(&cudie, DW_AT_comp_dir, false);
+
+		char *d = comp_dir;
+		char *f = dfile->name;
+		while (*d && *f && *d == *f) {
```
ekyooo, line 918: I think we can compare the string once per compilation unit DIE and reuse it for all children for efficiency.
Please give me your opinion.

honggyukim:
Thanks for doing this work. I will take a look when I have some time.

utils/dwarf.c
namhyung: I still prefer declaration at the beginning but there are some places violate it.  Please do not add more.

utils/dwarf.c
namhyung: I'm worry about a big program which might have more than one compilation directories.  In that case we may need to find shortest one and use it for all the places.  So my idea is that do not compare strings here and do it when all srcline info is collected - like in `build_dwarf_info()`.

namhyung: And I think we can simple use `strncmp()`?

utils/dwarf.c
```diff
@@ -907,6 +907,25 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
...
+			d++;
+			f++;
+		}
+		if (!*d) {
+			char *relative_name = xstrdup(dfile->name + (ptrdiff_t)(d - comp_dir + 1));
```
namhyung, line 923: `ptrdiff_t` seems not necessary.

utils/dwarf.c
ekyooo: > about a big program which might have more than one compilation directories. In that case we may need to find shortest one and use it for all the places. 

For understanding, what is shortest one in this case?
A program has 2 CU like as follows.
- first complication dir: /home/soft/uftrace/tests
- second complication dir: /bingbong/test/aaa

namhyung: Yes, that's a case what I'm worring about.  So my idea is like below:

1. save file location as is
2. collect all comp_dir info from all CUs
3. select most used one as a base dir

I think, in many cases, comp_dir will be a child directory of a base like "/home/xxx/uftrace", "/home/xxx/uftrace/tests", "/home/xxx/uftrace/libmcount", "/home/some/unrelated/path".  Then we can count "/home/xxx/uftrace" as 3 (it includes children) and select it.  I think we should not use a part of common directory (in this example "/home") and use a full comp_dir.

ekyooo: ok, you want to select 1 base dir for a binary to represent all src file name as a relative path of the base dir, right? (if it has common path with base dir)

If so, It'd be good to discuss the policy for situations below.

1. in case that there's no common full comp_dir.
> I think we should not use a part of common directory (in this example "/home") and use a full comp_dir.

There's no base dir in case below with the policy above. 
Do you agree with it for simplicity?
"/home/xxx/uftrace/cmds", "/home/xxx/uftrace/tests", "/home/xxx/uftrace/libmcount", "/home/some/unrelated/path"

2. in case that used count of dir is same.
in the case like below, do you think we have to use the number of functions of each CU for these case?
- first complication 
dir: /home/soft/uftrace/tests
number of child functions: 1
- second complication 
dir: /bingbong/test/aaa
number of child functions: 100


3. comp_dir of another binary
With https://github.com/namhyung/uftrace/pull/1161, source line of linked library to executable can be shown.
Do you think base dir can be selected independently for each binary?

namhyung: Thanks for sharing your thoughts!

1. if there's no common dir, we can show full path as of now.
2. looks like a good idea.
3. each binary will have a separate base dir.

ekyooo: To "select the mostly used one as a base dir", I think we need a routine to compare comp_dir with other comp_dir. 
now I have a version using array for it simply, but time complexity of that could makes delay in big program.
in case of node, the number of comp_dir of main executable has 1673. 
I think the number could be bigger in bigger program. 
```
~/soft/node$ readelf -wi node|grep comp_dir|wc
   1673   13268  167457
```
so, do you think that we need to try to use hash map or trie for this?
there are some light opensource hash tables. followings are used on RTOS device with under 1MB ram.
https://github.com/troydhanson/uthash
https://github.com/rxi/map
Trie can be alternatives, but it could have space complexity issue as condition.

Could you give your opinion?
If available, could you also share the max condition(number of comp_dir / length or dir) you assume or watched?

namhyung: Hmm.. I don't think 1673 is too big and I guess they more mostly repeated.  We can sort and count to have an unique set.. did you already try that?  Or else, while we don't have hash map yet, you can use low-level rbtree APIs or take a look at (and help to get merge) #1051.

ekyooo: yes. node is an example. I thought you maybe want uftrace be used in SpaceX.
yes. I have primitive version to see just working simply. I share it anyway for communication.  It's using just array to insert and to search now, it has to be replaced with better data structure.
maybe hashmap could be better than rbtree in performance. but if you agree I'll use your rbtree for this PR, and then replace it with hashmap when #1051 is finished. 

about #1051, it seems that it could to be divided to PR for applying hashmap and PR for using hashmap. it was just an opinion.

utils/dwarf.c
```diff
@@ -1016,6 +997,69 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+}
+
+struct cu* map_get(struct cu *cands, int n_cands, char *name) {
+	int i = 0;
+	for (i = 0; i < n_cands; i++) {
+		if (!strcmp(cands[i].name, name))
+			return &cands[i];
+	}
+	return NULL;
+}
```
ekyooo, line 1019: this part is related to future hashmap or rbtree.

utils/dwarf.c
```diff
@@ -1016,6 +997,69 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+	struct cu *cands = (struct cands*)malloc(sizeof(struct cu) * n_cu);
+	struct cu *cand;
+	int n_cands = 0;
+
+	qsort(cus, n_cu, sizeof(char*), func_cmp);
+
+	for (int i = 0; i < n_cu; i++) {
+		char *cu_prefix = xstrdup(cus[i]);
+		cand = NULL;
+		while (cu_prefix && *cu_prefix) {
```
ekyooo, line 1039: now I think rbtree could be better than hashmap without while loop above when using rbtree.

namhyung: Agreed.  And we can add/count each comp_dir in `build_dwarf_info()` and compare the result here.

utils/dwarf.c
```diff
@@ -1064,13 +1112,31 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
...
 
+		// collect CU dir
+		if (dwarf_hasattr(&cudie, DW_AT_comp_dir)) {
+			char *comp_dir = str_attr(&cudie, DW_AT_comp_dir, false);
+			if (n_cu == n_buf - 1) {
+				n_buf += 20;
+				cus = (char**)realloc(cus, sizeof(char*) * n_buf);
+			}
+			cus[n_cu++] = comp_dir;
+		}
```
namhyung, line 1123: Please add it to a rbtree directly.

utils/dwarf.c
```diff
@@ -1016,6 +997,69 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+	qsort(cus, n_cu, sizeof(char*), func_cmp);
+
+	for (int i = 0; i < n_cu; i++) {
+		char *cu_prefix = xstrdup(cus[i]);
+		cand = NULL;
+		while (cu_prefix && *cu_prefix) {
+			struct cu *cand = map_get(cands, n_cands, cus[i]);
+			if (cand) {
+				cand->n++;
+				break;
```
namhyung, line 1043: I'm afraid that it depends on the timing when a shorter entry appears.  So I think it's better collect it with full name, and then calculate/aggregate counts at last.

namhyung:
I think you can add a unit test instead.  You don't need to use dwarf code for test, just separate common prefix handling functions so that we can test them with test inputs.

All you need to do is adding test functions with `TEST_CASE` macro under `#ifdef UNIT_TEST` at the end of the file.  Please refer other tests like in utils/filter.c file.

utils/dwarf.c
```diff
@@ -1054,9 +1153,26 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
...
+
 		curr = next;
 	}
 
+	base_dir = get_base_dir(&dirs);
+	if (base_dir) {
+		pr_dbg3("base dir: %s\n", base_dir);
+		dinfo->base_dir = xstrdup(base_dir);
+	}
+	else {
```
ekyooo, line 1170: get_base_dir return first node if all cu has same count like first unit test.
if you agree with this, I'll delete this else case.
thank you.

ekyooo:
unit test results.
```
[039] srcline_remove_common_prefix  : PASS
[040] srcline_remove_common_prefix2 : PASS
[041] srcline_remove_common_prefix3 : PASS
```

utils/dwarf.c
```diff
@@ -1007,6 +1103,9 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 	struct uftrace_pattern *ret_patt;
 	char *s;
 	int i;
+	struct rb_root dirs = RB_ROOT;
+	char *comp_dir;
+	char *base_dir;
```
namhyung, line 1108: I think you can use a single `dir` variable instead..

utils/dwarf.c
namhyung: I'm ok with that.

utils/dwarf.c
```diff
@@ -996,6 +998,100 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+			   (prev->nr_used == max->nr_used && prev->nr_locs > max->nr_locs))
+				max = prev;
+			prev = e;
+		}
+
+		rbnode = rb_next(rbnode);
+	}
+
+	if (prev->nr_used > max->nr_used ||
+	   (prev->nr_used == max->nr_used && prev->nr_locs > max->nr_locs))
```
namhyung, line 1089: Please factor out the comparison to a separate function.

utils/dwarf.c
```diff
@@ -1054,9 +1153,26 @@ static void build_dwarf_info(struct debug_info *dinfo, struct symtab *symtab,
 
 		dwarf_getfuncs(&cudie, get_dwarfspecs_cb, &bd, 0);
 
+		if (dwarf_hasattr(&cudie, DW_AT_comp_dir)) {
+			comp_dir = str_attr(&cudie, DW_AT_comp_dir, false);
+			add_dir_entry(&dirs, comp_dir, dinfo->nr_locs_used);
```
namhyung, line 1158: Please copy the string here and free it in `free_dir_entry` so that the ownership can be explicit.

utils/dwarf.c
```diff
@@ -996,6 +998,100 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+	while (!RB_EMPTY_ROOT(root)) {
+		node = rb_first(root);
+		entry = rb_entry(node, typeof(*entry), node);
+
+		rb_erase(node, root);
+		free(entry);
+	}
+}
+
+char* get_base_dir(struct rb_root *dirs)
```
namhyung, line 1059: I think the names are too generic.  Also this can be a static function..

How about below?
* add_comp_dir()
* free_comp_dir()
* get_base_comp_dir()

utils/dwarf.c
```diff
@@ -1288,7 +1407,12 @@ static void save_debug_entries(struct debug_info *dinfo,
 			continue;
 
 		save_debug_file(fp, 'F', loc->sym->name, loc->sym->addr);
-		save_debug_file(fp, 'L', loc->file->name, loc->line);
+
+		if (dinfo->base_dir) {
+			if (strstr(loc->file->name, dinfo->base_dir))
+				idx = strlen(dinfo->base_dir) + 1;
+		}
+		save_debug_file(fp, 'L', loc->file->name + idx, loc->line);
```
namhyung, line 1415: Please add a comment that it skips the common part.

utils/dwarf.c
```diff
@@ -996,6 +998,100 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
...
+
+	rbnode = rb_first(dirs);
+	max = rb_entry(rbnode, typeof(*e), node);
+	prev = max;
+	rbnode = rb_next(rbnode);
+
+	while (rbnode != NULL) {
+		e = rb_entry(rbnode, typeof(*e), node);
+
+		if (e->name == strstr(e->name, prev->name)) {
```
namhyung, line 1074: I'd like to use `strncmp()` since it guarantees to match at the beginning.

utils/dwarf.c
ekyooo: ok. thank you.

utils/dwarf.c
```diff
@@ -996,6 +998,100 @@ static int get_dwarfspecs_cb(Dwarf_Die *die, void *data)
 	return DWARF_CB_OK;
 }
 
+struct dir_entry {
+	struct rb_node	node;
+	char		*name;
+	int		nr_locs;
+	int		nr_used;
```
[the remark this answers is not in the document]
ekyooo: ok. thank you.

utils/dwarf.c
ekyooo: ok. thank you. but I copied the str in add_comp_dir to make pair with free_comp_dir as you did in add_debug_entry. please let me know if you have another opinion. thank you.

utils/dwarf.c
ekyooo: I removed else case. but I couldn't find the information that DW_AT_comp_dir is mandatory field in dwarf spec. if it's not mandatory, I have to check code in related functions. do you have related information? if we are not sure, I'll update related code to handle that case. thank you.

namhyung:
> I couldn't find the information that DW_AT_comp_dir is mandatory field in dwarf spec.

I don't know.  But I think we'd better being conservative.  Let's assume it can have no comp_dir at all.

ekyooo:
It's updated.
Thank you.

PR #1167, opened 2020-05-08
test: Enable sub-case with -finstrument-functions option in srcline tests

ekyooo:
.dbg files are generated for binary files built with -g and -finstrument-functions
options together according to the latest changes.

It doesn't need to skip the test for case above in test cases for srcline anymore,
so it is deleted.

### DIFF
--- a/tests/t068_filter_time_A.py
+++ b/tests/t068_filter_time_A.py
@@ -15,13 +15,6 @@ class TestCase(TestBase):
    2.107 ms [23157] | } /* main */
 """)
 
-    def build(self, name, cflags='', ldflags=''):
-        # cygprof doesn't support return value now
-        if cflags.find('-finstrument-functions') >= 0:
-            return TestBase.TEST_SKIP
-
-        return TestBase.build(self, name, cflags, ldflags)
-
     def setup(self):
         self.option  = '-t 1ms -R mem_alloc@retval '
         self.option += '-A mem_free@arg1 -A usleep@plt,arg1'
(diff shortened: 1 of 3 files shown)


tests/t068_filter_time_A.py
```diff
@@ -15,13 +15,6 @@ def __init__(self):
...
 """)
 
-    def build(self, name, cflags='', ldflags=''):
-        # cygprof doesn't support return value now
-        if cflags.find('-finstrument-functions') >= 0:
-            return TestBase.TEST_SKIP
-
-        return TestBase.build(self, name, cflags, ldflags)
-
     def setup(self):
```
ekyooo, line 18: Actually, this file is not related to description of this commit. 
I found this test passed without that function. 
I'll exclude this file if you want. please let me know.
Thank you.

ekyooo:
Before:
```
235 report_srcline      : OK OK OK OK OK SK SK SK SK SK
236 replay_srcline      : OK OK OK OK OK SK SK SK SK SK
```

After:
```
235 report_srcline      : OK OK OK OK OK OK OK OK OK OK
236 replay_srcline      : OK OK OK OK OK OK OK OK OK OK
```

tests/t068_filter_time_A.py
honggyukim: Yes, it can be in another patch.

honggyukim:
Thanks for finding this test. You can write the commit message as follows.

.dbg files are generated for binary files built with -g and -finstrument-functions
options together according to the ~~latest changes~~ following commit.

  c9945fe dwarf: Produce srcline in .dbg for -finstrument-functions

It gives more information about the commit you're mentioning.

tests/t068_filter_time_A.py
ekyooo: ok. I divided it.

PR #1171, opened 2020-05-18
tui: Show source location with --srcline option

ekyooo:
Show source location of each function with loaded debug_info.

'-g' and '-pg' options are required when compiling an executable
to have source line information like this.

  $ gcc -o t-abc -g -pg tests/s-abc.c

The '--srcline' option is required when using record command.

  $ uftrace record --srcline t-abc

The '--srcline' option is also required when using tui command.

  $ uftrace tui --srcline
  TOTAL TIME : FUNCTION
    8.909 us : (1) t-abc
    1.260 us :  ├─(1) __monstartup
             :  │
    0.179 us :  ├─(1) __cxa_atexit
             :  │
    7.470 us :  └─(1) main [tests/s-abc.c:26]
    5.522 us :    (1) a [tests/s-abc.c:11]
    4.912 us :    (1) b [tests/s-abc.c:16]
    4.176 us :    (1) c [tests/s-abc.c:21]
    0.794 us :    (1) getpid

  $ uftrace tui --report --srcline
  *Total Time   Self Time       Calls  Function
     7.470 us    1.948 us           1  main [tests/s-abc.c:26]
     5.522 us    0.610 us           1  a [tests/s-abc.c:11]
     4.912 us    0.736 us           1  b [tests/s-abc.c:16]
     4.176 us    3.382 us           1  c [tests/s-abc.c:21]
     1.260 us    1.260 us           1  __monstartup
     0.794 us    0.794 us           1  getpid
     0.179 us    0.179 us           1  __cxa_atexit


cmds/tui.c
```diff
@@ -1212,8 +1231,17 @@ static void win_display_graph(struct tui_window *win, void *node)
...
 
 		if (width < COLS) {
-			w = COLS - width;
-			printw("%-*.*s", w, w, curr->n.name);
+			if (curr->n.loc) {
+				printw("%s", curr->n.name);
+				width += strlen(curr->n.name);
+
+				asprintf(&str_loc, " [%s:%d]", curr->n.loc->file->name, curr->n.loc->line);
+				printw("%-*.*s", COLS - width, COLS - width, str_loc);
```
ekyooo, line 1239: I couldn't find function to print as gray color for tui. do you have?

utils/graph.c
```diff
@@ -202,10 +205,11 @@ static int add_graph_event(struct uftrace_task_graph *tg, size_t node_size)
...
 
 int graph_add_node(struct uftrace_task_graph *tg, int type, char *name,
-		   size_t node_size)
+		   size_t node_size,
+		   struct debug_location* loc)
 {
 	if (type == UFTRACE_ENTRY)
-		return add_graph_entry(tg, name, node_size);
+		return add_graph_entry(tg, name, node_size, loc);
 	else if (type == UFTRACE_EXIT)
```
ekyooo, line 213: now I don't know what UFTRACE_EXIT means exactly. is it right it doesn't need loc?

uftrace.c
```diff
@@ -1324,10 +1324,6 @@ int main(int argc, char *argv[])
 	if (opts.use_pager && opts.mode != UFTRACE_MODE_LIVE)
 		start_pager(pager);
 
-	/* the srcline info is used for TUI status line by default */
-	if (opts.mode == UFTRACE_MODE_TUI)
-		opts.srcline = true;
-
```
ekyooo, line 1330: please check this change.

cmds/tui.c
```diff
@@ -431,22 +433,29 @@ static int build_tui_node(struct uftrace_task_reader *task,
...
-			update_report_node(task, name, tg);
+			update_report_node(task, name, tg, NULL);
 		else if (addr != EVENT_ID_PERF_SCHED_OUT)
 			return 0;
 	}
 	else  /* rec->type == UFTRACE_LOST */
 		return 0;
 
-	graph_add_node(tg, rec->type, name, sizeof(struct tui_graph_node));
+	if (opts->srcline && loc == (void*)-1)
```
ekyooo, line 455: I know it's weird. it's used to call task_find_loc_addr once by checking 1 variable.
could you give me your opinion?

cmds/tui.c
honggyukim: You can find the following code in `utils/debug.c`
```c
#define TERM_COLOR_RESET        "\033[0m"
#define TERM_COLOR_GRAY         "\033[90m"
```
`pr_gray` is basically like.
```c
pr_out(TERM_COLOR_GRAY "Hello" TERM_COLOR_RESET);
```

cmds/tui.c
```diff
@@ -1212,8 +1231,17 @@ static void win_display_graph(struct tui_window *win, void *node)
 		}
 
 		if (width < COLS) {
-			w = COLS - width;
-			printw("%-*.*s", w, w, curr->n.name);
+			if (curr->n.loc) {
+				printw("%s", curr->n.name);
+				width += strlen(curr->n.name);
+
+				asprintf(&str_loc, " [%s:%d]", curr->n.loc->file->name, curr->n.loc->line);
```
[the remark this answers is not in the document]
ekyooo: Did you get crash when running uftrace tui without --srcline option?
that case makes crash due to my fault of last update. I'll fix it. thank you.
if you get crash with --srcline option, please let me know. thank you.

cmds/tui.c
honggyukim: Sorry, it was wrong info. You can find the color setup at `init_colors` function that calls `init_pair`. Please see the following documents.
- https://linux.die.net/man/3/init_pair
- https://www.linuxjournal.com/content/programming-color-ncurses

cmds/tui.c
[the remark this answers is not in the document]
honggyukim: It crashes regardless of having `--srcline` option. I found that `curr->n.loc` value is wrong for reasons.
```
(gdb) p curr->n.loc
$3 = (struct debug_location *) 0xffffffffffffffff
```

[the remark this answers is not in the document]
honggyukim: I recommend you to record more complicated program rather than testing with `t-abc` only.

uftrace.c
honggyukim: This is needed because TUI always has to show source location info at the footer if available. We might have to change the type of `srcline` from `bool` to `int` and use this case as `1` and other explicit `srcline` option goes to `2`. It would be better to be some enum type that represents useful meaning.

cmds/tui.c
```diff
@@ -1413,7 +1442,17 @@ static void win_display_report(struct tui_window *win, void *node)
 	printw("%10lu", curr->n.call);
 	printw("  ");
 
-	printw("%-*.*s", COLS - width, COLS - width, curr->n.name);
+	if (curr->n.loc) {
+		printw("%s", curr->n.name);
+		width += strlen(curr->n.name);
+
+		asprintf(&str_loc, " [%s:%d]", curr->n.loc->file->name, curr->n.loc->line);
```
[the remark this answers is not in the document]
ekyooo: Thank you. 
it's updated.

cmds/tui.c
ekyooo: Thank you for your information!
I applied it. In case of terminal that doesn't support color change, white is used instead. 
Unfortunately, my terminal doesn't support change of color. so I couldn't see the gray color.
Could you check if it shows as gray color if your terminal support it? (of course if you have time :) )
Thank you.

cmds/tui.c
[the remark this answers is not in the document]
ekyooo: if you can't share information of the program you use, could you please check it with latest update again ?
I tested with node, but it's not produced with --srcline option. 
the address you found is -1, so I think It could be related with code that is removed in latest update. 
If it's still produced, I'll try to look up programs for producing it. 
Thank you.

uftrace.c
ekyooo: Do you mean change type of srcline in "struct opts" or apply it in tui?

namhyung:
Thanks for working on this.

In general, I want to make it possible to change the setting in runtime.  IOW, we can add a menu or hotkey to toggle srcline display.  Maybe the footer line always show srcline anyway but report and graph can be changed.

ekyooo:
It's updated to be toggled with key .
Please give your opinions for these.
1. current key for srcline is 'd'.
2. alternative srcline color is white when init_color is not supported in terminal.

utils/graph.c
namhyung: For graph processing, UFTRACE_ENTRY records add information about the function and UFTRACE_EXIT records update the execution time.  So I think it's ok to add the loc only to the ENTRY path.

cmds/tui.c
namhyung: I think we need to check whether current line has focus.

ekyooo:
> This is what I'm seeing now.. the gray color doesn't look good here..
ok. I'll look for terminal that support color change, and propose some colors like light gray, emerald, sky blue, hot pink....it maybe could be very retro. :(
if you have any color you want. please share it.

ekyooo:
I tested some colors. but there was no color that is displayed clearly in all terminal except white as I feel.
I tested many gray, yellow, silver, purple.. etc of page below. but a color that is displayed in a terminal is not displayed clearly in other terminal with different setting.
https://www.rapidtables.com/web/color/gray-color.html 

If It could be shown differently as user environment, how do you think using white for srcline, too?

test environment was as follows.
1. terminal in ubuntu gui with env variable TERM=xterm-256color
2. putty that is connected to 1.

namhyung:
Hmm.. frankly I don't see strong needs for this because tui already shows the srcline at the bottom.  But I'm not against your work - it'd be good for consistency.. and maybe we can display other info at the bottom instead.

That being said, I'm ok with just using white for srcline.

ekyooo:
I hope the master is clean with necessary codes.
It doesn't matter to close this if it's not needed.

@honggyukim has said needs to see srcline for all lines even if line doesn't have focus. also he said that it's good to check your opinion. so, if both agree, I'll simply close it.

Thank you.

cmds/tui.c
```diff
@@ -175,6 +179,15 @@ static void init_colors(void)
 	init_pair(C_GREEN,  COLOR_GREEN,  COLOR_BLACK);
 	init_pair(C_YELLOW, COLOR_YELLOW, COLOR_BLACK);
 	init_pair(C_RED,    COLOR_RED,    COLOR_BLACK);
+
+	if (init_color(COLOR_CYAN, 128, 128, 128) == ERR) {
```
honggyukim, line 183: It seems that this `init_color` always failed in my environment for some reasons.

ekyooo: Thank you for your opinion.
I tested it after changing environment variable (TERM=xterm-256color) as referring page below.
https://stackoverflow.com/questions/47686906/ncurses-init-color-has-no-effect
Thank you.

PR #1181, opened 2020-07-19
build: Add -Wdeclaration-after-statement compiler flag

honggyukim:
The -Wdeclaration-after-statement option makes a warning when a
declaration is found after a statement in a block.  It makes easier to
identify when breaking the coding style showing the following warning.
```
  warning: ISO C90 forbids mixed declarations and code
```
It makes some warnings in the existing code so this patch also fixes
those warnings together.


honggyukim:
I was worried that if `-Wdeclaration-after-statement` is not supported in old compilers, but confirmed that `gcc-4.6` is fine with it. I see some other compilation errors with `gcc-4.4` so couldn't test it.

namhyung:
We could add a config test to add the compiler option conditionally.  But I think it's fine for the most cases.  If we hear some complain about this later, we can make the change.

PR #1183, opened 2020-07-26
[tui] Add support for +line for neovim

lanza:
Unfortunately neovim doesn't pass this test as the name is `nvim`.
Simply add another test to check if `nvi` is the beginning of the
editor.


lanza:
Done.

PR #1185, opened 2020-08-27
chrome: Fix rust function name that has " inside

honggyukim:
In some rust functions, the name contains " mark and it makes uftrace
dump --chrome output cannot be loaded in chrome browsers due to parse
errors.

This patch fixes the problem by escaping it in the function name.

Before:
```
  {... ,"name":"_<extern "C" fn(A0) .> R as rusty_v8..support..CFnFrom<F>>::mapping::c_fn"},
```
After:
```
  {... ,"name":"_<extern \"C\" fn(A0) .> R as rusty_v8..support..CFnFrom<F>>::mapping::c_fn"},
```

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -855,11 +855,25 @@ static void dump_chrome_header(struct uftrace_dump_ops *ops,
 	chrome->last_comma = false;
 }
 
+static void escape_func_name(char *name_buf, const char *name)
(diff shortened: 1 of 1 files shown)


honggyukim:
The deno chrome can be found at https://uftrace.github.io/dump/deno.html.

Issue #1137, opened 2020-03-15
provide kind error messages for configure script


honggyukim:
As of now, the `configure` script fails on some newly installed systems as follows:
```
$ ./configure
./configure: line 168: make: command not found
./configure: line 169: make: command not found
uftrace detected system features:
...         prefix: /usr/local
...         libelf: [ OFF ] - more flexible ELF data handling
...          libdw: [ OFF ] - DWARF debug info support
...      libpython: [ OFF ] - python scripting support
...      libluajit: [ OFF ] - luajit scripting support
...    libncursesw: [ OFF ] - TUI support
...   cxa_demangle: [ OFF ] - full demangler support with libstdc++
...     perf_event: [ OFF ] - perf (PMU) event support
...       schedule: [ OFF ] - scheduler event support
...       capstone: [ OFF ] - full dynamic tracing support
```
The configuration is failed because the system doesn't have `make` installed.

We can provide better error messages on such cases.

It can also check whether `gcc` is installed as well and it fails silently even if `make` is installed.
```
$ ./configure
uftrace detected system features:
...         prefix: /usr/local
...         libelf: [ OFF ] - more flexible ELF data handling
...          libdw: [ OFF ] - DWARF debug info support
...      libpython: [ OFF ] - python scripting support
...      libluajit: [ OFF ] - luajit scripting support
...    libncursesw: [ OFF ] - TUI support
...   cxa_demangle: [ OFF ] - full demangler support with libstdc++
...     perf_event: [ OFF ] - perf (PMU) event support
...       schedule: [ OFF ] - scheduler event support
...       capstone: [ OFF ] - full dynamic tracing support
```

PR #1187, opened 2020-09-02
configure: check if make and gcc exist or not and change error message format

bavernet:
This is related issue #1137 

It checks if `make` and `${CC}` and adds `Error:` prefix to indicate explicit error.


namhyung (approved):
LGTM

PR #1192, opened 2020-09-13
tests: Fix tests t184_arg_enum and t246_report_srcline2

honggyukim:
This fixes broken tests for t184_arg_enum and t246_report_srcline2.

The test 184 arg_enum fails because the file descriptor number is
different from the expected number, but this patch does post processing
to make it just 'FD' for better comparison.
```
  t184_arg_enum: diff result of -pg -O0
  --- expect      2020-09-13 15:08:24.164255538 +0100
  +++ result      2020-09-13 15:08:24.164255538 +0100
  @@ -2,7 +2,7 @@
      foo() {
  -     open("/dev/zero", O_RDONLY) = 4;
  -     mmap(0, 4096, PROT_READ, MAP_ANON|MAP_PRIVATE, 4, 0) = 0xADDR;
  +     open("/dev/zero", O_RDONLY) = 6;
  +     mmap(0, 4096, PROT_READ, MAP_ANON|MAP_PRIVATE, 6, 0) = 0xADDR;
        mprotect(0xADDR, 4096, PROT_NONE) = 0;
        munmap(0xADDR, 4096) = 0;
  -     close(4) = 0;
  +     close(6) = 0;
      } /* foo */

  184 arg_enum            : NG
```
The test 246 fails because 'lib_a' is captured twice, but one of them is
captured by a library call, but summarized into the same function result
in report output.  So the order can be different for each execution
depending on the time spent on the library call, so let's just ignore
library call for the test.
```
  t246_report_srcline2: diff result of -pg -O0
  --- expect      2020-09-13 15:08:42.112217325 +0100
  +++ result      2020-09-13 15:08:42.112217325 +0100
  @@ -1,4 +1,4 @@
  -2 lib_a
   1 main s-libmain.c:16
   1 foo s-libmain.c:11
  +2 lib_a
   1 lib_b s-lib.c:15

  246 report_srcline2     : NG
```


namhyung (commented):
Looks good, but could you please split the commit?

honggyukim:
I thought that creating a merge commit only for two commits are not needed, but will split and update again.

honggyukim:
I've splitted it into two patches. Thanks.

Issue #1154, opened 2020-04-10
incorrect enum display problem


honggyukim:
Here is a simple program.
```c
$ cat enum.c
enum et {
        MINUS_ONE = -1,
        ZERO,
        ONE,
        TWO,
        TREE,
};

void foo(enum et e) { }

int main()
{
        foo(MINUS_ONE);
        return 0;
}
```
The enum type `et` begins with `-1`. In this case, argument display shows the enum incorrectly as follows.
```
$ gcc -pg -g enum.c

$ uftrace record -a a.out
$ uftrace replay
# DURATION     TID     FUNCTION
   1.134 us [ 17360] | __monstartup();
   0.427 us [ 17360] | __cxa_atexit();
            [ 17360] | main() {
   1.130 us [ 17360] |   foo(TREE|TWO|ONE|ZERO|MINUS_ONE+0xfffffffa);
   1.594 us [ 17360] | } = 0; /* main */
```
The `dump` output shows the problem as follows.
```
$ uftrace dump -F foo -v
        ...
reading 17360.dat
2602960.655323809  17360: [entry] foo(40061f) depth: 1
 <0000000000000000>: a1 92 39 01 61 3f 09 00  6c 00 1f 06 40 00 00 00
2602960.655323809  17360: [args ] length = 8
  args[0] enum et: TREE|TWO|ONE|ZERO|MINUS_ONE+0xfffffffa (reading 17360.dat
2602960.655323809  17360: [entry] foo(40061f) depth: 1
 <0000000000000000>: a1 92 39 01 61 3f 09 00  6c 00 1f 06 40 00 00 00
2602960.655323809  17360: [args ] length = 8
  args[0] enum et: TREE|TWO|ONE|ZERO|MINUS_ONE+0xfffffffa (4294967295)
 <0000000000000010>: ff ff ff ff 00 00 00 00)
 <0000000000000010>: ff ff ff ff 00 00 00 00
        ...
```
The enum is interpreted as `4294967295` instead of `-1` here so it needs to be fixed correctly.

honggyukim:
It might be a problem of the logic in the following code.
```c
char * convert_enum_val(struct enum_def *e_def, long val)
{
        struct enum_val *e_val;
        char *str = NULL;

        /* exact match? */
        list_for_each_entry(e_val, &e_def->vals, list) {
                if (e_val->val == val)
                        return xstrdup(e_val->str);
        }

        /* if not, try OR-ing bit flags */
[...]

PR #1194, opened 2020-09-13
dwarf: Fix the incorrect enum display problem

honggyukim:
The enum value can contain a negative value, but it's incorrectly
displayed in the current implementation.

Let's say there is a simple program as follows.
```c
  $ cat enum.c
  enum et {
          MINUS_ONE = -1,
          ZERO,
          ONE,
  };

  void foo(enum et e) {}

  int main() {
          foo(MINUS_ONE);
          return 0;
  }
```
The tracing output with auto-args is as follows.
```
  $ gcc -pg -g enum.c
  $ uftrace -a a.out
              [ 56519] | main() {
     4.446 us [ 56519] |   foo(ONE|ZERO|MINUS_ONE+0xffffffff);
     4.957 us [ 56519] | } = 0; /* main */
```
The size of enum by default is 4, but uftrace currently expects the
sizes is to be sizeof(long) so 8.

So the negative value -1 in enum type is interpreted as 8 bytes so the
it reads more bytes than stored, but it can be interpreted incorrectly.

This patch assumes the size of enum is 4 so type casted as int type and
it fixes the problem.
```
  $ uftrace -a a.out
              [ 22103] | main() {
     9.033 us [ 22103] |   foo(MINUS_ONE);
    11.284 us [ 22103] | } = 0; /* main */
```
Fixed: #1154

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -714,12 +714,12 @@ void get_argspec_string(struct uftrace_task_reader *task,
 
 			map = find_map(&s->symtabs, task->rstack->addr);
 			if (map == NULL || map->mod == NULL) {
-				print_args("<ENUM?> %lx", val.i);
+				print_args("<ENUM?> %x", (int)val.i);
 				goto next;
 			}
 
 			dinfo = &map->mod->dinfo;
-			estr = get_enum_string(&dinfo->enums, spec->enum_str, val.i);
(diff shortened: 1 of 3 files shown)


namhyung (approved):
While enum constants can be large as 8 bytes, I know that it's almost always 4 bytes.  I didn't carefully handle negative numbers so ok to change it for now.

PR #1204, opened 2020-09-26
build: Hide stderr if python3-config not found

honggyukim:
We get the following error messages when the system doesn't have
`python3-config` installed.
```
  $ ./configure
  /bin/sh: 1: python3-config: not found
  /bin/sh: 1: python3-config: not found
          ...
```
So it'd be better to hide the stderr output to `/dev/null`.


namhyung (approved):
LGTM

PR #636, opened 2018-12-27
Check/uftrace print

honggyukim:
uftrace is able to capture function arguments and return values and that is the core feature.

But sometimes, it's also necessary to see some internal variables or even user may want to put a marker to distinguish the point from a long sequence of function calls.

This PR is an implementation of #428 so please see the issue for more details.

The usage is as follows:
1. Include "uftrace_print.h" in the target source code.
2. Modify the source code by inserting uftrace_print() function with a
    single argument that you want to record with its function flow.
3. Run "uftrace record" for the target program.
4. Run "uftrace replay" to observe the values.

Here is an example. Let's say `tests/s-arg.c` is modified as follows:
```diff
$ diff -u tests/s-arg.c tests/s-arg-uftrace_print.c
--- tests/s-arg.c     2018-12-27 22:23:42.142927670 +0900
+++ tests/s-arg-uftrace_print.c       2018-12-27 21:12:46.590220776 +0900
@@ -6,6 +6,7 @@
 #include <stdarg.h>
 #include <stdlib.h>
 #include <string.h>
+#include <uftrace_print.h>

 const char *strs[] = { "a", "b", "c" };
 const unsigned nr_strs = sizeof(strs)/sizeof(strs[0]);
@@ -61,6 +62,13 @@

 int check(struct big val, struct big *ref)
 {
+       uftrace_print(val.c);
+       uftrace_print(ref->d);
+       uftrace_print(val.s);
+       uftrace_print(ref->ld);
+       uftrace_print(val.i);
+       uftrace_print(ref->ull);
+       uftrace_print(val.l);
        if (val.c != ref->c)
                return -1;
        if (val.d != ref->d)
@@ -81,6 +89,7 @@
 int pass(int n)
 {
        struct big b = { 'b', 3.14, -1, 2.71828, n, 987654321ULL, 12345L };
+       uftrace_print(&b);
        return check(b, &b);
 }
```
Then, if it's traced with uftrace, it records the parameter values of uftrace_print as follows:
```
$ gcc -pg -g tests/s-arg-uftrace_print.c
$ uftrace -a -F pass a.out
# DURATION     TID     FUNCTION
            [129490] | pass(3) {
 319.995 us [129490] |   uftrace_print_pv("&b", 0x7ffeddcd84c0);
            [129490] |   check(0x7ffeddcd84c0, 0x15ca470) {
   0.470 us [129490] |     uftrace_print_c("val.c", 'b');
[...]

PR #1205, opened 2020-09-29
script: Support native C/C++ shared object scripting

honggyukim:
This patch adds a native script support that can use a shared object as
a script plugin.

The basic interface is same as python and luajit, but the benefit of
this native script is that it doesn't add any overhead from the
scripting engines.  It just bypasses the execution flow into the
dynamically loaded shared object.

The target shared object has to be compiled from C/C++ programs that
provides the following functions.
```
  void uftrace_begin(void);
  void uftrace_entry(struct script_context *sc_ctx);
  void uftrace_exit(struct script_context *sc_ctx);
  void uftrace_end(void);
```
The 'struct script_context' is same as the struct used inside uftrace so
it's just passed without any manipulation.

The example C/C++ files are in sctipts directory and the shared objects
can be compiled as follows.
```
  $ make -C scripts/
  make: Entering directory '/home/honggyu/work/uftrace/scripts'
  gcc -fPIC -shared -o simple_c.so simple.c
  gcc -fPIC -shared -o simple_cpp.so simple.cpp
  make: Leaving directory '/home/honggyu/work/uftrace/scripts'
```
The basic usage is as follows.
```
  $ uftrace record -S scripts/simple_c.so --no-libcall t-abc
  program begins...
  entry : main()
  entry : a()
  entry : b()
  entry : c()
  exit  : c()
  exit  : b()
  exit  : a()
  exit  : main()
  program is finished
```
The output looks exactly same as python and luajit scripts.


namhyung:
Yes, I like it.  Actually I thought about this some time ago, but didn't find real use cases.  And we should consider argument and other info for proper usage.

scripts/simple.c
```diff
@@ -0,0 +1,37 @@
+#include <stdio.h>
+#include <stdint.h>
+
+struct script_context {
```
mattyclarkson, line 4: Might be worth putting this into a header that can be easily included in a plugin?

Would we need any versioning of the interface?

honggyukim: Yes, we can think about it. I've just put the struct as it looks very simple to copy and paste.

My concern is about having `struct list` for argument info because it adds a dependency on the internal list implementation. It's not always required so user can decide not to use and remove the dependency. I will think more about it. Thanks for the comment.

honggyukim:
I will also think about providing a way to disable the uftrace shmem record behaviour when we only need to use this native script interface.

scripts/simple.c
mattyclarkson: > looks very simple to copy and paste.

Depends if we're ever going to change the structure. I guess as long as we put new members at the end we will be fine.

Depends if you think it would be useful for plugins to know the argument types? Can always follow up with that later.

In the future we may want to have a `uftrace/plugin.h` (or similar) that includes all the necessary headers for parsing the argument and this structure. For now, it's likely fine to ignore the arguments.

honggyukim: Thanks. `uftrace/plugin.h` looks good. We have to decide how to provide the header to users. I also prototyped another feature that needs to provide a header to user, but it's also pending now at https://github.com/namhyung/uftrace/pull/636.

I will think more about how to provide those headers.

namhyung: I agree that we should provide a header for API/ABI compatibility and also it's good to use a versioning.  Then we should consider installing/loading the header and plugin binaries as well.

namhyung:
Yes, it's reasonable when we run it during record and want to reduce the overhead.  It might be handled in `uftrace_begin` or using `--disable` option.

honggyukim:
> And we should consider argument and other info for proper usage.

I didn't use `struct script_info` yet in this implementation because it introduces another internal data structure `struct strv` so it will make the user to have more complicated header.

How about replace `struct strv cmds` to just `const char *cmds` as a concatenated string? I think we don't have to make it complicated unnecessarily and the users can parse the info themselves. I've applied this change at fe80d3f2.

scripts/simple.c
honggyukim: I've made a draft change at c2c22b8c regarding versioning and providing a header `uftrace-script.h`.

The C/C++ scripts can be written by including `uftrace-script.h` and be compiled. The version can be checked before using the struct. We can change the major version when the layout is change and the minor version can be changed when we added a new members at the end of struct.

It's just a draft version, but would like to hear how you think about this approach.

namhyung (changes requested):

utils/uftrace-script.h
```diff
@@ -0,0 +1,54 @@
...
+#endif
+
+/* informantion passed during initialization */
+#define SCRIPT_INFO_MAJOR_VERSION 1
+#define SCRIPT_INFO_MINOR_VERSION 0
+struct script_info {
+	struct si_version {
+		uint8_t major;
+		uint8_t minor;
+	} ver;
```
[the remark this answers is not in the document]
honggyukim: I thought about handling the major version change doesn't allow the previously compiled shared objects cannot be used.  But minor version change allows that the shared objects can be used in a higher version of uftrace.

In summary
- If major version is different, then uftrace rejects loading the shared object.
- If uftrace minor version is higher than the shared object, then it can use the shared object without a problem.
- If uftrace minor version is lower than the shared object, then uftrace rejects loading the shared object.

But I also think that we can simply use a single integer version because the layout will not be changed often and we don't actually distribute the plugin shared object binaries.

cmds/record.c
```diff
@@ -323,7 +323,7 @@ static void setup_child_environ(struct opts *opts, int argc, char *argv[])
 		int i;
 
 		for (i = 0; i < argc; i++)
-			args = strjoin(args, argv[i], "\n");
+			args = strjoin(args, argv[i], " ");
```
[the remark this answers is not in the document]
honggyukim: Ah. That's why `\n' was used here. I will change it back to `\n'.

check-deps/Makefile.check
```diff
@@ -35,6 +35,8 @@ ifneq ($(wildcard $(srcdir)/check-deps/have_libluajit),)
   COMMON_CFLAGS += $(shell pkg-config --cflags luajit)
 endif
 
+COMMON_CFLAGS += -DHAVE_LIBNATIVE
+
```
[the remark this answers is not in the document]
honggyukim: Sure. I will remove this.

utils/script-native.c
```diff
@@ -0,0 +1,109 @@
...
+static void *native_handle;
+
+static void (*__uftrace_begin)(struct script_info *info);
+static void (*__uftrace_entry)(struct script_context *sc_ctx);
+static void (*__uftrace_exit)(struct script_context *sc_ctx);
+static void (*__uftrace_end)(void);
+
+static int native_uftrace_begin(struct script_info *info)
+{
+	pr_dbg("%s()\n", __func__);
```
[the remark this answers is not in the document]
honggyukim: I've just used for debugging. I will remove this.

scripts/Makefile
```diff
@@ -1,6 +1,6 @@
 script:
 	gcc -I.. -fPIC -shared -o simple_c.so simple.c
-	gcc -I.. -fPIC -shared -o simple_cpp.so simple.cpp
+	g++ -I.. -fPIC -shared -o simple_cpp.so simple.cpp
```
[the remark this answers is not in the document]
honggyukim: ack.

scripts/info.c
```diff
@@ -1,29 +1,5 @@
...
-	unsigned long		address;
-	char			*name;
-#if 0
-	/* for arguments and return value */
-	int			arglen;
-	void			*argbuf;
-	struct list_head	*argspec;
-#endif
-};
+#include "utils/uftrace-script.h"
```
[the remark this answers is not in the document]
honggyukim: I will make `include` directory and put the headers that will be installed together. If you want to put the headers under `/usr/include/uftrace` directory, the name can be simpler like `/usr/include/uftrace/script.h`.

utils/uftrace-script.h
```diff
@@ -0,0 +1,54 @@
...
+#define UFTRACE_UFTRACE_SCRIPT_H
+
+#ifndef __cplusplus
+#include <stdint.h>
+#include <stdbool.h>
+#include "utils/list.h"
+#else
+#include <cstdint>
+#include <cstdbool>
+#endif
```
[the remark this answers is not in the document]
honggyukim: Sure. I will use same headers for both C/C++.

utils/uftrace-script.h
```diff
@@ -0,0 +1,54 @@
...
+#define SCRIPT_INFO_MAJOR_VERSION 1
+#define SCRIPT_INFO_MINOR_VERSION 0
+struct script_info {
+	struct si_version {
+		uint8_t major;
+		uint8_t minor;
+	} ver;
+	char			*name;
+	char			*version;
+	bool			record;
```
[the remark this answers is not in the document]
honggyukim: Yes, that's also possible. I also thought that it's wasteful to use a bool type here.

utils/uftrace-script.h
```diff
@@ -0,0 +1,54 @@
...
+	/* for arguments and return value */
+	int			arglen;
+	void			*argbuf;
+	struct list_head	*argspec;
+#else
+	/*
+	 * struct list_head internally uses 'new' as a variable name so
+	 * it's not possible to use in C++.
+	 */
+#endif
```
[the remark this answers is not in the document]
honggyukim: We may have to convert the `struct list_head` to this type. Could you explain why you used `char data[]` here?

Otherwise, how about modifying `utils/list.h` that can be compilable in C++, then distribute it together because uftrace also uses GPLv2 same as `list.h` anyway.

honggyukim:
I've updated this PR again as follows.
* aabc21d3 uftrace: Do not use C++ keywords as variable names
* b1d3ff22 script: Change script_info.cmds from struct strv to char*
* 306b5fd9 script: Move script_info and script_context out to include/script.h
* edeb132a script: Support native C/C++ shared object scripting
* ab138715 script: Add version info to script_info
* cb05efc9 script: Make native script disable uftrace record

utils/uftrace-script.h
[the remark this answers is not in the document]
honggyukim: > Maybe we can have 2 integers as bitmasks for them. One indicates current status - only the record bit is available for now.

For now, I've just made the existing `record` variable can be disabled by the given native script. If it's disabled by script, then uftrace do not actually record the data.

> The other can control the behavior like disable, or skip arguments. I think we should keep original behavior unless script changes the bitmasks.

I don't see any use case for disabling argument record because we can already do this with uftrace command. I will just leave it as is, but can bump up the script version when it's needed.

utils/uftrace-script.h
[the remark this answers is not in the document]
honggyukim: I've just made it to use a single uftrace version for both struct.

utils/script-native.c
```diff
@@ -0,0 +1,106 @@
...
+
+static int native_uftrace_end(void)
+{
+	__uftrace_end();
+	return 0;
+}
+
+static int native_atfork_prepare(void)
+{
+	fflush(stdout);
```
namhyung, line 49: It's already done in the `atfork_prepare_handler()`.

utils/script-native.h
```diff
@@ -0,0 +1,28 @@
+#ifndef UFTRACE_SCRIPT_NATIVE_H
+#define UFTRACE_SCRIPT_NATIVE_H
+
+#include "utils/filter.h"
+
+struct script_info;
+
+#ifdef HAVE_LIBNATIVE
```
namhyung, line 8: This is not needed.

include/script.h
```diff
@@ -5,8 +5,22 @@
...
+
+/* used only by external script files */
+#ifndef UFTRACE_SCRIPT_H
+#if __cplusplus
+extern "C"
+#endif
+int uftrace_script_version(void)
+{
+	return SCRIPT_VERSION;
+}
```
namhyung, line 18: A function definition in the header?  I'm afraid that a plugin can consist of multiple files.

cmds/script.c
```diff
@@ -133,6 +133,7 @@ int command_script(int argc, char *argv[], struct opts *opts)
 	struct uftrace_data handle;
 	struct uftrace_task_reader *task;
 	struct script_info info = {
+		.script_version = SCRIPT_VERSION,
```
namhyung, line 136: API version?  I don't want to repeat "script" in the script_info..

utils/script-native.c
```diff
@@ -66,6 +68,13 @@ static int load_native_api_funcs(const char *so_pathname)
 	}
 	pr_dbg("%s is loaded\n", so_pathname);
 
+	INIT_NATIVE_API_FUNC(uftrace_script_version);
+	if (SCRIPT_VERSION != __uftrace_script_version()) {
+		pr_warn("incompatible script versions (uftrace %d != module %d)\n",
+			SCRIPT_VERSION, __uftrace_script_version());
```
namhyung, line 74: Well.. I think we should have backward compatibility so old version of script can still be run.

scripts/info.c
```diff
@@ -9,6 +9,9 @@ void uftrace_begin(struct script_info *sc_info)
 	printf("%s\n", sc_info->version);
 	printf("%s\n", sc_info->name);
 	printf("%s\n", sc_info->cmds);
+
+	/* disable actual data record by uftrace */
+	sc_info->record = false;
```
namhyung, line 14: As I said before, I'd like to have two bitmasks to control the behavior.

namhyung (commented):
Thanks for your work!

honggyukim:
I've just pushed only after rebase to make the "force-pushed" to be clearer in the next update. I will update this PR a bit later.

check-deps/Makefile.check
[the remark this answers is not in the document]
honggyukim: It's now removed, but also we have to always enable script feature so added ca0c6e1e0229c1aa509b3e20cc71e54d87f37ffe.

include/script.h
```diff
@@ -0,0 +1,30 @@
...
+/* informantion passed during initialization */
+struct script_info {
+	char			*name;
+	char			*version;
+	bool			record;
+	const char 		*cmds;
+};
+
+/* context information passed to script */
+struct script_context {
```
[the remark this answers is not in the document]
honggyukim: It's also done.

Makefile
```diff
@@ -187,6 +187,7 @@ LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/debug.c $(srcdir)/utils/regs.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/rbtree.c $(srcdir)/utils/filter.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/demangle.c $(srcdir)/utils/utils.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/script.c $(srcdir)/utils/script-python.c $(srcdir)/utils/script-luajit.c
+LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/script-native.c
```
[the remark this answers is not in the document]
honggyukim: done

scripts/info.c
```diff
@@ -0,0 +1,24 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include "include/script.h"
```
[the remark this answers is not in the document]
honggyukim: If so, maybe we should put `script.h` at `uftrace/include/uftrace/script.h`, otherwise we cannot compile this before installation.

scripts/simple.cpp
```diff
@@ -0,0 +1,31 @@
+#include <iostream>
+#include <cstdint>
+#include "include/script.h"
+
+#if __cplusplus
+extern "C" {
+#endif
```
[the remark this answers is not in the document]
honggyukim: Yes, It's removed.

utils/script-native.c
```diff
@@ -0,0 +1,106 @@
+#ifdef HAVE_LIBNATIVE
```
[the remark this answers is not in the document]
honggyukim: It's removed.

utils/script-native.c
honggyukim: Right. I removed it.

cmds/script.c
honggyukim: I've changed it to `api_version`.

utils/script-native.c
honggyukim: Yes, it's right. But that's why I made the major and minor version to handle this before. If we want to keep the backward compatibility then we must not change the layout of struct, but we can only append the data to the struct in `include/script.h`.

scripts/info.c
honggyukim: I can make it to use two bitmasks, but I don't see if it's useful to control argument recording so left it for now. I can update it in the next update.

utils/uftrace-script.h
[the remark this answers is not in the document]
honggyukim: Thanks for the suggestion. I see that if we introduce another way of handling arguments then we better use this in `script-python.c` and `script-luajit.c` as well without using the current `struct list_head`.

It may take more time to refine them all so I didn't update it for now.

utils/script-native.h
honggyukim: It's removed now.

include/script.h
honggyukim: Hmm.. Otherwise, the script writer has to implement this. Could you recommend if you have a better idea?

honggyukim:
@namhyung I've updated this patch again with the following changes.
- Split `uftrace_script_context` into `uftrace_script_base_ctx` and `uftrace_script_args`.
- Remove `list.h` dependency by passing `uftrace_script_base_ctx` to native script APIs without argument info.

In addition, I've already made comments for API version handling and record bits so please have a look again. I actually don't have a better idea yet so please let me know if you have better ideas.

utils/script.h
```diff
@@ -31,20 +31,29 @@ struct uftrace_script_info {
...
+/* arguments and return value passed to script */
+struct uftrace_script_args {
 	int			arglen;
 	void			*argbuf;
 	struct list_head	*argspec;
 };
 
+/* context and args information passed to script */
+struct uftrace_script_context {
+	struct uftrace_script_base_ctx ctx;
```
namhyung, line 36: I'd like to rename it to 'base' so that I can refer to 'ctx->base.tid' or so.

Makefile
```diff
@@ -338,6 +340,9 @@ endif
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-fast.so $(DESTDIR)$(libdir)/libmcount-fast.so
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-single.so $(DESTDIR)$(libdir)/libmcount-single.so
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-fast-single.so $(DESTDIR)$(libdir)/libmcount-fast-single.so
+	$(call QUIET_INSTALL, headers)
+	$(Q)$(INSTALL) -m 644 $(srcdir)/include/script.h $(DESTDIR)$(incdir)/script.h
+	$(Q)$(INSTALL) -m 644 $(srcdir)/utils/list.h $(DESTDIR)$(incdir)/list.h
```
namhyung, line 346: We don't use it anymore, right?

utils/script.h
```diff
@@ -14,8 +14,6 @@
 #include "utils/script-luajit.h"
 #include "utils/utils.h"
 
-#define SCRIPT_ENABLED (SCRIPT_LUAJIT_ENABLED || SCRIPT_PYTHON_ENABLED)
-
```
namhyung, line 17: Maybe we can just make this 'true', then no need to change other code.

scripts/info.c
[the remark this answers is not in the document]
namhyung: I'm ok to put the header file there.  This script is an example that users can compile even without the uftrace source IMHO.  So it should work with the system-installed header.

scripts/Makefile
```diff
@@ -0,0 +1,7 @@
+script:
+	gcc -I.. -fPIC -shared -g -o simple_c.so simple.c
+	g++ -I.. -fPIC -shared -g -o simple_cpp.so simple.cpp
```
namhyung, line 3: No need to implement same one twice.  Please add other one like 'count.so'?

include/script.h
```diff
@@ -0,0 +1,25 @@
...
+struct uftrace_script_base_ctx {
+	int			tid;
+	int			depth;
+	uint64_t		timestamp;
+	uint64_t		duration;	/* exit only */
+	unsigned long		address;
+	char			*name;
+};
+
+#endif	/* UFTRACE_INCLUDE_SCRIPT_H */
```
namhyung, line 39: Don't we need 'extern "C"' here?

include/script.h
namhyung: I think it's script writer's responsibility to check the version number.

libmcount/mcount.c
```diff
@@ -74,6 +74,9 @@ static int __maybe_unused mcount_depth = MCOUNT_DEFAULT_DEPTH;
 /* boolean flag to turn on/off recording */
 static bool __maybe_unused mcount_enabled = true;
 
+/* boolean flag to turn on/off recording, but keep script working */
+bool __maybe_unused record_enabled = true;
```
namhyung, line 78: IIRC, mcount_enable can be used for this (hmm.. but it'd need this too)

```diff
diff --git a/libmcount/mcount.c b/libmcount/mcount.c
index a2310798..60b10945 100644
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -1232,7 +1232,7 @@ void mcount_exit_filter_record(struct mcount_thread_data *mtdp,
                        mtdp->record_idx--;
 
                if (!mcount_enabled)
-                       return;
+                       goto out;
 
                if (!(rstack->flags & MCOUNT_FL_RETVAL))
                        retval = NULL;
@@ -1276,6 +1276,7 @@ void mcount_exit_filter_record(struct mcount_thread_data *mtdp,
                                mtdp->nr_events = k;  /* invalidate sync events */
                }
 
+out:
                /* script hooking for function exit */
                if (SCRIPT_ENABLED && script_str)
                        script_hook_exit(mtdp, rstack);
```

scripts/info.c
namhyung: Let's do this in a separate PR.  I think it's better to focus on enabling the native script here.  Also other scripting like Python might use this feature too.

utils/script.h
```diff
@@ -32,20 +32,29 @@ struct uftrace_script_info {
...
+struct uftrace_script_args {
 	int arglen;
 	void *argbuf;
 	struct list_head *argspec;
 };
 
+/* context and args information passed to script */
+struct uftrace_script_context {
+	struct uftrace_script_base_ctx base;
+	struct uftrace_script_args args;
```
namhyung, line 38: I don't think this is needed.  Maybe we can add a separate script context definition for native script and leave this as is.

utils/script.h
```diff
@@ -29,7 +29,7 @@ struct uftrace_script_info {
 	char *name;
 	char *version;
 	bool record;
-	struct strv cmds;
+	const char *cmds;
```
namhyung: Let's not expose this to users.  Native scripts can have a separate header with your change.  Maybe argc + argv[] are more appropriate for native scripts.

utils/script-luajit.c
```diff
@@ -223,6 +223,9 @@ static int luajit_uftrace_begin(struct uftrace_script_info *info)
 {
 	int i;
 	char *s;
+	struct strv sv = {
```
namhyung, line 226: Please use `STRV_INIT`.

utils/script-python.c
```diff
@@ -572,6 +572,9 @@ int python_uftrace_begin(struct uftrace_script_info *info)
 	PyObject *ctx;
 	int i;
 	char *s;
+	struct strv sv = {
+		0,
+	};
```
namhyung, line 577: Ditto.

uftrace.c
```diff
@@ -1217,7 +1217,12 @@ void parse_script_opt(struct uftrace_opts *opts)
 	size_t len = 0;
 	static const char optname[] = "uftrace-option";
 	enum script_type_t script_type;
-	const char *comments[SCRIPT_TYPE_COUNT] = { "", "#", "--" };
+	const char *comments[SCRIPT_TYPE_COUNT] = {
+		"",
+		"#",
+		"--",
+		"//",
```
namhyung, line 1224: I think the trailing comma makes line break.

scripts/simple.cpp
```diff
@@ -0,0 +1,26 @@
...
+void uftrace_exit(struct uftrace_script_base_ctx *sc_ctx)
+{
+	std::cout << "exit  : " << sc_ctx->name << "()\n";
+}
+
+void uftrace_end(void)
+{
+	std::cout << "program is finished\n";
+}
+}
```
namhyung, line 26: This can be confusing.  Please add a comment like `// extern "C"`.

scripts/Makefile
```diff
@@ -0,0 +1,23 @@
...
+CXX := g++
+CFLAGS   := -fPIC -shared -g -I$(incdir)
+CXXFLAGS := $(CFLAGS)
+
+CC	= gcc
+SRCS	= simple.c info.c count.c dump.c simple.cpp
+CSO     = $(SRCS:%.c=%_c.so)
+CPPSO   = $(SRCS:%.cpp=%_cpp.so)
+
+all: $(CSO) $(CPPSO)
```
namhyung, line 14: This needs to be called from the top level Makefile.  But I think you should split the addition of examples to a separate commit.

namhyung (commented):
Thanks for working on this.

paranlee:
I thought it would be very useful to write a script to view CPU register information, and I'll work on that feature if it needs to be enhanced.

paranlee:
I've also rebased @honggyukim branch and test the rust plugin works well!
```
$ uftrace record -S  scripts/simple_rs.so --no-libcall tests/t-abc
uftrace_begin called scripts/simple_rs.so
uftrace_entry called with name: main
uftrace_entry called with name: a
uftrace_entry called with name: b
uftrace_entry called with name: c
uftrace_exit called with name: c
uftrace_exit called with name: b
uftrace_exit called with name: a
uftrace_exit called with name: main
uftrace_end called
```
- https://github.com/paranlee/uftrace/tree/review/native-script

```rust
use std::os::raw::c_char;
use std::ffi::CStr;

// Information passed during initialization
#[repr(C)] // Ensure C-compatible layout
pub struct UftraceScriptInfo {
	pub api_version: i32,
	pub name: *const c_char, // Use raw pointer for name and version
	pub version: *const c_char,
	pub record: bool,
	pub cmds: *const c_char,  // Use raw pointer for commands string
}

// Base context information passed to script
#[repr(C)] // Ensure C-compatible layout
pub struct UftraceScriptBaseCtx {
	pub tid: i32,
	pub depth: i32,
	pub timestamp: u64,
	pub duration: u64,
	pub address: usize, // Use usize for address on most platforms
	pub name: *const c_char, // Use raw pointer for name
}

#[no_mangle]
pub extern "C" fn uftrace_begin(info: &UftraceScriptInfo) {
	println!("uftrace_begin called {}", unsafe { CStr::from_ptr(info.name) }.to_string_lossy());
}

#[no_mangle]
pub extern "C" fn uftrace_entry(ctx: &UftraceScriptBaseCtx) {
	println!("uftrace_entry called with name: {}", unsafe { CStr::from_ptr(ctx.name) }.to_string_lossy());
}

#[no_mangle]
pub extern "C" fn uftrace_exit(ctx: &UftraceScriptBaseCtx) {
	println!("uftrace_exit called with name: {}", unsafe { CStr::from_ptr(ctx.name) }.to_string_lossy());
}

#[no_mangle]
pub extern "C" fn uftrace_end() {
	println!("uftrace_end called");
}
```

namhyung:
> I thought it would be very useful to write a script to view CPU register information

I'm not sure if it's useful.  The register content is volatile - IOW it's changed whenever you call a function (at least for some registers).

honggyukim:
> I've also rebased @honggyukim branch and test the rust plugin works well!

Thanks!

> I'll work on that feature if it needs to be enhanced.

I think current status is good but this PR is pending because we have many things to consider to keep the backward compatibility especially for argument handling. Maybe we better drop argument handling for now and let this merged without it.

namhyung:
> Maybe we better drop argument handling for now and let this merged without it.

Then it needs to be specified with API version.  I thought converting all data into a string and pass them once.  This would reduce the complexity to handle different types and number of arguments.  But the script should parse the argument (which may be fine since it can control the argument when recording).

yskelg:
This is especially great for uncertain post-mortem analysis or when we are unsure about which filtering parameters to narrow down.

In my opinion, the strength of this feature lies in the fact that uftrace currently cannot dynamically observe information at runtime, but with this, we can script specific parts as needed. Additionally, if we can implement such modules using Rust, C, or C++, which have lower performance overhead compared to other scripting languages, it allows for more flexible observation of the program's behavior at runtime. And once the source code is written as a module, I think it's great that anyone can share and reuse it.

@honggyukim's slides were a great inspiration for me. Thank you so much!
- https://uftrace.github.io/talks/uftrace_BHAS23_Arsenal.pdf
image
image
image
image

I believe that people will find better use cases than the ones I come up with, and that's how it will be.

honggyukim:
Thanks @yskelg. I would like to complete this feature but I haven't had much time for working on this. It's in my TODO list for a long time but didn't get a strong request as of yet.

I can't promise I will continue in the near future but will come back to this when I can find some time. Thanks.

PR #1211, opened 2020-10-18
Support patternized module name for dynamic patch

ParkHanbum:
uftrace provides pattern matching on function names However, that is not supported for module names.
so, I added it.

### DIFF
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -310,6 +310,13 @@ struct patt_list {
 	bool positive;
 };
 
+struct module_patt_list {
+	struct list_head list;
+	struct list_head func_patt;
+	struct uftrace_pattern module_patt;
+	char *module;
+};
+
 static bool match_pattern_list(struct list_head *patterns,
 			       struct uftrace_mmap *map,
 			       char *sym_name)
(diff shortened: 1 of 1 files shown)


honggyukim:
Hi @ParkHanbum, it'd be better if you post the way how we can reproduce the difference with a simple example using "Before" and "After".

ParkHanbum:
@honggyukim I should have done it, thank you for telling me. it was updated at first comment.

libmcount/dynamic.c
```diff
@@ -537,13 +537,22 @@ static int do_dynamic_update(struct symtabs *symtabs, char *patch_funcs,
...
+			free(pl->module);
+			free(pl);
+		}
 
-		list_del(&pl->list);
-		free(pl->module);
-		free(pl);
+		list_del(&ml->list);
+		free(ml->module);
+		free(ml);
```
namhyung, line 555: In general, I think it's better to put the release code along with the allocation code.  So we can easily find where they are and make sure not to create any leaks.

Also please consider adding an unit test for new functionalities.  It's also good to check memory errors with ASAN.

namhyung (changes requested):
I think we can simply change type of patt_list.module to struct uftrace_pattern, no?

Issue #189, opened 2017-10-20
Addition of a build system generator


elfring:
I suggest to reuse a higher level build system than [your current make script](https://github.com/namhyung/uftrace/blob/574f97660f8ea9eec39fcd91d74ea87a6fa9524c/Makefile#L1 "Update candidate") so that powerful checks for software features will become easier.
- [CMake](https://cmake.org/ "Building software with CMake")
- [Autotools](https://www.gnu.org/software/autoconf/autoconf.html#introduction "Building software with Autoconf")

namhyung:
Thanks for the suggestion.  But I'm not familiar with those tools so I worry about the cost of rewriting and maintenance.

elfring:
- Does [a tool like “autoscan”](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/autoscan-Invocation.html "Using autoscan to create configure.ac") show any software dependencies that you find worth for further considerations?
- How do you think about to support more development environments by [the reuse of build system generators](https://cmake.org/cmake/help/v3.10/manual/cmake-generators.7.html#cmake-generators "Listing of CMake generators")?

namhyung:
Yes, those are great features and I definitely need them in the long run.  But maintenace is also important and I don't have much time to study them for now.  Are you willing to volunteer? ;-)

gpollo:
Hi, I've been working with uftrace as part of my research. My code was using `meson` as the build system. Since the current build system is basically raw makefile, I though I'd give a try about converting uftrace to `meson`. The [meson](https://mesonbuild.com/) build system is relatively new, but multiple projects are or has switchted to it like [systemd](https://github.com/systemd/systemd), [pulseaudio](https://github.com/pulseaudio/pulseaudio), [radare2](https://github.com/radareorg/radare2) and [much more](https://mesonbuild.com/Users.html). Having previously worked with `cmake` extensively before, I was very pleased with `meson` clear syntax and how simple it is to use.

Anyway, I have an attempt [here on this branch](https://github.com/gpollo/uftrace/commit/2398086b794c1a6dae967e4f8cf96108ee954f9d). It doesn't support all of the original build options as of yet, but it shouldn't be hard. If there is interest, I wouldn't mind working on it more and sending a PR.

honggyukim:
Hi @gpollo, Thanks very much for bringing this issue again.

[...]

PR #1214, opened 2020-10-25
WIP: add initial support for meson build system

gpollo:
Ref: #189

In theory, all options from the old build system are supported, even cross-compilation as the is built-in `meson`.

An example build commands could be:
```
$ cd uftrace/ && mkdir build && cd build
$ meson ..
$ ninja
```

You can list build options and configure them:
```
$ cd uftrace/ && mkdir build && cd build
$ meson ..
$ meson configure
$ meson reconfigure -Dwithout-libpython=enabled -Dwithout-schedule=enabled
```

You can run tests:
```
$ cd uftrace/ && mkdir build && cd build
$ meson ..
$ ninja
$ ninja check-unit      # executes unit tests
$ ninja check-uftrace # executes runtest.py
```

For cross-compilation, you can read ([`meson`](https://mesonbuild.com/Cross-compilation.html) documentation. It involves creating a file with cross-compilation options. For example, this file can cross-compile ARM for the Raspberry-Pi 4.
```
[binaries]
c = ['/data/toolchain/arm-unknown-linux-gnueabihf/bin/gcc', '--sysroot', '/data/sysroot/rpi4-arm']
strip = ['/data/toolchain/arm-unknown-linux-gnueabihf/bin/strip']

[properties]
sys_root = '/data/sysroot/rpi4-arm'

[host_machine]
system = 'linux'
cpu_family = 'arm'
cpu = 'armv8'
endian = 'little'
```
If the cross-build libraries are in `sys_root`, the build will find them. The application can be configured and build:
```
$ cd uftrace/ && mkdir build && cd build
$ meson .. --cross ../meson_arm.txt \
          -Dwithout-libstdcxx=enabled \
          -Dwithout-perf=enabled \
          -Dwithout-schedule=enabled \
          -Dwithout-tests=enabled \
          -Dwithout-libpython=enabled
$ ninja
```

Finally, it should be possible to install:
```
$ cd uftrace/ && mkdir build && cd build
$ meson .. --prefix=/usr/local
$ ninja # can be skipped
$ meson install
```

I marked the PR as WIP as I believe more testing could be done on it. It is possible that I forgot some compiler flags for certain targets and cross-compiling supports could be improved/tested more.


namhyung:
Thanks a lot for doing this!

But the first impression is that it looks complicated.  Maybe we can have a wrapper Makefile calling meson + ninja internally for the basic usecases?

honggyukim:
@gpollo Thanks for doing this work! I've tried to build it with meson, but see the following build error.
```
$ mkdir build && cd build
$ meson ..
$ ninja
[54/227] Linking target libmcount.so
FAILED: libmcount.so
cc  -o libmcount.so libmcount.so.p/libmcount_dynamic.c.o libmcount.so.p/libmcount_event.c.o libmcount.so.p/libmcount_mcount.c.o libmcount.so.p/libmcount_misc.c.o libmcount.so.p/libmcount_plthook.c.o libmcount.so.p/libmcount_pmu.c.o libmcount.so.p/libmcount_record.c.o libmcount.so.p/libmcount_wrap.c.o libmcount.so.p/utils_argspec.c.o libmcount.so.p/utils_auto-args.c.o libmcount.so.p/utils_debug.c.o libmcount.so.p/utils_demangle.c.o libmcount.so.p/utils_dwarf.c.o libmcount.so.p/utils_filter.c.o libmcount.so.p/utils_hashmap.c.o libmcount.so.p/utils_rbtree.c.o libmcount.so.p/utils_regs.c.o libmcount.so.p/utils_script-luajit.c.o libmcount.so.p/utils_script-python.c.o libmcount.so.p/utils_script.c.o libmcount.so.p/utils_symbol-libelf.c.o libmcount.so.p/utils_symbol-rawelf.c.o libmcount.so.p/utils_symbol.c.o libmcount.so.p/utils_utils.c.o -Wl,--as-needed -Wl,--no-undefined -shared -fPIC -Wl,--start-group -Wl,-soname,libmcount.so -lstdc++ /usr/lib/x86_64-linux-gnu/libcapstone.so -ldl -ldw -lelf /usr/lib/x86_64-linux-gnu/libluajit-5.1.so -lpthread -lrt -Wl,--whole-archive libmcount-arch.a -Wl,--end-group -Wl,--no-whole-archive
/usr/bin/ld: libmcount-arch.a(arch_x86_64_dynamic.S.o): relocation R_X86_64_PC32 against symbol `mcount_entry' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
[59/227] Compiling C object libmcount-fast.so.p/utils_demangle.c.o
ninja: build stopped: subcommand failed.
```
It seems that the related part is as follows.
```meson
libmcount_arch = static_library(
    'mcount-arch', [libmcount_arch_sources, libmcount_arch_asm_sources, version_h],
    include_directories: [libmcount_arch_include],
    c_args: [common_cflags],
    pic: true,
)

libmcount_arch_dep = declare_dependency(
    include_directories: [libmcount_arch_include],
    # we link whole archive to overwrite weak symbols
    link_args: ['-Wl,--whole-archive', 'libmcount-arch.a' , '-Wl,--no-whole-archive']
)
```
It seems that it already builds `libmcount-arch.a` with pic enabled, but doesn't work as expected.

gpollo:
@namhyung That's a good idea, I'll try to add some basic target in the current Makefile.

@honggyukim Building the arch stuff in a separate lib should not longer be needed. I think I did it this way initially because `meson` do not supports [per-object compilation flags](https://github.com/mesonbuild/meson/issues/3345). I though that these files needed specific flags. I'll do some test and probably move them directly in the `shared_library` where they are needed.

namhyung:
Also you can change misc/install-deps.sh to have meson and ninja.

gpollo:
Rebased the branch and remove the intermediate `libmcount_entry` static library.

@honggyukim Turns out your build error was caused by `-fvisibility=hidden` not used. I believe that Arch Linux has this option activated by default. When I tested on a Ubuntu 18.04 container, I had the same error as you. I compared the old build system commands with the new one and I found this missing option. I should work now.

honggyukim:
@gpollo Thanks for the update. I've tested it again and see that the build is fine this time. But I see the default build generates `unittests` binary unnecessarily.

And please properly rebase the patch on top of the current master. It seems that your last patch is on top of a wrong commit hash.

gpollo:
* Rebased on `master`.
* Optionally build `dbginfo` and `demangler` binary with `-Dbuild-dbginfo=enabled` and `-Dbuild-demangler=enabled`
* Don't build tests by default. If `ninja check-unit` or `ninja check-uftrace` is ran, a message is written to tell to reconfigure the build with `-Dwith-tests=enabled`.

honggyukim:
Thanks a lot for the update! It looks good now, but I have a few more requests.
1. I think `with-tests`, `build-dbginfo`, and `build-demangler` don't have to be in `meson_options.txt`. Rather than that, they can be `ninja` target using `run_target`. They can be compiled like `ninja unittests`, `ninja dbginfo` and `ninja demangler`.
2. We also need to handle runtime variables just like we did `make ASAN=1`, `make COVERAGE=1`. They are build options that enable address sanitizer and gcov for code coverage.

honggyukim:
In other words, I think `meson` command should be the replacement of `configure` script and `ninja` should be for the current `make` command. So `ninja` is better to accept most of the usage that we provide with `make`.

gpollo:
Alright, did some modifications again, hopefully, I didn't broke anything, I'll be testing more the upcoming weeks. Mainly, 

* You are right, I remove `with-tests`, `build-dbginfo` and `build-demangler`. They are optional when you run `ninja dbginfo` or `ninja demangler`.
* Unit tests can be executed with `ninja tests` and uftrace tests can be executed with `ninja uftrace-tests`. Should they both be executed at the same time?
* I also renamed all options like `-Dwithout-capstone=` (default: disabled) to `-Dwith-capstone=` (default: auto). Doesn't change much except it makes more sense.
* For `make DEBUG=1`, `make ASAN=1` or `make COVERAGE=1`, I don't think we can have the equivalent with `ninja`. I put those at configuration time (e.g. `meson -Ddebug=true -Dasan=true -Dcoverage=false`, etc.).

honggyukim:
Thanks for the update. I see `version.h` missing error.
```
$ mkdir build && cd build
$ meson ..
      ...
$ ninja
[1/188] Generating gen_version_h with a custom command.
FAILED: version.h
/usr/bin/sh -c '/home/honggyu/work/uftrace/misc/version.sh version.h $(git describe --tags 2> /dev/null || echo v0.9.4) . > /dev/null'
[8/188] Compiling C object 'traceevent@sta/libtraceevent_event-parse.c.o'.
ninja: build stopped: subcommand failed.
```
I will have a look at the details and your comment when I have more time. Thanks.

gpollo:
I fixed it, commit 31d89ea added a new argument to `version.sh` and the generation failed. When I rebased against `master`. It kept the old one and my build wouldn't failed. Should be good now.

honggyukim:
Thanks for the update. It's getting closer to the complete set.

> You are right, I remove with-tests, build-dbginfo and build-demangler. They are optional when you run ninja dbginfo or ninja demangler.

There is `symbols` binary in `misc` so could you add `ninja symbols` as well? And I think we better build those binaries by default because `make` does it.

> Unit tests can be executed with ninja tests and uftrace tests can be executed with ninja uftrace-tests. Should they both be executed at the same time?

To make the behavior same as `make`, it'd be better to make `ninja unittest` does same thing with `make unittest`, `ninja runtest` runs same as `make runtest` and `ninja test` to run both unittest and runtest.

> I also renamed all options like -Dwithout-capstone= (default: disabled) to -Dwith-capstone= (default: auto). Doesn't change much except it makes more sense.

I'm fine with that, but is there any way to list up all the configurations to user? I'd expected `meson --help` to show it but there must be something else.

> For make DEBUG=1, make ASAN=1 or make COVERAGE=1, I don't think we can have the equivalent with ninja. I put those at configuration time (e.g. meson -Ddebug=true -Dasan=true -Dcoverage=false, etc.).

I see that `-Dsanitize=address`, `-Dcoverage=true` and other options do this work. ~But I prefer `-Ddebug=true` rather than `-Ddebugging=true` because we use `make DEBUG=1`.~ I found that `debug` is reserved so cannot use.

In addition, `meson -Dsanitize=address` hits a linking error so we have to add link flag as well.

So I would like to apply the following change although it didn't address all the points I mentioned.
```diff
diff --git a/meson.build b/meson.build
index 660e8562..c72e4f84 100644
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,8 @@ common_ldflags = []
 uftrace_cflags    = []
 libmcount_cflags  = []
 demangler_cflags  = []
+symbols_cflags    = []
+dbginfo_cflags    = []
 traceevent_cflags = []
 test_cflags       = []

@@ -41,6 +43,7 @@ if get_option('trace')
     trace_cflags       = ['-pg', '-fno-omit-frame-pointer']
     uftrace_cflags    += trace_cflags
     demangler_cflags  += trace_cflags
+    dbginfo_cflags    += trace_cflags
     traceevent_cflags += trace_cflags
     test_cflags       += trace_cflags
     # cannot add -pg to libmcount_cflags because mcount() is not reentrant
@@ -55,15 +58,29 @@ libmcount_cflags
     test_cflags      += coverage_cflags
 endif

+if get_option('asan')
+    asan_cflags = ['-O0', '-g', '-fsanitize=address,leak']
+    common_ldflags   += ['-fsanitize=address,leak']
+
+    uftrace_cflags    += asan_cflags
+    demangler_cflags  += asan_cflags
+    dbginfo_cflags    += asan_cflags
+    traceevent_cflags += asan_cflags
+    test_cflags       += asan_cflags
+endif
+
 if get_option('sanitize') != ''
     if get_option('sanitize') == 'all'
         sanitize_cflags = ['-O0', '-g', '-fsanitize=address,leak,undefined']
+        common_ldflags   += ['-fsanitize=address,leak,undefined']
     else
         sanitize_cflags = ['-O0', '-g', '-fsanitize=@0@'.format(get_option('sanitize'))]
+        common_ldflags   += ['-fsanitize=@0@'.format(get_option('sanitize'))]
     endif

     uftrace_cflags    += sanitize_cflags
     demangler_cflags  += sanitize_cflags
+    dbginfo_cflags    += sanitize_cflags
     traceevent_cflags += sanitize_cflags
     test_cflags       += sanitize_cflags
 endif
@@ -521,9 +538,9 @@ dbginfo_deps = [
 dbginfo = executable(
     'dbginfo', [dbginfo_sources],
     dependencies: dbginfo_deps,
-    c_args: [common_cflags],
+    c_args: [common_cflags, dbginfo_cflags],
     link_args: [common_ldflags],
-    build_by_default: false,
+    build_by_default: true,
 )

 #############
@@ -545,7 +562,41 @@ demangler = executable(
     dependencies: demangler_deps,
     c_args: [common_cflags, demangler_cflags],
     link_args: [common_ldflags],
-    build_by_default: false,
+    build_by_default: true,
+)
+
+###########
+# symbols #
+###########
+
+symbols_sources = [
+    'misc/symbols.c',
+    'utils/session.c',
+    'utils/demangle.c',
+    'utils/rbtree.c',
+    'utils/utils.c',
+    'utils/debug.c',
+    'utils/filter.c',
+    'utils/dwarf.c',
+    'utils/auto-args.c',
+    'utils/regs.c',
+    'utils/argspec.c',
+    'utils/symbol.c',
+    'utils/symbol-libelf.c',
+]
+symbols_deps = [
+    cxa_demangle_dep,
+    libdl_dep,
+    libdw_dep,
+    libelf_dep,
+]
+
+symbols = executable(
+    'symbols', [symbols_sources],
+    dependencies: symbols_deps,
+    c_args: [common_cflags, symbols_cflags],
+    link_args: [common_ldflags],
+    build_by_default: true,
 )

 #########
diff --git a/meson_options.txt b/meson_options.txt
index b7acf797..d088940d 100644
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -112,9 +112,16 @@ option(
     description: 'Enable code coverage',
 )

+option(
+    'asan',
+    type : 'boolean',
+    value : 'false',
+    description: 'Enable address sanitizer',
+)
+
 option(
     'sanitize',
     type : 'string',
     value : '',
     description: 'Specify address and thread sanitizer options',
-)
\ No newline at end of file
+)
```
@namhyung It'd be better if you could provide how you think.

honggyukim:
In addition, `-Dwith-libelf=disabled` has to disable `with-libdw` as well. So the following cannot be compiled.
```
$ meson -Dwith-libelf=disabled ../
      ...
Message:            libdw: [ on  ] - dwarf support
Message:           libelf: [ off ] - more flexible ELF data handling
      ...

$ ninja
      ...
build/../utils/dwarf.c:215: undefined reference to `gelf_getehdr'
collect2: error: ld returned 1 exit status
```

honggyukim:
I found one more. The `libmcount*.so` has to be installed directly under `lib` directory rather than `lib/x86_64-linux-gnu` to make it same as current status.
```
$ meson -Dprefix=`pwd`/out ..
 
$ ninja install

$ ls -R out/lib/
out/lib/:
x86_64-linux-gnu

out/lib/x86_64-linux-gnu:
libmcount-fast-single.so  libmcount-fast.so  libmcount-nop.so  libmcount-single.so  libmcount.so
```
@namhyung I actually think it's better structure and can find the `libmcount*.so` under its arch name as I suggested at #1032. How do you think about it?

gpollo:
* Applied your patch.
* Renamed `ninja tests` to `ninja unittest`, `ninja uftrace-tests` to `ninja runtest`.
* Cannot use `ninja test` for running both since it is a reserved keyword. I use `ninja tests` for now.

> Can we make the value to boolean? I think -Dwith-capstone=false is easier than -Dwith-capstone=disabled because it might be confusing between disabled and disable.

The problem with using `boolean` type is that it only allows 2 states (`true`/`false`). By using feature, there is 3 states (`enabled`/`disabled`/`auto`). By default, they are all `auto`. It allows us to enable the feature if the library is present on the system or disable the feature if it is not present. By forcing `-Dwith-capstone=enabled`, the build will fail if the library is not found.

An alternative solution using `boolean` would be to have the options like `-Dwithout-capstone=true` which would disable the feature and `-Dwithout-capstone=false` (e.g. by default) which would only enable the feature if it is the library is present. The difference is that we can't "force" the build to enable a feature and fail if the library is not present (it will only do so if the library is present).

> In addition, is there any way to list up all the configurations to user? I'd expected meson --help to show it but there must be something else.

You can use `meson configure` to list the options.

gpollo:
Sorry, I forgot some points.

> In addition, -Dwith-libelf=disabled has to disable with-libdw as well. So the following cannot be compiled.
Fixed it.

> I found one more. The libmcount*.so has to be installed directly under lib directory rather than lib/x86_64-linux-gnu to make it same as current status.
I think it depends on your system. For me (Arch Linux), the installation location was `lib`.

honggyukim:
@gpollo Thanks very much for the update. I have two more requests.
1. Could you add `meson -Dcc=clang` or `meson -Dcompiler=clang` so that we can select the compiler to gcc or clang?
2. Please add more descriptive commit message for your work with `Signed-off-by`. You can refer to other commits with `git log`.

I think it's almost done so @namhyung can take a look at the final review. I appreciate your work!

gpollo:
> Could you add meson -Dcc=clang or meson -Dcompiler=clang so that we can select the compiler to gcc or clang?

Turns out that it can be done using `CC=clang meson ..` during the build configuration. See this [page](https://mesonbuild.com/howtox.html#set-compiler).

> Please add more descriptive commit message for your work with Signed-off-by. You can refer to other commits with git log. It'd be better to show the usage for build options. I think your first message in this PR is a good example.

I didn't list the build options, be the command to list them is in the commit message.

meson.build
```diff
@@ -0,0 +1,800 @@
...
+    'uftrace-recv',
+    'uftrace-replay',
+    'uftrace-report',
+    'uftrace-script',
+    'uftrace-tui',
+    'uftrace',
+]
+
+if get_option('without-doc').disabled()
+    program_pandoc = find_program('pandoc', required: true)
```
namhyung, line 734: Can we make it optional and skip installing document if missing?

namhyung (commented):
Thanks for working on this!

It looks way better than before.  I have some comments and also have a problem when running tests.  Not sure if it's because the python version, my tests stopped at 170.  The master works fine.  Can you please take a look?

meson.build
```diff
@@ -0,0 +1,800 @@
...
+    libluajit_dep,
+    libncurses_dep,
+    libpthread_dep,
+    libpython3_dep,
+    librt_dep,
+]
+
+libmcount = shared_library(
+    'mcount', [libmcount_arch_sources, libmcount_arch_asm_sources, libmcount_sources, version_h],
+    include_directories: [libmcount_arch_include, libmcount_include],
```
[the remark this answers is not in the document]
gpollo: The top-level directory (where the `meson.build` is located) is always included implicitly in the include directories. Quick test shown that `libmcount_include` was not needed. I'll remove it.

meson.build
gpollo: Done. Also, replaced `without-doc` with `with-doc` (default: auto).

meson.build
```diff
@@ -0,0 +1,800 @@
...
+        common_cflags += '-DHAVE_LIBDW'
+    endif
+endif
+
+##############################
+# -Dwith-libelf build option #
+##############################
+
+libelf_is_enabled = false
+libelf_dep = cc.find_library('elf', required: get_option('with-libelf'))
```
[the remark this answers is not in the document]
gpollo: By default, it does not check for headers. But it is pretty easy by adding the `has_headers` argument, as specified in the [documentation](https://mesonbuild.com/Reference-manual.html#compiler-object).

By `grep`-ing in the code, `gelf.h`, `libelf.h` and `elfutils/libdw.h` seem required. I'll add them.

Should I also do the same for the other dependencies?

meson.build
```diff
@@ -0,0 +1,800 @@
...
+
+cc = meson.get_compiler('c')
+
+if get_option('arch') == 'auto'
+    arch = host_machine.cpu_family()
+elif get_option('arch') == host_machine.cpu_family()
+    arch = host_machine.cpu_family()
+else
+    error('host compiler `@0@` doesn\'t match host architecture `@1@`'.format(
+        host_machine.cpu_family(), get_option('arch')))
```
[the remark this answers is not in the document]
gpollo: I'm not entirely sure what you mean. I tested on a 32-bit machine, `host_machine.cpu_family()` gives `x86` which makes the build fail because we want `i386`.

Before the above checks, I rename `x86` to `i383` (see the commits when I'll push them).

gpollo:
I've pushed the changes in separate commit if you want to see exactly the change I made. When it is ready, I'll squash all of them into one.

meson.build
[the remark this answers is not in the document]
namhyung: I think so.  Btw how does it check the dependency?  Is it possible to use pkg-config to check the availability and to list needed flags?

meson.build
```diff
@@ -20,13 +20,18 @@ program_git = find_program('git', required: true)
 
 cc = meson.get_compiler('c')
 
+host_machine_cpu_family = host_machine.cpu_family()
+if host_machine_cpu_family == 'x86'
+    host_machine_cpu_family = 'i386'
```
namhyung, line 25: It'd be nice if 'x86' can cover both 'i386' and 'x86_64' (preferred) depending on the system (word-size).

meson.build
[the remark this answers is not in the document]
gpollo: I'm not sure how `cc.find_library()` works, I believe it does a test compile.

But I searched a little. Turns out I'm using `dependency` for other dependencies rather than `cc.find_library()`. The `dependency()` function will use `auto` method by default. It may choose `pkg-config` or `cmake` to find library. See the [docs](https://mesonbuild.com/Dependencies.html#dependency-method).

The `dependency()` can't verify for header existance. I don't think it is necessary is the dependency is found using `pkg-config`. If the library has a `pkg-config` it means it has been installed for development purpose and headers should be present.

Still, I'll change `libelf` and `libdw` to `dependency()` as it seems more robust.

meson.build
gpollo: If the system is 64-bit, `host_machine.cpu_family()` will return `x86_64`. If the system is 32-bit, it will return `x86`. Thus, I don't think it is necessary. Also, `host_machine` doesn't seem to have word size in it.

The standard `host_machine.cpu_family()` values for Meson are listed in this [page](https://mesonbuild.com/Reference-tables.html#cpu-families).

namhyung: I was thinking about user-given arch names like for cross-compilation.

gpollo: The `cpu_family` should only use standard values. If the user wants to be more specific, he will define `cpu` field in the cross compilation file:

```INI
[host_machine]
system = 'linux'
cpu_family = 'x86'
cpu = 'i586'
endian = 'little'
```

Since we use `cpu_family`, it shouldn't matter. See this [page](https://mesonbuild.com/Cross-compilation.html#machine-entries).

honggyukim:
@namhyung Is there any reason to hold this PR? I think it's mature enough and we can merge this then maintain it gradually. We can keep both `make` and `meson` for a while until we're sure meson can fully replace the current `make` build.

I see some values in terms of generating `compile_commands.json` and it can be used for using `clang-format` and `clang-tidy` in the future.

namhyung:
Oh, I'm really sorry about that.  I'll take a look and consider merging.

Yes I agree that we can keep the two build system for a while.  And I'd like to maintain the simple way of build/install/testing with `make`.

meson.build
```diff
@@ -0,0 +1,819 @@
+# TODO: cross-compile supports is rudimentary
+
+project('uftrace', 'c', version: '0.9.4')
```
namhyung, line 3: Need to update the version.

meson.build
```diff
@@ -0,0 +1,819 @@
...
+python_module = import('python')
+
+common_cflags = ['-D_GNU_SOURCE', '-D_DEFAULT_SOURCE', '-D_XOPEN_SOURCE=600']
+common_cflags += '-fvisibility=hidden'
+common_cflags += '-fno-omit-frame-pointer'
```
namhyung, line 8: Actually these two flags are for libmcount.

Common flags should be: -W -Wall -Wdeclaration-after-statement -Wno-unused-parameter -Wno-missing-field-initializers

meson.build
```diff
@@ -0,0 +1,819 @@
...
+# libmcount (arch specific) #
+#############################
+
+if arch == 'aarch64'
+    libmcount_arch_include = include_directories('arch/aarch64')
+    libmcount_arch_sources = [
+        'arch/aarch64/cpuinfo.c',
+        'arch/aarch64/mcount-dynamic.c',
+        'arch/aarch64/mcount-insn.c',
+        'arch/aarch64/mcount-support.c',
```
namhyung, line 300: Is it possible to use a wildcard match like arch/aarch64/mcount-*.c?

gpollo: Sadly, no. Using a glob pattern is not supported by Meson since it is generally considered a bad practice and may lead to build inefficiency or the unintended inclusion of files.

gpollo:
> I see some values in terms of generating `compile_commands.json` and it can be used for using `clang-format` and `clang-tidy` in the future.

It's also very useful for language server. I've been using a lot and it allows me navigate very quickly in the source code. (Without it, you need to use Build Ear, `bear make` to generate it)

meson.build
```diff
@@ -0,0 +1,820 @@
...
+    uftrace_cflags    += asan_cflags
+    demangler_cflags  += asan_cflags
+    symbols_cflags    += asan_cflags
+    dbginfo_cflags    += asan_cflags
+    traceevent_cflags += asan_cflags
+    test_cflags       += asan_cflags
+endif
+
+if get_option('coverage')
+libmcount_cflags
```
honggyukim, line 72: Please drop this.

meson.build
```diff
@@ -0,0 +1,820 @@
...
+
+################################
+# additional compilation flags #
+################################
+
+if get_option('debugging')
+    common_cflags += ['-O0', '-g']
+else
+    common_cflags += ['-O2', '-g']
+endif
```
honggyukim, line 46: Sorry, we need another update again because of the recent change at [here](https://github.com/namhyung/uftrace/commit/76027ad368a3d2ec2485d9a1e52ac8078e05cbff?branch=76027ad368a3d2ec2485d9a1e52ac8078e05cbff&diff=unified#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52). Could you please apply this change?
```diff
diff --git a/meson.build b/meson.build
index 04055ba4..98dd0c5a 100644
--- a/meson.build
+++ b/meson.build
@@ -40,9 +40,11 @@ endif
 ################################

 if get_option('debugging')
-    common_cflags += ['-O0', '-g']
+    common_cflags += ['-O0', '-g', '-DDEBUG_MODE=1']
+    common_ldflags += ['-lunwind']
 else
-    common_cflags += ['-O2', '-g']
+    common_cflags += ['-O2', '-g', '-DDEBUG_MODE=0']
+    common_ldflags += ['-rdynamic']
 endif

 if get_option('trace')
```

honggyukim (commented):
@gpollo Thanks for the update!

> It's also very useful for language server. I've been using a lot and it allows me navigate very quickly in the source code.

agreed.

honggyukim:
@gpollo It would be even better if you could update [INSTALL.md](https://github.com/namhyung/uftrace/blob/master/INSTALL.md) for meson build as well.

honggyukim (approved):
@gpollo Thanks for your effort and patience.

namhyung:
I'm seeing more test failures (mostly about arguments) with `ninja runtest`.

honggyukim:
> I'm seeing more test failures (mostly about arguments) with ninja runtest.

I also see the problem with the following change.
```diff
diff --git a/meson.build b/meson.build
index 98dd0c5a..d4a85c4d 100644
--- a/meson.build
+++ b/meson.build
@@ -719,7 +719,7 @@ run_target('unittest',
 )

 run_target('runtest',
-    command: ['sh', '-c', 'cd @0@ && objdir=@1@ ./runtest.py'.format(tests_dir, meson.current_build_dir())],
+    command: ['sh', '-c', 'cd @0@ && objdir=@1@ ./runtest.py arg_'.format(tests_dir, meson.current_build_dir())],
     depends: [
         libmcount,
         libmcount_fast,
```
The result is as follows.
```
Start 33 tests with 33 worker
058 arg_int             : OK OK OK OK OK SK SK SK SK SK
059 arg_str             : OK OK OK OK OK SK SK SK SK SK
060 arg_fmt             : OK OK OK OK OK SK SK SK SK SK
061 arg_plt             : OK OK OK OK OK OK OK OK OK OK
062 arg_char            : OK OK OK OK OK SK SK SK SK SK
065 arg_order           : OK OK OK OK OK SK SK SK SK SK
082 arg_many            : OK OK OK OK OK SK SK SK SK SK
083 arg_float           : NG NG NG NG NG SK SK SK SK SK
084 arg_mixed           : NG NG NG NG NG SK SK SK SK SK
085 arg_reg             : NG NG NG NG NG SK SK SK SK SK
086 arg_stack           : OK OK OK OK OK SK SK SK SK SK
087 arg_variadic        : NG NG NG NG NG SK SK SK SK SK
126 arg_regex           : OK OK OK OK OK SK SK SK SK SK
127 arg_module          : OK OK OK OK OK SK SK SK SK SK
128 arg_module2         : OK OK OK OK OK SK SK SK SK SK
146 arg_std_string      : OK OK OK OK OK SK SK SK SK SK
176 arg_fptr            : OK OK OK OK OK OK OK OK OK OK
178 arg_auto1           : OK OK OK OK OK OK OK OK OK OK
179 arg_auto2           : OK OK OK OK OK OK OK OK OK OK
180 arg_auto3           : OK OK OK OK OK OK OK OK OK OK
184 arg_enum            : OK OK OK OK OK OK OK OK OK OK
198 lib_arg_float       : NZ NZ NZ NZ NZ SK SK SK SK SK
201 arg_dwarf1          : SK SK SK SK SK SK SK SK SK SK
202 arg_dwarf2          : SK SK SK SK SK SK SK SK SK SK
203 arg_dwarf3          : SK SK SK SK SK SK SK SK SK SK
204 arg_dwarf4          : SK SK SK SK SK SK SK SK SK SK
205 arg_auto4           : SK SK SK SK SK SK SK SK SK SK
206 arg_enum2           : SK SK SK SK SK SK SK SK SK SK
213 arg_symbol          : SK SK SK SK SK SK SK SK SK SK
231 arg_bound           : OK OK OK OK OK SK SK SK SK SK
254 arg_dwarf5          : SK SK SK SK SK SK SK SK SK SK
255 arg_dwarf6          : SK SK SK SK SK SK SK SK SK SK
256 arg_dwarf7          : SK SK SK SK SK SK SK SK SK SK
```
It looks like all of the failed tests are related to floating point argument handling, but I see that the problem comes from our code base as I reported the bug in #1285.

honggyukim (commented):
Besides that, I see that the uftrace is built without proper features enabled.
```
$ cd build

$ meson ..
      ...
Message:           prefix: /usr/local
Message:      libcapstone: [ on  ] - full dynamic tracing support
Message:            libdw: [ on  ] - dwarf support
Message:           libelf: [ on  ] - more flexible ELF data handling
Message:        libluajit: [ on  ] - luajit scripting support
Message:        libpython: [ on  ] - python scripting support
Message:       libncurses: [ on  ] - terminal UI support
Message:     cxa_demangle: [ on  ] - full demangler support with libstdc++
Message:       perf event: [ on  ] - perf event support
Message:    perf schedule: [ on  ] - perf scheduler event support
Build targets in project: 27

$ ninja

$ ./uftrace --version
uftrace v0.10-17-g19c05165 ( x86_64 )
```
It didn't enable dwarf, script, and other features at all. I expected this to be printed as follows. (the version scheme looks a bit different as well)
```
$ ./uftrace  --version
uftrace v0.10-17-g19c0 ( x86_64 dwarf python luajit tui perf sched dynamic )
```

honggyukim:
@gpollo If you want you can squash all the changes from [build/meson-v1](https://github.com/namhyung/uftrace/compare/build/meson-v1) and update this PR again.

gpollo:
Sorry about the delay. I squashed your changes and pushed them.

> The remaining problem is not showing features in version string.

I don't quite understand, a quick test on `master` with the original build system gives me the same string. Or am I doing something wrong?

meson.build
```diff
@@ -0,0 +1,822 @@
...
+libpython_is_enabled = false
+libpython_dep = []
+python_cflags = []
+libpython2_dep = dependency('python2', required: get_option('with-libpython'))
+libpython3_dep = dependency('python3', required: get_option('with-libpython'))
+if libpython3_dep.found()
+    libpython_dep = libpython3_dep
+    libpython_is_enabled = true
+    common_cflags += '-DHAVE_LIBPYTHON3'
+    common_cflags += '-DLIBPYTHON_VERSION=3.8'
```
honggyukim, line 162: Hmm.. I forgot to change this part by detecting the python3 version yet. I don't know how to do that in meson but will think about it more.

meson.build
```diff
@@ -0,0 +1,822 @@
...
+###########
+# version #
+###########
+
+version_sh = find_program('misc/version.sh')
+version_sh_cmd = '$(git describe --tags --abbrev=4 --match="v[0-9].[0-9]*" 2> /dev/null || echo v' + meson.project_version() + ')'
+version_h = custom_target(
+    'gen_version_h',
+    output: 'version.h',
+    command: [program_shell, '-c', version_sh.path() + ' @OUTPUT@' + ' ' + version_sh_cmd + ' @0@ .'.format(arch) + ' > /dev/null'],
```
honggyukim, line 282: We only put arch info to version string, but we have to add other feature info as well.

honggyukim (commented):
> Sorry about the delay. I squashed your changes and pushed them.

Thanks for the quick update.

> I don't quite understand, a quick test on master with the original build system gives me the same string. Or am I doing something wrong?

Maybe it's because you already generated `version.h` from the original build system. If you remove `version.h` then start building with meson then you will see the output as follows.
```
$ ./uftrace --version
uftrace v0.10-58-g5a0a ( x86_64 )
```
But the original build shows the version string as follows.
```
$ uftrace --version
uftrace v0.10-58-g5a0a ( x86_64 dwarf python luajit tui perf sched dynamic )
```

meson.build
honggyukim: Related part in the original build is at https://github.com/namhyung/uftrace/blob/master/check-deps/Makefile.check#L26.

meson.build
honggyukim: The version string generation is done at https://github.com/namhyung/uftrace/blob/master/misc/version.sh#L33-L55.

gpollo:
It's a dependency order problem. When I used the original build system, I compiled with `make -j16` and `version.h` gets generated before the dependencies are checked.
```
  GEN      .config
  GEN      version.h
...         libelf: [ on  ] - more flexible ELF data handling
...          libdw: [ on  ] - DWARF debug info support
...      libpython: [ on  ] - python scripting support
...      libluajit: [ on  ] - luajit scripting support
...    libncursesw: [ on  ] - TUI support
...   cxa_demangle: [ on  ] - full demangler support with libstdc++
...     perf_event: [ on  ] - perf (PMU) event support
...       schedule: [ on  ] - scheduler event support
...       capstone: [ on  ] - full dynamic tracing support
...      libunwind: [ on  ] - libunwind support
```
When I use a single thread (e.g. simple `make`), I get:
```
  GEN      .config
...         libelf: [ on  ] - more flexible ELF data handling
...          libdw: [ on  ] - DWARF debug info support
...      libpython: [ on  ] - python scripting support
...      libluajit: [ on  ] - luajit scripting support
...    libncursesw: [ on  ] - TUI support
...   cxa_demangle: [ on  ] - full demangler support with libstdc++
...     perf_event: [ on  ] - perf (PMU) event support
...       schedule: [ on  ] - scheduler event support
...       capstone: [ on  ] - full dynamic tracing support
...      libunwind: [ on  ] - libunwind support
  GEN      version.h
```
While `ninja` uses all available thread by default, it's not the only problem. The meson build system won't generate the files in `check-deps/` directory. I'll think more about it later in the day.

honggyukim:
Yeah, I also think that it's tricky to handle all the difference at once. I'm okay to accept the patch if we can manage the python3 version handling. We can revisit version string maybe in the follow up patch because it's an experimental build option as of now. Thanks for working on this anyway.

gpollo:
I've modified `misc/version.sh` to check for environment variables. Meson sets the environments variables when executing the script to tell what has been built into.

meson.build
gpollo: I it like this
```
find_python_version = 'import sysconfig; print(sysconfig.get_config_vars()["LDVERSION"],end="")'
...
    python_version = run_command(['python3', '-c', find_python_version])
    common_cflags += '-DLIBPYTHON_VERSION=@0@'.format(python_version.stdout())
```

gpollo:
Fixed the python version.

honggyukim:
Thanks. I see that python version is fixed, but `build: fix version string in when building with Meson` breaks the original build system because `UFTRACE_HAVE_LIBDW` and other variables are not defined there.

It even shows some errors `misc/version.sh: 34: [: unexpected operator`.

gpollo:
That's weird, the original build system still seems to work here, see below. Maybe there's some configurations in your shell?

[asciicast](https://asciinema.org/a/BjwzZJgr97fQ3hvBidGmwbEUI)

honggyukim:
It's because my `/bin/sh` is a soft link to `dash`.
```
$ ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Feb 17  2016 /bin/sh -> dash
```
Maybe we should explicitly specify shell name to `bash`. It's okay with the following change.
```diff
diff --git a/misc/version.sh b/misc/version.sh
index 51b01b51..8a309463 100755
--- a/misc/version.sh
+++ b/misc/version.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash

 VERSION_FILE=$1

```

namhyung:
Thanks for working on this!  I think we should change the logic in the check-deps to generate files under the build directory.  Please check the [check/build-dep-v1](https://github.com/namhyung/uftrace/compare/check/build-dep-v1) branch.

Also I think it'd be nice if it can work with any shell.

misc/version.sh
```diff
@@ -31,26 +31,33 @@ if test -z "${GIT_VERSION}" -a -n "${FILE_VERSION}"; then
...
     DEPS="${DEPS} dwarf"
 fi
-if test -f ${SRCDIR}/check-deps/have_libpython2.7 ||
+if [ "$UFTRACE_HAVE_LIBPYTHON" == "on" ] ||
+   test -f ${SRCDIR}/check-deps/have_libpython2.7 ||
    test -f ${SRCDIR}/check-deps/have_libpython3; then
     DEPS="${DEPS} python"
 fi
-if test -f ${SRCDIR}/check-deps/have_libluajit; then
+if [ "$UFTRACE_HAVE_LIBDW" == "on" ] ||
```
eli-schwartz, line 43: ^^

misc/version.sh
```diff
@@ -31,26 +31,33 @@ if test -z "${GIT_VERSION}" -a -n "${FILE_VERSION}"; then
...
+   test -f ${SRCDIR}/check-deps/have_libluajit; then
     DEPS="${DEPS} luajit"
 fi
-if test -f ${SRCDIR}/check-deps/have_libncurses; then
+if [ "$UFTRACE_HAVE_LIBNCURSES" == "on" ] ||
```
eli-schwartz, line 47: ^^

misc/version.sh
```diff
@@ -31,26 +31,33 @@ if test -z "${GIT_VERSION}" -a -n "${FILE_VERSION}"; then
...
+   test -f ${SRCDIR}/check-deps/have_libncurses; then
     DEPS="${DEPS} tui"
 fi
-if test -f ${SRCDIR}/check-deps/perf_clockid; then
+if [ "$UFTRACE_HAVE_PERF_CLOCKID" == "on" ] ||
```
eli-schwartz, line 51: ^^

misc/version.sh
```diff
@@ -31,26 +31,33 @@ if test -z "${GIT_VERSION}" -a -n "${FILE_VERSION}"; then
...
+   test -f ${SRCDIR}/check-deps/perf_clockid; then
     DEPS="${DEPS} perf"
 fi
-if test -f ${SRCDIR}/check-deps/perf_context_switch; then
+if [ "$UFTRACE_HAVE_PERF_CONTEXT_SWITCH" == "on" ] ||
```
eli-schwartz, line 55: ^^

misc/version.sh
```diff
@@ -31,26 +31,33 @@ if test -z "${GIT_VERSION}" -a -n "${FILE_VERSION}"; then
...
+   test -f ${SRCDIR}/check-deps/perf_context_switch; then
     DEPS="${DEPS} sched"
 fi
-if test -f ${SRCDIR}/check-deps/have_libcapstone; then
+if [ "$UFTRACE_HAVE_LIBCAPSTONE" == "on" ] ||
```
eli-schwartz, line 59: ^^

honggyukim:
@eli-schwartz Thanks very much for your review in details. @gpollo Could you please have a look? If it takes a long time to address all the comments, maybe we can finalize the current PR and revisit addressing @eli-schwartz's comments in the follow up patch.

gpollo:
Thank you @eli-schwartz for your detailed review! @honggyukim I think I'll leave the current commit as is since it replicated as much as possible the old build system. When transitionning to the build-in meson options, we can document the changes in the commit. I've already started on a few, they'll be in separate commits.

eli-schwartz:
Well, please at least fix the /bin/sh syntax errors :p

And several other of my review comments pertained to using e.g. meson in-process string replacement rather than sed string replacement which is just all around good.

meson.build
```diff
@@ -0,0 +1,834 @@
...
+        libmcount_fast,
+        libmcount_single,
+        libmcount_fast_single,
+        libmcount_fast_nop,
+        uftrace,
+    ],
+)
+
+run_target('tests',
+    command: ['sh', '-c', 'cd @0@ && ninja unittest; ninja runtest'.format(meson.current_build_dir())],
```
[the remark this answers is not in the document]
gpollo: My understanding of the `test()` feature of Meson is to setup all your test individually using this command. In Uftrace, all tests are part of a single binary. Hence, I don't see the point of using the `test()` feature. In other projects I did, tests were separate binaries and it made sense to use `test()`. Anyway, I think it would be a good idea in the future to eventually switch to this Meson feature.

gpollo:
I've fix most of @eli-schwartz comments. Again, thanks a lot!

I've tried cross-compiling again (last time was before submitting this PR...). Removed the `-Darch` option and I had to change `cc.run()` to `cc.compiles()` when checking for some dependencies, but it's working *okay*. I've put a warning saying that it was still experimental.

misc/version.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
eli-schwartz, line 1: Since the bash-specific `==` was removed, this should work using /bin/sh again.

meson.build
```diff
@@ -0,0 +1,797 @@
...
+endif
+
+#################################
+# -Dwith-libstdcxx build option #
+#################################
+
+cxa_demangle_is_enabled = false
+cxa_demangle_dep = []
+have_cxa_demangle_code = run_command('cat', files('check-deps/__cxa_demangle.c')).stdout().strip()
+have_cxa_demangle_compiles = cc.compiles(have_cxa_demangle_code, name: 'cxa_demangle check', args: ['-lstdc++'])
```
eli-schwartz, line 146: `cc.compiles(files('check-deps/__cxa_demangle.c'))` should work, without needing to first cat it into a meson variable.

meson.build
```diff
@@ -0,0 +1,797 @@
...
+if get_option('with-libstdcxx').auto()
+    if have_cxa_demangle_compiles
+        cxa_demangle_dep = cc.find_library('stdc++', required: true)
+        cxa_demangle_is_enabled = true
+        common_cflags += '-DHAVE_CXA_DEMANGLE'
+        message('Enabling libstdc++')
+    else
+        message('Disabling libstdc++')
+    endif
+elif get_option('with-libstdcxx').enabled()
```
eli-schwartz, line 156: This is duplicating additional logic around. What about:

```meson
cxa_demangle_dep = cc.find_library('stdc++', required: get_option('with-libstdcxx'))

if get_option('with-libstdcxx').enabled() and not have_cxa_demangle_compiles
    error('Failed to compile and run check-deps/__cxa_demangle.c')
elif cxa_demangle_dep.found() and have_cxa_demangle_compiles
    cxa_demangle_is_enabled = true
    common_cflags += '-DHAVE_CXA_DEMANGLE'
    msg = 'Enabling libstdc++'
    if get_option('with-libstdcxx').enabled()
        msg += ' (enabled by: with-libstdcxx)'
    endif
else
    msg = 'Disabling libstdc++'
    if get_option('with-libstdcxx').disabled()
        msg += ' (disabled by: with-libstdcxx)'
    elif cxa_demangle_dep.found()
        msg = ' (disabled by: failure to compile and run check-deps/__cxa_demangle.c)'
    endif
endif
message(msg) 
```

3 fewer lines, I added additional logging in the disabling case, the coding flow is less duplicative, and honestly I'd skip the last 13 lines:

```meson
cxa_demangle_dep = cc.find_library('stdc++', required: get_option('with-libstdcxx'))

if get_option('with-libstdcxx').enabled() and not have_cxa_demangle_compiles
    error('Failed to compile and run check-deps/__cxa_demangle.c')
elif cxa_demangle_dep.found() and have_cxa_demangle_compiles
    cxa_demangle_is_enabled = true
    common_cflags += '-DHAVE_CXA_DEMANGLE'
endif
```

meson.build
```diff
@@ -0,0 +1,797 @@
...
+    message('Disabling libstdc++ (disabled by: with-libstdcxx)')
+endif
+
+############################
+# -Dwith-perf build option #
+############################
+
+perf_clockid_is_enabled = false
+have_perf_clockid_code = run_command('cat', files('check-deps/__perf_clockid.c')).stdout().strip()
+have_perf_clockid_compiles = cc.compiles(have_perf_clockid_code, name: 'perf clockid check')
```
eli-schwartz, line 175: more cc.`compiles(files('...'))`

meson.build
```diff
@@ -0,0 +1,797 @@
...
+)
+
+#############################
+# libmcount (arch specific) #
+#############################
+
+if arch == 'aarch64'
+    libmcount_arch_include = include_directories('arch/aarch64')
+    libmcount_arch_sources = [
+        'arch/aarch64/cpuinfo.c',
```
eli-schwartz, line 259: Could be potentially rewritten

```
fs_mod = import('fs')
if fs_mod.is_dir('arch' / arch)
    libmcount_arch_include = include_directories('arch' / arch)
    subdir('arch' / arch)
else
     error('target architecture `@0@` is not supported'.format(arch))
endif
```

inside arch/aarch64/meson.build

```meson
libmcount_arch_sources = files(
    'cpuinfo.c',
    'mcount-dynamic.c',
    'mcount-insn.c',
    'mcount-support.c',
)
libmcount_arch_asm_sources = files(
    'dynamic.S',
    'mcount.S',
    'plthook.S',
)
```

meson.build
```diff
@@ -0,0 +1,797 @@
...
+            libmcount,
+            libmcount_fast,
+            libmcount_single,
+            libmcount_fast_single,
+            libmcount_fast_nop,
+            uftrace,
+        ],
+    )
+
+    run_target('tests', command: ['ninja', 'unittest', 'runtest'])
```
eli-schwartz, line 717: cf. https://github.com/mesonbuild/meson/commit/dc51740e2cdee171ac6ebc7ed261e5c08730c9a9

In meson 0.60 you will be able to do this instead:

```meson
unittest = run_target(...)
runtest = run_target(...)
alias_target('tests', unittest, runtest)
```

qpakzk:
@honggyukim I want to keep going this feature.

honggyukim:
@qpakzk thanks but you should address all the comments given by @eli-schwartz here.

PR #1215, opened 2020-10-26
dynamic: Support soname for patch option

ParkHanbum:
some libraries that required to run executable are linked with symbolic
link. so, user who want to trace shared object that required for
executable need to know actually loaded file name of shared object.

this works can be omitted by use soname field in dynamic section of
shared object. because soname represent its symbolic linked name.

Signed-off-by: hanbum <hbpark@stealien.com>


ParkHanbum:
it useful when want to trace shared object to symbolic linked. likely 'libc.so.6'

PR #1216, opened 2020-10-31
Support soname

ParkHanbum:
some libraries that required to run executable are linked with symbolic
link. so, user who want to trace shared object that required for
executable need to know actually loaded file name of shared object.

this works can be omitted by use soname field in dynamic section of
shared object. because soname represent its symbolic linked name.

here is the usage case:
```
m@LAPTOP-G3M5JJUQ:~/uftrace_soname$ ./uftrace -v -L. -P.@libc.so.6 test
uftrace: checking binary test
uftrace: skipping perf event due to error: Invalid argument
uftrace: using ./libmcount/libmcount.so library for tracing
dynamic: dynamic patch type: test: 0 (none)
dynamic: dynamic patch type: libc-2.27.so: 1 (pg)
dynamic: dynamic patch type: librt-2.27.so: 0 (none)
dynamic: dynamic patch type: libdl-2.27.so: 0 (none)
dynamic: dynamic patch type: libelf-0.170.so: 0 (none)
dynamic: dynamic patch type: libdw-0.170.so: 0 (none)
dynamic: dynamic patch type: libcapstone.so.3: 0 (none)
dynamic: dynamic patch type: libpthread-2.27.so: 0 (none)
dynamic: dynamic patch type: ld-2.27.so: 0 (none)
dynamic: dynamic patch type: libm-2.27.so: 0 (none)
dynamic: dynamic patch type: libgcc_s.so.1: 0 (none)
dynamic: dynamic patch type: libz.so.1.2.11: 0 (none)
dynamic: dynamic patch type: liblzma.so.5.2.2: 0 (none)
dynamic: dynamic patch type: libbz2.so.1.0.4: 0 (none)
dynamic: libc.so.6 libc.so.6 libc-2.27.so __libc_init_first matched
dynamic: libc.so.6 libc.so.6 libc-2.27.so __libc_start_main matched
dynamic: libc.so.6 libc.so.6 libc-2.27.so __errno_location matched
dynamic: libc.so.6 libc.so.6 libc-2.27.so iconv_open matched
dynamic: libc.so.6 libc.so.6 libc-2.27.so iconv matched
```

as you can see, actually libc-2.27.so has been loaded instead libc.so.6. but it keep the libc.so.6 in the soname field. so, user can use soname which will be printed by ldd.

### DIFF
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -368,7 +368,8 @@ static bool match_pattern_module(char *pathname)
 	return ret;
 }
 
-static bool match_pattern_list(struct uftrace_mmap *map, char *sym_name)
+static bool match_pattern_list(struct uftrace_mmap *map, char *sym_name,
+			       char* soname)
 {
 	struct patt_list *pl;
 	bool ret = false;
@@ -420,7 +421,7 @@ static void patch_func_matched(struct mcount_dynamic_info *mdi,
 		    sym->type != ST_GLOBAL_FUNC)
 			continue;
 
-		if (!match_pattern_list(map, sym->name)) {
+		if (!match_pattern_list(map, sym->name, soname)) {
 			if (mcount_unpatch_func(mdi, sym, &disasm) == 0)
 				stats.unpatch++;
 			continue;
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -399,6 +399,7 @@ static void patch_func_matched(struct mcount_dynamic_info *mdi,
 		"__libc_csu_init",
 		"__libc_csu_fini",
 	};
+	char *soname = get_soname(map->libname);
 
 	symtab = &map->mod->symtab;
 
@@ -442,6 +443,9 @@ static void patch_func_matched(struct mcount_dynamic_info *mdi,
 
 	if (!found)
 		stats.nomatch++;
+
+	if (soname != NULL)
+		free(soname);
 }
 
 static int do_dynamic_update(struct symtabs *symtabs, char *patch_funcs,
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -376,10 +376,16 @@ static bool match_pattern_list(struct uftrace_mmap *map, char *sym_name,
 	char *libname = basename(map->libname);
 
 	list_for_each_entry(pl, &patterns, list) {
-		if (strncmp(libname, pl->module, strlen(pl->module)))
-			continue;
+		int len = strlen(pl->module);
+		bool matched = false;
 
-		if (match_filter_pattern(&pl->patt, sym_name))
+		if (soname != NULL)
+			matched = (strncmp(soname, pl->module, len) == 0);
+
+		if (!matched)
+			matched = (strncmp(libname, pl->module, len) == 0);
+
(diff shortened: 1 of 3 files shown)


utils/symbol.c
```diff
@@ -83,6 +83,33 @@ static int namefind(const void *a, const void *b)
...
 
+char *get_soname(const char *filename)
+{
+	struct uftrace_elf_data elf;
+	struct uftrace_elf_iter iter;
+	char *soname = NULL;
+
+	if (elf_init(filename, &elf) < 0) {
+		pr_dbg("error during open symbol file: %s: %m\n", filename);
+		return false;
```
namhyung, line 94: You should return NULL here.

utils/symbol.c
```diff
@@ -83,6 +83,33 @@ static int namefind(const void *a, const void *b)
...
+	elf_for_each_shdr(&elf, &iter) {
+		if (iter.shdr.sh_type == SHT_DYNAMIC)
+			break;
+	}
+
+	elf_for_each_dynamic(&elf, &iter) {
+		if (iter.dyn.d_tag != DT_SONAME)
+			continue;
+
+		soname = xstrdup(elf_get_name(&elf, &iter, iter.dyn.d_un.d_ptr));
```
namhyung, line 106: Please add 'break'.

libmcount/dynamic.c
```diff
@@ -442,6 +443,9 @@ static void patch_func_matched(struct mcount_dynamic_info *mdi,
 
 	if (!found)
 		stats.nomatch++;
+
+	if (soname != NULL)
+		free(soname);
```
namhyung, line 455: It's ok to call free with a NULL argument.

By the way, this change itself has no meaning and you should move it to the actual use case.

libmcount/dynamic.c
```diff
@@ -420,7 +421,7 @@ static void patch_func_matched(struct mcount_dynamic_info *mdi,
 		    sym->type != ST_GLOBAL_FUNC)
 			continue;
 
-		if (!match_pattern_list(map, sym->name)) {
+		if (!match_pattern_list(map, sym->name, soname)) {
```
namhyung, line 430: Same comment as before.  Please move the code, you don't need to separate commits this way.

libmcount/dynamic.c
```diff
@@ -376,10 +376,16 @@ static bool match_pattern_list(struct uftrace_mmap *map, char *sym_name,
...
-		if (strncmp(libname, pl->module, strlen(pl->module)))
-			continue;
+		int len = strlen(pl->module);
+		bool matched = false;
+
+		if (soname != NULL)
+			matched = (strncmp(soname, pl->module, len) == 0);
+
+		if (!matched)
+			matched = (strncmp(libname, pl->module, len) == 0);
```
namhyung, line 386: How about this?

```c
if (strncmp(libname, pl->module, len) &&
    (!soname || strncmp(soname, pl->module, len)))
        continue;
```

Maybe you can factor out this into a function and call it for dlopen as well.

namhyung:
I made a few changes and pushed it to [check/soname-v1](https://github.com/namhyung/uftrace/compare/check/soname-v1).  Please take a look.

Also I think we can apply the SONAME to regular filters/triggers too.

Issue #1169, opened 2020-05-13
new requirement: support for filter one function itself instead of whole subtree.


simpzan:
I am tracing wayfire, a window compositor for linux. the tracing log is very huge for just 20 second. and a large part of the log is from std template. I tried to filter out all std function with `-F std::.`, it turns out most user functions are filtered out as well. because some user function is called in the subtree of a std function like cpp lambda. 
I just want to filter out all std function and keep all user functions, how to achieve this? one feature to achieve this is using filter based on one function instead of a whole subtree, but the feature seems not available in uftrace now. Any advice? thanks!

namhyung:
Hi, thanks for using uftrace.

The `-F std::.` will show std functions (and their subtrees) only, what about `-N std::.` instead?

simpzan:
Oh, sorry, I made a mistake. I do use -N std::. and it filtered out all std
functions, meanwhile many user functions in the sub tree of std functions
are also filtered out. For example, user functions called from cpp
 lambda.  That’s not what I want. I want to hide all std functions and keep
all user functions.

namhyung:
OK, we don't have such filter just for single function.

Instead, you can try depth trigger like `-T std::.@depth=0`, it'll hide all functions started with std namesapce.  But you need to restore depth for user functions manually.  If your program starts with a common prefix or contains some unique name, you can use `-T prefix@depth=100` for example.

namhyung:
So I've played with it. Please test [check/hide-filter](https://github.com/namhyung/uftrace/compare/check/hide-filter).  `--hide` option was added for testing.

Normal run:
```
$ uftrace a.out | head
# DURATION     TID     FUNCTION
            [ 86842] | main() {
            [ 86842] |   Map::Map() {
            [ 86842] |     std::unordered_map::unordered_map() {
            [ 86842] |       std::_Hashtable::_Hashtable() {
            [ 86842] |         std::__detail::_Hashtable_alloc::_Hashtable_alloc() {
            [ 86842] |           std::__detail::_Hashtable_ebo_helper::_Hashtable_ebo_helper() {
            [ 86842] |             std::allocator::allocator() {
   0.230 us [ 86842] |               __gnu_cxx::new_allocator::new_allocator();
   2.549 us [ 86842] |             } /* std::allocator::allocator */
```

To hide std library functions:
```
$ uftrace --hide ^std a.out | head
# DURATION     TID     FUNCTION
            [ 86895] | main() {
[...]

PR #1217, opened 2020-11-07
filter: Add -H/--hide option to hide functions without subtree

honggyukim:
filter: Add -H/--hide option to hide functions without subtree

The -H/--hide option is used to apply notrace filter to some functions without
affecting their subtree functions.  In other words, it simply hides
single depth of function and continue to trace child functions.

This option works only for replay time and 'hide' can be used as a
trigger just like other triggers such as filter.
```
  $ uftrace -H main -T c@hide t-abc
              [ 13037] | a() {
              [ 13037] |   b() {
     2.600 us [ 13037] |     getpid();
     8.500 us [ 13037] |   } /* b */
    11.400 us [ 13037] | } /* a */
```
This can be useful when user functions are called from a library, but
don't want to trace functions in the library as the following example.
```
  $ cat call_once.cc
  #include <mutex>
  std::once_flag flag;
  void foo() {}
  int main() {
      std::call_once(flag, foo);
  }

  $ g++ -pg -O2 -pthread call_once.cc
  $ uftrace record a.out
  $ uftrace replay
              [ 19661] | main() {
              [ 19661] |   pthread_once() {
              [ 19661] |     std::call_once::$_1::_FUN() {
     1.900 us [ 19661] |       foo();
     6.700 us [ 19661] |     } /* std::call_once::$_1::_FUN */
    50.900 us [ 19661] |   } /* pthread_once */
    56.100 us [ 19661] | } /* main */
```
We can simply hide functions in C++ std namespace easily as follows.
```
  $ uftrace replay -H ^std:: a.out
              [ 19666] | main() {
              [ 19666] |   pthread_once() {
     2.300 us [ 19666] |     foo();
    31.400 us [ 19666] |   } /* pthread_once */
    38.000 us [ 19666] | } /* main */
```
It also supports 'hide' trigger for the same purpose.

Closes: #1169


honggyukim:
@namhyung I still have to add a test and also need to update the document, but I see that the name **unpatch** is a bit weird so didn't add it yet. Could you tell me how you think about it?

namhyung:
Well I don't know.. but basically it **reuses** the option and there's nothing different if we hide the option from the documentation.

honggyukim:
Okay. I will update  the document using the existing name.

honggyukim:
I've updated the patch with a test and document. Please have a look. Thanks.

namhyung:
Sorry for late reply.  I've checked the code and I think we should implement it only for the replay part now.  I plan to apply it for dynamic tracing later..

namhyung:
Also we might consider reusing -P option too.  If it's used we can think all other functions should be hidden (this is the same way we use it for dynamic tracing).

honggyukim:
> I think we should implement it only for the replay part now. I plan to apply it for dynamic tracing later..

Sure. I've removed record part from the patch and updated it again.

> Also we might consider reusing -P option too. If it's used we can think all other functions should be hidden (this is the same way we use it for dynamic tracing). But this can be done later..

Right. I will think it may be done by setting `-F FUNC@depth=1`, but need to think about it more. Thanks.

namhyung:
Thanks for the update... but I'm still hesitating about the naming, sorry.

Maybe we can add 'hide' as an alias for the trigger and keep the internal naming like it (ex. TRIGGER_FL_HIDE).  It's essentially for hide but we are reusing the -U option.  And then 'unpatch' trigger should work too.

doc/uftrace-replay.md
```diff
@@ -70,6 +70,11 @@ COMMON OPTIONS
 -D *DEPTH*, \--depth *DEPTH*
 :   Set trace limit in nesting level.  See *FILTERS*.
 
+-U *FUNC*, \--unpatch=*FUNC*
+:   Set filter not to trace selected functions.  It doesn't affects their
+    subtrees, but hides only the given functions.
+    This option can be used more than once.
+
```
namhyung, line 77: I think we should update the trigger part too.

uftrace.c
```diff
@@ -198,7 +198,7 @@ __used static const char uftrace_help[] =
 "      --time                 Print time information\n"
 "  -T, --trigger=FUNC@act[,act,...]\n"
 "                             Trigger action on those FUNCs\n"
-"  -U, --unpatch=FUNC         Don't apply dynamic patching for FUNCs\n"
+"  -U, --unpatch=FUNC         Hide FUNCs from trace or don't apply dynamic patching for FUNCs\n"
```
namhyung, line 201: I'd like to put it after the dynamic tracing as it's the primary feature.

uftrace.c
```diff
@@ -630,6 +630,7 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		break;
 
 	case 'U':
+		opts->unpatch = opt_add_string(opts->unpatch, arg);
 		opts->patch = opt_add_prefix_string(opts->patch, "!", arg);
```
namhyung, line 634: This is redundant and I think we can use opts->patch (I know it's annoying).  You can ignore ones without the '!' prefix later.

utils/filter.c
```diff
@@ -640,6 +647,7 @@ static const struct trigger_action_parser actions[] = {
 	{ "depth=",    parse_depth_action,        TRIGGER_FL_FILTER, },
 	{ "time=",     parse_time_action,         TRIGGER_FL_FILTER, },
 	{ "caller",    parse_caller_action,       TRIGGER_FL_FILTER, },
+	{ "unpatch",   parse_unpatch_action,      TRIGGER_FL_FILTER, },
```
namhyung, line 650: Can you please add "hide" too?

utils/filter.h
```diff
@@ -35,6 +35,7 @@ enum trigger_flag {
 	TRIGGER_FL_AUTO_ARGS	= (1U << 14),
 	TRIGGER_FL_CALLER	= (1U << 15),
 	TRIGGER_FL_SIGNAL	= (1U << 16),
+	TRIGGER_FL_UNPATCH	= (1U << 17),
```
namhyung, line 38: We can make it _HIDE.

utils/fstack.c
```diff
@@ -327,6 +338,17 @@ static int setup_fstack_filters(struct uftrace_data *handle, char *filter_str,
 		pr_dbg("setup caller filters for %d function(s)\n", count);
 	}
 
+	if (unpatch_str) {
+		setting->private = unpatch_str;
+		walk_sessions(sessions, setup_unpatch, setting);
+		walk_sessions(sessions, count_filters, &count);
```
namhyung, line 344: I think you should add count_unpatch (or count_hide).

honggyukim:
@namhyung I've renamed `-f`/`--output-fields` to `-O`/`--output-fields`, then used `-f` and `-n` for filter and notrace for the single function without affecting subtree.

`-n` option works both for record and replay, but `-f` works only for replay time yet. I need to update it to make `-f` work for record time as well.

namhyung:
OK, first we should update PR title accordingly.  And we are now not reusing the -P/-U options, so I'm thinking about this again... (sorry).

1. we only need this for the 'hide' case.  The single depth filter option can be specified like `-F foo@depth=1` : this will show foo function only and not its child functions.

2. the consistent filter experience is good, but I'm not sure we'll use 'filter-single' and 'notrace-single' for record.  Since it's better to use -P and -U (well, they'll support -pg binaries eventually) : they have same semantics and perform better.

3. so I'd like to implement `--hide` option for replay and take the `-H` from the `--host` (as it's seldom used).

honggyukim:
@namhyung Thanks for your feedback. Can we have a call during weekends to talk about this?

namhyung:
Sure!

yezz123 (approved):

honggyukim:
Hi @namhyung, I've just updated as you requested. The changes are as follows.
- Renive `-H` option from `--host`
- Use `-H`/`--hide` option
- It only works for replay time

The only problem left is to handle `{}` during replay as follows.
```
$ uftrace replay -H getpid
   0.793 us [ 21720] | __monstartup();
   0.415 us [ 21720] | __cxa_atexit();
            [ 21720] | main() {
            [ 21720] |   a() {
            [ 21720] |     b() {
            [ 21720] |       c() {
   0.803 us [ 21720] |       } /* c */
   1.044 us [ 21720] |     } /* b */
   1.298 us [ 21720] |   } /* a */
   1.541 us [ 21720] | } /* main */
```
Please have a look. Thanks.

uftrace.c
```diff
@@ -143,7 +144,7 @@ __used static const char uftrace_help[] =
 "  -f, --output-fields=FIELD  Show FIELDs in the replay or graph output\n"
 "  -F, --filter=FUNC          Only trace those FUNCs\n"
 "      --graphviz             Dump recorded data in DOT format\n"
-"  -H, --host=HOST            Send trace data to HOST instead of write to file\n"
+"      --host=HOST            Send trace data to HOST instead of write to file\n"
```
namhyung, line 148: We should remove it from the doc too.

tests/t252_filter_H.py
```diff
@@ -0,0 +1,24 @@
...
+  14.500 us [ 30217] | } /* main */
+""")
+
+    def prepare(self):
+        self.subcmd = 'record'
+        return self.runcmd()
+
+    def setup(self):
+        self.subcmd = 'replay'
+        self.option = '-H c'
```
namhyung, line 24: It should work for 'live' command too.  Then we can just set self.option in the setup function and remove the preprare function.

utils/fstack.c
```diff
@@ -327,6 +338,17 @@ static int setup_fstack_filters(struct uftrace_data *handle, char *filter_str,
 		pr_dbg("setup caller filters for %d function(s)\n", count);
 	}
 
+	if (hide_str) {
+		setting->info_str = hide_str;
+		walk_sessions(sessions, setup_hide, setting);
+		walk_sessions(sessions, count_filters, &count_hide);
```
namhyung, line 344: I don't think we can use count_filters here.  It just counts number of any entries, we need to count number of entries having the hide trigger.

utils/fstack.c
```diff
@@ -442,13 +464,15 @@ int fstack_setup_filters(struct opts *opts, struct uftrace_data *handle)
...
+					 opts->caller, opts->hide, &setting) < 0) {
+			pr_use("failed to set filter or trigger: %s%s%s%s%s%s\n",
 			       opts->filter ?: "",
 			       (opts->filter && opts->trigger) ? " or " : "",
 			       opts->trigger ?: "",
 			       ((opts->filter || opts->trigger) && opts->caller) ? " or " : "",
-			       opts->caller ?: "");
+			       opts->caller ?: "",
+			       (((opts->filter || opts->trigger) && opts->caller) && opts->hide) ? " or " : "",
+			       opts->hide ?: "");
```
namhyung, line 475: IIRC you suggested a cleaner way.. no?

honggyukim:
Thanks for the review. I've updated it again based on your comments.

utils/fstack.c
honggyukim: Please let me know if the updated patch is not what you wanted.

namhyung (commented):
Looks good.  For better display I think we should merge functions contain hidden children properly.  Something like below:

```diff
diff --git a/utils/fstack.c b/utils/fstack.c
index 47d91f4c..1a717f32 100644
--- a/utils/fstack.c
+++ b/utils/fstack.c
@@ -813,7 +813,7 @@ static int fstack_check_skip(struct uftrace_task_reader *task,
        if (tr.flags & (TRIGGER_FL_DEPTH | TRIGGER_FL_TRACE_ON))
                return 1;
 
-       if (tr.flags & TRIGGER_FL_TRACE_OFF || depth <= 0)
+       if (tr.flags & (TRIGGER_FL_TRACE_OFF | TRIGGER_FL_HIDE) || depth <= 0)
                return -1;
 
        return 0;
```

honggyukim:
@namhyung Thanks for the tip. The change is applied and the output looks good as follows.
```
$ uftrace -H getpid t-abc
  41.600 us [  8009] | __monstartup();
   4.300 us [  8009] | __cxa_atexit();
            [  8009] | main() {
            [  8009] |   a() {
            [  8009] |     b() {
   5.900 us [  8009] |       c();
   8.800 us [  8009] |     } /* b */
  11.800 us [  8009] |   } /* a */
  14.700 us [  8009] | } /* main */
```

namhyung (approved):
LGTM

namhyung:
@yezz123 though it's a bit late, thanks for your review. :)

PR #1221, opened 2020-12-12
uftrace: Do not use C++ keywords as variable names

honggyukim:
Since list.h and rbtree.h internally uses 'new' as a variable name, it
cannot be compiled by C++ compilers.

Besides those, there are more places that use C++ keywords so this
patch also renames those names as follows.
```
  s/new/new_entry/g
  s/old/old_entry/g
  s/class/elf_class/g
  s/private/info_str/g
```


namhyung (approved):
LGTM

PR #1224, opened 2020-12-13
uftrace: Suggest to report a bug when segfault happens

honggyukim:
This patch suggests to report a bug when the program gets segfault.

The output contains uftrace version info and it looks as follows.
```
  $ uftrace record -l a.out
  WARN: Segmentation fault: address not mapped (addr: 0x523e66c8)
  WARN:  if this happens only with uftrace, please consider
  -e/--estimate-return option.

  WARN: Backtrace from uftrace v0.9.4-174-g0fce3 ( perf sched )
  WARN: [1] (___tls_get_addr[f7dc18d0] <= __cxa_get_globals[f7dd15eb])
  WARN: [0] (__cxa_get_globals[f7dc2720] <= __cxa_throw[f7dd27d5])

```

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -719,7 +719,7 @@ static void segv_handler(int sig, siginfo_t *si, void *ctx)
 			" please consider -e/--estimate-return option.\n\n");
 	}
 
-	pr_warn("Backtrace from uftrace:\n");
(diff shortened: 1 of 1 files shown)


honggyukim:
It might be better to show uftrace version number at the beginning of `-v` log output.

namhyung (approved):
LGTM

PR #1225, opened 2020-12-29
uftrace: Add arch info to uftrace version and show it

honggyukim:
This patch adds arch info to uftrace version string as follows.

Before:
```
  uftrace v0.9.4-201-g6801 ( dwarf python luajit tui perf sched dynamic )
```
After:
```
  uftrace v0.9.4-201-g6801 ( x86_64 dwarf python luajit tui perf sched dynamic )
```
It also adds a debugging output that shows version info at the
beginning of execution.


honggyukim:
It will be useful to get more info with verbose option rather than asking external users to show the uftrace version and architecture info.

namhyung (approved):
LGTM

Issue #1129, opened 2020-02-13
C++11 user defined string literals break the json file


kobi-ca:
and it cannot be loaded.
C++11 user defined string literals looks like the following:
```
ret operator "" _name(some_type var) {
// ...
}

// to make it work with
auto v = 10_name;
```
in the json, the name of the func would be
```
"operator ""'
```
those 3 trailing quotes makes it impossible to load this into chrome

honggyukim:
@kobi-ca Thanks for the report, but could you please provide us a simple and complete example program that shows the problem?

kobi-ca:
sure. will do. And I also need to get back to you on another issue that you resolved. sorry for the delay. Will make it happen this week. take care!

kobi-ca:
Here is an example:
```cpp
#include <iostream>

namespace my_user_literals {

long double operator"" _kg(long double x) { return x * 1000; }

} // namespace my_user_literals

// Driver code
int main() {
  using namespace my_user_literals;
  long double weight = 3.6_kg;
  std::cout << weight << std::endl;
  return 0;
}

```

cmake:
```cmake
cmake_minimum_required(VERSION 3.15)
project(cpp_user_literals_example)

set(CMAKE_CXX_STANDARD 14)

add_executable(cpp_user_literals_example main.cpp)

target_compile_options(cpp_user_literals_example PRIVATE -finstrument-functions -fno-omit-frame-pointer -ggdb)
```

running
```
uftrace record ./cpp_user_literals_example
```

dumping
```
uftrace dump --chrome
{"traceEvents":[
{"ts":0,"ph":"M","pid":32733,"name":"process_name","args":{"name":"[32733] cpp_user_litera"}},
{"ts":0,"ph":"M","pid":32733,"name":"thread_name","args":{"name":"[32733] cpp_user_litera"}},
{"ts":1124421380161.629,"ph":"B","pid":32733,"name":"_GLOBAL__sub_I_my_user_literals::operator"""},
{"ts":1124421380161.963,"ph":"B","pid":32733,"name":"__static_initialization_and_destruction_0"},
{"ts":1124421380162.589,"ph":"B","pid":32733,"name":"std::ios_base::Init::Init"},
{"ts":1124421380284.084,"ph":"E","pid":32733,"name":"std::ios_base::Init::Init"},
{"ts":1124421380285.454,"ph":"B","pid":32733,"name":"__cxa_atexit"},
[...]

Issue #632, opened 2018-12-21
replay: display static and dynamic symbol


namhyung:
To support variables, uftrace replay (and others too, later) should display symbol name rather than memory address.  Showing static symbols (in file) is easy and I extended it to save data symbols in the symbol table/file.

With following example, you can see how the local variable (L) and global variable (G) are displayed when the memory addresses are used.

```c
$ cat name-test.c
static int L;
int G = 1;

int bar(int *p, int v)
{
  *p = v;
  return v + 1;
}

int main(int argc, char *argv[])
{
  bar(&L, L);
  bar(&G, G);

  return 0;
}
```

Now uftrace will show the symbol names.
```
$ uftrace -a name-test
# DURATION     TID     FUNCTION
            [ 23365] | main(1, 0x7ffe1ba07348) {
   0.081 us [ 23365] |   bar(&L, 0) = 1;
   0.074 us [ 23365] |   bar(&G, 1) = 2;
   9.039 us [ 23365] | } = 0; /* main */
```

This is good.  But we mostly need to deal with dynamic object which can be allocated at runtime.  There's no symbol for those objects and we need keep track of them manually.

So I added 'name' trigger to make symbol name and display them.  With some care I hope it'll help users to identify some objects in interest more easily.

Extend the above example to have dynamic objects like below:

```c
$ cat name-test2.c
#include <stdio.h>
#include <stdlib.h>

int G = 1;
static int L;

int foo(void)
{
  free(malloc(0));
  return 1;
}

int bar(int *p, int v)
{
  *p = v;
  return v + 1;
}

int baz(int n)
{
  int i;

  for (i = 0; i < n; i++) {
    int *p = malloc(sizeof(*p));
    bar(p, i);
    free(p);
  }

  bar(&L, L);
  bar(&G, G);

  return n + 2;
}

int main(int argc, char *argv[])
{
  int n = 3;

  if (argc > 1)
    n = atoi(argv[1]);

  foo();
  baz(n);

  return 0;
}
```

The syntax is `name=STR` and it'll work on return value, if you want to do it for arguments do like `name=STR:arg1`.  It can be used in any of `-A`, `-R`, or `-T` options.  Below is normal output.

```
$ uftrace record -a name-test2
$ uftrace replay
# DURATION     TID     FUNCTION
            [ 23475] | main(1, 0x7ffed220f8e8) {
            [ 23475] |   foo() {
   2.160 us [ 23475] |     malloc(0) = 0x563d3dde6400;
[...]

PR #1230, opened 2021-01-16
uftrace: Refactor replay print functions and reuse it in dump

honggyukim:
This patch refactors replay print macros to normal functions and reuse
it in dump command to replace existing escape_func_name() to
print_json_escaped_char() in replay.c.

In addition, the struct type info brings another useful information so
this PR also highlights the type name in yellow color.

Fixed: #1129

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -885,18 +885,7 @@ static void dump_chrome_header(struct uftrace_dump_ops *ops,
 	chrome->last_comma = false;
 }
 
-static void escape_func_name(char *name_buf, const char *name)
-{
-	size_t namelen = strlen(name);
-	size_t i, j;
-
-	for (i = 0, j = 0; i < namelen; i++) {
-		if (name[i] == '"')
-			name_buf[j++] = '\\';
-		name_buf[j++] = name[i];
-	}
-	name_buf[j] = '\0';
-}
+void print_json_escaped_char(char **args, size_t *len, const char c);
 
 static void dump_chrome_task_rstack(struct uftrace_dump_ops *ops,
 				     struct uftrace_task_reader *task, char *name)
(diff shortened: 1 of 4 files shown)


cmds/dump.c
```diff
@@ -909,6 +898,10 @@ static void dump_chrome_task_rstack(struct uftrace_dump_ops *ops,
 	struct uftrace_chrome_dump *chrome = container_of(ops, typeof(*chrome), ops);
 	bool is_process = task->t->pid == task->tid;
 	int rec_type = frs->type;
+	size_t namelen = strlen(name);
+	char *name_buf_ptr = name_buf;
```
namhyung, line 902: I prefer short names for temp local variables, like just 'p' or 'np'?

cmds/dump.c
```diff
@@ -930,7 +923,9 @@ static void dump_chrome_task_rstack(struct uftrace_dump_ops *ops,
 	}
 
 	/* escape the function name */
-	escape_func_name(name_buf, name);
+	for (i = 0; i < namelen; i++)
+		name_buf_ptr += print_json_escaped_char(name_buf_ptr, &len, name[i]);
+	*name_buf_ptr = '\0';
```
namhyung, line 928: I think it should be ok but code analysis tools may concern about the overflow of the name_buf.  We can initialize the len to sizeof(name_buf) - 1 to make them happy.

honggyukim:
@namhyung Thanks for the review. I've updated the PR as you requested.

honggyukim:
I've uploaded one more patch 67398ad5d88dfbd88f9773b61d76846d43cb9097 to highlight the struct type name in yellow color as follows.

**1. `std::shared_ptr` example execution**
image

**2. `node` program execution**
image

utils/debug.c
```diff
@@ -38,6 +38,7 @@ const char *color_reset   = TERM_COLOR_RESET;
 const char *color_bold    = TERM_COLOR_BOLD;
 const char *color_string  = TERM_COLOR_MAGENTA;
 const char *color_symbol  = TERM_COLOR_CYAN;
+const char *color_struct  = TERM_COLOR_YELLOW;
```
namhyung, line 41: I'm ok with it but I'd like to reserve the yellow for a later use (like a user-defined object name in #632).  What about using CYAN for this too?

honggyukim: Hmm.. OK. I can change the color but CYAN is already used by function pointer or global variables so I'm not sure if it's good to add one more category to the same color. How about just using red color instead?

**1. CYAN**
image

**1. RED**
image

namhyung: Well, personally I prefer cyan (and don't want to make it too colorful).  Maybe we can allow users to customize it later..

honggyukim: OK. I've updated it using cyan color.

namhyung (approved):
LGTM

PR #1239, opened 2021-02-23
[WIP] Resolve indirect jumps when patching a function

gpollo:
Hi, I've been working on recovering indirect jumps for dynamic tracing application. For testing purposes, I've used uftrace for experimenting my work. This is more as a proof-of-concept and sharing my work as the requirements on how indirect jumps are handled in uftrace are not as severe as other dynamic instrumentation implementations (e.g. Kprobes) since it only instruments function entry/exit.

The merge request is simple by itself. It depends on the Meson build system. The changes done in uftrace are minimal, but not optimal, since most of the work is done by my library [1]. A simple test is added that checks if the recovered jumps are valid. For more information, see [1].

[1] https://github.com/gpollo/libresolver


namhyung:
Thanks for working on this.

Your work in resolving indirect jump address is interesting.  While uftrace only applies the dynamic tracing to function entry (and exit sometimes), it still wants to resolve indirect jumps in the function body because they can jump back to the prologue (which is replaced by uftrace).

Right now, we just ignore such indirect jumps but I know it's dangerous.  It'd be nice if we can improve this.

Also I'm not sure about the deployment of your library.  Does it build statically?

gpollo:
The library is fetched by the Meson build system then built and linked statically (the `subprojects/libresolver.wrap` file). It could could be built as a shared library, but it's easier to install/use for now with Meson.

namhyung:
Thanks for the confirmation!

Issue #1013, opened 2019-10-23
tui: support a key 'f' for field change in graph mode


honggyukim:
The current tui graph mode only shows "total" by default, but it could show more fields from the command line option `-f`.

It'd be better to provide a way to customize fields in graph mode in interactive mode. I think `f` key can be mapped for this purpose. It could show a new box window just like `h` key for help message, then provide check boxes for each field.

nasafix-nasser:
Hi,
May I work on this?

nasafix-nasser:
Thank you.

honggyukim:
@nasafix-nasser Do you have any update? Please let me know if you have any questions working on this.

hjongkim:
Hello. If there is no update, can I try?

honggyukim:
@hjongkim It's good to see you again. It's thankful if you work on it.

honggyukim:
In addition to this, you can modify tui sort key to set the cursor to the same function. In the current implementation, when `s` key is pressed, the report window is sorted and the cursor goes to the top. But I'd rather want to make the cursor to be set to the same function.

qpakzk:
Hello @honggyukim 
No one seems to have work on this issue, so I tried it.

Is this right what you intended?

https://user-images.githubusercontent.com/20490285/109505716-3f686000-7ae0-11eb-8b23-819837ab7d73.mp4

honggyukim:
@qpakzk Thanks a lot for doing this!

> Is this right what you intended?

Exactly. Could you please send a PR?

namhyung:
@qpakzk nice work!  Thanks for doing this and I'll review the code.

PR #1240, opened 2021-03-02
tui: Customize fields in graph mode interactively

qpakzk:
Hello!

This PR supports users to customize fields for graph mode in interactive mode.

When you press f key, a menu is displayed like h key for help message.

You can see three fields, and each field has a checkbox.
If a field is checked, then the checkbox is checked with 'x'.

If you run uftrace tui by default, 'total' field is already checked.
Depending on using -f option, fields can be already checked differently.

If you press SPACE key, you can (un) check the field.

If you want to move to another field, press DOWN/UP arrow keys or j/k keys.

If you want to apply your customization, press f or ENTER key.

If you want to revoke your customization, press q key.

Fixed: #1013

Signed-off-by: Sangwon Hong <qpakzk@gmail.com>

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -156,8 +156,23 @@ static const char *help[] = {
 	"/             Search",
 	"</>/N/P       Search next/prev",
 	"v             Show debug message",
+	"f             Customize fields in graph mode",
 	"h/?           Show this help",
 	"q             Quit",
+};
+
+#define NUM_GRAPH_FIELD 3
+
+static const char *graph_field_names[NUM_GRAPH_FIELD] = {
+	"TOTAL TIME", "SELF TIME", "ADDRESS",
+};
+
+static const char *graph_field_help[] = {
+	"DOWN/UP ARROW Move down/up",
+	"j/k           Move down/up",
+	"Enter         Apply checked fields",
+	"SPACE         Check or uncheck a field",
+	"f/q           Close the window without any changes",
 };
 
 static const char *report_sort_key[] = { OPT_SORT_KEYS, "self", "call" };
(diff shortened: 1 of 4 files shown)


cmds/tui.c
```diff
@@ -2346,12 +2362,149 @@ static void tui_window_help(void)
...
+			if (field->id == i && !graph_field_flag[i]) {
+				exist = true;
+				del_field(field);
+				break;
+			}
+		}
+
+		if (!exist && graph_field_flag[i])
+			add_field(&graph_output_fields, graph_field_table[i]);
+	}
```
namhyung, line 2394: This code can make the output fields displayed in a different order.  Say, 1. select total and self at first (press ENTER), 2. uncheck total (press ENTER), and finally add total back (press ENTER).  Now total time field comes after the self time.

To preserve the field order, how about this? (not tested)

```c
struct display_field *field, *tmp;
int i;

list_for_each_entry_safe(field, tmp, &graph_output_fields, list)
    del_field(field);

for (i = 0; i < NUM_GRAPH_FIELD; i++) {
    if (graph_field_flags[i])
        add_field(...);
}
```

cmds/tui.c
```diff
@@ -2346,12 +2362,149 @@ static void tui_window_help(void)
...
+	box(win, 0, 0);
+
+	mvwprintw(win, 1, 2, "Customize fields in graph mode");
+
+	for (i = 0; i < ARRAY_SIZE(graph_field_help); i++)
+		mvwprintw(win, i + 3, 2, "%-*.*s", w-3, w-3, graph_field_help[i]);
+
+	display_graph_field(win, selected_field, graph_field_flag);
+
+
```
namhyung, line 2421: Unnecessary blank line.  Please remove.

cmds/tui.c
```diff
@@ -2346,12 +2362,149 @@ static void tui_window_help(void)
...
+	mvwprintw(win, h-1, w-1, "");
+	wrefresh(win);
+
+	bool done = false;
+	while (!done) {
+		int k = wgetch(win);
+
+		switch (k) {
+			case 'k':
+			case KEY_UP:
```
namhyung, line 2431: We don't indent the case label.  Please put it in the sample column as the switch.

cmds/tui.c
```diff
@@ -2346,12 +2362,149 @@ static void tui_window_help(void)
...
+
+			pr_dbg2("check field \"%s\"\n", field->name);
+			if (!check_field_name(field, p))
+				continue;
+
+			graph_field_flag[i] = true;
+			break;
+		}
+	}
+	strv_free(&strv);
```
namhyung, line 2505: I think we can simply do

```c
for (i = 0; i < NUM_GRAPH_FIELD; i++)
    graph_field_flags[i] = graph_field_table[i]->used;
```

cmds/tui.c
```diff
@@ -156,10 +156,26 @@ static const char *help[] = {
...
 	"</>/N/P       Search next/prev",
 	"v             Show debug message",
+	"f             Customize fields in graph mode",
 	"h/?           Show this help",
 	"q             Quit",
 };
 
+#define NUM_GRAPH_FIELD 3
+
+static const char *fields[NUM_GRAPH_FIELD] = {
```
namhyung, line 166: It's too generic.  What about 'graph_field_names'?

cmds/tui.c
```diff
@@ -2578,6 +2735,10 @@ static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
 		case 'v':
 			tui_debug = !tui_debug;
 			break;
+		case 'f':
+			tui_window_graph_field(graph_field_flag, &selected_field);
```
namhyung, line 2739: Do you think we need to keep the selected_field here?  Also graph_field_flag can be initialized inside the function so we can make it not take any arguments.

namhyung (commented):
Thanks for your contribution!  The patches generally look good, and I just made some comments below.  Actually I think we should have same for report fields too.  If you're interested, please consider working on it as a next step. :)

qpakzk:
@namhyung Thanks for your comments.
I'll apply your comments to the code.
And after this commit I'll work on report fields as well.

cmds/tui.c
qpakzk: I agree with initializing `graph_field_flag` inside `tui_window_graph_field` function.

But the reason I kept `selected_field` in `tui_main_loop` function is to remember the location a user last moved.
How do you think about it?

namhyung: Well.. it might be helpful but I think we can simply start with the first item always.

qpakzk: I see. I'll change it!

qpakzk:
@namhyung I applied your reviews to the modified commit.
Could you check it again please?

qpakzk:
@namhyung I inserted `void` in the parameter for `tui_window_graph_field` function I implemented.
I also put `void` in the parameter for functions missing it such as `cancel_search` function and `display_loading_msg` function.

namhyung (approved):
LGTM, thank you so much!

PR #1243, opened 2021-03-08
test: Add -c, --compiler option for clang

iamyooon:
Add an option to specify the compiler type so that runtest.py
can work with nor only gcc but also clang.

Signed-off-by: SeokHoon Yoon <iamyooon@gmail.com>


honggyukim (approved):
Thanks for doing this work!

namhyung (approved):
LGTM

Issue #771, opened 2019-05-30
sched event with -F or -C filters


honggyukim:
When caller filter is used, we expect to see only the related functions, but it shows too many sched events as follows:
```
$ uftrace replay -C QQmlEnginePrivate::QQmlEnginePrivate
# DURATION     TID     FUNCTION
            [  4967] | QQmlEngine::QQmlEngine() {
            [  4967] |   QQmlEnginePrivate::QQmlEnginePrivate() {
  34.250 us [  4967] |     /* linux:schedule */
  12.666 us [  4967] |     /* linux:schedule */
  12.542 us [  4967] |     /* linux:schedule */
  11.750 us [  4967] |     /* linux:schedule */
  16.041 us [  4967] |     /* linux:schedule */
  11.541 us [  4967] |     /* linux:schedule */
  12.000 us [  4967] |     /* linux:schedule */
  11.292 us [  4967] |     /* linux:schedule */
  17.250 us [  4967] |     /* linux:schedule */
  11.583 us [  4967] |     /* linux:schedule */
  11.959 us [  4967] |     /* linux:schedule */
  11.334 us [  4967] |     /* linux:schedule */
  30.042 us [  4967] |     /* linux:schedule */
  12.209 us [  4967] |     /* linux:schedule */
  12.041 us [  4967] |     /* linux:schedule */
  12.042 us [  4967] |     /* linux:schedule */
  14.542 us [  4967] |     /* linux:schedule */
  11.167 us [  4967] |     /* linux:schedule */
  11.667 us [  4967] |     /* linux:schedule */
  11.416 us [  4967] |     /* linux:schedule */
  15.500 us [  4967] |     /* linux:schedule */
  11.334 us [  4967] |     /* linux:schedule */
  11.875 us [  4967] |     /* linux:schedule */
  37.083 us [  4967] |     /* linux:schedule */
  15.833 us [  4967] |     /* linux:schedule */
  26.584 us [  4967] |     /* linux:schedule */
   3.139 ms [  4967] |     /* linux:schedule */
   6.233 ms [  4967] |   } /* QQmlEnginePrivate::QQmlEnginePrivate */
  35.750 us [  4967] |   /* linux:schedule */
  24.750 us [  4967] |   /* linux:schedule */
  45.875 us [  4967] |   /* linux:schedule */
...
```
It seems that we better hide sched event when caller filter is used or the sched event needs to be combined into a group.

honggyukim:
[...]

Issue #1173, opened 2020-05-23
analysis commands still shows linux:schedule event for caller filter


honggyukim:
`linux:schedule` event is disabled for caller filter during record as follows.
- d908312 record: Disable linux:schedule event for caller filter

However, it's still shown in analysis commands such as replay, etc. `linux:schedule` also has to be disabled for analysis commands.

In addition, it has to be applied when caller trigger is enabled as well.

namhyung:
It should be closed by 55ab685b.

PR #1245, opened 2021-03-28
uftrace: Add --no-sched option to disable schedule event

honggyukim:
uftrace: Add --no-sched option to disable schedule event

There is --no-event option but it also disables other useful events that
we don't want to miss out such as task name events, etc.

So we need to disable only schedule events and this patch adds
--no-sched option to disable schedule event only while other events are
still available.

In addition, there was no way to disable schedule event for analysis commands
but --no-sched can do this now.  Let's simply enable it for both record and
other analysis commands when caller filter is used.

Fixed: #771, #1173


honggyukim:
@namhyung Could you please have a look at this?

honggyukim:
Rebase is done. Please have a look again. Thanks.

honggyukim:
Rebase is done again.

cmds/record.c
```diff
@@ -1674,7 +1674,13 @@ static void check_perf_event(struct opts *opts)
 	bool found = false;
 	enum uftrace_pattern_type ptype = opts->patt_type;
 
-	has_perf_event = has_sched_event = !opts->no_event;
+	if (opts->no_event) {
+		has_perf_event = false;
+		has_sched_event = false;
+	}
```
namhyung, line 1680: I think this part (and the change above) looks unnecessary.  Just adding the below is enough, no?

utils/fstack.c
```diff
@@ -1033,6 +1033,11 @@ bool fstack_check_opts(struct uftrace_task_reader *task, struct opts *opts)
 	if (opts->no_event && rec->type == UFTRACE_EVENT)
 		return false;
 
+	if (opts->no_sched && (rec->addr == EVENT_ID_PERF_SCHED_IN ||
+			       rec->addr == EVENT_ID_PERF_SCHED_OUT ||
+			       rec->addr == EVENT_ID_PERF_SCHED_BOTH))
```
namhyung, line 1038: This is repeating.. you can consider adding a helper function like `is_sched_event(u64 addr)`.

honggyukim: I made a helper `is_addr_sched_event` for this.

cmds/record.c
honggyukim: > I think this part (and the change above) looks unnecessary.

Do you mean we don't have to have 2 if statements for `opts->no_event` and `opts->no_sched`?

> Just adding the below is enough, no?

Which below is enough?

namhyung: Your code right below here: `if (opts->no_sched) has_sched_event =false;`.  You can omit the brace too.

honggyukim: > I think this part (and the change above) looks unnecessary. Just adding the below is enough, no?

1) Do you mean this?
```diff
-        if (opts->no_event) {
-                has_perf_event = false;
-                has_sched_event = false;
-        }
         if (opts->no_sched) {
                 has_sched_event = false;
         }
```
> Your code right below here: if (opts->no_sched) has_sched_event =false;. You can omit the brace too.

2) Do you mean this?
```diff
         if (opts->no_event) {
                 has_perf_event = false;
                 has_sched_event = false;
         }
-        if (opts->no_sched) {
+        if (opts->no_sched)
                 has_sched_event = false;
-        }
```

3) Otherwise this? (How do we switch off schedule event then?)
```diff
         if (opts->no_event) {
                 has_perf_event = false;
                 has_sched_event = false;
         }
-        if (opts->no_sched) {
-                has_sched_event = false;
-        }
```

namhyung: Yes, 1 + 2.  Also defaults back to false.

honggyukim: Hmm.. I'm getting confused. Do you mean this?
```diff
diff --git a/cmds/record.c b/cmds/record.c
index c3fbc9cb..447d02f9 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -59,7 +59,7 @@ static pthread_mutex_t write_list_lock = PTHREAD_MUTEX_INITIALIZER;
 static bool buf_done;
 static int thread_ctl[2];

-static bool has_perf_event = true;
+static bool has_perf_event = false;
 static bool has_sched_event = true;
 static bool finish_received;

@@ -1674,13 +1674,8 @@ static void check_perf_event(struct opts *opts)
        bool found = false;
        enum uftrace_pattern_type ptype = opts->patt_type;

-       if (opts->no_event) {
-               has_perf_event = false;
-               has_sched_event = false;
-       }
-       if (opts->no_sched) {
+       if (opts->no_sched)
                has_sched_event = false;
-       }

        if (opts->event == NULL)
                return;
```

namhyung: Something like this?

```c
static bool has_perf_event;
static bool has_sched_event;

...

void check_perf_event(struct opts *opts)
{
   ...

   has_perf_event = has_sched_event = !opts->no_event;

   if (opts->no_sched)
      has_sched_event = false;
```

honggyukim: Thanks. updated it again.

namhyung (approved):
LGTM

PR #1249, opened 2021-04-18
aarch64: dynamic: Add debug messages for patched function

honggyukim:
Unlike x86_64, aarch64 doesn't print debug messages for patched
function so this patch adds it same as x86_64.
```
  $ gcc -o t-abc tests/s-abc.c

  $ uftrace record -P. --debug-domain dynamic:2 t-abc
  dynamic: patch normal func: call_weak_fn (patch size: 8)
  dynamic: patch normal func: a (patch size: 8)
  dynamic: patch normal func: b (patch size: 8)
  dynamic: patch normal func: c (patch size: 8)
  dynamic: patch normal func: main (patch size: 8)
  dynamic: patched all (5) functions in 't-abc'
```


namhyung (approved):
LGTM

PR #381, opened 2018-06-15
uftrace: Add .clang-format file for better code styling

honggyukim:
clang-format describes configurable formatting style options.
It'd be useful when asking someone to follow coding-style guide.

Since uftrace almost follow Linux kernel coding style guide, this patch
copies .clang-format from Linux kenel source and adopt it to follow
uftrace coding style.

Link: https://clang.llvm.org/docs/ClangFormat.html
Link: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
Signed-off-by: Honggyu Kim <honggyu.kp@gmail.com>

PR #1255, opened 2021-05-10
Copy Linux kernel's clang-format configuration

gpollo:
Since we are following the Kernel coding style, it may be a good idea to use the Kernel `clang-format` configuration. This commit does that and also formats the current source code based on the configuration. I copied the kernel's configuration and removed kernel-specific `ForEachMacros` macros.

You can format uftrace source code with this command executed from the root directory:
```
$ clang-format $(find . -type f -iname "*.c") $(find . -type f -iname "*.h") -i
```

Also, would it be a good idea to add this to the build system (e.g. `make format`)?

NOTE: There is a lot of changes, done by clang-format. I had to remove formatting in some areas (search for  `/* clang-format off */` and `/* clang-format on */`) to either make the code compile or make some part clearer to understand. We can adjust the `.clang-format` file if needed.


arch/arm/mcount-support.c
```diff
@@ -429,7 +416,9 @@ void mcount_get_stack_arg(struct mcount_arg_context *ctx,
 			if (spec->size <= 4)
 				offset = spec->idx - ARCH_MAX_FLOAT_REGS;
 			else
-				offset = (spec->idx - ARCH_MAX_DOUBLE_REGS) * 2 - 1;
+				offset =
+					(spec->idx - ARCH_MAX_DOUBLE_REGS) * 2 -
+					1;
```
namhyung, line 421: I really don't like that, can we suppress such kind of transformation?

arch/aarch64/mcount-arch.h
```diff
@@ -1,38 +1,38 @@
 #ifndef MCOUNT_ARCH_H
 #define MCOUNT_ARCH_H
 
-#define mcount_regs  mcount_regs
+#define mcount_regs mcount_regs
```
namhyung, line 4: Actually I prefer 2 spaces after define.

.clang-format
```diff
@@ -0,0 +1,134 @@
...
+  AfterControlStatement: false
+  AfterEnum: false
+  AfterFunction: true
+  AfterNamespace: true
+  AfterObjCDeclaration: false
+  AfterStruct: false
+  AfterUnion: false
+  #AfterExternBlock: false # Unknown to clang-format-5.0
+  BeforeCatch: false
+  BeforeElse: false
```
namhyung, line 43: Does it places a closing brace on a different line before the following else?  It's the main difference of our style.

.clang-format
```diff
@@ -0,0 +1,134 @@
...
+#SortUsingDeclarations: false # Unknown to clang-format-4.0
+SpaceAfterCStyleCast: false
+SpaceAfterTemplateKeyword: true
+SpaceBeforeAssignmentOperators: true
+#SpaceBeforeCtorInitializerColon: true # Unknown to clang-format-5.0
+#SpaceBeforeInheritanceColon: true # Unknown to clang-format-5.0
+SpaceBeforeParens: ControlStatements
+#SpaceBeforeRangeBasedForLoopColon: true # Unknown to clang-format-5.0
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
```
namhyung, line 125: Is it for comments in the same line? Then I'd prefer 2.

arch/aarch64/mcount-insn.c
```diff
@@ -249,9 +248,9 @@ int disasm_check_insns(struct mcount_disasm_engine *disasm,
 		if (state) {
 			if (!modify_instruction(disasm, &insn[i], mdi, info))
 				goto out;
-		}
-		else {
-			memcpy(info->insns + info->copy_size, insn[i].bytes, insn[i].size);
+		} else {
```
namhyung, line 251: I'd like to keep the original style here.

.clang-format
```diff
@@ -0,0 +1,134 @@
...
+  #SplitEmptyNamespace: true # Unknown to clang-format-4.0
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+#BreakBeforeInheritanceComma: false # Unknown to clang-format-4.0
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializersBeforeComma: false
+#BreakConstructorInitializers: BeforeComma # Unknown to clang-format-4.0
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: false
+ColumnLimit: 80
```
namhyung, line 56: It seems this limit is too hard.  I mean that I'd like to see most of the code would keep this limit but sometimes it's better to have them in a same line.  Afaik the kernel recently increased this limit to 100 maybe we can do the same while recommending 80 column limit as a common case.

namhyung (commented):
Thank you for doing this!  I'm not satisfied with the whole result but I understand the automatic tool's limitation.  A couple of comments below..

.clang-format
gpollo: Right now, it currently puts the brace on the same line as the `else`, but I'll change this settings.

.clang-format
gpollo: Yeah, it is for comment on the same line, I'll change it to 2, the diff looks like this for some lines.
```diff
-                       exename = ptr + 8 + 1; // skip double-quote
+                       exename = ptr + 8 + 1;  // skip double-quote
```

.clang-format
gpollo: I agree, 80 is a small amout. The code looks better IMO with a 100 characters limit.

arch/arm/mcount-support.c
gpollo: With the 100 characters limit, we get the following diff:
```diff
-                               offset =
-                                       (spec->idx - ARCH_MAX_DOUBLE_REGS) * 2 -
-                                       1;
+                               offset = (spec->idx - ARCH_MAX_DOUBLE_REGS) * 2 - 1;
```

honggyukim:
Thanks for your work! I've actually raised this issue at #381, but didn't finish the work. I hope this PR can be split into 3 parts.
1. copy `.clang-format` file from Linux kernel
2. modify `.clang-format` file for uftrace coding style and create a make target or a script
3. apply the changes to uftrace code base with a prepared command at step 2

I don't work for it these days so you can take over the ownership if you want. Thanks.

namhyung:
So I think it's done with #381 and close this.  Please reopen it if you have additional issues.

Issue #873, opened 2019-09-16
Enhance bash-completion?


DanielTimLee:
I think the Bash completion can help new users get a full picture of the various uftrace options.
Well, `uftrace -h` could help though, but it's size is huge and hard to read.

Furthermore, I'm not sure it's working correctly?
I think it's stuck since 1cad398dc 2017-11-24.

What I thought was kind of like this.
Auto-completion of commands will help to get know about required arguments, or etc..
Reference for bash-completion (bpftool). 

[![asciicast](https://asciinema.org/a/268714.svg)](https://asciinema.org/a/268714)

https://github.com/torvalds/linux/blob/master/tools/bpf/bpftool/bash-completion/bpftool#L1-L809

DanielTimLee:
Sure!
I'm not working on this right now, so feel free to go ahead.

2021년 5월 11일 (화) 12:38, Kang Minchul ***@***.***>님이 작성:

PR #1264, opened 2021-05-16
build: Relocate bash completion script

kangtegong:
This commit fixes bash completion feature,
which is not working right now.

misc/bash_completion.sh should be relocated, and
we can get recommended location of the script wq!:with
`pkg-config --variable=completionsdir bash-completion`.

Other directories, including /usr/local/etc/bash_completions.d
are only present for backwords compatibility, so its usage is
no longer recommended.

Fixed: #873

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,7 +322,7 @@ $(objdir)/misc/dbginfo: $(DBGINFO_OBJS)
 install: all
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(bindir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(libdir)
-	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(etcdir)/bash_completion.d
+	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(compldir)
 ifneq ($(wildcard $(elfdir)/lib/libelf.so),)
 ifeq ($(wildcard $(prefix)/lib/libelf.so),)
 	# install libelf only when it's not in the install directory.
@@ -339,7 +339,7 @@ endif
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-single.so $(DESTDIR)$(libdir)/libmcount-single.so
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-fast-single.so $(DESTDIR)$(libdir)/libmcount-fast-single.so
 	$(call QUIET_INSTALL, bash-completion)
-	$(Q)$(INSTALL) -m 644 $(srcdir)/misc/bash-completion.sh $(DESTDIR)$(etcdir)/bash_completion.d/uftrace
+	$(Q)$(INSTALL) -m 644 $(srcdir)/misc/bash-completion.sh $(DESTDIR)$(compldir)/uftrace
 	@$(MAKE) -sC $(docdir) install DESTDIR=$(DESTDIR)$(mandir)
 	@if [ `id -u` = 0 ]; then ldconfig $(DESTDIR)$(libdir) || echo "ldconfig failed"; fi
 
@@ -357,7 +357,7 @@ uninstall:
 	$(call QUIET_UNINSTALL, libmcount-fast-single)
 	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount-fast-single.so
 	$(call QUIET_UNINSTALL, bash-completion)
-	$(Q)$(RM) $(DESTDIR)$(etcdir)/bash_completion.d/uftrace
+	$(Q)$(RM) $(DESTDIR)$(compldir)/uftrace
 	@$(MAKE) -sC $(docdir) uninstall DESTDIR=$(DESTDIR)$(mandir)
 
 test: all
--- a/configure
+++ b/configure
@@ -139,6 +139,7 @@ bindir=${bindir:-${prefix}/bin}
 libdir=${libdir:-${prefix}/lib}
 etcdir=${etcdir:-${prefix}/etc}
 mandir=${mandir:-${prefix}/share/man}
+compldir=${compldir:-${prefix}/share/bash-completion/completions}
 
 if [ "$etcdir" = /usr/etc ]; then
     etcdir=/etc
@@ -268,6 +269,7 @@ override bindir := $bindir
 override libdir := $libdir
 override mandir := $mandir
 override etcdir := $etcdir
+override compldir := $compldir
 EOF
 
 if [ ! -z $with_elfutils ]; then


kangtegong:
Reference: https://github.com/scop/bash-completion#faq

Q. I author/maintain package X and would like to maintain my own completion code for this package. Where should I put it to be sure that interactive bash shells will find it and source it?

A. The recommended directory is completionsdir, which you can get with pkg-config --variable=completionsdir bash-completion.

honggyukim:
I cannot install with this PR when `configure --prefix` is used.
```
$ ./configure --prefix=$HOME/usr
$ make
$ make install
  INSTALL  uftrace
  INSTALL  libmcount
  INSTALL  bash-completion
install: cannot create regular file '/usr/share/bash-completion/completions/uftrace': Permission denied
Makefile:324: recipe for target 'install' failed
make: *** [install] Error 1
```
In addition, I don't remember how to add this bash complete file when it's installed in a custom path.

namhyung:
Sure, all installation directories should honor the prefix setting and users sometimes want to install it in a non-system directory.

kangtegong:
how do you think about this commit then?
This one will guarantee bash completion work correctly 
when `sudo make install`ed or installed by root at least.

```
ifeq ($(shell id -u),0)	# check if make install was ran by root, or sudo make installed
  completionsdir = $(shell pkg-config --variable=completionsdir bash-completion)
else			# other installations, including ./configure --prefix
  completionsdir = $(etcdir)/bash-completion.d
endif
```

Unfortunately, I don't know how to auto complete commands which are not in $PATH.
`complete -o filenames -F _uftrace uftrace` in misc/bash-completion.sh
works only when `uftrace` command is in $PATH or uftrace is aliased, etc.

+ Thanks for code review :) You don't have to merge if this doesn't look good enough.

namhyung:
I think it's better to check if the prefix is `/usr` instead of uid being 0.  For some cases, they want to install it in a different location.

Currently it install the completion script to $(etcdir)/bash_completion.d, but you can change it to $(prefix)/$(remaining-completion-dir) if pkg-config has it.

kangtegong:
Ok, I just amended the commit like below

```
ifeq ($(shell echo $(DESTDIR)$(prefix) | head -c 5),/usr/)  # check if prefix of installation location is /usr/
ifneq ($(shell pkg-config --variable=completionsdir bash-completion),)  # check pkg-config has bash-completion directory
	$(Q)$(INSTALL) -m 644 $(srcdir)/misc/bash-completion.sh $(shell pkg-config --variable=completionsdir bash-completion)/uftrace
endif
endif
```

kangtegong:
just amended commit for code readability. changed `head` to `cut`

```Makefile
 # check if prefix of installation location is /usr/
ifeq ($(shell echo $(DESTDIR)$(prefix) | cut -c 1-5),/usr/)
# check pkg-config has bash-completion directory
ifneq ($(shell pkg-config --variable=completionsdir bash-completion),)
	$(Q)$(INSTALL) -m 644 $(srcdir)/misc/bash-completion.sh $(shell pkg-config --variable=completionsdir bash-completion)/uftrace
endif
endif
```

kangtegong:
Thanks for reply :) 
just rebased and amended commit as you mentioned.

Makefile
```diff
@@ -322,7 +322,6 @@ $(objdir)/misc/dbginfo: $(DBGINFO_OBJS)
 install: all
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(bindir)
 	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(libdir)
-	$(Q)$(INSTALL) -d -m 755 $(DESTDIR)$(etcdir)/bash_completion.d
```
namhyung, line 325: I think you still need to create $(DESTDIR)$(compldir)

Makefile
```diff
@@ -338,8 +337,10 @@ endif
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-fast.so $(DESTDIR)$(libdir)/libmcount-fast.so
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-single.so $(DESTDIR)$(libdir)/libmcount-single.so
 	$(Q)$(INSTALL) $(objdir)/libmcount/libmcount-fast-single.so $(DESTDIR)$(libdir)/libmcount-fast-single.so
+ifneq ($(compldir),)
```
namhyung, line 340: Let's make this unconditional by setting compledir as the original value.

Makefile
```diff
@@ -356,8 +357,10 @@ uninstall:
 	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount-single.so
 	$(call QUIET_UNINSTALL, libmcount-fast-single)
 	$(Q)$(RM) $(DESTDIR)$(libdir)/libmcount-fast-single.so
+ifneq ($(compldir),)
```
namhyung, line 360: Ditto.

configure
```diff
@@ -138,13 +138,23 @@ bindir=${bindir:-${prefix}/bin}
...
 mandir=${mandir:-${prefix}/share/man}
+compldir=
 
 if [ "$etcdir" = /usr/etc ]; then
     etcdir=/etc
 fi
 if [ -n "$sysconfdir" ]; then
     etcdir=$sysconfdir
 fi
+if [ ${prefix:0:10} = /usr/local ]; then
```
namhyung, line 149: We care only if it installs to the system directory (i.e. prefix is "/usr").  So it should be:

```
if [ "$prefix" = /usr ]; then
```

configure
```diff
@@ -138,13 +138,23 @@ bindir=${bindir:-${prefix}/bin}
 libdir=${libdir:-${prefix}/lib}
 etcdir=${etcdir:-${prefix}/etc}
 mandir=${mandir:-${prefix}/share/man}
+compldir=
```
namhyung, line 141: Please remove this.

configure
```diff
@@ -138,13 +138,23 @@ bindir=${bindir:-${prefix}/bin}
...
 if [ -n "$sysconfdir" ]; then
     etcdir=$sysconfdir
 fi
+if [ ${prefix:0:10} = /usr/local ]; then
+    if [ -d "$(pkg-config --variable=completionsdir bash-completion)" ]; then
+        compldir=$(pkg-config --variable=completionsdir bash-completion)
+    elif [ -d "$(pkg-config --variable=compatdir bash-completion)" ]; then
+        compldir=$(pkg-config --variable=compatdir bash-completion)
+    else
+        continue
```
namhyung, line 155: continue ?  Better to remove it.

configure
```diff
@@ -138,13 +138,23 @@ bindir=${bindir:-${prefix}/bin}
...
 etcdir=${etcdir:-${prefix}/etc}
 mandir=${mandir:-${prefix}/share/man}
+compldir=
 
 if [ "$etcdir" = /usr/etc ]; then
     etcdir=/etc
 fi
 if [ -n "$sysconfdir" ]; then
     etcdir=$sysconfdir
 fi
```
namhyung, line 148: I think we need to set the default value of compldir here like:

```
compldir=${compldir:-${etcdir}/bash_completion.d}
```

namhyung (commented):
I'm really sorry about being too late...

kangtegong:
i just amend the commit as @namhyung  said, and clarified `compldir` path in configure script in simpler way.

Thanks for feedbacks!

kangtegong:
if this commit is ok, I want to leave these other issue and amend misc/bash-completion.sh.

1. auto-suggest uftrace arguments hierarchically

2. enable bash completion when uftrace is built with prefix given, and installed binary is not in $PATH (e.g. ./configure --prefix=$HOME/usr) because as far as I know, currently `misc/bash-completion.sh` is supported only when uftrace binary is in $PATH

configure
```diff
@@ -139,6 +139,7 @@ bindir=${bindir:-${prefix}/bin}
 libdir=${libdir:-${prefix}/lib}
 etcdir=${etcdir:-${prefix}/etc}
 mandir=${mandir:-${prefix}/share/man}
+compldir=${compldir:-${prefix}/share/bash-completion/completions}
```
namhyung, line 142: It should be `compldir=${compldir:-${etcdir}/bash_completion.d}.

namhyung (commented):
I cannot find pkg-config usage anymore.

namhyung:
> 2. enable bash completion when uftrace is built with prefix given, and installed binary is not in $PATH (e.g. ./configure --prefix=$HOME/usr) because as far as I know, currently `misc/bash-completion.sh` is supported only when uftrace binary is in $PATH

But I don't think it's a good idea.  For the custom build version, users need to update the script.  I don't want to touch it unnecessarily.

PR #1269, opened 2021-05-21
dynamic: Add client command, to control uftrace at runtime

clementguidi:
Hi,

In order to achieve full dynamic tracing, we need to be able to control the running uftrace instance.

This patch implements the new **client** command, designed to send commands to a **daemon run inside libmcount**.
At the moment, the following features are supported:
- depth (`-D`) and time filter (`-t`)
- filters (`-F` and `-N`) and caller-filter (`-C`)
- triggers (`-T`)
- argument and return value info (`-A` and `-R`)

The client can be extended to support dynamic patching options `-P` and `-U`.

By default, a daemon is spawned inside libmcount as a new thread, listening on the `/var/run/user/$UID/uftrace/$PID.socket` socket for incoming commands. The client uses the `-p <pid>` option to locate the daemon.

Example usage:

On terminal **1**:
```
$ uftrace record t-abc
```

On terminal **2** (suppose the command is sent _before_ function `b` is entered):
```
$ uftrace -p $(pidof t-abc) -D 2 client
```

The replay command yields:
```
$ uftrace replay
            [  3654] | main() {
   2.445 us [  3654] |   a();
   4.579 us [  3654] | } /* main */
```

See more examples in the client man page.

Edit: the daemon has to be enabled at startup using the `--daemon` option.

Signed-off-by: Clément Guidi <clementguidi@gmail.com>


namhyung:
Thanks for contributing this work!  The idea of changing options and settings at runtime looks interesting.

But in general, I don't want to affect target processes behavior (like creating a new thread) as much as possible.  Also I'm afraid that it slows down the fast path by taking locks (even though it usually just add a few atomic instructions) for the now-mutable filters and triggers.

So I think it should be disabled by default, and might have a separate library with a conditional compilation (like libmcount-single.so).  Then we can (pre)load the library only if user asks it and keep the normal cases without unnecessary performance degradation.

Also I found a bunch of coding style issues in the code.  Please fix them!

gpollo:
If there is only a single socket in `/var/run/user/$UID/uftrace/` (only one instance of `uftrace` running), maybe we can omit the `--pid PID` parameter and automatically choose it. This can be done later I think.

honggyukim:
Thanks very much for doing this! I have tested this PR and it works great.

I couldn't read all the details but how about changing the command name from `client` to `apply`? I think this feature applies filters at runtime so `apply` might be a better name. I know naming is always difficult :)

gpollo:
I believe it was initialy named `client` because it's a client-server architecture. `libmcount.so` is the server waiting on socket connections and `uftrace client` is the client since it connects to the server's socket. I agree that `client` is not that descriptive from the user's perspective, but `apply` is also no quite descriptive because in #1274 we also dynamically patch using this daemon.

honggyukim:
> I believe it was initialy named client because it's a client-server architecture.

Yes, I thought `client` comes from it.

> I agree that client is not that descriptive from the user's perspective, but apply is also no quite descriptive because in #1274 we also dynamically patch using this daemon.

Hmm.. We might have different thoughts but I think patching and unpatching looks like applying `-P`/`-U` options at runtime. Let's see how @namhyung thinks about this.

yottapulsky:
I rearranged the code so the daemon builds as a separate library `libmcount-daemon.so`. It is used when the `--daemon` option is specified.

As gpollo suggested, the client is now able to guess the daemon's PID from the open sockets, so there is no need to specify the PID when there is a single daemon instance running.

> Hmm.. We might have different thoughts but I think patching and unpatching looks like applying `-P`/`-U` options at runtime. Let's see how @namhyung thinks about this.

Let me know what name is more appropriate, I agree that `apply` fits better.

honggyukim:
@yottapulsky Thanks very much for your update. But I think this PR requires deeper discussion so I think it might be good to have a video call together with @namhyung and @gpollo if possible. We can also talk about general discussion not only for this PR. How do you think?

We can talk more at the chatting room https://gitter.im/uftrace/uftrace.

Issue #1268, opened 2021-05-20
Compiler warning on utils/script-python.c


namhyung:
Now I'm seeing a repeated compiler warning from the python code.

```
  CC FPIC  libmcount/script-python.op
uftrace/utils/script-python.c:39:1: warning: ‘visibility’ attribute ignored [-Wattributes]
   39 | static PyAPI_FUNC(void) (*__Py_Initialize)(void);
      | ^~~~~~
uftrace/utils/script-python.c:40:1: warning: ‘visibility’ attribute ignored [-Wattributes]
   40 | static PyAPI_FUNC(void) (*__Py_Finalize)(void);
      | ^~~~~~
uftrace/utils/script-python.c:41:1: warning: ‘visibility’ attribute ignored [-Wattributes]
   41 | static PyAPI_FUNC(void) (*__PySys_SetPath)(char *);
      | ^~~~~~
uftrace/utils/script-python.c:42:1: warning: ‘visibility’ attribute ignored [-Wattributes]
   42 | static PyAPI_FUNC(PyObject *) (*__PyImport_Import)(PyObject *name);
      | ^~~~~~
...
```

I guess the attribute is hidden in the PyAPI_FUNC macro.  Also I don't think we need the macro from the beginning..

kangtegong:
I'll remove the macro and perform tests, if you don't mind.
If it's ok to remove, I'll send the patch.

namhyung:
It'd be nice if you work on it, but I'd wait for @honggyukim in case I missed something.

honggyukim:
I don't see such warnings. Which compiler do you use?

Anyway, `PyAPI_FUNC` is defined at [here](https://github.com/python/cpython/blob/4d4be477055d0f9e415407bcd193788fae04dc11/Include/pyport.h#L699) as follows.
```c
/* If no external linkage macros defined by now, create defaults */
#ifndef PyAPI_FUNC
#       define PyAPI_FUNC(RTYPE) Py_EXPORTED_SYMBOL RTYPE
#endif
```
The `Py_EXPORTED_SYMBOL` is defined at [here](https://github.com/python/cpython/blob/4d4be477055d0f9e415407bcd193788fae04dc11/Include/exports.h#L21) as follows.
```c
#if defined(_WIN32) || defined(__CYGWIN__)
    ...
#else
    ...
    #if (defined(__GNUC__) && (__GNUC__ >= 4)) ||\
        (defined(__clang__) && __has_attribute(visibility))
[...]

PR #1271, opened 2021-05-31
utils: Remove unnecessary python macro causing compiler warning

kangtegong:
This commit removes `PyAPI_FUNC` in `utils/script-python.c`,
which causes compiler warning.

This macro is used only for python script within this file,
and doesn't affect in any way, it's ok to remove.

Fixed: #1268

### DIFF
--- a/utils/script-python.c
+++ b/utils/script-python.c
@@ -36,45 +36,45 @@ static bool python_error_reported = false;
 /* whether script_init() was done successfully */
 static bool python_initialized;
 
-static PyAPI_FUNC(void) (*__Py_Initialize)(void);
-static PyAPI_FUNC(void) (*__Py_Finalize)(void);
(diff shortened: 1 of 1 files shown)


honggyukim (changes requested):
It doesn't compile. We just remove `PyAPI_FUNC` macro but leaves the declarations for the function pointers. For example,
```
static PyAPI_FUNC(void) (*__Py_Initialize)(void);
```
should be
```
static void (*__Py_Initialize)(void);
```

kangtegong:
just amended the commit! thanks for feedback :)

namhyung (approved):
LGTM

namhyung:
@honggyukim can you approve the change now?

honggyukim (approved):
Yes, LGTM.

PR #1274, opened 2021-06-08
[WIP] [RFC] Implement runtime patching for x86_64 architecture

gpollo:
Hi, [our group](https://www.dorsal.polymtl.ca/en/) has been working for a bit on runtime patching for `uftrace`. The first patches has already been submitted by #1269 and adds a command daemon inside `libmcount.so`. This pull request, based on #1269, implement safe runtime patching for the x86_64 architecture.

The command daemon is currently started automatically when running `uftrace` on your program:
```
$ uftrace live --force <prog>
```

You can now insert/remove or activate/desactivate dynamic tracepoints using one of the following commands:
```
$ uftrace client --pid $(pidof <prog>) -P <func>
$ uftrace client --pid $(pidof <prog>) -U <func>
```

* The first few commits are a copy of #1269. I'll rebase this PR when #1269 will be updated/merged.
* The next few arch-independent commits changes some of the internal of `libmcount.so` to make runtime patching work.
* The last 3 arch-dependent commits implement safe runtime patching/unpatching for x86_64.

_NOTE_: While this work has been tested quite a lot, it is still a work-in-progress and there may be issues. It also changes some behaviours of `uftrace` (mainly prepending of match-all patterns), so comments/suggestions are welcome.


namhyung:
Wow, thanks for the great work!  I'll take a look.

arch/x86_64/mcount-dynamic.c
```diff
@@ -373,95 +760,124 @@ static int patch_xray_func(struct mcount_dynamic_info *mdi, struct sym *sym)
...
+	 * it will check if the next instruction pointer is in the patching
+	 * region. If it is, will move the next instruction pointer to the
+	 * equivalent modified instruction.
+	 */
+
+	if (map_move_instructions(original_code_addr, modified_code_addr, original_code_size) < 0) {
+		pr_dbg("failed to map instructions to trampoline\n");
+	}
+
+	send_sigusr1_to_all_threads();
```
namhyung, line 962: It seems that it sends the SIGUSR1 whenever it (un)patches a function to check the current instruction.  Looks like too heavy..

gpollo: This is one way of moving threads away from the patching region. There are other methods.

# stop-the-world method
In this method, all the threads of the program are stopped. The patching thread then check all the program-counter of each thread. If a thread is in the patching region, we move it to the equivalent instruction in the out-of-line execution (OLX) buffer. Then, the threads can be restarted.
* The main advantage of this method is that it's less intrusive than the current methods, except that it may slow to program considerably.

# current method
Our current way is to send a signal to each thread. In the signal handler, we check our current thread program counter. If it is in the patching region, we move it to the equivalent instruction in the OLX buffer.
* This method is more intrusive since it uses a signal, here `SIGUSR1`, that the traced application might be using. That said, it is usually faster than the first stop-the-world approach.
* There is other corner cases with the nested signal handlers.
* Usually okay fast to the patching thread and the other threads.

# sleep method
If we don't want to send a signal or stop all threads, then there is a much simpler way. We could simply wait for a specific amount of time, let's say 1 second, and assume that any thread that was executing the patching region has exited it.
* This method is not safe since we can't guarantee that a thread that was in the patching region was resumed by the scheduler and has exited the patching region.

# sleep method (safer)
If we want a safer approach using the wait method, we can technically poll other threads' program counters by reading `/proc/self/task/%d/stat`. If we see that all program counter are outside the patching region, we can safely continue with the patching process.
* We do not use signals, so it is less intrusive.
* This patching process, from the user's perspective would probably be slower, albeit in most cases probably unnoticeable.

# how it's done in the kernel
Anybody familiar with RCU will know that all those methods a very similar to how a grace period might be done. That's probably why the Linux kernel [optimized Kprobe](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/kprobes/opt.c#L483) uses `synchronize_rcu_tasks` in order to wait for other threads. Without going too much into details, the kernel is "aware" of context switches and can exploit this fact to know when a thread has exited a critical section. I don't think it is possible in userspace, the [userspace-rcu](https://github.com/urcu/userspace-rcu#usage-of-liburcu-memb) project's grace period requires reader-side (other threads) critical sections, which we don't have.

# changes
I was thinking of maybe implementing the safe sleep method. I think there are some cases where it can be very slow. For example, if a thread is in the patching region and the system is congested, the patching thread may wait for a long time. That said, I think it's statistically unlikely and I believe that any method on a congested system will be pretty slow anyway.

I implemented the signal method mainly because of it's simplicity but I can concede that there are problems with it.

gpollo: Actually, I'll change `SIGUSR1` to a free `SIGRT`. On x86_64, there should be about 32 `SIGRT`s. The usual contract when something is using a `SIGRTMIN+n`, where `n` is between 0 and 32, is to check if it isn't used already, if it is used, then increase `n` and check again.

As for sending a signal being considered "heavy", we shouldn't need to send one when the program hasn't started (I'll change it). When the program is running, we still need to send one. I think it is unlikely that the user want to "patch" or "unpatch" all functions after the program has started. It may one to patch or unpatch a fews of interest.

It should be possible to only send a single signal when the user want to unpatch a single function, but it would require changing the interface between arch-specific and arch-independent dynamic code, as well as changing some logic elsewhere:
1. insert the trap for all functions that need patching/unpatching
2. send a single signal
3. insert/remove the jump for all functions

namhyung: Yes, I was thinking about the last method - sending a signal just once.  I know it needs more work but I think that's the way to go.

PR #1277, opened 2021-06-13
misc: Replace libpython2.7 to libpython3 in install-deps.sh

honggyukim:
The current misc/install-deps.sh installs libpython2.7-dev but
libpython3-dev is better to be used so this patch changes it.


namhyung (approved):
Looks ok.

Issue #1278, opened 2021-06-15
TUI report mode with option `-f`


qpakzk:
The current `tui` command provides the `-f` option only for the graph mode.

How about providing the `-f` option for the report mode as well when adding the `--report` option together?

```sh
$ uftrace tui --report -f +total-avg,total-min
```

honggyukim:
Yes, looks like a good idea. :)

PR #1279, opened 2021-06-16
tui: Enable TUI report mode to use option -f

qpakzk:
Hello.

Before this commit, the `tui` command provided the `-f` option only for the graph mode.
This commit allows the `tui` command to provide the `-f` option when adding the `--report` option together.

Fixed: #1278.

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -417,9 +417,16 @@ static void tui_setup(struct uftrace_data *handle, struct opts *opts)
 
 	tui_report.name_tree = RB_ROOT;
 
-	setup_field(&graph_output_fields, opts, setup_default_graph_field,
-		    graph_field_table, ARRAY_SIZE(graph_field_table));
-	setup_default_report_field(&report_output_fields, opts, report_field_table);
+	if (opts->report) {
+		setup_field(&report_output_fields, opts, setup_default_report_field,
+				report_field_table, ARRAY_SIZE(report_field_table));
+		setup_default_graph_field(&graph_output_fields, opts, graph_field_table);
+	}
+	else {
+		setup_field(&graph_output_fields, opts, setup_default_graph_field,
+				graph_field_table, ARRAY_SIZE(graph_field_table));
+		setup_default_report_field(&report_output_fields, opts, report_field_table);
+	}
 }
 
 static void tui_cleanup(void)
@@ -1402,14 +1409,39 @@ static void * win_parent_no(struct tui_window *win, void *node)
 	return NULL;
 }
 
+static void report_sort_key_init(void)
+{
+	int i, j = 0;
+
+	for (i = 0; i < NUM_REPORT_FIELD; i++) {
+		if (report_field_table[i]->used)
+			selected_report_sort_key[j++] = report_sort_key[i];
+	}
+}
+
+static int count_selected_report_sort_key(void)
+{
+	int count = 0;
+	int i;
+
+	for (i = 0; i < NUM_REPORT_FIELD; i++) {
+		if (report_field_table[i]->used)
+			count++;
+	}
+
+	return count;
+}
+
 /* per-window operations for report window */
 static struct tui_report * tui_report_init(struct opts *opts)
 {
 	struct tui_window *win = &tui_report.win;
 
-	report_setup_sort(OPT_SORT_KEYS);
-	report_sort_nodes(&tui_report.name_tree, &tui_report.sort_tree);
 	report_calc_avg(&tui_report.name_tree);
+	report_sort_key_init();
+	if (count_selected_report_sort_key())
+		report_setup_sort(selected_report_sort_key[curr_sort_key]);
+	report_sort_nodes(&tui_report.name_tree, &tui_report.sort_tree);
 
 	tui_window_init(win, &report_ops);
 
@@ -2649,25 +2681,7 @@ static inline void cancel_search(void)
 	tui_search = NULL;
 }
 
-static int count_selected_report_sort_key(void)
-{
-	int count = 0;
-	int i;
 
-	for (i = 0; i < NUM_REPORT_FIELD; i++) {
-		if (report_field_table[i]->used)
-			count++;
-	}
-
-	return count;
-}
-
-static inline void report_sort_key_init()
-{
-	selected_report_sort_key[0] = report_sort_key[REPORT_F_TOTAL_TIME];
-	selected_report_sort_key[1] = report_sort_key[REPORT_F_SELF_TIME];
-	selected_report_sort_key[2] = report_sort_key[REPORT_F_CALL];
-}
 
 static void tui_main_loop(struct opts *opts, struct uftrace_data *handle)
 {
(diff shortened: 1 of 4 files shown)


qpakzk:
@honggyukim Thanks for your feedback. I updated your comment.
Also, I wrote the more specific commit message.

doc/uftrace-tui.md
```diff
@@ -23,10 +23,16 @@ the initial data loading.
 TUI OPTIONS
 ===========
 -f *FIELD*, \--output-fields=*FIELD*
-:   Customize field in the output.  Possible values are: total, self and addr.
-    Multiple fields can be set by using comma.  Special field of 'none' can be
-    used (solely) to hide all fields.  Default is 'total'.
-    See `uftrace-graph`(1) for an explanation of fields.
+:   Customize fields in the output. This option basically indicates graph fields.
+    Possible values are total, self and addr. The default value is 'total'.
+    But if this option is combined with the --report option,
```
honggyukim, line 28: I think "used" is better than "combined" here and don't need to use "the".
```suggestion
    But if this option is used with --report option,
```

doc/uftrace-tui.md
```diff
@@ -23,10 +23,16 @@ the initial data loading.
 TUI OPTIONS
 ===========
 -f *FIELD*, \--output-fields=*FIELD*
-:   Customize field in the output.  Possible values are: total, self and addr.
-    Multiple fields can be set by using comma.  Special field of 'none' can be
-    used (solely) to hide all fields.  Default is 'total'.
-    See `uftrace-graph`(1) for an explanation of fields.
+:   Customize fields in the output. This option basically indicates graph fields.
```
honggyukim, line 26: By convention in uftrace docs, we give 2 white spaces after the period.
```suggestion
:   Customize fields in the output.  This option basically indicates graph fields.
```
Please apply the same for the rest of other sentences.

cmds/tui.c
```diff
@@ -317,45 +317,46 @@ static struct display_field *graph_field_table[] = {
...
-	.list    = LIST_HEAD_INIT(report_field_##_func.list)		\
+#define REPORT_FIELD_STRUCT(_id, _name, _alias, _func, _header, _length)	\
+static struct display_field report_field_##_func = {				\
+	.id      = _id,								\
+	.name    = #_name,							\
+	.alias   = #_alias,							\
+	.header  = _header,							\
+	.length  = _length,							\
+	.print   = print_report_##_func,					\
+	.list    = LIST_HEAD_INIT(report_field_##_func.list)			\
```
honggyukim, line 328: We don't need to add `_alias` here. Please revert this change.

cmds/tui.c
```diff
@@ -317,45 +317,46 @@ static struct display_field *graph_field_table[] = {
...
+
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME, total, total, total.sum, total, "TOTAL TIME");
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_AVG, total-avg, total-avg, total.avg, total_avg, "TOTAL AVG");
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MIN, total-min, total-min, total.min, total_min, "TOTAL MIN");
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MAX, total-max, total-max, total.max, total_max, "TOTAL MAX");
+REPORT_FIELD_TIME(REPORT_F_SELF_TIME, self, self, self.sum, self, "SELF TIME");
+REPORT_FIELD_TIME(REPORT_F_SELF_TIME_AVG, self-avg, self-avg, self.avg, self_avg, "SELF AVG");
+REPORT_FIELD_TIME(REPORT_F_SELF_TIME_MIN, self-min, self-min, self.min, self_min, "SELF MIN");
+REPORT_FIELD_TIME(REPORT_F_SELF_TIME_MAX, self-max, self-max, self.max, self_max, "SELF MAX");
+REPORT_FIELD_CALL(REPORT_F_CALL, call, call, call, call, "CALL");
```
honggyukim, line 359: By removing `_alias` change, we can remove all these changes in the diff.

honggyukim (commented):
One of the unit test fails with this change.
```
$ make unittest
  TEST     test_unit
Running 69 test cases
    ...
[025] field_setup_list              : FAIL
```
Please see the test at https://github.com/namhyung/uftrace/blob/master/utils/field.c#L224-L239.

qpakzk:
@honggyukim Thanks for your review.

```sh
$ uftrace tui -f +call
uftrace: Unknown field name 'call'
uftrace:   Possible fields are: total-time self-time address
```

The possible values and the error message values for `-f` option are different in TUI graph mode.
Users might make a mistake of using `total-time`, `self-time` or `address` when adding values for `-f` option.
How about adding `.alias` and showing `.alias` instead?

```diff
diff --git a/utils/field.c b/utils/field.c
index f19da6ba..94666969 100644
--- a/utils/field.c
+++ b/utils/field.c
@@ -165,7 +165,7 @@ void setup_field(struct list_head *output_fields, struct opts *opts,
                        pr_out("uftrace: Unknown field name '%s'\n", p);
                        pr_out("uftrace:   Possible fields are:");
                        for (i = 0; i < field_table_size; i++)
-                               pr_out(" %s", field_table[i]->name);
+                               pr_out(" %s", field_table[i]->alias);
                        pr_out("\n");
                        exit(1);
                }
```

qpakzk:
@honggyukim Thanks for your review. I misunderstood something.
I applied your advice to my commit.
Also, I updated tests.

qpakzk:
I found a problem when printing an error message.

When my shell is on the bottom, it works fine.
Screen Shot 2021-06-19 at 19 00 46

However, when my shell isn't on the bottom, my shell overlaps the error message.
Screen Shot 2021-06-19 at 19 01 04

honggyukim (commented):
Sorry for the late response and thanks for the update again. Could you split this patch into two patches? Otherwise looks good to me.
1. field: Keep the predefined field order when using -f option
2. tui: Allow tui report mode to use option -f

The field keeps ordering not only in report command but also the other commands such as graph and replay commands.

qpakzk:
@honggyukim Thanks for your feedback.
I updated your requirement.

honggyukim (approved):
LGTM. Thanks for your work!

honggyukim:
@qpakzk We can also use `-s`/`--sort` option when `--report` option is used with this change. If you want to work more about TUI, then this can be a possible next work item. :)

qpakzk:
@honggyukim I'm gonna try it!

utils/field.c
```diff
@@ -129,11 +130,17 @@ void setup_field(struct list_head *output_fields, struct opts *opts,
 	if (!strcmp(opts->fields, "none"))
 		return;
 
+	field_flags = calloc(field_table_size, sizeof(bool));
```
namhyung, line 133: Please use `xcalloc` to check error.  Also I prefer using `sizeof(*field_flags)` than the type name (bool) because it's easier to deal with future (type) changes.

namhyung (commented):
I have a nitpick, but otherwise looks ok.

namhyung (approved):
LGTM

PR #1282, opened 2021-07-10
script: Fix a crash for struct type handling in script

honggyukim:
We didn't handle struct type for scripting, so it crashes as follows.
```
  $ g++ -pg -g -O2 tests/s-dwarf7.cpp
  $ uftrace record -a --no-libcall a.out
  $ uftrace script --no-pager -S scripts/dump.py
        ...
  4981.341259000   4788: [entry] main(7fd7928006cd) depth: 0
  WARN: invalid argument format: 10
  WARN: invalid argument format: 10
  4981.341264100   4788: [entry] compare_iters(7fd79280088a) depth: 1
```
This patch fixes this problem for both python and luajit, but print the
struct type name in string type as follows.
```
  $ uftrace script -S scripts/dump.py
        ...
  4981.341259000   4788: [entry] main(7fd7928006cd) depth: 0
  4981.341264100   4788: [entry] compare_iters(7fd79280088a) depth: 1
    args[0] <type 'str'>: __normal_iterator
    args[1] <type 'str'>: __normal_iterator
  4981.342013200   4788: [exit ] compare_iters(7fd79280088a) depth: 1
    retval  <type 'long'>: 0
  4981.342024300   4788: [exit ] main(7fd7928006cd) depth: 0
    retval  <type 'long'>: 0
```

### DIFF
--- a/tests/s-struct.c
+++ b/tests/s-struct.c
@@ -1,0 +1,35 @@
+// The sizeof(Option) is 11 in 64-bit, 7 in 32-bit.
+struct __attribute__((packed)) Option {
+        char m;     // 1
+        long n;     // 8 or 4
+        short k;    // 2
+};
+
+// empty struct
+struct StringRef {};
+
+// global variables to suppress compiler optimization
+struct Option g_opt;
+struct StringRef g_sr;
+unsigned g_index;
+int g_value1;
+int g_value2;
+
+__attribute__((noinline))
+void foo(const struct Option Opt, struct StringRef S, unsigned Index,
+         const int Value1, const int Value2) {
+        g_opt = Opt;
+        g_sr = S;
+        g_index = Index;
+        g_value1 = Value1;
+        g_value2 = Value2;
+}
+
+int main(void)
+{
+        struct Option Opt = { 11, 22, 33 };
+        struct StringRef S;
+
+        foo(Opt, S, 44, 55, 66);
+        return 0;
+}
--- a/tests/t259_arg_struct_script.py
+++ b/tests/t259_arg_struct_script.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+
+from runtest import TestBase
+import subprocess as sp
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'struct', """
+uftrace_begin(ctx)
+  record  : False
+  version : v0.10-17-g8d1b ( x86_64 dwarf python luajit tui perf sched dynamic )
+  cmds    : t-struct
+
+11332966.196463691  50529: [entry] main(4006ef) depth: 0
+11332966.196464540  50529: [entry] foo(40068f) depth: 1
+  args[0] <class 'str'>: struct: Option{}
+  args[1] <class 'str'>: struct: StringRef{}
+  args[2] <class 'int'>: 44
+  args[3] <class 'int'>: 55
+  args[4] <class 'int'>: 66
+11332966.196670289  50529: [exit ] foo(40068f) depth: 1
+11332966.196671664  50529: [exit ] main(4006ef) depth: 0
+  retval  <class 'int'>: 0
+
+uftrace_end()
+""", cflags='-g', sort='dump')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not 'dwarf' in self.feature or not 'python' in self.feature:
+            return TestBase.TEST_SKIP
+        # cygprof doesn't support arguments now
+        if cflags.find('-finstrument-functions') >= 0:
+            return TestBase.TEST_SKIP
+        return TestBase.build(self, name, cflags, ldflags)
+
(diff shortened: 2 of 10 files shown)


utils/script-luajit.c
```diff
@@ -197,6 +197,16 @@ static void setup_argument_context(bool is_retval, struct script_context *sc_ctx
 			data += 4;
 			break;
 
+		case ARG_FMT_STRUCT:
+			if (spec->type_name) {
```
namhyung, line 201: What if it doesn't have a type name?

utils/script-luajit.c
```diff
@@ -197,6 +197,16 @@ static void setup_argument_context(bool is_retval, struct script_context *sc_ctx
...
+				slen = strlen(spec->type_name);
+				dllua_pushinteger(L, 1 + count++);
```
namhyung, line 203: `++count` instead?

namhyung (commented):
I think we should give a hint to the scripts as it's just not a string.  Maybe we can add a prefix like "struct: TYPENAME" or add surrounding braces like "{ TYPENAME }".

honggyukim:
@namhyung Thanks for the review. I changed the output as follows.
```
$ g++ -pg -g -o t-struct tests/s-struct.c

$ uftrace script -S scripts/dump.py --no-libcall --no-event --record t-struct
uftrace_begin(ctx)
  record  : False
  version : v0.10-17-g8d1b ( x86_64 dwarf python luajit tui perf sched dynamic )
  cmds    : t-struct

11336060.26574470  26208: [entry] main(4006ef) depth: 0
11336060.26575270  26208: [entry] foo(40068f) depth: 1
  args[0] <class 'str'>: struct: Option{}
  args[1] <class 'str'>: struct: StringRef{}
  args[2] <class 'int'>: 44
  args[3] <class 'int'>: 55
  args[4] <class 'int'>: 66
11336060.26780955  26208: [exit ] foo(40068f) depth: 1
11336060.26782239  26208: [exit ] main(4006ef) depth: 0
  retval  <class 'int'>: 0

uftrace_end()
```
I also added 3 more tests for struct argument handling.

tests/s-struct.c
```diff
@@ -0,0 +1,34 @@
...
+void foo(const struct Option Opt, struct StringRef S, unsigned Index,
+         const int Value1, const int Value2) {
+        g_opt = Opt;
+        g_sr = S;
+        g_index = Index;
+        g_value1 = Value1;
+        g_value2 = Value2;
+}
+
+int main()
```
namhyung, line 28: C requires void in the signature.

tests/s-struct.c
```diff
@@ -0,0 +1,34 @@
...
+{
+        struct Option Opt = { 11, 22, 33 };
+        struct StringRef S;
+
+        foo(Opt, S, 44, 55, 66);
+}
```
namhyung, line 34: Please add 'return 0'

tests/t259_arg_struct_script.py
```diff
@@ -0,0 +1,50 @@
...
+11332966.196670289  50529: [exit ] foo(40068f) depth: 1
+11332966.196671664  50529: [exit ] main(4006ef) depth: 0
+  retval  <class 'int'>: 0
+
+uftrace_end()
+""", cflags='-g', sort='dump')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not 'dwarf' in self.feature:
+            return TestBase.TEST_SKIP
```
namhyung, line 30: I think it should also skip if 'python' is not in the feature.

tests/t260_arg_struct_luajit.py
```diff
@@ -0,0 +1,45 @@
...
+514662126427.20138  29459: [exit ] foo(40061f) depth: 1
+514662126427.21036  29459: [exit ] main(40067f) depth: 0
+  retval  number: 0
+
+uftrace_end()
+""", cflags='-g', sort='dump')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not 'dwarf' in self.feature:
+            return TestBase.TEST_SKIP
```
namhyung, line 30: Ditto for 'luajit'.

honggyukim: done

tests/t259_arg_struct_script.py
honggyukim: done

tests/s-struct.c
honggyukim: done

tests/s-struct.c
honggyukim: I changed this example from C++ to C, so didn't add it because C++ doesn't require to have explicit `return 0` for main function. But I changed this to C again, so you're right. I just added it :)

honggyukim (commented):
Thanks for the review. I just updated again.

namhyung (approved):
LGTM

Issue #1286, opened 2021-07-22
segfault in 150 recv_event


honggyukim:
```
$ ./runtest.py -vdp -O0 150
Start 1 tests with 1 worker
Test case                 pg
------------------------: O0
build command: gcc -o t-sdt -fno-inline -fno-builtin -fno-ipa-cp -fno-omit-frame-pointer -D_FORTIFY_SOURCE=0  -pg -O0  s-sdt.c
prerun command: /home/honggyu/work/uftrace/uftrace recv --no-pager --no-event -L/home/honggyu/work/uftrace -d xxx --port 41144
prerun command: /home/honggyu/work/uftrace/uftrace record --no-pager --no-event -L/home/honggyu/work/uftrace --host localhost --port 41144 -E uftrace:event t-sdt
WARN: child terminated by signal: 11: Segmentation fault
uftrace: /home/honggyu/work/uftrace/cmds/recv.c:207:send_trace_metadata
 ERROR: open uftrace.data/task.txt failed: No such file or directory
uftrace: /home/honggyu/work/uftrace/cmds/recv.c:562:handle_client_sock
 ERROR: message recv failed: No such file or directory
test command: /home/honggyu/work/uftrace/uftrace replay --no-pager --no-event -L/home/honggyu/work/uftrace -d xxx/uftrace.data
t150_recv_event: -pg -O0 returns 255

150 recv_event          : NZ
```
Here is the backtrace.
```
Thread 2.1 "t-sdt" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7fc4800 (LWP 54867)]
do_lookup_x (undef_name=undef_name@entry=0x7ffff7b8cb34 "gelf_getphdr", new_hash=new_hash@entry=1599266512, old_hash=old_hash@entry=0x7fffffffcc20, ref=0x7ffff7b8be98, result=resul
402     in dl-lookup.c
#0  do_lookup_x (undef_name=undef_name@entry=0x7ffff7b8cb34 "gelf_getphdr", new_hash=new_hash@entry=1599266512, old_hash=old_hash@entry=0x7fffffffcc20, ref=0x7ffff7b8be98, result=result@entry=0x7fffffffcc30, scope=<optimized out>, i=0, version=0x7ffff7fc8950, flags=1, skip=0x0, type_class=1, undef_map=0x7ffff7ff6000) at dl-lookup.c:402
[...]

PR #1287, opened 2021-07-31
event: Fix a crash by adding PROT_READ along with PROT_EXEC

honggyukim:
Currently, uftrace gets segfault when patching SDT events into the
target binary.
```
  $ gcc -pg -o t-sdt s-sdt.c

  $ uftrace -E uftrace:* --match glob t-sdt
  WARN: cannot open record data: /tmp/uftrace-live-CCoVbF: No data available
```
The reason of the segfault is because the access permission of the
SDT memory region becomes PROT_EXEC only without having PROT_READ.

It might be okay in some environments but some are not so this patch
explicitly adds PROT_READ along with PROT_EXEC.

Before:
```
  147 event_sdt           : SG SG SG SG SG SG SG SG SG SG
  150 recv_event          : NZ NZ NZ NZ NZ NZ NZ NZ NZ NZ
```
After:
```
  147 event_sdt           : OK OK OK OK OK OK OK OK OK OK
  150 recv_event          : OK OK OK OK OK OK OK OK OK OK
```
Fixed: #1286

### DIFF
--- a/arch/x86_64/mcount-event.c
+++ b/arch/x86_64/mcount-event.c
@@ -50,10 +50,11 @@ int mcount_arch_enable_event(struct mcount_event_info *mei)
 		return -1;
 	}
 
-	/* replace NOP to an invalid OP so that it can catch SIGILL */
+	/* replace NOP to an invalid OP so that it can catch SIGILL,
+	   then it will fall into sdt_handler() above. */
 	memset((void *)mei->addr, INVALID_OPCODE, 1);
 
-	if (mprotect(PAGE_ADDR(mei->addr), PAGE_SIZE, PROT_EXEC))
+	if (mprotect(PAGE_ADDR(mei->addr), PAGE_SIZE, PROT_READ | PROT_EXEC))
 		pr_err("cannot setup event due to protection");
 
 	return 0;
--- a/utils/fstack.c
+++ b/utils/fstack.c
@@ -1045,7 +1045,7 @@ bool fstack_check_opts(struct uftrace_task_reader *task, struct opts *opts)
 			return false;
 	}
 
-	if (opts->no_event && rec->type == UFTRACE_EVENT)
+	if (opts->no_event && !opts->event && rec->type == UFTRACE_EVENT)
 		return false;
 
 	if (opts->no_sched && is_sched_event(rec->addr))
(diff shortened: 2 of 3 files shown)


utils/fstack.c
```diff
@@ -1045,7 +1045,7 @@ bool fstack_check_opts(struct uftrace_task_reader *task, struct opts *opts)
 			return false;
 	}
 
-	if (opts->no_event && rec->type == UFTRACE_EVENT)
+	if (opts->no_event && !opts->event && rec->type == UFTRACE_EVENT)
```
namhyung, line 1048: This looks like a separate fix.  Can you please split into another commit?

honggyukim: OK. I just split it into a new commit.

arch/x86_64/mcount-event.c
```diff
@@ -50,10 +50,11 @@ int mcount_arch_enable_event(struct mcount_event_info *mei)
 		return -1;
 	}
 
-	/* replace NOP to an invalid OP so that it can catch SIGILL */
+	/* replace NOP to an invalid OP so that it can catch SIGILL,
+	   then it will fall into sdt_handler() above. */
 	memset((void *)mei->addr, INVALID_OPCODE, 1);
 
-	if (mprotect(PAGE_ADDR(mei->addr), PAGE_SIZE, PROT_EXEC))
+	if (mprotect(PAGE_ADDR(mei->addr), PAGE_SIZE, PROT_READ | PROT_EXEC))
```
namhyung, line 57: Could you please take a look at the other callsites of mproctect() and fix them all?

honggyukim: I also fixed it in i386 as well.

honggyukim:
I found that we can enable `-mfentry -mnop-mcount` build in i386 as well so made it work together.

namhyung:
Oh, good. But I don't want to put it here together.  It's a separate change and would belong to another PR.

honggyukim:
I put this together because the i386 patch requires `uftrace: Add PROT_READ permission along with PROT_EXEC` but any of them needs to be merged first. Otherwise, I have to put the all together. Same for the test fix PR as well. As the PR grows, some of the patches are needed together so any of the PR needs to be merged first.

honggyukim:
I removed `i386` changes in #1296 separately.

namhyung (approved):
LGTM

honggyukim:
I will rebase other PRs.

Issue #1298, opened 2021-08-10
detect libexecinfo on musl platforms


cz172638:
on musl you are getting following errors:
undefined reference to `backtrace'
undefined reference to `backtrace_symbols'

```./configure --prefix=/usr --ldflags="-lexecinfo"``` is workaround.

**Comment by cz172638** [comment]
on alpine 3.14 i'm getting segmentation fault with ```make unittest``` and with ```make runtest``` 2300 build failures:
```
runtime test stats
====================
total  2600  Tests executed (success: 1.85%)
  OK:    48  Test succeeded
  OK:     0  Test succeeded (with some fixup)
  NG:     2  Different test result
  NZ:    10  Non-zero return value
  SG:     0  Abnormal exit by signal
  TM:    50  Test ran too long
  BI:  2300  Build failed
  LA:     0  Unsupported Language
  SK:   190  Skipped
```

**Comment by honggyukim** [comment]
Hi @cz172638, musl seems to be different from glibc. Can you test this change?
```diff
diff --git a/utils/utils.c b/utils/utils.c
index 31726794..6e502eb3 100644
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -888,6 +888,10 @@ char *absolute_dirname(const char *path, char *resolved_path)
        return resolved_path;
 }

+#if !__GLIBC__
+/* we support stacktrace only in glibc. */
+void stacktrace(void) {}
+#else
 void stacktrace(void)
 {
 #if !DEBUG_MODE
@@ -948,6 +952,7 @@ void stacktrace(void)
 #endif
        pr_out("\n");
 }
+#endif

 #ifdef UNIT_TEST
 TEST_CASE(utils_parse_cmdline)
```
I wanted to detect musl but it doesn't have a macro as written in https://wiki.musl-libc.org/faq.html so we can enable `stacktrace` only in glibc.

honggyukim:
> on alpine 3.14 i'm getting segmentation fault with make unittest and with make runtest 2300 build failures:

I don't have Alpine Linux environment so can't test it. Do you know if the older version of uftrace worked fine? For example, `v0.10`, `0.9.4`, etc.

namhyung:
I think we should check libunwind in the check-deps as well.

honggyukim:
Do you mean to use libunwind by default and just disable stacktrace if not exists? That might be also good.

cz172638:
[...]

PR #1301, opened 2021-08-12
utils: Use only libunwind without using execinfo

honggyukim:
The current stacktrace in release mode uses backtrace() in execinfo.h,
but it can't show static function names and it doesn't work reliable in
libmcount*.so libraries during record time.

Moreover, execinfo.h is not well supported in some environments such as
alpine Linux.

So we better use only libunwind when it's available. If not, we just do
not print anything in stacktrace().

This PR also contains build check routine for libunwind as follows.
```
  $ ./configure
                ...
  ...      libunwind: [ on  ] - libunwind support
```
Fixed: #1298

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,8 @@ TEST_LDFLAGS       = $(COMMON_LDFLAGS) -L$(objdir)/libtraceevent -ltraceevent
 
 ifeq ($(DEBUG), 1)
   COMMON_CFLAGS += -O0 -g -DDEBUG_MODE=1
-  COMMON_LDFLAGS += -lunwind
 else
-  COMMON_CFLAGS += -O2 -g -DDEBUG_MODE=0 -rdynamic
+  COMMON_CFLAGS += -O2 -g -DDEBUG_MODE=0
 endif
 
 ifeq ($(TRACE), 1)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -89,9 +89,8 @@ TEST_LDFLAGS       = $(COMMON_LDFLAGS) -L$(objdir)/libtraceevent -ltraceevent
 
 ifeq ($(DEBUG), 1)
   COMMON_CFLAGS += -O0 -g -DDEBUG_MODE=1
-  COMMON_LDFLAGS += -lunwind
```
namhyung, line 92: You can add it conditionally.

honggyukim:
@namhyung Could you please have a look? We might also be able to use `libunwind` for generating trace records without actual hooking maybe in the future.

namhyung (commented):
Please don't add libunwind for non-debug build.

honggyukim:
I update as you requested. Thanks for the review.

check-deps/Makefile.check
```diff
@@ -64,6 +64,13 @@ ifneq ($(wildcard $(srcdir)/check-deps/have_libdw),)
   COMMON_LDFLAGS += $(shell pkg-config --libs libdw 2> /dev/null || echo "-ldw")
 endif
 
+ifneq ($(wildcard $(srcdir)/check-deps/have_libunwind),)
+ifeq ($(DEBUG), 1)
+  COMMON_CFLAGS  += -DHAVE_LIBUNWIND
+  COMMON_LDFLAGS += -lunwind
```
namhyung, line 70: This should be same as {C,LD}FLAGS_have_libunwind in the check-deps/Makefile.

honggyukim: Thanks. I've just updated it.

namhyung (approved):
LGTM

PR #1305, opened 2021-08-14
debug: Fix to properly set end character to TERM_COLOR_RESET

honggyukim:
'sc' means start character and is set to TERM_COLOR_NORMAL initially,
but it has to be set to TERM_COLOR_RESET when color mode is on.

This patch fixes missing reset 'ec' in print_diff_percent and
print_diff_count.

### DIFF
--- a/utils/debug.c
+++ b/utils/debug.c
@@ -323,18 +323,28 @@ void print_time_unit(uint64_t delta_nsec)
 	__print_time_unit(delta_nsec, false);
 }
 
+static const char *get_color(const char *color)
+{
+	if (out_color == COLOR_ON)
+		return color;
+	else
+		return TERM_COLOR_NORMAL;
+}
+
 void print_diff_percent(uint64_t base_nsec, uint64_t pair_nsec)
 {
 	double percent = 999.99;
 	const char *sc = TERM_COLOR_NORMAL;
-	const char *ec = TERM_COLOR_NORMAL;
+	const char *ec = get_color(TERM_COLOR_RESET);
 
 	if (base_nsec == 0) {
-		pr_out("%s%7s%s ", TERM_COLOR_RED, "N/A", ec);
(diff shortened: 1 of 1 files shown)


honggyukim:
Here is the image screenshot between before and after applying this patch.
Before:
image
After:
image

utils/debug.c
```diff
@@ -329,6 +329,9 @@ void print_diff_percent(uint64_t base_nsec, uint64_t pair_nsec)
 	const char *sc = TERM_COLOR_NORMAL;
 	const char *ec = TERM_COLOR_NORMAL;
 
+	if (out_color == COLOR_ON)
+		ec = TERM_COLOR_RESET;
+
 	if (base_nsec == 0) {
 		pr_out("%s%7s%s ", TERM_COLOR_RED, "N/A", ec);
```
namhyung, line 336: I think we should check COLOR_ON before using the RED color.  And below BLUE too.

utils/debug.c
```diff
@@ -376,15 +379,13 @@ void print_diff_count(uint64_t base, uint64_t pair)
...
 	if (out_color == COLOR_ON) {
 		sc = diff_colors[sign_idx];
 		ec = TERM_COLOR_RESET;
 	}
 
-	pr_out("%s%+9"PRId64"%s", sc, diff, ec);
+	if (diff != 0)
+		pr_out("%s%+9"PRId64"%s", sc, diff, ec);
+	else
+		pr_out("%9s", "+0");
```
namhyung, line 390: It seems this function doesn't need the change?

honggyukim: I just simplified removing extra return statement.

utils/debug.c
honggyukim: Right. I've fixed it.

namhyung (approved):
LGTM

PR #1306, opened 2021-08-15
uftrace: Make option struct 'opts' as a global variable

honggyukim:
We currently pass 'struct opts *opts` many places around but sometimes
it's annoying or tricky to add an extra arguments for option handling.

Rather than doing this, we better make 'struct opts' a global variable
and use everywhere.

Test results looks same as before and the failed tests are known issues.
```
  $ make test
        ...
  unit test stats
  ====================
   74 ran successfully
    0 failed
    0 skipped
    0 signal caught
    0 unknown result
        ...
  runtime test stats
  ====================
  total  2600  Tests executed (success: 80.15%)
    OK:  2019  Test succeeded
    OK:    65  Test succeeded (with some fixup)
    NG:    36  Different test result
    SG:     0  Abnormal exit by signal
    TM:     0  Test ran too long
    BI:    10  Build failed
    SK:   470  Skipped
```

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -20,7 +20,7 @@
 struct uftrace_dump_ops {
 	/* this is called at the beginning */
 	void (*header)(struct uftrace_dump_ops *ops,
-		       struct uftrace_data *handle, struct opts *opts);
+		       struct uftrace_data *handle);
 	/* this is called when a task starts */
 	void (*task_start)(struct uftrace_dump_ops *ops,
 			   struct uftrace_task_reader *task);
@@ -59,7 +59,7 @@ struct uftrace_dump_ops {
 			   struct uftrace_record *frs);
 	/* this is called at the end */
 	void (*footer)(struct uftrace_dump_ops *ops,
-		       struct uftrace_data *handle, struct opts *opts);
+		       struct uftrace_data *handle);
 };
 
 struct uftrace_raw_dump {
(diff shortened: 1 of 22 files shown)


honggyukim:
The failed tests are as follows.
```
117 time_range          : OK OK OK OK OK OK OK NG OK OK
125 report_range        : OK OK OK OK OK OK OK OK NG OK
135 trigger_time2       : OK OK OK OK OK OK NG OK OK OK
151 recv_runcmd         : NG NG NG NG NG NG NG NG NG NG
162 pltbind_now_pie     : BI BI BI BI BI BI BI BI BI BI
202 arg_dwarf2          : NG NG NG NG NG SK SK SK SK SK
203 arg_dwarf3          : OK OK NG NG NG SK SK SK SK SK
204 arg_dwarf4          : OK OK NG NG OK SK SK SK SK SK
222 external_data       : OK OK OK OK OK OK OK NG OK OK
248 dynamic_dlopen      : OK OK OK OK NG OK OK OK OK NG
251 exception4          : NG NG NG NG NG NG NG NG NG NG
```
But I confirmed those failures are not because of this option change.

namhyung:
One thing I worry is that some codes are shared between uftrace binary and libmcount.  Then the opts should exist and set properly for libmcount too.  And I prefer having explicit arguments to control behaviors rather than using global options internally.

honggyukim:
> Then the opts should exist and set properly for libmcount too.

The `struct opts opts` global variable is defined in `uftrace.c` so won't be accessible in libmcount libraries.

> I prefer having explicit arguments to control behaviors rather than using global options internally.

I understand that, but the current `opts` argument passing is not well structured and it makes function signatures unnecessarily complicated in many areas.

namhyung:
It'd be confusing if some code in utils/ start to use the global variable as it's shared for libmcount too.

honggyukim:
OK. I can drop this change then.

Issue #1308, opened 2021-08-16
Provide a way to print output in HTML format


honggyukim:
It'd be useful if we can print uftrace output in HTML format as follows.

![image](https://user-images.githubusercontent.com/19642328/129522016-3681cabe-6538-4667-b1ff-72d1215f543a.png)

It can be useful especially when sharing uftrace result to someone who doesn't have uftrace installed. We can copy & paste the text but it might be too verbose to paste the entire result and we also lose all the colored output.

Same applies to uftrace replay, graph output as well.

The WIP is uploaded at [check/html-option](https://github.com/honggyukim/uftrace/commit/7010d2f9370dc9d64717e9d87467a95d42a9ede6). The current implementation is just for PoC so please don't blame too much on the repeated patterns of code :)

In the long term, it'd be more useful if we can make the page more interactive just like we do in `tui` command.

honggyukim:
We can upload some example output in https://uftrace.github.io later.

honggyukim:
I uploaded the initial PR at #1311. I made `--color=html` mode instead of adding an extra `--html` option. Please have a look.

PR #1311, opened 2021-08-18
uftrace: Introduce --format=html option to print html output

honggyukim:
This patch adds --format=html option to print the output in html format
that can be opened in a web browser.  The html page output looks exactly
same as the output in console.

This mode can be used in replay, report, graph commands.

Closes: #1308

### DIFF
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -703,7 +703,10 @@ void get_argspec_string(struct uftrace_task_reader *task,
 
 			if (sym) {
 				print_args(&args, &len, "%s", color_symbol);
-				print_args(&args, &len, "&%s", sym->name);
+				if (format_mode == FORMAT_HTML)
+					print_args(&args, &len, "&amp;%s", sym->name);
+				else
+					print_args(&args, &len, "&%s", sym->name);
 				print_args(&args, &len, "%s", color_reset);
 			}
 			else if (val.p)
@@ -1218,6 +1221,9 @@ int command_replay(int argc, char *argv[], struct opts *opts)
 	setup_field(&output_fields, opts, &setup_default_field,
 		    field_table, ARRAY_SIZE(field_table));
 
+	if (format_mode == FORMAT_HTML)
+		pr_out(HTML_HEADER);
+
 	if (!opts->flat && peek_rstack(&handle, &task) == 0)
 		print_header(&output_fields, "#", "FUNCTION", 1, false);
 	if (!list_empty(&output_fields)) {
@@ -1254,6 +1260,9 @@ int command_replay(int argc, char *argv[], struct opts *opts)
 	}
 
 	print_remaining_stack(opts, &handle);
+
+	if (format_mode == FORMAT_HTML)
+		pr_out(HTML_FOOTER);
 
 	close_data_file(opts, &handle);
 
--- a/uftrace.c
+++ b/uftrace.c
@@ -77,6 +77,7 @@ enum options {
 	OPT_graphviz,
 	OPT_sample_time,
 	OPT_diff,
+	OPT_format,
 	OPT_sort_column,
 	OPT_tid_filter,
 	OPT_num_thread,
@@ -143,6 +144,7 @@ __used static const char uftrace_help[] =
 "      --flame-graph          Dump recorded data in FlameGraph format\n"
 "      --flat                 Use flat output format\n"
 "      --force                Trace even if executable is not instrumented\n"
+"      --format=FORMAT        Use FORMAT for output: normal, html (default: normal)\n"
 "  -f, --output-fields=FIELD  Show FIELDs in the replay or graph output\n"
 "  -F, --filter=FUNC          Only trace those FUNCs\n"
 "      --graphviz             Dump recorded data in DOT format\n"
@@ -268,6 +270,7 @@ static const struct option uftrace_options[] = {
 	NO_ARG(flame-graph, OPT_flame_graph),
 	REQ_ARG(sample-time, OPT_sample_time),
 	REQ_ARG(diff, OPT_diff),
+	REQ_ARG(format, OPT_format),
 	REQ_ARG(sort-column, OPT_sort_column),
 	REQ_ARG(num-thread, OPT_num_thread),
 	NO_ARG(no-comment, OPT_no_comment),
@@ -826,6 +829,20 @@ static int parse_option(struct opts *opts, int key, char *arg)
 
 	case OPT_diff_policy:
 		opts->diff_policy = arg;
+		break;
+
+	case OPT_format:
+		if (!strcmp(arg, "normal"))
+			format_mode = FORMAT_NORMAL;
+		else if (!strcmp(arg, "html")) {
+			format_mode = FORMAT_HTML;
+			if (opts->color == COLOR_AUTO)
+				opts->color = COLOR_ON;
+		}
+		else {
+			pr_use("invalid format argument: %s\n", arg);
+			format_mode = FORMAT_NORMAL;
+		}
 		break;
 
 	case OPT_sort_column:
--- a/utils/debug.c
+++ b/utils/debug.c
@@ -25,11 +25,23 @@
 #define TERM_COLOR_CYAN		"\033[36m"
 #define TERM_COLOR_GRAY		"\033[90m"    /* bright black */
 
+#define HTML_COLOR_NORMAL	"<span>"
+#define HTML_COLOR_RESET	"</span>"
+#define HTML_COLOR_BOLD		"<span style='font-weight:bold'>"
+#define HTML_COLOR_RED		"<span style='color:red'>"    /* bright red */
+#define HTML_COLOR_GREEN	"<span style='color:green'>"
+#define HTML_COLOR_YELLOW	"<span style='color:yellow'>"
+#define HTML_COLOR_BLUE		"<span style='color:blue'>"    /* bright blue */
+#define HTML_COLOR_MAGENTA	"<span style='color:magenta'>"
+#define HTML_COLOR_CYAN		"<span style='color:cyan'>"
+#define HTML_COLOR_GRAY		"<span style='color:gray'>"    /* bright black */
+
 int debug;
 FILE *logfp;
 FILE *outfp;
 enum color_setting log_color;
 enum color_setting out_color;
+enum format_mode format_mode;
 int dbg_domain[DBG_DOMAIN_MAX];
 
 /* colored output for argspec display */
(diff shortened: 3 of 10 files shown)


honggyukim:
It looks empty lambda function shows incorrect color for this.
```
std::condition_variable::wait(0x7fff34204e20, 0x7fff34204e50, {...})
```
It's printed as follows.
```html
std::condition_variable::wait(0x7fff34204e20, 0x7fff34204e50, <span style='color:cyan'><lambda</span>{...})
```
The `<lambda` part breaks the html parsing.

honggyukim:
This can be reproduced as follows.
```
$ g++ -pg -g -o t-dwarf2 tests/s-dwarf2.cpp

$ uftrace --color=html -a -F std::sort -D 1 t-dwarf2
<html>
<head></head>
<body style='background-color:black;color:white;'>
<pre>
   7.847 us</span> [   887] | std::sort(0x7ffd19b87b50, 0x7ffd19b87b64, <span style='color:cyan'>&ampmyless</span>);
   3.267 us</span> [   887] | std::sort(0x7ffd19b87b50, 0x7ffd19b87b64, <span style='color:cyan'>less</span>{...});
 118.365 us</span> [   887] | std::sort(0x7ffd19b87b50, 0x7ffd19b87b64, <span style='color:cyan'><lambda</span>{...});
</pre>
</body>
</html>
```

honggyukim:
We may need this to avoid this problem.
```diff
diff --git a/cmds/replay.c b/cmds/replay.c
index fadf53ac..a14c7149 100644
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -751,9 +751,14 @@ void get_argspec_string(struct uftrace_task_reader *task,
                        free(estr);
                }
                else if (spec->fmt == ARG_FMT_STRUCT) {
-                       if (spec->type_name)
-                               print_args(&args, &len, "%s%s%s", color_struct,
-                                               spec->type_name, color_reset);
+                       if (spec->type_name) {
+                               /* gcc puts "<lambda" to annoymous lambda */
+                               if (strcmp(spec->type_name, "<lambda")) {
+                                       print_args(&args, &len, "%s%s%s",
+                                                  color_struct, spec->type_name,
+                                                  color_reset);
+                               }
+                       }
                        if (spec->size)
                                print_args(&args, &len, "{...}");
                        else
```

namhyung (commented):
It's not just for color, it also needs HTML header and footer.  Maybe need a separate option?

namhyung:
> The <lambda part breaks the html parsing.

It's not just lambda, I think we should escape the "<" for HTML mode.

honggyukim:
> It's not just for color, it also needs HTML header and footer. Maybe need a separate option?

I thought about having `--html` option, but it gets complicated because we can't use `opts` in `debug.c` file, so we need a global variable any way something like `html_mode`. The initial draft is at https://github.com/honggyukim/uftrace/commit/22b8aea89c7646f556b460a93bbdc065617ad274.

So I wanted to make `opts` a global variable, but dropped this change now. I think the implementation is tightly coupled with color handling so chose to add `--color=html` in addition to having html header and footer.

In the long term, I'm thinking of generating another html page that can be similar to the current `tui` that allows user interaction. I would like to leave `--html` option for this purpose.

honggyukim:
> It's not just lambda, I think we should escape the "<" for HTML mode.

This makes the current test fails since some specific gcc version and doesn't look consistent with clang so I think we better remove this part from the output entirely.
```diff
t202_arg_dwarf2: diff result of -pg -O0
--- expect      2021-08-19 07:47:41.109006691 +0100
+++ result      2021-08-19 07:47:41.109006691 +0100
@@ -4,3 +4,3 @@
    std::sort(0xADDR, 0xADDR, less{...});
-   std::sort(0xADDR, 0xADDR, {...});
+   std::sort(0xADDR, 0xADDR, <lambda{...});
  } = 0; /* main */

202 arg_dwarf2          : NG
```
This test is okay with clang.
```
$ ./runtest.py -c clang 202
Start 1 tests with 1 worker
202 arg_dwarf2          : OK OK OK OK OK SK SK SK SK SK
```

honggyukim:
> It's not just lambda, I think we should escape the "<" for HTML mode.

This is needed anyway for string argument output as well to complete this feature.

honggyukim:
@namhyung I've added `--format=html` option for this feature. If you're okay with this approach, then I will update man pages in the following update.

utils/debug.c
```diff
@@ -376,14 +427,17 @@ void print_diff_time_unit(uint64_t base_nsec, uint64_t pair_nsec)
...
 	};
 	int sign_idx = (pair < base);
 	int64_t diff = pair - base;
-	const char *sc = get_color(diff_colors[sign_idx]);
-	const char *ec = get_color(TERM_COLOR_RESET);
+	const char *sc;
+	const char *ec;
+
+	sc = get_color(diff_colors[sign_idx]);
+	ec = get_color(COLOR_CODE_RESET);
```
namhyung, line 440: Please minimize the change.

utils/debug.c
```diff
@@ -356,12 +408,11 @@ void print_diff_percent(uint64_t base_nsec, uint64_t pair_nsec)
...
-		sc = percent > 30 ? TERM_COLOR_RED :
-			percent > 3 ? TERM_COLOR_MAGENTA :
-			percent < -30 ? TERM_COLOR_BLUE :
-			percent < -3 ? TERM_COLOR_CYAN : TERM_COLOR_NORMAL;
-	}
+	sc = percent > 30 ? get_color(COLOR_CODE_RED) :
+		percent > 3 ? get_color(COLOR_CODE_MAGENTA) :
+		percent < -30 ? get_color(COLOR_CODE_BLUE) :
+		percent < -3 ? get_color(COLOR_CODE_CYAN) : get_color(COLOR_CODE_NORMAL);
+	ec = get_color(COLOR_CODE_RESET);
```
namhyung, line 415: You don't need to set 'ec' here, right?

utils/debug.c
```diff
@@ -44,24 +56,29 @@ const char *color_enum_or = TERM_COLOR_RESET TERM_COLOR_BOLD "|" TERM_COLOR_RESE
...
 };
 
 static void color(const char *code, FILE *fp)
 {
 	size_t len = strlen(code);
 
-	if ((fp == logfp && log_color == COLOR_OFF) ||
-	    (fp == outfp && out_color == COLOR_OFF))
-		return;
+	if (format_mode != FORMAT_HTML) {
```
namhyung, line 77: Well, I think color and output format is independent.  We might have a html without color..

utils/debug.c
```diff
@@ -136,7 +153,16 @@ void setup_color(enum color_setting color, char *pager)
 		out_color = color;
 	}
 
-	if (out_color != COLOR_ON) {
+	if (format_mode == FORMAT_HTML) {
```
namhyung, line 156: Ditto.

utils/debug.c
honggyukim: But we can just dump the plain text if someone wants to print without color. So I think this is needed.

Especially, we can't use redirection to save html format into another file without this. This makes output file contain comments in html gray color tag.

utils/debug.c
honggyukim: As mentioned, we can just dump the plain text if someone wants to print without color so I think this is okay.

cmds/replay.c
```diff
@@ -533,9 +533,9 @@ void get_argspec_string(struct uftrace_task_reader *task,
 	ASSERT(arg_list && !list_empty(arg_list));
 
 	if (needs_paren)
-		print_args(&args, &len, "%s(%s", color_bold, color_reset);
+		print_args(&args, &len, "(");
```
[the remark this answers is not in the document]
honggyukim: I think this doesn't make much difference in console output, but it makes too verbose in html output printing lots of span tags. So I just removed it.

I will split this change into a separate patch If you want.

utils/debug.c
namhyung: I think you can set opts->color to yes if format is HTML and user didn't set the color option.

honggyukim: Yeah, I can do that. just updated as you mentioned.

utils/debug.c
```diff
@@ -288,12 +336,20 @@ static void __print_time_unit(int64_t delta_nsec, bool needs_sign)
...
+			HTML_COLOR_CYAN    "-",
+			HTML_COLOR_NORMAL  "-",
+		};
 		int sign_idx = (delta_nsec > 0);
 		int indent = (delta >= 100) ? 0 : (delta >= 10) ? 1 : 2;
 		const char *sign = signs[sign_idx];
 		const char *ends = TERM_COLOR_NORMAL;
 
-		if (out_color == COLOR_ON) {
+		if (out_color == COLOR_ON || format_mode == FORMAT_HTML) {
```
namhyung, line 352: Please don't change this.

utils/debug.c
```diff
@@ -147,6 +171,24 @@ void setup_color(enum color_setting color, char *pager)
...
+	unsigned i;
+
+	for (i = 0; i < ARRAY_SIZE(colors); i++) {
+		if (code == colors[i].code) {
+			if (out_color == COLOR_ON) {
+			        if (format_mode == FORMAT_HTML)
+				        return colors[i].html_color;
+                                else
+				        return colors[i].color;
+                        }
```
namhyung, line 185: Indentation was broken.  Also I prefer having fewer blocks in a loop.  Please consider using 'continue' more frequently.

uftrace.c
```diff
@@ -825,6 +828,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
...
 		break;
 
+	case OPT_format:
+		if (!strcmp(arg, "html")) {
+			format_mode = FORMAT_HTML;
+			if (opts->color == COLOR_AUTO)
+				opts->color = COLOR_ON;
+		}
+		else
+			format_mode = FORMAT_NORMAL;
```
namhyung, line 838: I'm not sure about this part.  This will accept any invalid strings..  should be better if it shows a warning.

utils/debug.c
```diff
@@ -273,8 +317,12 @@ static void __print_time_unit(int64_t delta_nsec, bool needs_sign)
...
 		delta = delta_small = 999;
 
-	if (out_color == COLOR_ON)
-		unit = color_units[idx];
+	if (out_color == COLOR_ON) {
+		if (format_mode == FORMAT_HTML)
+			unit = html_color_units[idx];
+		else
+			unit = color_units[idx];
+        }
```
namhyung, line 325: weird indentation.

uftrace.c
honggyukim: I put a warning message in that case.

cmds/replay.c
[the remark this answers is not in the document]
honggyukim: Please see #1333.

cmds/replay.c
```diff
@@ -703,7 +703,10 @@ void get_argspec_string(struct uftrace_task_reader *task,
 
 			if (sym) {
 				print_args(&args, &len, "%s", color_symbol);
-				print_args(&args, &len, "&%s", sym->name);
+				if (format_mode == FORMAT_HTML)
+					print_args(&args, &len, "&amp%s", sym->name);
```
namhyung, line 707: Isn't it `&amp;`?

utils/debug.c
```diff
@@ -147,6 +171,25 @@ void setup_color(enum color_setting color, char *pager)
...
+
+	for (i = 0; i < ARRAY_SIZE(colors); i++) {
+		if (code == colors[i].code) {
+			if (format_mode == FORMAT_HTML)
+				return colors[i].html_color;
+			else
+				return colors[i].color;
+		}
+	}
+	pr_warn("get_color('%c'): invalid color code\n", code);
```
namhyung, line 189: I'm not sure we want to add a warning here... it'd generate too many warnings if something is broken.

cmds/replay.c
honggyukim: The output looked fine, but changed it to `&amp;` anyway.

utils/debug.c
honggyukim: Yeap, removed it.

namhyung (approved):
LGTM

Issue #1303, opened 2021-08-12
compiler warning in alpine Linux build


honggyukim:
We can reproduce the problem as follows.
```
$ docker run -v $PWD:/uftrace -w /uftrace --rm -it alpine:3.13

/uftrace # apk add build-base linux-headers git bash htop less libunwind-dev

/uftrace # ./configure && make
        ...
/uftrace/utils/debug.c: In function '__pr_err_s':
/uftrace/utils/debug.c:190:21: warning: format '%s' expects argument of type 'char *', but argument 3 has type 'int' [-Wformat=]
  190 |  fprintf(logfp, ": %s\n", strerror_r(saved_errno, buf, sizeof(buf)));
      |                    ~^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                     |     |
      |                     |     int
      |                     char *
      |                    %d
```
man page says.
```c
$ man strerror_r
        ...
       int strerror_r(int errnum, char *buf, size_t buflen);
                   /* XSI-compliant */

       char *strerror_r(int errnum, char *buf, size_t buflen);
                   /* GNU-specific */
        ...
```

kangtegong:
I want to resolve this issue!

kangtegong:
i think this warning comes from alpine linux, not from uftrace.

Althogh man page of strerror_r does says `strerror_r` has two return values depending on the system like below,

```
int strerror_r(int errnum, char *buf, size_t buflen);
                   /* XSI-compliant */

char *strerror_r(int errnum, char *buf, size_t buflen);
                   /* GNU-specific */
```

if i `apk add build-base`, and then see /usr/include/string.h,

```
#if defined(_POSIX_SOURCE) || defined(_POSIX_C_SOURCE) \
 || defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE) \
 || defined(_BSD_SOURCE)
int strerror_r (int, char *, size_t);
```

https://stackoverflow.com/questions/42190156/compatibility-of-strerror-r-on-different-platform

namhyung:
I think we can add a wrapper of the function like `uftrace_strerror()` and check the return value of the strerror_r().  If it's 0, then it's XSI version and the funciton succeeded.  We can return the buf then.  If it's XSI version and failed, the return value would be a positive integer less than 4096.  We might fill the buffer with a default message like "error: %d" and return.

[...]

PR #1313, opened 2021-08-20
utils: Change strerror_r into wrapper function

kangtegong:
This commit change `strerror_r` into wrapper function called
`uftrace_strerror`, because return value of the function can be
either string, or int depending on the system.

Fixed: #1303

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -886,6 +886,23 @@ char *absolute_dirname(const char *path, char *resolved_path)
 	return resolved_path;
 }
 
+char *uftrace_strerror(int errnum, char *buf, size_t buflen)
+{
+	long result = (long) strerror_r(errnum, buf, buflen);
+
+	if (result == 0)
+		/* XSI-compliant strerror_r succeed */
+		return buf;
+	else if (result < 4096) {
+		/* XSI-compliant strerror_r failed */
+		snprintf(buf, buflen, "error: %d", errnum);
+		return buf;
+	}
+	else
+		/* GNU-specific strerror_r */
+		return (char *)result;
+}
+
 void stacktrace(void)
 {
 #ifdef HAVE_LIBUNWIND
--- a/libtraceevent/event-parse.c
+++ b/libtraceevent/event-parse.c
@@ -32,6 +32,7 @@
 #include <stdint.h>
 #include <limits.h>
 
+#include "utils/utils.h"
 #include "event-parse.h"
 #include "event-utils.h"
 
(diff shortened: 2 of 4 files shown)


utils/utils.c
```diff
@@ -886,6 +886,23 @@ char *absolute_dirname(const char *path, char *resolved_path)
...
+char *uftrace_strerror_r(int errnum, char *buf, size_t buflen)
+{
+	long strerror_r_result = (long) strerror_r(errnum, buf, buflen);
+
+	if (strerror_r_result == 0)
+		/* XSI-compliant strerror_r succeed */
+		return buf;
+	else if (strerror_r_result > 0 && strerror_r_result < 4096) {
+		/* XSI-compliant strerror_r failed */
+		sprintf(buf, "error: %ld", strerror_r_result);
```
namhyung, line 898: Please use `snprintf()` rather than plain `sprintf()`.  Also I'd like to print the errnum here (but printing strerror_r_result too looks ok).

utils/utils.c
```diff
@@ -886,6 +886,23 @@ char *absolute_dirname(const char *path, char *resolved_path)
 	return resolved_path;
 }
 
+char *uftrace_strerror_r(int errnum, char *buf, size_t buflen)
+{
+	long strerror_r_result = (long) strerror_r(errnum, buf, buflen);
```
namhyung, line 891: For a local variable, I prefer a short name like 'result' or even 'ret'.

utils/utils.c
```diff
@@ -886,6 +886,23 @@ char *absolute_dirname(const char *path, char *resolved_path)
 	return resolved_path;
 }
 
+char *uftrace_strerror_r(int errnum, char *buf, size_t buflen)
```
namhyung, line 889: Can we simply call it `uftrace_strerror`?  The `_r` part in the `strerror_r()` was an ugly addition since it could not replace the original one.

utils/utils.c
```diff
@@ -886,6 +886,23 @@ char *absolute_dirname(const char *path, char *resolved_path)
...
+{
+	long strerror_r_result = (long) strerror_r(errnum, buf, buflen);
+
+	if (strerror_r_result == 0)
+		/* XSI-compliant strerror_r succeed */
+		return buf;
+	else if (strerror_r_result > 0 && strerror_r_result < 4096) {
```
namhyung, line 896: This can be simply `else if (strerror_r_result < 4096)` as userspace pointer never goes to negative.

kangtegong:
Just updated the commit as said. Always thank you for your kind feedbacks!

namhyung (commented):
Please update the commit message.  And did you actually test it in the alpine?

kangtegong:
just changed the commit message 

`uftrace_strerror_r` to `uftrace_strerror`.

And yes, I tested this code on alpine 3.13 as following order.

```
$ docker run -v $PWD:/uftrace -w /uftrace --rm -it alpine:3.13
/uftrace # apk add build-base linux-headers git bash htop less libunwind-dev
/uftrace # ./configure && make
```

namhyung:
Thanks, and you don't need to check on a host.

namhyung (approved):
LGTM

PR #1333, opened 2021-09-03
replay: Do not make bold for parenthesis and comma

honggyukim:
We don't have to make the replay output unnecessarily complicated so
this patch removes the terminal output bold for parenthesis and comma.


namhyung (approved):
LGTM

PR #1334, opened 2021-09-05
replay: Do not print "<lambda" as struct type name

honggyukim:
gcc puts "<lambda" to annoymous lambda but it'd be better to ignore
to make it same as clang.

It would make less troublesome in format html output.


namhyung (approved):
LGTM

PR #1341, opened 2021-09-11
doc: Add doc/run-spell.sh for spell check in the doc

honggyukim:
spell is a spell checking program which prints each misspelled word on a
line of its own.

We need to generate dictionary file used in the project first and it was
generated as follows.
```
  $ spell *.md doc/*.md | sort | uniq --ignore-case > doc/dictionary.txt
```
The spell checker can be run as follows.
```
  $ ./doc/run-spell.sh
  /home/honggyu/work/uftrace/doc/uftrace-record.md:633: excutables
```
It detects an incorrect word from the document.

This PR also fixes the above incorrect word as well.


namhyung:
It seems 'aspell' is more popular than 'spell'.  At least my system has 'aspell' already.

honggyukim:
My system doesn't have 'aspell' but anyway, I need to check 'aspell' works to update the script.

honggyukim:
I should spend some time to make the script run with `aspell`. If you want to make release before then could you apply the second patch for an incorrect word?

namhyung:
OK, I will apply the second patch only.

honggyukim:
It looks the usage of `aspell` is basically interactive mode, which is different from `spell`. 

I'm not sure how to use and create a custom dictionary for uftrace as of now. I'm not sure if it's useful enough using `aspell` instead of `spell`.

> OK, I will apply the second patch only.

Please go ahead.

namhyung:
Done.

honggyukim:
We don't need this because #1507 does code spell check in pre-commit.

PR #1342, opened 2021-09-11
doc: Change NEWS to NEWS.md and add issue/pr links

honggyukim:
It can be preferable to having markdown NEWS and add github
issue/pr number to NEWS.md.


namhyung:
Do we really need to add the links manually?

honggyukim:
I think it'd be clearer if we have links for each item. If you don't want, I can drop this.

namhyung:
Yeah, I think keeping the issue/pr number is enough.

honggyukim:
> Yeah, I think keeping the issue/pr number is enough.

Then we can just use plain text `README` with issue/pr numbers is enough.

honggyukim:
I will close this then.

PR #1344, opened 2021-09-12
build: Use pkg-config to get compiler options of libpython3

jetswayss:
In a cross-compilation scenario, it might not have python executable.
Even if it exists, the compiler options are for the host environment.
Not for the build target environment.

Using pkg-config to get compiler options of libpython3 fixes
libpython in the cross-compilation scenario.

Signed-off-by: Kun-Chuan Hsieh <jetswayss@gmail.com>

### DIFF
--- a/check-deps/Makefile
+++ b/check-deps/Makefile
@@ -22,9 +22,9 @@ CHECK_LIST += cc_has_minline_all_stringops
 CHECK_CFLAGS  = $(CFLAGS)  $(CFLAGS_$@) -Werror
 CHECK_LDFLAGS = $(LDFLAGS) $(LDFLAGS_$@)
 
-# python3.8 added --embed option
-ifneq ($(shell python3-config 2> /dev/null | grep embed),)
-  EMBED := --embed
+# libpython3 provides an embed verison of pkg-config file since python3.8
+ifeq ($(shell pkg-config python3-embed --exists 2> /dev/null; echo $$?), 0)
+  EMBED := -embed
 endif
 
 CFLAGS_cc_has_mfentry = -mfentry
@@ -32,8 +32,8 @@ LDFLAGS_cxa_demangle = -lstdc++
 LDFLAGS_have_libelf = -lelf
 CFLAGS_cc_has_mno_sse2 = -mno-sse2
 LDFLAGS_have_libpython2.7 = -lpython2.7
-CFLAGS_have_libpython3 = $(shell python3-config --cflags 2> /dev/null)
-LDFLAGS_have_libpython3 = $(shell python3-config $(EMBED) --libs 2> /dev/null)
+CFLAGS_have_libpython3 = $(shell pkg-config python3$(EMBED) --cflags 2> /dev/null)
+LDFLAGS_have_libpython3 = $(shell pkg-config python3$(EMBED) --libs 2> /dev/null)
 CFLAGS_have_libluajit = $(shell pkg-config --cflags luajit 2> /dev/null)
 LDFLAGS_have_libluajit = $(shell pkg-config --libs  luajit 2> /dev/null)
 CFLAGS_have_libncurses  = $(shell pkg-config --cflags ncursesw 2> /dev/null)
--- a/check-deps/Makefile.check
+++ b/check-deps/Makefile.check
@@ -16,14 +16,13 @@ ifneq ($(wildcard $(srcdir)/check-deps/cc_has_mno_sse2),)
 endif
 
 ifneq ($(wildcard $(srcdir)/check-deps/have_libpython3),)
-  # python3.8 added --embed option
-  ifneq ($(shell python3-config | grep embed),)
-    EMBED := --embed
+  # libpython3 provides an embed verison of pkg-config file since python3.8
+  ifeq ($(shell pkg-config python3-embed --exists 2> /dev/null; echo $$?), 0)
+    EMBED := -embed
   endif
-  PYTHON3 := $(shell python3-config $(EMBED) --libs | sed 's/-l\(python[0-9\.m]\+\) .*/\1/g')
   COMMON_CFLAGS += -DHAVE_LIBPYTHON3
-  COMMON_CFLAGS += -I $(shell $(PYTHON3) -c 'import sysconfig; print(sysconfig.get_config_vars()["INCLUDEPY"])')
-  COMMON_CFLAGS += -DLIBPYTHON_VERSION=$(shell $(PYTHON3) -c 'import sysconfig; print(sysconfig.get_config_vars()["LDVERSION"])')
+  COMMON_CFLAGS += $(shell pkg-config python3$(EMBED) --cflags)
+  COMMON_CFLAGS += -DLIBPYTHON_VERSION=$(shell pkg-config python3$(EMBED) --modversion)
 else ifneq ($(wildcard $(srcdir)/check-deps/have_libpython2.7),)
   COMMON_CFLAGS += -DHAVE_LIBPYTHON2
   COMMON_CFLAGS += -I/usr/include/python2.7


check-deps/Makefile
```diff
@@ -23,17 +23,17 @@ CHECK_CFLAGS  = $(CFLAGS)  $(CFLAGS_$@) -Werror
 CHECK_LDFLAGS = $(LDFLAGS) $(LDFLAGS_$@)
 
 # python3.8 added --embed option
```
namhyung, line 25: Could you please update this comment too?

check-deps/Makefile
```diff
@@ -23,17 +23,17 @@ CHECK_CFLAGS  = $(CFLAGS)  $(CFLAGS_$@) -Werror
...
-ifneq ($(shell python3-config 2> /dev/null | grep embed),)
-  EMBED := --embed
+ifeq ($(shell pkg-config python3-embed --exists 2> /dev/null; echo $$?), 0)
+  EMBED := -embed
```
namhyung, line 27: Well I'm not familiar with the python development but it seems reasonable.  Can we assume `pkg-config python3-embed` is more available (or right way) than `python3-config --embed`?  I mean if there's a case where it has `python3-config` but no `pkg-config python3`.  If so, I feel like we should fall back to the existing way.

namhyung (commented):
Thanks for your contribution!

check-deps/Makefile
jetswayss: updated~

check-deps/Makefile
jetswayss: After doing some research, I found that the both ways are the right ways(since `pkg-config` has been supported from [2008](https://bugs.python.org/issue3585)).
In terms of availability, `pkg-config` is often more available than `python3-config`. For example, the `libpython3-dev` package in debian only provide `pkg-config` files and does not provide `python3-config`. Therefore, if a user who is using debian 11 and does not install `python3-dev` manually and install `uftrace` by following `QUICK GUIDE` in `INSTALL.md`, he won't be able to use `uftrace script` with python (since `install-deps.sh` only installs `libpython3-dev`).
Also in a cross-compilation scenario, using `python3-config` might accidentally get the wrong compiler options since the options are for host not target.

namhyung: Fair enough.  @honggyukim do you have any comments?

honggyukim: Hmm.. In Ubuntu 18.04, `pkg-config` doesn't give a proper output when installing only `libpython3-dev` without `python3-dev`. I tested it on docker as follows.
```
$ docker pull ubuntu:18.04

$ docker run -it ubuntu:18.04 bash

# apt update

# apt install -y pkg-config

# apt install -y libpython3-dev

# pkg-config python3 --cflags
Package python3 was not found in the pkg-config search path.
Perhaps you should add the directory containing `python3.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python3' found

# python3-config --cflags
bash: python3-config: command not found
```
I found that `python3-dev` depends on `libpython3-dev` so it looks installing only `python3-dev` is enough for our purpose.
```
$ docker pull ubuntu:18.04

$ docker run -it ubuntu:18.04 bash

# apt update

# apt install -y pkg-config

# apt install -y python3-dev

# pkg-config python3 --cflags
-I/usr/include/python3.6m -I/usr/include/x86_64-linux-gnu/python3.6m

# python3-config --cflags
-I/usr/include/python3.6m -I/usr/include/python3.6m  -Wno-unused-result -Wsign-compare -g -fdebug-prefix-map=/build/python3.6-1chSKv/python3.6-3.6.9=. -specs=/usr/share/dpkg/no-pie-compile.specs -fstack-protector -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O3 -Wall
```
How about applying only the following change?
```diff
diff --git a/misc/install-deps.sh b/misc/install-deps.sh
index 92e22e0e..a60be6c7 100755
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -13,7 +13,7 @@ distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')

 case $distro in
     "ubuntu" | "debian" | "raspbian" | "kali" | "linuxmint")
-        apt-get $OPT install pandoc libdw-dev libpython3-dev libncursesw5-dev pkg-config
+        apt-get $OPT install pandoc libdw-dev python3-dev libncursesw5-dev pkg-config
         apt-get $OPT install libluajit-5.1-dev || true
         apt-get $OPT install libcapstone-dev || true ;;
     "fedora")
```
I haven't tested other distros though, but `libpython3-dev` is only shown for ubuntu in `install-deps.sh`.

jetswayss: Thanks for checking the issue on ubuntu 18.04.
I found that on ubuntu 18.04 both `pkg-config python3` and `python3-config` won't work since both of them are provided by [python3-dev](https://packages.ubuntu.com/bionic/amd64/python3-dev/filelist).
However, after ubuntu 20.04, `pkg-config` is provided by [libpython3-dev](https://packages.ubuntu.com/focal/amd64/libpython3-dev/filelist) and `python3-config` is still provided by [python3-dev](https://packages.ubuntu.com/focal/amd64/python3-dev/filelist).
Since `python3-dev` depends on `libpython3-dev`, and `libpython3-dev` does not depends on `python3-dev`, which makes `pkg-config` more available than `python3-dev` in this case.


The patch @honggyukim provided can fix `install-deps.sh` issue on ubuntu 18.04.
Mine patch can fix the issues that one might encounter in a cross-compilation scenario. 

Could we apply the both changes? (both mine and @honggyukim).

honggyukim: > after ubuntu 20.04, pkg-config is provided by libpython3-dev

You're right. More precisely `/usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc` is provided by `libpython3-dev` so `pkg-config` can find it in Ubuntu 20.04.

But it looks a bit different using `pkg-config` and `python3-config` as follows.
```
# pkg-config python3 --cflags
-I/usr/include/python3.8 -I/usr/include/x86_64-linux-gnu/python3.8

# pkg-config python3 --libs
(... print nothing ...)

# python3-config --cflags
-I/usr/include/python3.8 -I/usr/include/python3.8  -Wno-unused-result -Wsign-compare -g -fdebug-prefix-map=/build/python3.8-CoVRmP/python3.8-3.8.10=. -specs=/usr/share/dpkg/no-pie-compile.specs -fstack-protector -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O3 -Wall

# python3-config --libs
 -lcrypt -lpthread -ldl  -lutil -lm -lm
```
If it doesn't make difference, then I'm fine having @jetswayss's change together with my change in `install-deps.sh`.

jetswayss: After python3.8, the [embed](https://bugs.python.org/issue36721) versions of `pkg-config` and `python3-config` have been added.
The `pkg-config python3-embed` is for `embedding Python into an application`. And the `pkg-config python3` is for `building a C extension for Python`.
Therefore, if an embed version exists, we should use it. `uftrace` has already handled this case.

However, the results which `pkg-config python3-embed` and `python3-config --embed` produced still have a some differences.

```
# pkg-config python3-embed --cflags
I/usr/include/python3.8

# python3-config --cflags --embed
-I/usr/include/python3.8 -I/usr/include/python3.8  -Wno-unused-result -Wsign-compare -g -fdebug-prefix-map=/build/python3.8-CoVRmP/python3.8-3.8.10=. -specs=/usr/share/dpkg/no-pie-compile.specs -fstack-protector -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O3 -Wall

# pkg-config python3-embed --libs
lpython3.8

# python3-config --libs --embed
-lpython3.8 -lcrypt -lpthread -ldl  -lutil -lm -lm
```

For `--cflags`, `python3-config --cflags --embed` generates the same result as `python3-config --cflags` which contains lots of unnecessary flags.
For `--libs`, `python3-config --libs --embed` contains some system libs which is also unnecessary.

In my opinion, both `python3-config` and `pkg-config` are the right ways. But the result `pkg-config` generated is more concise.

namhyung: Agreed, I like the pkg-config output.

@honggyukim can you please send a PR for your change?

namhyung (approved):
LGTM

check-deps/Makefile
honggyukim: > In my opinion, both python3-config and pkg-config are the right ways. But the result pkg-config generated is more concise.

@jetswayss I also think using `pkg-config` is better and your patch works fine. Thanks for your work!

> @honggyukim can you please send a PR for your change?

@namhyung I've uploaded it at #1349.

honggyukim: Hmm... `python-3.8.pc` doesn't provide `-lpython3.8` unlike `python-3.5.pc`.
```
$ pkg-config python-3.5 --cflags
-I/usr/include/python3.5m -I/usr/include/x86_64-linux-gnu/python3.5m

$ pkg-config python-3.5 --libs
-lpython3.5m

$ pkg-config python-3.8 --cflags
-I/usr/include/python3.8 -I/usr/include/x86_64-linux-gnu/python3.8

$ pkg-config python-3.8 --libs
( ... print nothing ...)
```
The `python-3.5.pc` is as follows.
```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5.pc
      ...
Name: Python
Description: Python library
Requires:
Version: 3.5
Libs.private: -lpthread -ldl  -lutil
Libs: -lpython3.5m
Cflags: -I${includedir}/python3.5m -I${includedir}/x86_64-linux-gnu/python3.5m
```
It has `-lpython3.5m` part in `Libs`.
```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8.pc
Name: Python
Description: Build a C extension for Python
Requires:
Version: 3.8
Libs.private: -lcrypt -lpthread -ldl  -lutil -lm
Libs:
Cflags: -I${includedir}/python3.8 -I${includedir}/x86_64-linux-gnu/python3.8
```
But `python-3.8.pc` does have nothing in `Libs`. The result shows a problem because of this difference. I had to use `python3
```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8-embed.pc
      ...
Name: Python
Description: Embed Python into an application
Requires:
Version: 3.8
Libs.private: -lcrypt -lpthread -ldl  -lutil -lm
Libs: -L${libdir} -lpython3.8
Cflags: -I${includedir}/python3.8
```
But it's also tricky because of the following difference. It looks `python3.5` and `python3.8` has to be used in difference ways.
```
$ pkg-config python3-embed --libs
Package python3-embed was not found in the pkg-config search path.
Perhaps you should add the directory containing `python3-embed.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python3-embed' found

$ pkg-config python-3.5-embed --libs
Package python-3.5-embed was not found in the pkg-config search path.
Perhaps you should add the directory containing `python-3.5-embed.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python-3.5-embed' found

$ pkg-config python-3.8-embed --libs
-lpython3.8
```
Because of this problem, `check-deps/have_libpython3` is not created in my environment.

jetswayss: As mentioned in the previous response, after python 3.8, if the embed version exists, it should be used.
We test whether it exists by executing `pkg-config python3-embed --exists 2> /dev/null; echo $?`.

No sure why `check-deps/have_libpython3` was not created in your environment.
It seems you have installed more than one version of python on your environment.

Could you test the following commands on your environment?

`pkg-config python3 --exists 2> /dev/null; echo $?`
`pkg-config python3 --modversion`
`pkg-config python3-embed --exists 2> /dev/null; echo $?`
`pkg-config python3-embed --modversion`
`python3-config --cflags`

honggyukim: In my environment `pkg-config python3-embed --exists` is false unexpectedly.
```
$ pkg-config python-3.8-embed --libs
-lpython3.8
```
And the package name is not `python3.8-embed`, but `python-3.8-embed`. Maybe is this my environmental problem? I just worry if someone else can have the similar issue.

honggyukim: > No sure why check-deps/have_libpython3 was not created in your environment.

It's not created because `pkg-config python3-embed --exists` is false so it uses `pkg-config python3 --libs` instead of `pkg-config python3-embed --libs` and it doesn't provide `-lpython3.8` so `__have_libpython3.c` can't be compiled.

jetswayss: The reason why I want to know the above mentioned 5 commands is that I'm trying to figure out whether this is an old issue or is an issue created by this commit.

`pkg-config python3-embed --exists` returning `false` is okay as long as `pkg-config python3 --modversion` do not print `3.8`.

If the following three all returning `0` or not found, then this sholud be an old issue. (`check-deps/have_libpython3` won't be created on your environment even if this commit was not applied).
`pkg-config python3 --exists 2> /dev/null; echo $?`
`pkg-config python3-embed --exists 2> /dev/null; echo $?`
`python3-config --cflags`

Cloud you check the following commands on your environment?
`pkg-config python3 --exists 2> /dev/null; echo $?`
`pkg-config python3 --modversion`
`python3-config --cflags`

Thank you!

honggyukim: Sorry, I didn't answer your question before. Here are the results you asked.
```
$ pkg-config python3 --exists 2> /dev/null; echo $?
0

$ pkg-config python3 --modversion
3.8

$ pkg-config python3-embed --exists 2> /dev/null; echo $?
1

$ pkg-config python3-embed --modversion
Package python3-embed was not found in the pkg-config search path.
Perhaps you should add the directory containing `python3-embed.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python3-embed' found

$ python3-config --cflags
-I/usr/include/python3.8 -I/usr/include/python3.8  -Wno-unused-result -Wsign-compare -g -fstack-protector -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O3 -Wall
```

honggyukim: The current master doesn't create `have_libpython3` as follows.
```
$ make -siC check-deps/ check-clean

$ make -siC check-deps/ check-build

$ ls check-deps/have_libpython*
check-deps/have_libpython2.7
```
But I was able to get `have_libpython3` after reverting your commit.
```
$ make -siC check-deps/ check-clean

$ make -siC check-deps/ check-build

$ ls check-deps/have_libpython*
check-deps/have_libpython2.7  check-deps/have_libpython3
```
That's the current problem in my side.

honggyukim: > It seems you have installed more than one version of python on your environment.

Yes, I use Ubuntu 16.04 so default is python3.5 and I install python3.8 manually long before and it has worked fine so far.

jetswayss: I'm so sorry that mine changes break the libpython support on your computer.

I thought that if `python3.pc` exists, `python-embed.pc` should exist as well after python 3.8.
But it looks like my presumption was wrong.


I'm still trying to figure out why this happened by trying to reproducing the environment.
I've tried two ways to install python 3.8 on ubuntu 16.04 using docker.
But still not able to produce the same result in your environment.


## Install python 3.8 via ppa:deadsnakes
```
docker run -it ubuntu:16.04 bash
apt update
apt install -y pkg-config
apt install -y python3-dev

apt install software-properties-common
add-apt-repository ppa:deadsnakes/ppa
apt install python3.8-dev
```

```
# ls -l /usr/lib/x86_64-linux-gnu/pkgconfig/python*
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5m.pc -> python-3.5.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8-embed.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc -> python-3.5.pc
```

Which result looks fine, the default python version 3.5. Therefore `uftrace` will use libpython3.5.


## Install python 3.8 via building from the source
```
docker run -it ubuntu:16.04 bash
apt update
apt install -y pkg-config
apt install -y python3-dev

mkdir /download
cd /download
apt install build-essential zlib1g-dev libncurses5-dev libgdbm-dev libnss3-dev libssl-dev libreadline-dev libffi-dev wget
wget https://www.python.org/ftp/python/3.8.0/Python-3.8.0.tgz
tar -xf Python-3.8.0.tgz
cd Python-3.8.0
./configure --enable-shared
make
make install
```

```
# ls -l /usr/lib/x86_64-linux-gnu/pkgconfig/python*
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5m.pc -> python-3.5.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc -> python-3.5.pc

# ls -l /usr/local/lib/pkgconfig/python*
-rw-r--r-- 1 root root 313 Sep 19 11:19 /usr/local/lib/pkgconfig/python-3.8-embed.pc
-rw-r--r-- 1 root root 287 Sep 19 11:19 /usr/local/lib/pkgconfig/python-3.8.pc
lrwxrwxrwx 1 root root  19 Sep 19 11:19 /usr/local/lib/pkgconfig/python3-embed.pc -> python-3.8-embed.pc
lrwxrwxrwx 1 root root  13 Sep 19 11:19 /usr/local/lib/pkgconfig/python3.pc -> python-3.8.pc
```

Which result looks fine, the default python version 3.8 and `python3-embed.pc` exists.


Sorry for the inconvenience.
Do you remember how python3.8 was installed on your environment?
I'm tring to figure out who created symbolic link `python3.pc` for python3.8 and why it didn't created one for `python3-embed.pc`

honggyukim: @jetswayss Sorry. I found the problem. It's because of missing link `python3-embed.pc` to `python-3.8-embed.pc`. I installed `python3.8` from ppa but it looks like setting up the soft link has to be done by the user manually.

I didn't do because `python3.5` didn't have such soft link `python3-embed.pc` at all, but it's required by `python3.8`. Sorry for making you confused. I think the problem is from my environment.

But I worry a bit about handling too many soft links when changing python versions. Maybe I should bind more soft links for my `python3.8` like this way.
```
$ sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 2 --slave /usr/bin/python3-config python3-config /usr/bin/python3.8-config --slave /usr/lib/x86_64-linux-gnu/pkgconfig/python3-embed.pc python3-embed.pc python-3.8-embed.pc
```

honggyukim: The current python3 setup is
```
$ python3 --version
Python 3.8.7

$ python3-config --embed --libs
-lpython3.8 -lcrypt -lpthread -ldl  -lutil -lm -lm

$ pkg-config python3-embed --libs
-lpython3.8
```
Then I can switch python3 versions as follows.
```
$ sudo update-alternatives --config python3
There are 2 choices for the alternative python3 (providing /usr/bin/python3).

  Selection    Path                Priority   Status
------------------------------------------------------------
  0            /usr/bin/python3.8   2         auto mode
  1            /usr/bin/python3.5   1         manual mode
* 2            /usr/bin/python3.8   2         manual mode

Press <enter> to keep the current choice[*], or type selection number: 1
update-alternatives: using /usr/bin/python3.5 to provide /usr/bin/python3 (python3) in manual mode
```
Then the python3 is updated to `python3.5`.
```
$ python3 --version
Python 3.5.2

$ python3-config --libs
-lpython3.5m -lpthread -ldl  -lutil -lm

$ pkg-config python3-embed --libs
Package python3-embed was not found in the pkg-config search path.
Perhaps you should add the directory containing `python3-embed.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python3-embed' found

$ pkg-config python3 --libs
-lpython3.5m
```
I confirmed that both `python3.5` and `python3.8` can create `have_libpython3` without a problem.

@jetswayss I've learned many from this. Thanks for your help!

honggyukim: > Install python 3.8 via ppa:deadsnakes
> ```
> # ls -l /usr/lib/x86_64-linux-gnu/pkgconfig/python*
> /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5.pc
> /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.5m.pc -> python-3.5.pc
> /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8-embed.pc
> /usr/lib/x86_64-linux-gnu/pkgconfig/python-3.8.pc
> /usr/lib/x86_64-linux-gnu/pkgconfig/python3.pc -> python-3.5.pc
> ```
> Which result looks fine, the default python version 3.5. Therefore uftrace will use libpython3.5.

This is my problem because there is no such `/usr/lib/x86_64-linux-gnu/pkgconfig/python3-embed.pc`. The problem happens when I switch the other setup to `python3.8`, which is not default in the system.

PR #1346, opened 2021-09-14
tests: Sort import order of libraries using isort

kangtegong:
I think it's better to stanzdarize import orders
of libraries in python scripts (in tests/ and scripts/, etc)
to prevent circular imports (although happens rarely),
or for faster import, or for better readabilities.

We can reformat import orders with isort tools,
and this commit contains the result of the tool.

reference: https://pypi.org/project/isort/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,8 @@
+[tool.isort]
+profile = "black"
+lines_between_types = 0
+lines_after_imports = 1
+combine_as_imports = true
+ignore_whitespace = true
+skip_gitignore = true
+
--- a/misc/gen-autoargs.py
+++ b/misc/gen-autoargs.py
@@ -9,8 +9,9 @@
 #
 
 from __future__ import print_function
-import sys
+
 import re
+import sys
 
 # The syntax of C in Backus-Naur Form
 #  https://cs.wmich.edu/~gupta/teaching/cs4850/sumII06/The%20syntax%20of%20C%20in%20Backus-Naur%20form.htm
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python
 
-import random
-import os, sys
-import tempfile
-import glob, re
-import subprocess as sp
+import glob
 import multiprocessing
+import os
+import random
+import re
+import subprocess as sp
+import sys
+import tempfile
 import time
 
 class TestBase:
--- a/tests/t022_filter_kernel.py
+++ b/tests/t022_filter_kernel.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
-from runtest import TestBase
 import os
+
+from runtest import TestBase
 
 class TestCase(TestBase):
     def __init__(self):
--- a/tests/t067_report_diff.py
+++ b/tests/t067_report_diff.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
-from runtest import TestBase
 import subprocess as sp
+
+from runtest import TestBase
 
 XDIR='xxx'
 YDIR='yyy'
--- a/tests/t079_replay_kernel_D.py
+++ b/tests/t079_replay_kernel_D.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 
-from runtest import TestBase
-import subprocess as sp
 import os
+import subprocess as sp
+
+from runtest import TestBase
 
 class TestCase(TestBase):
     def __init__(self):
--- a/tests/t080_replay_kernel_D2.py
+++ b/tests/t080_replay_kernel_D2.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 
-from runtest import TestBase
-import subprocess as sp
 import os
+import subprocess as sp
+
+from runtest import TestBase
 
 TDIR='xxx'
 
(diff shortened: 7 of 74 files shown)


kangtegong:
This patch can be reproduced by 

```
isort --ignore-whitespace [PATH TO tests/ and scripts/ directory]
```

or

```
isort --ws [PATH TO tests/ and scripts/ directory]
```

However, if you run this command directly, this will reorder library imports (just like the patch I sent)
**AND** make line break before `from` keyword,
**AND** make two line breaks defore class declaration, like below.

```
# t022_filter_kernel.py
import subprocess as sp
[newline]
from runtest import TestBase
[newline]
[newline]
class TestCase(TestBase):
```

This commit only contains reordered imports of library,
and did not included linebreaks that isort make.

If you want to make this patch same as the output of `isort --ws` command,
let me know. I'll amend the commit right away.

honggyukim:
It looks `--ws` might have difference across isort versions. My isort doesn't ignore whitespace when fixing `.py` files.

The help message says `--ws` is only meaningful when it's used with `--check-only`, which doesn't change the code.
```
  -c, --check-only, --check
                        Checks the file for unsorted / unformatted imports and
                        prints them to the command line without modifying the
                        file. Returns 0 when nothing would change and returns
                        1 when the file would be reformatted.
  --ws, --ignore-whitespace
                        Tells isort to ignore whitespace differences when
                        --check-only is being used.
```
What is your isort version? Can you show us the help message?

kangtegong:
Hmm, my isort is `5.9.3` too, and help message is same as you. 
But in my case, when used `  --ws, --ignore-whitespace` option, 
(even without using `--check-only` option) insignificant whitespaces are ignored.

For instance, when I run `isort .` in tests/ directory,
256 files are fixed including the whitespace fixes


But when I run the command below in tests/ directory,
`isort --ws .` , only 67 files (whose import order should be reordered) are fixed,
and other files that only need whitespace fixes are ignored.

namhyung (approved):
Besides the isort issue, I'm ok with the change itself.

honggyukim:
> Are the results of isort . and isort --ws . same in your computer?

I tested again based on Ubuntu 20.04 docker in a clean status. I also see some difference as follows.
```
root@c44b763aa376:/usr/src/uftrace/tests# isort *.py | wc -l
263

root@c44b763aa376:/usr/src/uftrace/tests# git reset --hard HEAD
HEAD is now at 751c1375 build: Use pkg-config to get compiler options of libpython3

root@c44b763aa376:/usr/src/uftrace/tests# isort --ws *.py | wc -l
71
```
The difference comes from the change whether to apply the fix only for whitespaces or not. For example `--ws` doesn't apply change something like this.
```diff
diff --git a/tests/t001_basic.py b/tests/t001_basic.py
index 88aab387..649c3f07 100644
--- a/tests/t001_basic.py
+++ b/tests/t001_basic.py
@@ -2,6 +2,7 @@

 from runtest import TestBase

+
 class TestCase(TestBase):
     def __init__(self):
         TestBase.__init__(self, 'abc', """
```
But if `isort` has to fix the order of `import`s then it also gives empty lines as follows.
```diff
diff --git a/tests/t222_external_data.py b/tests/t222_external_data.py
index 0c293a86..4adc3fba 100644
--- a/tests/t222_external_data.py
+++ b/tests/t222_external_data.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python

-from runtest import TestBase
-import subprocess as sp
 import os.path
+import subprocess as sp
+
+from runtest import TestBase
+

 class TestCase(TestBase):
     def __init__(self):
```
I don't see empty spaces from your changes.

honggyukim:
I think our `isort` result has to be same. Otherwise, my `isort` run will fix your changes again and your `isort` run will change mine and it's never ending formatting fixes. Please check it again.

In addition, can you apply `isort` to `misc/gen-autoargs.py` as well?

honggyukim (changes requested):
I think our `isort` results have to be same. Otherwise, my `isort` run will fix your changes again and your `isort` run will change mine and it's never ending formatting fixes. Please check it again.

In addition, can you apply `isort` to `misc/gen-autoargs.py` as well?

honggyukim:
Maybe we can consider adding `pyproject.toml` and write the configuration in it.
https://pycqa.github.io/isort/docs/configuration/config_files.html

kangtegong:
just changed pr as @honggyukim mentioned :)
Can you take a look?

Also, I think importing local module and python library should be distinguished with one line like below.

https://github.com/microsoft/debugpy/blob/main/tests/logs.py

honggyukim (approved):
Sorry for coming back late. I was away for a while. I also get the same output when running `isort .` so it looks fine, but could you please fix your commit title from `tets: Sort ...` to `python: Sort ...`?

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.isort]
+profile = "black"
+lines_between_types = 0
+lines_after_imports = 1
+combine_as_imports = true
+ignore_whitespace = true
+skip_gitignore = true
```
honggyukim, line 7: It shows `No newline at end of file` and it looks because of your editor setting. Please make sure it's not shown.

kangtegong:
I just fixed typo in commit message and added newline in pyproject.toml in order to prevent `No newline at end of file`. Thank you for your feedback! :)

honggyukim (approved):
LGTM

PR #1349, opened 2021-09-16
misc: Install python3-dev instead of libpython3-dev

honggyukim:
python3-config can be installed by python3-dev, which depends on
libpython3-dev.

We no longer require to use python3-config but it'd be better to install
python3-dev instead of libpython3-dev.


namhyung (approved):
LGTM

Issue #1238, opened 2021-02-10
Investigate gcc -fpatchable-function-entry


namhyung:
Recent GCC added the -fpatchable-function-entry option to better support dynamic tracing.  We should investigate it and support the binaries too.

honggyukim:
It would be really helpful if we have `-fpatchable-function-exit` as well in the future.

honggyukim:
I also see that both GCC and Clang have `-fpatchable-function-entry` in the recent versions.

honggyukim:
This option is also supported both in x86_64 and aarch64 as well.

honggyukim:
alpine doesn't support `-pg` nor `-finstrument-functions`, but it supports `-fpatchable-function-entry`.
```
$ docker run -v $PWD:/uftrace -w /uftrace --rm -it alpine:3.13

/uftrace # apk add build-base

/uftrace # gcc -pg tests/s-abc.c
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find gcrt1.o: No such file or directory
collect2: error: ld returned 1 exit status

/uftrace # gcc -finstrument-functions tests/s-abc.c
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /tmp/cceEeFpD.o: in function `a':
s-abc.c:(.text+0x18): undefined reference to `__cyg_profile_func_enter'
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: s-abc.c:(.text+0x33): undefined reference to `__cyg_profile_func_exit'
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /tmp/cceEeFpD.o: in function `b':
s-abc.c:(.text+0x57): undefined reference to `__cyg_profile_func_enter'
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: s-abc.c:(.text+0x72): undefined reference to `__cyg_profile_func_exit'
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: /tmp/cceEeFpD.o: in function `c':
[...]

PR #1367, opened 2021-10-03
dynamic: Support -fpatchable-function-entry=N tracing

honggyukim:
Recnet gcc and clang provides a way to have NOP sleds at the beginning
of functions with new option '-fpatchable-function-entry=N[,M]'.

The explanation in gcc[1] is as follows.
```
  -fpatchable-function-entry=N[,M]
    Generate N NOPs right at the beginning of each function, with the
    function entry point before the Mth NOP. If M is omitted, it
    defaults to 0 so the function entry points to the address just at
    the first NOP. The NOP instructions reserve extra space which can be
    used to patch in any desired instrumentation at run time, provided
    that the code segment is writable. The amount of space is
    controllable indirectly via the number of NOPs; the NOP instruction
    used corresponds to the instruction emitted by the internal GCC
    back-end interface gen_nop. This behavior is target-specific and may
    also depend on the architecture variant and/or other compilation
    options.

    For run-time identification, the starting addresses of these areas,
    which correspond to their respective function entries minus M, are
    additionally collected in the __patchable_function_entries section
    of the resulting binary.

    Note that the value of __attribute__ ((patchable_function_entry (N,M)))
    takes precedence over command-line option -fpatchable-function-entry=N,M.
    This can be used to increase the area size or to remove it completely
    on a single function. If N=0, no pad location is recorded.

    The NOP instructions are inserted at—and maybe before, depending on
    M—the function entry address, even before the prologue.

    The maximum value of N and M is 65535.
```
This patch makes uftrace use -fpatchable-function-entry=N in x86_64.
The N should be 5(or greater) because The call instruction requires 5
bytes.  The example usage is as follows.
```
  $ gcc -fpatchable-function-entry=5 tests/s-abc.c

  $ uftrace -P. --debug-domain dynamic:3 a.out
  dynamic: dynamic patch type: a.out: 3 (fentry-nop)
  dynamic: update function 'a' dynamically to call __fentry__
  dynamic: update function 'b' dynamically to call __fentry__
  dynamic: update function 'c' dynamically to call __fentry__
  dynamic: update function 'main' dynamically to call __fentry__
  dynamic: patched all (4) functions in 'a.out'
              [ 30535] | main() {
              [ 30535] |   a() {
              [ 30535] |     b() {
              [ 30535] |       c() {
     0.937 us [ 30535] |         getpid();
     4.143 us [ 30535] |       } /* c */
     4.618 us [ 30535] |     } /* b */
     5.078 us [ 30535] |   } /* a */
     5.769 us [ 30535] | } /* main */
```
Note that this feature reuses __fentry__ hook so it doesn't require
extra trampoline to handle original instructions as it does in full
dynamic tracing.

Closes: #1238

[1] https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fpatchable-function-entry

### DIFF
--- a/arch/aarch64/mcount-dynamic.c
+++ b/arch/aarch64/mcount-dynamic.c
@@ -16,6 +16,8 @@
 
 #define CODE_SIZE  8
 
+static const unsigned int patchable_nop_patt[] = { 0xd503201f, 0xd503201f };
+
 static void save_orig_code(struct mcount_disasm_info *info)
 {
 	uint32_t jmp_insn[6] = {
@@ -175,7 +177,6 @@ static int patch_code(struct mcount_dynamic_info *mdi, struct sym *sym)
 
 static int patch_patchable_func(struct mcount_dynamic_info *mdi, struct sym *sym)
 {
-	unsigned int patchable_nop_patt[] = { 0xd503201f, 0xd503201f };
 	void *insn = (void *)sym->addr + mdi->map->start;
 
 	/* only support calls to 2 NOPs at the beginning */
--- a/arch/aarch64/mcount-dynamic.c
+++ b/arch/aarch64/mcount-dynamic.c
@@ -38,6 +38,7 @@ static void save_orig_code(struct mcount_disasm_info *info)
 int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 {
 	uintptr_t dentry_addr = (uintptr_t)(void *)&__dentry__;
+	uintptr_t fentry_addr = (uintptr_t)(void *)&__fentry__;
 	/*
 	 * trampoline assumes {x29,x30} was pushed but x29 was not updated.
 	 * make sure stack is 8-byte aligned.
@@ -49,6 +50,11 @@ int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 		dentry_addr,
 		dentry_addr >> 32,
 	};
+
+	if (mdi->type == DYNAMIC_PATCHABLE) {
+		trampoline[3] = fentry_addr;
+		trampoline[4] = fentry_addr >> 32;
+	}
 
 	/* find unused 16-byte at the end of the code segment */
 	mdi->trampoline  = ALIGN(mdi->text_addr + mdi->text_size, PAGE_SIZE);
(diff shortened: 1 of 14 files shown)


honggyukim:
@namhyung It's a simple implementation, but I would like to make sure if it's okay to reuse `__fentry__`.

it also requires some updates as follows.
1. add a test
2. update document
3. support aarch64 (it doesn't have `__fentry__` though)

arch/x86_64/mcount-dynamic.c
```diff
@@ -190,6 +190,7 @@ void mcount_arch_find_module(struct mcount_dynamic_info *mdi,
 	struct arch_dynamic_info *adi;
 	unsigned char fentry_nop_patt1[] = { 0x67, 0x0f, 0x1f, 0x04, 0x00 };
 	unsigned char fentry_nop_patt2[] = { 0x0f, 0x1f, 0x44, 0x00, 0x00 };
+	unsigned char fpatchable_nop_patt[] = { 0x90, 0x90, 0x90, 0x90, 0x90 };
```
namhyung, line 193: How about moving them to file level static const variables?

namhyung (commented):
I think it's ok to reuse the `__fentry__` here.  I'm not sure about AArch64, but we can probably use `__dentry__`.

honggyukim:
It's also interesting `-fpatchable-function-entry` option is supported in i386 and arm 32-bit as well.
https://godbolt.org/z/crarfTTK4

Here shows the oldest (confirmed) supported compiler versions.

|  **arch**  |    **gcc**      | **clang** |
|:---------:|:-----------:|:-------:|
| x86_64  | gcc-8.1 | clang-10 |
| aarch64 | gcc-8.1 | clang-10 |
| i386    | gcc-8.1 | clang-10 |
| arm     | gcc-8.2 | not supported |

honggyukim:
I just updated dynamic tracing section in the document as well.
https://github.com/honggyukim/uftrace/blob/review/fpatchable-function-entry/doc/uftrace-record.md#patchable-function-entry

libmcount/dynamic.c
```diff
@@ -527,8 +527,61 @@ static bool skip_sym(struct sym *sym, struct mcount_dynamic_info *mdi,
...
+	 */
+	for (i = 0; i < mdi->nr_patch_target; i++) {
+		uint64_t rel_addr = patchable_loc[i] - mdi->map->start;
+		struct sym fake_sym = {
+			.addr = rel_addr,
+			.size = UINT_MAX,
+		};
+		struct sym *searched_sym =
+			bsearch(&rel_addr, symtab->sym, symtab->nr_sym,
+				sizeof(*sym), addrfind);
```
namhyung, line 552: Why not use `find_sym`?

libmcount/dynamic.c
```diff
@@ -527,8 +527,61 @@ static bool skip_sym(struct sym *sym, struct mcount_dynamic_info *mdi,
...
+			xasprintf(&sym->name, "<%"PRIx64">", patchable_loc[i]);
+		}
+		else {
+			sym = searched_sym;
+			if (skip_sym(sym, mdi, map))
+				continue;
+		}
+
+		switch (mcount_patch_func(mdi, sym, &disasm, min_size)) {
+			case INSTRUMENT_FAILED:
```
namhyung, line 566: We put case statements as same indent level as switch.

libmcount/dynamic.c
```diff
@@ -527,8 +527,61 @@ static bool skip_sym(struct sym *sym, struct mcount_dynamic_info *mdi,
...
+			.size = UINT_MAX,
+		};
+		struct sym *searched_sym =
+			bsearch(&rel_addr, symtab->sym, symtab->nr_sym,
+				sizeof(*sym), addrfind);
+		struct sym *sym;
+
+		if (searched_sym == NULL) {
+			sym = &fake_sym;
+			xasprintf(&sym->name, "<%"PRIx64">", patchable_loc[i]);
```
namhyung, line 557: Maybe we can use a local buffer to avoid repeatitive alloc and free.

honggyukim (commented):
@namhyung Thanks for the review. I've addressed all your comments.

arch/x86_64/mcount-dynamic.c
honggyukim: It requires another separate change so I made it in https://github.com/namhyung/uftrace/pull/1367/commits/e50b2b459a5d13fe0976cc6188ed1ec17796b751.

arch/aarch64/mcount-dynamic.c
```diff
@@ -76,47 +82,168 @@ int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
...
+		if (sym->type != ST_LOCAL_FUNC && sym->type != ST_GLOBAL_FUNC)
+			continue;
+
+		/* dont' check special functions */
+		if (sym->name[0] == '_')
+			continue;
+
+		/* only support calls to __fentry__ at the beginning */
+		if (!memcmp(code_addr, fpatchable_nop_patt, CODE_SIZE)) {
+			mdi->type = DYNAMIC_PATCHABLE;
```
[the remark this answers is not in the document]
honggyukim: I got a bit confused. The current logic loops over the patchable locations from the section so it won't work without having the section even if we set the type to `DYNAMIC_PATCHABLE` here. I will check more if there is really a case where the section is removed.

So I just removed this logic in both x86_64 and aarch64.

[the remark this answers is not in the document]
honggyukim: I just checked again, but the `__patchable_function_entries` section can be removed by `-Wl,--gc-sections`.
```
       --gc-sections
       --no-gc-sections
           Enable garbage collection of unused input sections.  It is
           ignored on targets that do not support this option.  The default
           behaviour (of not performing this garbage collection) can be
           restored by specifying --no-gc-sections on the command line.
           Note that garbage collection for COFF and PE format targets is
           supported, but the implementation is currently considered to be
           experimental.

           --gc-sections decides which input sections are used by examining
           symbols and relocations.  The section containing the entry symbol
           and all sections containing symbols undefined on the command-line
           will be kept, as will sections containing symbols referenced by
           dynamic objects.  Note that when building shared libraries, the
           linker must assume that any visible symbol is referenced.  Once
           this initial set of sections has been determined, the linker
           recursively marks as used any section referenced by their
           relocations.  See --entry and --undefined.

           This option can be set when doing a partial link (enabled with
           option -r).  In this case the root of symbols kept must be
           explicitly specified either by an --entry or --undefined option
           or by a "ENTRY" command in the linker script.
```
Maybe we better handle this case without saving original instruction, but I will leave it as a future work.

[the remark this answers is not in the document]
honggyukim: I'm not sure but it might be difficult to keep only the `__patchable_function_entries` when `--gc-sections` is enabled. We might need to have a way to apply patching only to the `NOP`s rather than force applying all the full dynamic tracing patching. In this case, we don't have to worry about not having the section.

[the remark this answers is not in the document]
honggyukim: This case is handled in https://github.com/namhyung/uftrace/pull/1367/commits/92ef7a86c5062a8432579ce570ad14e5ebb5ff4e and this PR is updated again.

libmcount/dynamic.c
```diff
@@ -475,46 +475,57 @@ static void release_pattern_list(void)
...
-	bool csu_skip;
-	unsigned i, k;
-	struct sym *sym;
 	/* skip special startup (csu) functions */
 	const char *csu_skip_syms[] = {
 		"_start",
 		"__libc_csu_init",
 		"__libc_csu_fini",
 	};
 	char *soname = get_soname(map->libname);
```
namhyung, line 487: Now, the soname is allocated and free'd for every symbol.

arch/aarch64/mcount-dynamic.c
```diff
@@ -38,6 +38,7 @@ static void save_orig_code(struct mcount_disasm_info *info)
 int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 {
 	uintptr_t dentry_addr = (uintptr_t)(void *)&__dentry__;
+	uintptr_t fentry_addr = (uintptr_t)(void *)&__fentry__;
```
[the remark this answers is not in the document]
namhyung: Maybe some compilers would complain it.  Better to do it here.

libmcount/dynamic.c
```diff
@@ -511,8 +511,63 @@ static bool skip_sym(struct sym *sym, struct mcount_dynamic_info *mdi,
...
+static void patch_patchable_func_matched(struct mcount_dynamic_info *mdi,
+					 struct uftrace_mmap *map)
+{
+	struct symtab *symtab;
+	unsigned long *patchable_loc = mdi->patch_target;
+	unsigned i;
+	struct sym *sym;
+	char namebuf[BUFSIZ];
+	struct sym fake_sym = {
+		.size = UINT_MAX,
```
namhyung, line 540: `.name = namebuf,`

Then you don't need to do it in the loop.

arch/x86_64/mcount-dynamic.c
```diff
@@ -216,6 +219,17 @@ void mcount_arch_find_module(struct mcount_dynamic_info *mdi,
...
 			continue;
 
+		/*
+		 * there might be some chances of not having patchable section
+		 * '__patchable_function_entries' but shows the NOPs pattern.
+		 * this can be treated in the same way.
+		 */
+		if (!memcmp(code_addr, patchable_gcc_nop, CALL_INSN_SIZE) ||
+		    !memcmp(code_addr, patchable_clang_nop, CALL_INSN_SIZE)) {
+			mdi->type = DYNAMIC_PATCHABLE;
```
namhyung, line 229: Maybe we can set it as `DYNAMIC_FENTRY_NOP` and check the NOP pattern differently?

arch/aarch64/mcount-dynamic.c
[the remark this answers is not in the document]
honggyukim: Do you mean we have to remove the externs and put `__fentry__` and others into arch specific source?

But I think `libmcount/dynamic.h` is included in all the arch, so it wouldn't be a problem. Could you explain more about your concern?

arch/x86_64/mcount-dynamic.c
honggyukim: Is there any reason to do this? I actually don't get your point.

libmcount/dynamic.c
honggyukim: right. I changed it again. thanks.

libmcount/dynamic.c
honggyukim: ack

arch/aarch64/mcount-dynamic.c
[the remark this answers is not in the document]
namhyung: I mean having the declaration in the header is fine, but it needs to be done in this commit.

arch/x86_64/mcount-dynamic.c
namhyung: If the section is not there, no need to special case the binary.  I think we can handle it same as fentry + nop.  It should only be handled separately if it comes with the section, right?

libmcount/dynamic.c
```diff
@@ -591,7 +591,17 @@ static void patch_normal_func_matched(struct mcount_dynamic_info *mdi,
...
+	 * In some cases, the __patchable_function_entries section can be
+	 * removed.  For example, -Wl,--gc-sections strips this section.
+	 * In this case, mdi->nr_patch_target becomes 0 so it's not possible to
+	 * find the patchable locations from the section.
+	 *
+	 * However, we can still try again by looping over all the symbols
+	 * available and check if the function begins with the NOP patterns for
+	 * patchable function entry inside patch_normal_func_matched().
+	 */
+	if (mdi->type == DYNAMIC_PATCHABLE && mdi->nr_patch_target > 0)
```
namhyung, line 604: Then we can keep this condition as is.

arch/aarch64/mcount-dynamic.c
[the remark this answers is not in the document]
honggyukim: I mean the change for declaration in the header was already merged before.

> I moved the declaration in libmcount/dynamic.h at 2b76fac.

honggyukim:
@namhyung Sorry for late update. I think I mostly updated it based on your comments now. Please have a look again when you have time. Thanks.

arch/aarch64/mcount-dynamic.c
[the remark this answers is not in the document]
namhyung: Oops, I forgot about it.  Then it's ok. :)

namhyung (approved):
LGTM

PR #1380, opened 2021-11-04
dynamic: Set dynamic min_size as patched code size + 1

honggyukim:
uftrace has to skip the function target if the function size is same as
min_size given.  This is needed in the following case.
```
  00000000000005f5 <foo>:
   5f5:   e9 26 ff ff ff          jmpq   520 <AAA::bar(int)@plt>

  00000000000005fa <AAA::bar(int)>:
   5fa:   48 8b 05 ef 09 20 00    mov    0x2009ef(%rip),%rax        # 200ff0 <_DYNAMIC+0x198>
   601:   89 38                   mov    %edi,(%rax)
   603:   c3                      retq
```
In this case, the call instruction is overwritten at the entry of 'foo',
but the return address for the entry of foo is misinterpreted as the
address 0x5fa, which is the address of 'AAA::bar(int)>' so uftrace gets
confused about the function name and it fails as follows.

We should avoid patching when symbol size is less than or equal to
min_size.

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -596,8 +596,8 @@ int mcount_patch_func(struct mcount_dynamic_info *mdi, struct sym *sym,
 {
 	int result = INSTRUMENT_SKIPPED;
 
-	if (min_size < CALL_INSN_SIZE)
-		min_size = CALL_INSN_SIZE;
+	if (min_size < CALL_INSN_SIZE + 1)
+		min_size = CALL_INSN_SIZE + 1;
 
 	if (sym->size < min_size)
 		return result;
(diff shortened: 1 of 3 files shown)


arch/x86_64/mcount-dynamic.c
```diff
@@ -541,7 +541,7 @@ int mcount_patch_func(struct mcount_dynamic_info *mdi, struct sym *sym,
 	if (min_size < CALL_INSN_SIZE)
 		min_size = CALL_INSN_SIZE;
 
-	if (sym->size < min_size)
+	if (sym->size <= min_size)
```
namhyung, line 544: But this changes what users might want if it's bigger than the call size.  I think it's better to change the above condition like

```c
if (min_size < CALL_INSN_SIZE + 1)
    min_size = CALL_INSN_SIZE + 1;

if (sym->size < min_size)
    return result;
```

honggyukim: Right. it's better to apply dynamic patching in such cases. I just updated the patch. In AArch64, it was already written as `<=` before so switched it back.

namhyung (approved):
LGTM

Issue #1384, opened 2021-11-19
-fpatchable-function-entry compiled binary doesn't patch full functions


honggyukim:
If there is a simple example as follows.
```cpp
$ cat shared_ptr.cc
#include <memory>

int main()
{
        std::shared_ptr<int> p(new int);
}
```
`-pg` compiled binary traces functions as follows.
```
$ g++ -pg -o shared_ptr.pg shared_ptr.cc

$ uftrace --no-libcall shared_ptr.pg
# DURATION     TID     FUNCTION
            [ 53226] | main() {
            [ 53226] |   std::shared_ptr::shared_ptr() {
            [ 53226] |     std::__shared_ptr::__shared_ptr() {
            [ 53226] |       std::__shared_count::__shared_count() {
            [ 53226] |         std::__shared_count::__shared_count() {
            [ 53226] |           std::_Sp_counted_ptr::_Sp_counted_ptr() {
   1.637 us [ 53226] |             std::_Sp_counted_base::_Sp_counted_base();
   6.045 us [ 53226] |           } /* std::_Sp_counted_ptr::_Sp_counted_ptr */
   9.608 us [ 53226] |         } /* std::__shared_count::__shared_count */
  12.838 us [ 53226] |       } /* std::__shared_count::__shared_count */
   1.557 us [ 53226] |       std::__shared_ptr::_M_enable_shared_from_this_with();
  19.431 us [ 53226] |     } /* std::__shared_ptr::__shared_ptr */
  22.753 us [ 53226] |   } /* std::shared_ptr::shared_ptr */
            [ 53226] |   std::shared_ptr::~shared_ptr() {
            [ 53226] |     std::__shared_ptr::~__shared_ptr() {
            [ 53226] |       std::__shared_count::~__shared_count() {
            [ 53226] |         std::_Sp_counted_base::_M_release() {
            [ 53226] |           __gnu_cxx::__exchange_and_add_dispatch() {
   1.573 us [ 53226] |             __gthread_active_p();
   1.564 us [ 53226] |             __gnu_cxx::__exchange_and_add_single();
   8.513 us [ 53226] |           } /* __gnu_cxx::__exchange_and_add_dispatch */
   6.102 us [ 53226] |           std::_Sp_counted_ptr::_M_dispose();
            [ 53226] |           __gnu_cxx::__exchange_and_add_dispatch() {
   1.546 us [ 53226] |             __gthread_active_p();
[...]

PR #1386, opened 2021-11-20
dynamic: Do not skip weak symbol functions for dynamic tracing

honggyukim:
If there is a simple example as follows.
```cpp
  $ cat shared_ptr.cc
  #include <memory>

  int main()
  {
          std::shared_ptr<int> p(new int);
  }
```
We can compile it with -pg and -fpatchable-function-entry=5 as follows.
```
  $ g++ -pg -o shared_ptr.pg shared_ptr.cc
  $ g++ -fpatchable-function-entry=5 -o shared_ptr.fpatchable shared_ptr.cc
```
Tracing result of shared_ptr.fpatchable shows much less functions than
shared_ptr.pg example.
```
  $ uftrace --no-libcall shared_ptr.pg | wc -l
  39

  $ uftrace -P. --no-libcall shared_ptr.fpatchable | wc -l
  11
```
It's because many functions are shown as weak symbols.
```
  $ nm -C shared_ptr.fpatchable | grep "std::shared"
  0000000000400b0e W std::shared_ptr<int>::shared_ptr<int, void>(int*)
  0000000000400b0e W std::shared_ptr<int>::shared_ptr<int, void>(int*)
  0000000000400aee W std::shared_ptr<int>::~shared_ptr()
  0000000000400aee W std::shared_ptr<int>::~shared_ptr()
```
But our logic in skip_sym() skips such weak symbols so this patch makes
dynamic tracing not to skip such weak symbols.

Fixed: #1384

### DIFF
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -491,7 +491,9 @@ static bool skip_sym(struct sym *sym, struct mcount_dynamic_info *mdi,
 			return true;
 	}
 
-	if (sym->type != ST_LOCAL_FUNC && sym->type != ST_GLOBAL_FUNC)
+	if (sym->type != ST_LOCAL_FUNC &&
+	    sym->type != ST_GLOBAL_FUNC &&
+	    sym->type != ST_WEAK_FUNC)
 		return true;
 
 	if (!match_pattern_list(map, soname, sym->name)) {


namhyung (approved):
LGTM

PR #1389, opened 2021-11-28
cmds/records: replace ADDR_NO_RANDOMIZE by its value

ffontaine:
uClibc-ng lacks the definition of ADDR_NO_RANDOMIZE, causing a build
failure. A patch was submitted to upstream uClibc-ng to address this
issue, but in the mean time, use an hardcoded value.

Using a #ifdef ... #endif test doesn't work as this value is defined
through an enum in glibc.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>
[Retrieved from:
https://git.buildroot.net/buildroot/tree/package/uftrace/0001-cmds-records-replace-ADDR_NO_RANDOMIZE-by-its-value.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -2105,7 +2105,7 @@ int do_child_exec(int ready, struct opts *opts,
 
 	if (opts->no_randomize_addr) {
 		/* disable ASLR (Address Space Layout Randomization) */
-		if (personality(ADDR_NO_RANDOMIZE) < 0)
+		if (personality(0x0040000 /* ADDR_NO_RANDOMIZE */) < 0)
 			pr_dbg("disabling ASLR failed\n");
 	}
 


namhyung (commented):
I don't think we need this change since the upstream uClibc already accepted the patch.  Please keep the interim fix in the buildroot in the mean time.

ffontaine:
Indeed, we'll keep the patch in buildroot but I assume that some of your users could use old uclibc-ng toolchains (toolchains on embedded systems can be quite old)

namhyung:
Right, someone may use old toolchains but I didn't get any report yet. :)

honggyukim:
It might be helpful if there is a macro definition for uClibc-ng toolchain, but not sure if it has.

honggyukim:
I'm not sure if `__UCLIBC__` macro is correct, but how about applying this?
```diff
diff --git a/cmds/record.c b/cmds/record.c
index 4738e6da..102f95c2 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -29,6 +29,11 @@
 #include "utils/kernel.h"
 #include "utils/perf.h"
 
+#if defined(__UCLIBC__)
+/* disable randomization of VA space */
+#define ADDR_NO_RANDOMIZE 0x0040000
+#endif
+
 #define SHMEM_NAME_SIZE (64 - (int)sizeof(struct list_head))
 
 struct shmem_list {
```

namhyung:
Looks correct, but I'm not still sure if we need this actually.

https://git.uclibc.org/uClibc/tree/include/features.h#n175

honggyukim:
Thanks for the confirmation. I think there is no reason not to support uClibc-ng if the above `__UCLIBC__` part is the only thing we need to support.

honggyukim:
I saw they keep the local patch to replace `ADDR_NO_RANDOMIZE` to the actual value to avoid a build error.

PR #1394, opened 2021-12-12
uftrace: Add --clock option to set clock source

honggyukim:
This patch adds a new option --clock to specify the clock source for
timestamp read.

The current available clock sources are written as follows.
```
  static const struct {
         const char *name;
         clockid_t clock_id;
  } trace_clocks[] = {
         { "mono",       CLOCK_MONOTONIC         },
         { "mono_raw",   CLOCK_MONOTONIC_RAW     },
         { "boot",       CLOCK_BOOTTIME          },
  };
```


honggyukim:
@namhyung I actually do not fully understand the difference between `CLOCK_MONOTONIC` and `CLOCK_MONOTONIC_RAW`, but is there a reason we use `CLOCK_MONOTONIC` instead of `CLOCK_MONOTONIC_RAW`?
>        CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
>              Similar  to  CLOCK_MONOTONIC,  but provides access to a raw hardware-based time
>              that is not subject to NTP adjustments or the incremental adjustments performed
>              by adjtime(3).
https://coderedirect.com/questions/211222/what-is-the-difference-between-clock-monotonic-clock-monotonic-raw

utils/kernel.c
```diff
@@ -491,7 +494,8 @@ static int reset_tracing_files(void)
 	return 0;
 }
 
-static int __setup_kernel_tracing(struct uftrace_kernel_writer *kernel)
+static int __setup_kernel_tracing(struct uftrace_kernel_writer *kernel,
+                                  struct opts *opts)
```
namhyung, line 498: Can you please add a `clock` field in the `struct uftrace_kernel_writer` and use it instead?  I'd like to have all info to setup the kernel tracing in the struct.

namhyung (commented):
In my understanding, the difference between the two is the NTP support.  Any local clock source can have an error or skew so it needs to be corrected regularly by NTP (or similar).  This can be important if we deal with more than one machines.  Usually we use MONOTONIC as it includes the correction but it also means time can go in a different speed sometimes. 

MONOTONIC_RAW might be more precise (in some sense) when it calculates time delta in a single machine (and no need to correlate it with data from other machines).

honggyukim:
I've just updated based on the comments.

> MONOTONIC_RAW might be more precise (in some sense) when it calculates time delta in a single machine (and no need to correlate it with data from other machines).

If so, how about changing the default clock source to `CLOCK_MONOTONIC_RAW`?

namhyung:
I still think the monotonic clock should be default.  I believe there's not much difference in most cases.

honggyukim:
> I believe there's not much difference in most cases.

If there's not much difference, then how about using it? I internally have a strong reason to use `CLOCK_MONOTONIC_RAW` by default. Otherwise, I always have to ask users to add `--clock=mono_raw`.

namhyung:
Maybe we can support some config or env variable to change the default.  But I think the sane default should be plain monotonic.

namhyung (approved):
LGTM

honggyukim:
> Maybe we can support some config or env variable to change the default. But I think the sane default should be plain monotonic.

I'm okay with having a build config to change the default.

> I think the sane default should be plain monotonic.

But could you please explain more about the reason? I also think `CLOCK_MONOTONIC` is useful when we collect trace record from multiple machines as you mentioned as follows.

> Any local clock source can have an error or skew so it needs to be corrected regularly by NTP (or similar). This can be important if we deal with more than one machines.

I think `CLOCK_MONOTONIC` can be used by default when we expand uftrace to collect trace data from multiple remote machines. But for now, we don't have this feature yet.

namhyung:
Even not considering multi-machine sync, I think `CLOCK_MONOTONIC` is correct.  The NTP corrects time progress due to hardware errors so that it can keep accurate time.

Also I think most of other time keeping system uses the monotonic clock as default.  So I believe it's more compatible option than `CLOCK_MONOTONIC_RAW`, but I could be wrong.

honggyukim:
OK. I won't insist more. I will use this option to change the clock source.

We can merge this as is or if you prefer to change the option name same as perf, then we can also change it to `--clockid`. https://man7.org/linux/man-pages/man1/perf-record.1.html

namhyung:
I think `--clock` is just fine.

Issue #1378, opened 2021-11-03
source location filter


ekyooo:
This is issue to discuss the details of the request below.

> (source) location filter 

listed at https://github.com/namhyung/uftrace/wiki/TODO#9-filter-improvement

For now, I've only selected the available options for testing. 
-Y option is to set filter to trace selected source file.
-y option is to set filter not to trace selected source file.

Before:
--srcline option is used to show source location to you.

```
$ uftrace replay --srcline
# DURATION     TID     FUNCTION [SOURCE]
            [ 11746] | main() { /* /home/soft/uftrace/tests/s-libmain.c:16 */
            [ 11746] |   foo() { /* /home/soft/uftrace/tests/s-libmain.c:11 */
            [ 11746] |     lib_a() {
            [ 11746] |       lib_a() { /* /home/soft/uftrace/tests/s-lib.c:10 */
            [ 11746] |         lib_a() { /* /home/soft/uftrace/tests/s-lib.c:10 */
            [ 11746] |           lib_b() { /* /home/soft/uftrace/tests/s-lib.c:15 */
            [ 11746] |             lib_b() { /* /home/soft/uftrace/tests/s-lib.c:15 */
            [ 11746] |               lib_c() { /* /home/soft/uftrace/tests/s-lib.c:20 */
   1.963 us [ 11746] |                 lib_c(); /* /home/soft/uftrace/tests/s-lib.c:20 */
   3.590 us [ 11746] |               } /* lib_c at /home/soft/uftrace/tests/s-lib.c:20 */
   4.091 us [ 11746] |             } /* lib_b at /home/soft/uftrace/tests/s-lib.c:15 */
   4.584 us [ 11746] |           } /* lib_b at /home/soft/uftrace/tests/s-lib.c:15 */
```

After:

set filter to trace requested source file.
```
$ uftrace replay --srcline -Y /home/soft/uftrace/tests/s-libmain.c
# DURATION     TID     FUNCTION [SOURCE]
            [ 29804] | main() { /* /home/soft/uftrace/tests/s-libmain.c:16 */
            [ 29804] |   foo() { /* /home/soft/uftrace/tests/s-libmain.c:11 */
```

set filter not to trace requested source file.
```
$ uftrace replay --srcline -y /home/soft/uftrace/tests/s-libmain.c
# DURATION     TID     FUNCTION [SOURCE]
            [ 11746] | lib_a() {
[...]

PR #1395, opened 2021-12-13
filter: Add -L/--loc-filter option to filter by source location 

ekyooo:
The -O option supports a filter that traces or does not trace the selected
source location with the debug_info loaded.

These are usage example.

The '--srcline' option is required when using record command.
```
  $ uftrace record --srcline t-lib
```
The '-O' option is required when using replay command to set a filter to
trace selected source files.
```
  $ uftrace replay -O s-libmain.c --srcline
  # DURATION     TID     FUNCTION [SOURCE]
              [  4636] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
     4.119 us [  4636] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
     4.544 us [  4636] | } /* main at /home/uftrace/tests/s-libmain.c:16 */
```

You can set a filter to not trace selected source files with the '@hide' suffix.
```
  $ uftrace replay -O s-libmain.c@hide
              [  5327] | lib_a() {
              [  5327] |   lib_b() {
     1.226 us [  5327] |     lib_c();
     2.419 us [  5327] |   } /* lib_b */
     2.789 us [  5327] | } /* lib_a */
```
Fixed: #1378

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -71,7 +71,7 @@ static bool can_use_fast_libmcount(struct uftrace_opts *opts)
 	if (getenv("UFTRACE_FILTER") || getenv("UFTRACE_TRIGGER") || getenv("UFTRACE_ARGUMENT") ||
 	    getenv("UFTRACE_RETVAL") || getenv("UFTRACE_PATCH") || getenv("UFTRACE_SCRIPT") ||
 	    getenv("UFTRACE_AUTO_ARGS") || getenv("UFTRACE_WATCH") || getenv("UFTRACE_CALLER") ||
-	    getenv("UFTRACE_SIGNAL") || getenv("UFTRACE_AGENT"))
+	    getenv("UFTRACE_SIGNAL") || getenv("UFTRACE_AGENT") || getenv("UFTRACE_LOCATION"))
 		return false;
 	return true;
 }
@@ -172,6 +172,16 @@ static void setup_child_environ(struct uftrace_opts *opts, int argc, char *argv[
 		if (filter_str) {
 			setenv("UFTRACE_FILTER", filter_str, 1);
 			free(filter_str);
+		}
+	}
+
+	if (opts->loc_filter) {
+		char *loc_str = uftrace_clear_kernel(opts->loc_filter);
+
+		if (loc_str) {
+			setenv("UFTRACE_LOCATION", loc_str, 1);
+			setenv("UFTRACE_SRCLINE", "1", 1);
+			free(loc_str);
 		}
 	}
 
--- a/doc/uftrace-dump.md
+++ b/doc/uftrace-dump.md
@@ -78,6 +78,10 @@ COMMON OPTIONS
     explicitly have the 'trace' trigger applied, those are always traced
     regardless of execution time.
 
+-L *LOCATION*, \--loc-filter=*LOCATION*
+:   Set filter to trace selected source locations.
+    This option can be used more than once.  See *FILTERS*.
+
 \--no-libcall
 :   Do not show library calls.
 
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -146,6 +146,10 @@ COMMON OPTIONS
     explicitly have the 'trace' trigger applied, those are always traced
     regardless of execution time.  See *FILTERS*.
 
+-L *LOCATION*, \--loc-filter=*LOCATION*
+:   Set filter to trace selected source locations.
+    This option can be used more than once. Implies --srcline. See *FILTERS*.
+
 \--no-libcall
 :   Do not record library function invocations.  Library calls are normally
     traced by hooking calls to the resolver function of dynamic linker in the PLT.
@@ -367,6 +371,25 @@ example will show all user functions and the (kernel) page fault handler.
       3.340 us [14721] |   } /* a */
      79.086 us [14721] | } /* main */
 
+In addition, you can set filter to record selected source locations with `-L` option.
+
+  $ uftrace record -L s-libmain.c t-lib
+  $ uftrace replay --srcline
+  # DURATION     TID     FUNCTION [SOURCE]
+              [  5043] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
+     6.998 us [  5043] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
+     9.393 us [  5043] | } /* main */
+
+You can set filter with the `@hide` suffix not to record selected source locations.
+
+  $ uftrace record -L s-libmain.c@hide t-lib
+  $ uftrace replay --srcline
+  # DURATION     TID     FUNCTION [SOURCE]
+              [ 14688] | lib_a() { /* /home/uftrace/tests/s-lib.c:10 */
+              [ 14688] |   lib_b() { /* /home/uftrace/tests/s-lib.c:15 */
+     1.505 us [ 14688] |     lib_c(); /* /home/uftrace/tests/s-lib.c:20 */
+     2.816 us [ 14688] |   } /* lib_b */
+     3.181 us [ 14688] | } /* lib_a */
 
 TRIGGERS
 ========
--- a/doc/uftrace-replay.md
+++ b/doc/uftrace-replay.md
@@ -48,6 +48,9 @@ REPLAY OPTIONS
 \--libname
 :   Show libname name along with function name.
 
+\--srcline
+:   Show source location of each function if available.
+
 \--format=*TYPE*
 :   Show format style output. Currently, normal and html styles are supported.
 
@@ -80,6 +83,10 @@ COMMON OPTIONS
 :   Do not show functions which run under the time threshold.  If some functions
     explicitly have the 'trace' trigger applied, those are always traced
     regardless of execution time.  See *FILTERS*.
+
+-L *LOCATION*, \--loc-filter=*LOCATION*
+:   Set filter to trace selected source locations.
+    This option can be used more than once.  See *FILTERS*.
 
 \--no-libcall
 :   Do not show library calls.
@@ -313,6 +320,26 @@ as well as DURATION and TID.
       42.124 us   0.220 us [ 6126] |   fgets();
       42.529 us            [ 6126] |   get_values_from() {
       42.645 us   0.236 us [ 6126] |     strdup();
+
+In addition, you can set filter to trace selected source locations with `-L` option.
+For this option, the `--srcline` option is required when using record command.
+
+    $ uftrace record --srcline t-lib
+    $ uftrace replay --srcline -L s-libmain.c
+    # DURATION     TID     FUNCTION [SOURCE]
+                [  5043] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
+       6.998 us [  5043] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
+       9.393 us [  5043] | } /* main */
+
+You can set filter with the `@hide` suffix not to trace selected source locations.
+
+    $ uftrace replay -L libmain*@hide
+    # DURATION     TID     FUNCTION
+                [   866] | lib_a() {
+                [   866] |   lib_b() {
+       1.576 us [   866] |     lib_c();
+       2.833 us [   866] |   } /* lib_b */
+       3.132 us [   866] | } /* lib_a */
 
 You can also set triggers on filtered functions.  See *TRIGGERS* section below
 for details.
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -79,6 +79,9 @@ static bool __maybe_unused mcount_enabled = true;
 /* function filtering mode - inclusive or exclusive */
 static enum filter_mode __maybe_unused mcount_filter_mode = FILTER_MODE_NONE;
 
+/* location filtering mode - inclusive or exclusive */
+static enum filter_mode __maybe_unused mcount_loc_mode = FILTER_MODE_NONE;
+
 /* tree of trigger actions */
 static struct rb_root __maybe_unused mcount_triggers = RB_ROOT;
 
@@ -366,6 +369,7 @@ static void mcount_filter_init(enum uftrace_pattern_type ptype, bool force)
 	char *retval_str = getenv("UFTRACE_RETVAL");
 	char *autoargs_str = getenv("UFTRACE_AUTO_ARGS");
 	char *caller_str = getenv("UFTRACE_CALLER");
+	char *loc_str = getenv("UFTRACE_LOCATION");
 
 	struct uftrace_filter_setting filter_setting = {
 		.ptype = ptype,
@@ -403,6 +407,11 @@ static void mcount_filter_init(enum uftrace_pattern_type ptype, bool force)
 			      &filter_setting);
 	uftrace_setup_argument(argument_str, &mcount_sym_info, &mcount_triggers, &filter_setting);
 	uftrace_setup_retval(retval_str, &mcount_sym_info, &mcount_triggers, &filter_setting);
+
+	if (needs_debug_info) {
+		uftrace_setup_loc_filter(loc_str, &mcount_sym_info, &mcount_triggers,
+					 &mcount_loc_mode, &filter_setting);
+	}
 
 	if (caller_str) {
 		uftrace_setup_caller_filter(caller_str, &mcount_sym_info, &mcount_triggers,
@@ -869,6 +878,15 @@ enum filter_result mcount_entry_filter_check(struct mcount_thread_data *mtdp, un
 	else {
 		/* not matched by filter */
 		if (mcount_filter_mode == FILTER_MODE_IN && mtdp->filter.in_count == 0)
+			return FILTER_OUT;
+	}
+
+	if (tr->flags & TRIGGER_FL_LOC) {
+		if (tr->lmode == FILTER_MODE_OUT)
+			return FILTER_OUT;
+	}
+	else {
+		if (mcount_loc_mode == FILTER_MODE_IN)
 			return FILTER_OUT;
 	}
 
--- a/uftrace.c
+++ b/uftrace.c
@@ -18,6 +18,7 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
+#include <dirent.h>
 #include <fcntl.h>
 #include <getopt.h>
 #include <stdio.h>
@@ -102,6 +103,8 @@ enum uftrace_short_options {
 	OPT_usage,
 	OPT_libmcount_path,
 	OPT_mermaid,
+	OPT_library_path,
+	OPT_loc_filter,
 };
 
 /* clang-format off */
@@ -238,7 +241,7 @@ __used static const char uftrace_footer[] =
 "\n";
 
 static const char uftrace_shopts[] =
-	"+aA:b:C:d:D:eE:f:F:ghH:kK:lN:p:P:r:R:s:S:t:T:U:vVW:Z:";
+	"+aA:b:C:d:D:eE:f:F:ghH:kK:lL:N:p:P:r:R:s:S:t:T:U:vVW:Z:";
 
 #define REQ_ARG(name, shopt) { #name, required_argument, 0, shopt }
 #define NO_ARG(name, shopt)  { #name, no_argument, 0, shopt }
@@ -328,6 +331,8 @@ static const struct option uftrace_options[] = {
 	REQ_ARG(signal, OPT_signal),
 	NO_ARG(srcline, OPT_srcline),
 	REQ_ARG(hide, 'H'),
+	REQ_ARG(loc-filter, OPT_loc_filter),
+	REQ_ARG(loc-filter-warning, 'L'), /* the long option is dummy, will change later */
 	REQ_ARG(clock, OPT_clock),
 	NO_ARG(help, 'h'),
 	NO_ARG(usage, OPT_usage),
@@ -547,8 +552,33 @@ static char *remove_trailing_slash(char *path)
 	return path;
 }
 
+static bool is_libmcount_directory(const char *path)
+{
+	DIR *dp = NULL;
+	struct dirent *ent;
+	int ret = false;
+
+	dp = opendir(path);
+	if (dp == NULL)
+		return false;
+
+	while ((ent = readdir(dp)) != NULL) {
+		if ((ent->d_type == DT_DIR && !strcmp(ent->d_name, "libmcount")) ||
+		    ((ent->d_type == DT_LNK || ent->d_type == DT_REG) &&
+		     !strcmp(ent->d_name, "libmcount.so"))) {
+			ret = true;
+			break;
+		}
+	}
+
+	closedir(dp);
+	return ret;
+}
+
 static int parse_option(struct uftrace_opts *opts, int key, char *arg)
 {
+	char *pos;
+
 	switch (key) {
 	case 'F':
 		opts->filter = opt_add_string(opts->filter, arg);
@@ -581,6 +611,25 @@ static int parse_option(struct uftrace_opts *opts, int key, char *arg)
 
 	case 'H':
 		opts->hide = opt_add_string(opts->hide, arg);
+		break;
+
+	case 'L':
+		if (is_libmcount_directory(arg))
+			pr_warn("--libmcount-path option should be used to set libmcount path.\n");
+		/* fall through */
+	case OPT_loc_filter:
+		pos = strstr(arg, "@hide");
+		if (!pos)
+			opts->loc_filter = opt_add_string(opts->loc_filter, arg);
+		else {
+			*pos = '\0';
+			opts->loc_filter = opt_add_prefix_string(opts->loc_filter, "!", arg);
+		}
+		/*
+		 * location filter focuses onto a given location,
+		 * displaying sched event with it is annoying.
+		 */
+		opts->no_sched = true;
 		break;
 
 	case 'v':
@@ -1246,6 +1295,7 @@ static void free_opts(struct uftrace_opts *opts)
 	free(opts->caller);
 	free(opts->watch);
 	free(opts->hide);
+	free(opts->loc_filter);
 	free_parsed_cmdline(opts->run_cmd);
 }
 
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -365,6 +365,43 @@ struct {
 	{ PATT_GLOB, "glob" },
 };
 
+void init_locfilter_pattern(enum uftrace_pattern_type type, struct uftrace_pattern *p, char *str)
+{
+	char *converted;
+
+	if (strpbrk(str, REGEX_CHARS) == NULL) {
+		/* remove trailing '/' */
+		if (str[strlen(str) - 1] == '/')
+			str[strlen(str) - 1] = '\0';
+
+		/* remove preceding '/' */
+		if (str[0] == '/')
+			str += 1;
+
+		xasprintf(&converted, "%s%s%s", "((.*/)*)", str, "($|(/.*))");
+
+		p->type = PATT_REGEX;
+		p->patt = converted;
+	}
+	else {
+		/* remaining PATT_REGEX and PATT_GLOB cases */
+		p->type = type;
+		p->patt = xstrdup(str);
+	}
+
+	if (p->type == PATT_REGEX) {
+		/* to handle full demangled operator new and delete specially */
+		const char *str_operator = "operator ";
+		if (!strncmp(p->patt, str_operator, 9)) {
+			p->type = PATT_SIMPLE;
+		}
+		else if (regcomp(&p->re, p->patt, REG_NOSUB | REG_EXTENDED)) {
+			pr_dbg("regex pattern failed: %s\n", p->patt);
+			p->type = PATT_SIMPLE;
+		}
+	}
+}
+
 void init_filter_pattern(enum uftrace_pattern_type type, struct uftrace_pattern *p, char *str)
 {
 	if (strpbrk(str, REGEX_CHARS) == NULL)
@@ -398,6 +435,19 @@ bool match_filter_pattern(struct uftrace_pattern *p, char *name)
 	default:
 		return false;
 	}
+}
+
+bool match_location_filter(struct uftrace_pattern *p, struct uftrace_dbg_info *dinfo,
+			   size_t loc_idx)
+{
+	char *loc;
+
+	if (!dinfo || loc_idx >= dinfo->nr_locs || !dinfo->locs[loc_idx].file)
+		return false;
+
+	loc = dinfo->locs[loc_idx].file->name;
+
+	return match_filter_pattern(p, loc);
 }
 
 void free_filter_pattern(struct uftrace_pattern *p)
@@ -790,22 +840,12 @@ static int add_trigger_entry(struct rb_root *root, struct uftrace_pattern *patt,
 	struct uftrace_symbol *sym;
 	size_t i;
 	int ret = 0;
-	char *pos;
 
 	for (i = 0; i < symtab->nr_sym; i++) {
 		sym = &symtab->sym[i];
 
 		if (tr->flags == TRIGGER_FL_LOC) {
-			if (!dinfo || i >= dinfo->nr_locs || !dinfo->locs[i].file)
-				continue;
-
-			if (patt->type == PATT_SIMPLE && strchr(patt->patt, '/') == NULL &&
-			    (pos = strrchr(dinfo->locs[i].file->name, '/')))
-				pos = pos + 1;
-			else
-				pos = dinfo->locs[i].file->name;
-
-			if (!match_filter_pattern(patt, pos))
+			if (!match_location_filter(patt, dinfo, i))
 				continue;
 		}
 		else {
@@ -881,7 +921,11 @@ static void setup_trigger(char *filter_str, struct uftrace_sym_info *sinfo, stru
 
 		/* use demangled name for triggers (some auto-args need it) */
 		name = demangle(name);
-		init_filter_pattern(setting->ptype, &patt, name);
+
+		if (flags & TRIGGER_FL_LOC)
+			init_locfilter_pattern(setting->ptype, &patt, name);
+		else
+			init_filter_pattern(setting->ptype, &patt, name);
 		free(name);
 
 		if (module) {
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -1688,4 +1688,542 @@ TEST_CASE(trigger_setup_args)
 	return TEST_OK;
 }
 
+static struct uftrace_mmap *locfilter_test_load_mmap()
+{
+	static struct uftrace_symbol syms[] = {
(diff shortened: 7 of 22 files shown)


uftrace.c
```diff
@@ -303,6 +303,7 @@ static const struct option uftrace_options[] = {
 	REQ_ARG(signal, OPT_signal),
 	NO_ARG(srcline, OPT_srcline),
 	REQ_ARG(hide, 'H'),
+	REQ_ARG(loc-filter, 'O'),
```
ekyooo, line 306: This is not an agreed option.
Is it ok or not?

uftrace.c
```diff
@@ -558,6 +561,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		opts->hide = opt_add_string(opts->hide, arg);
 		break;
 
+	case 'O':
+		pos = strchr(arg, '@');
```
ekyooo, line 565: Do you want me to implement '@hide' suffix with trigger instead of this?

utils/filter.h
```diff
@@ -143,8 +145,11 @@ void uftrace_setup_caller_filter(char *filter_str, struct symtabs *symtabs,
 				 struct rb_root *root,
 				 struct uftrace_filter_setting *setting);
 void uftrace_setup_hide_filter(char *filter_str, struct symtabs *symtabs,
-				 struct rb_root *root,
-				 struct uftrace_filter_setting *setting);
+			       struct rb_root *root,
+			       struct uftrace_filter_setting *setting);
+void uftrace_setup_loc_filter(char *filter_str, struct symtabs *symtabs,
+			      struct rb_root *root, enum filter_mode *mode,
```
ekyooo, line 151: Sorry, I'll recheck the indentation of this block.

ekyooo (commented):
Could you help me to find test cases those might fail with this patch?
All comments are welcome and I would appreciate your feedback on any logical flaws. 
Thank you.

uftrace.c
namhyung: It'd be nice if we can use 'L' instead..

uftrace.c
namhyung: Yeah, I think that's the way to go.  But we could have this first and then convert it later..

utils/filter.h
namhyung: Thanks!

uftrace.c
ekyooo: I agree. This function is more intuitive with the 'L' option.
As you know, the L option is already in use for the library path, but mostly for manual testing purposes.

What do you think to exclude 'L' option and to keep just '--library-path' option for setting the library path?
Or, do you have any other suggestion?

Before:
```
-L PATH, --library-path=PATH : Load necessary internal libraries from this path. This is mostly for testing purposes.
```

After:
```
--library-path=PATH : Load necessary internal libraries from this path. This is mostly for testing purposes.
```

honggyukim: I also prefer to use `-L` option for this feature, but we should drop it from `--library-path` as you mentioned.

utils/fstack.c
```diff
@@ -804,7 +852,13 @@ static int fstack_check_skip(struct uftrace_task_reader *task,
 
 		depth = task->h->depth;
 	}
-	else if (fstack_filter_mode == FILTER_MODE_IN &&
+	else if (tr.flags & TRIGGER_FL_LOC) {
+		if (tr.fmode == FILTER_MODE_OUT) {
+			return -1;
+		}
```
honggyukim, line 858: We don't need curly-brace here.

doc/uftrace-replay.md
```diff
@@ -48,9 +48,15 @@ REPLAY OPTIONS
...
 
+\--srcline
+:   Show source location of each function if available.
+
 \--format=*TYPE*
 :   Show format style output. Currently, normal and html styles are supported.
 
+-O *LOCATION*, \--loc-filter=*LOCATION*
+:   Set filter to trace selected source locations.
+    This option can be used more than once.  See *FILTERS*.
```
honggyukim, line 59: Could you add the same description to live, report, graph, dump, script, and tui as well?

honggyukim (commented):
Thanks very much for doing this work! I haven't gone through the details but just tested this PR with more programs and I see this feature works fine. I have a few more requests.
- It'd be better to make the same feature work in record command.
- maybe better to disable `linux:sched` event when `--loc-filter` option is used.

utils/fstack.c
ekyooo: It's fixed. thank you.

ekyooo:
@honggyukim 
Thank you for your tests.
For your this comment, could you let me know how to test if available?

> maybe better to disable linux:sched event when --loc-filter option is used.

honggyukim:
> maybe better to disable linux:sched event when --loc-filter option is used.

@ekyooo You can refer to this commit at 55ab685b8e05deb6d3dddcd565a7fc8deee3bc0d.

uftrace.c
```diff
@@ -555,6 +558,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		opts->hide = opt_add_string(opts->hide, arg);
 		break;
 
+	case 'L':
+		pos = strchr(arg, '@');
+		if (!pos)
+			opts->loc_filter = opt_add_string(opts->loc_filter, arg);
```
namhyung, line 564: It'd be nice if we could add a warning for the invalid use - like for library path by old users.

I think we can check if the given path is a directory and has 'libmcount' subdirectory or 'libmcount.so' file.

cmds/record.c
```diff
@@ -177,6 +178,15 @@ static void setup_child_environ(struct opts *opts, int argc, char *argv[])
...
 	}
 
+	if (opts->loc_filter) {
+		char *loc_str = uftrace_clear_kernel(opts->loc_filter);
+
+		if (loc_str) {
+			setenv("UFTRACE_LOCATION", loc_str, 1);
+			free(loc_str);
+		}
+	}
```
namhyung, line 188: What if user gave `-L` without `--srcline`?  Maybe we could set `--srcline` implicitly for record.

namhyung (commented):
Thanks for doing this, mostly looks good!

uftrace.c
```diff
@@ -217,13 +218,13 @@ __used static const char uftrace_help[] =
...
 static const char uftrace_shopts[] =
-	"+aA:b:C:d:D:eE:f:F:hH:kK:lL:N:P:r:R:s:S:t:T:U:vVW:Z:";
+	"+aA:b:C:d:D:eE:f:F:hH:kK:lL:N:P:r:R:s:S:t:T:U:vVW:Z";
 
 #define REQ_ARG(name, shopt) { #name, required_argument, 0, shopt }
 #define NO_ARG(name, shopt)  { #name, no_argument, 0, shopt }
 
 static const struct option uftrace_options[] = {
-	REQ_ARG(library-path, 'L'),
+	REQ_ARG(library-path, OPT_library_path),
```
honggyukim, line 227: I believe not many people directly use `--library-path` option for now, so I would also like to change this option to `--libmcount-path` to be more meaningful. Users might not understand what the library path is.

uftrace.c
```diff
@@ -558,6 +558,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		opts->hide = opt_add_string(opts->hide, arg);
 		break;
 
+	case 'L':
+		pos = strchr(arg, '@');
+		if (!pos)
+			opts->loc_filter = opt_add_string(opts->loc_filter, arg);
+		else {
+			pos = '\0';
```
honggyukim, line 566: This shows a warning in clang compiler.
```
/home/honggyu/uftrace/uftrace.c:566:10: warning: expression which evaluates to zero treated as a null pointer constant of type 'char *' [-Wnon-literal-null-conversion]
                        pos = '\0';
                              ^~~~
1 warning generated.
```
Maybe do you want `*pos = '\0'`?

uftrace.c
```diff
@@ -558,6 +558,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		opts->hide = opt_add_string(opts->hide, arg);
 		break;
 
+	case 'L':
```
honggyukim, line 561: You can add `opts->no_sched = true` to suppress showing too many scheduling events just like https://github.com/namhyung/uftrace/commit/55ab685b8e05deb6d3dddcd565a7fc8deee3bc0d did.

cmds/record.c
honggyukim: @ekyooo You would need `setenv("UFTRACE_SRCLINE", "1", 1)` for this.

Please also specify this in `record` man page, but also specify in `live` page that `--srcline` is applied only for record not for replay.
```
Implies --srcline.
```
You can refer to https://github.com/namhyung/uftrace/blob/master/doc/uftrace-record.md#record-options.

honggyukim:
And please modify the commit messages about source line info at function return.
```
  $ uftrace replay -O s-libmain.c --srcline
  # DURATION     TID     FUNCTION [SOURCE]
              [  4636] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
     4.119 us [  4636] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
     4.544 us [  4636] | } /* main at /home/uftrace/tests/s-libmain.c:16 */
```
This has to be changed based on your recent patch https://github.com/namhyung/uftrace/commit/e088a3f4dc7588d78608edd95b63f0627934fb37.
```
  $ uftrace replay -O s-libmain.c --srcline
  # DURATION     TID     FUNCTION [SOURCE]
              [  4636] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
     4.119 us [  4636] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
     4.544 us [  4636] | } /* main */
```

uftrace.c
```diff
@@ -217,13 +218,13 @@ __used static const char uftrace_help[] =
 "\n";
 
 static const char uftrace_shopts[] =
-	"+aA:b:C:d:D:eE:f:F:hH:kK:lL:N:P:r:R:s:S:t:T:U:vVW:Z:";
+	"+aA:b:C:d:D:eE:f:F:hH:kK:lL:N:P:r:R:s:S:t:T:U:vVW:Z";
```
honggyukim, line 221: `Z:` shouldn't be modified.

uftrace.c
```diff
@@ -558,6 +558,16 @@ static int parse_option(struct opts *opts, int key, char *arg)
 		opts->hide = opt_add_string(opts->hide, arg);
 		break;
 
+	case 'L':
+		pos = strchr(arg, '@');
+		if (!pos)
+			opts->loc_filter = opt_add_string(opts->loc_filter, arg);
+		else {
```
honggyukim, line 565: You should check if the later part of `@` is "hide".

uftrace.c
ekyooo: If the purpose of the option is libmcount only, I agree.
I changed it.

cmds/record.c
ekyooo: It's modified. Thank you.

uftrace.c
ekyooo: I added it. please check the conditions.

ekyooo:
> In addition, I would like to use this feature in `-P`/`--patch` to patch only functions from specific files. The same thing can be applied to `-U`/`--unpatch` .
> 
> It could be used as follows.
> 
> ```
> $ uftrace record -a -L s-libmain.c@patch t-lib
> ```

I'll consider it. Thank you.

doc/uftrace-record.md
```diff
@@ -359,6 +362,25 @@ example will show all user functions and the (kernel) page fault handler.
...
      79.086 us [14721] | } /* main */
 
+In addition, you can set filter to record selected source locations with `-L` option.
+
+  $ uftrace record -L s-libmain.c t-lib
+  $ uftrace replay --srcline
+  # DURATION     TID     FUNCTION [SOURCE]
+              [  5043] | main() { /* /home/uftrace/tests/s-libmain.c:16 */
+     6.998 us [  5043] |   foo(); /* /home/uftrace/tests/s-libmain.c:11 */
+     9.393 us [  5043] | } /* main at /home/uftrace/tests/s-libmain.c:16 */
```
honggyukim, line 372:
```
/* main at /home/uftrace/tests/s-libmain.c:16 */
```
This should be changed to
```
/* main */
```

honggyukim (commented):
> I'll consider it. Thank you.

It's good to have now, but we can support maybe next time regarding `@patch` and `@unpatch` with `-L` option.

honggyukim:
And please rebase this PR again.

doc/uftrace-record.md
namhyung: Right, the return path used to have a different location info than the beginning of the function and it's possible to have multiple return locations.  As we don't know exactly where it returns from, we can just print function name here.

doc/uftrace-dump.md
```diff
@@ -71,6 +71,10 @@ COMMON OPTIONS
     explicitly have the 'trace' trigger applied, those are always traced
     regardless of execution time.
 
+-L *LOCATION*, \--loc-filter=*LOCATION*
+:   Set filter to trace selected source locations.
+    This option can be used more than once.  See *FILTERS*.
```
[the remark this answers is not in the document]
ekyooo: The -L option is currently only supported for record and replay command.
How about keeping the man page changes for record and replay only?

ekyooo (commented):
It's rebased.

> It's good to have now, but we can support maybe next time regarding @patch and @unpatch with -L option.

Unfortunately I don't know much about @patch and @unpatch, so I need to know first.
Thanks for letting me to consider it.

doc/uftrace-dump.md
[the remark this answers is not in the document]
honggyukim: It magically works for all the other commands. Please check it again :)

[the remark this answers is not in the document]
ekyooo: It's real magic. I didn't know that.
I am lucky from the new year.
I fixed it. Thank you.

uftrace.c
ekyooo: I updated it, but it may not be a good way.
Please review it.
Thank you.

uftrace.c
```diff
@@ -235,13 +236,13 @@ __used static const char uftrace_footer[] =
...
+	"+aA:b:C:d:D:eE:f:F:hH:kK:lN:P:r:R:s:S:t:T:U:vVW:Z:";
 
 #define REQ_ARG(name, shopt) { #name, required_argument, 0, shopt }
 #define NO_ARG(name, shopt)  { #name, no_argument, 0, shopt }
 
 static const struct option uftrace_options[] = {
-	REQ_ARG(library-path, 'L'),
+	REQ_ARG(libmcount-path, OPT_libmcount_path),
 	REQ_ARG(filter, 'F'),
 	REQ_ARG(notrace, 'L'),
```
namhyung, line 250: Unrelated, but I think the short option for notrace should 'N'.  Not sure how it worked so far..

libmcount/mcount.c
```diff
@@ -381,6 +384,7 @@ static void mcount_filter_init(enum uftrace_pattern_type ptype, char *dirname,
 	char *retval_str    = getenv("UFTRACE_RETVAL");
 	char *autoargs_str  = getenv("UFTRACE_AUTO_ARGS");
 	char *caller_str    = getenv("UFTRACE_CALLER");
+	char *loc_str       = getenv("UFTRACE_LOCATION");
```
namhyung, line 387: It also needs to set `needs_debug_info`.

namhyung (commented):
Sorry for the late review.  I think it's almost good to merge and we can improve on it like for more actions like patch. Thanks!

uftrace.c
ekyooo: :) It's another magic.. I fixed it in split patch for unexpected results.
Please let me know if you want squash it to first patch.
Thank you.

libmcount/mcount.c
ekyooo: `needs_debug_info` is set when -L is applied in the record.c code below as I understand and tested.
Could you please let me know if I've missed anything?

```
		if (loc_str) {
			setenv("UFTRACE_LOCATION", loc_str, 1);
			setenv("UFTRACE_SRCLINE", "1", 1);
```

uftrace.c
namhyung: Do we keep this `--library-path` for backward compatibility?  Then it should be `REQ_ARG(library-path, OPT_libmcount_path)`, isn't it?

uftrace.c
namhyung: Actually I got an idea.  So we want to check this only if short option `-L` is used, if they used `--library-path` then it should be fine.  Maybe we can separate it from the `--loc-filter` for a while so that existing uses can be warned.  How about adding a dummy (undocumented) option just for warning?

```c
static const struct option uftrace_options[] = {
	REQ_ARG(libmcount-path, OPT_libmcount_path),
	REQ_ARG(library-path, OPT_libmcount_path),
 	...
	REQ_ARG(loc-filter, OPT_loc_filter),
	REQ_ARG(loc-filter-warning, 'L'),  // the long option is dummy, will change later
	...
};

static int parse_option(struct opts *opts, int key, char *arg)
{
	switch (key) {
	case OPT_libmcount_path:
		/* ... */
	case 'L':
		if (is_libmcount_directory(arg))
			pr_warn(...);
		/* fall through */
	case OPT_loc_filter:
		/* ... */
```

utils/filter.h
```diff
@@ -143,8 +145,11 @@ void uftrace_setup_caller_filter(char *filter_str, struct symtabs *symtabs,
 				 struct rb_root *root,
 				 struct uftrace_filter_setting *setting);
 void uftrace_setup_hide_filter(char *filter_str, struct symtabs *symtabs,
-				 struct rb_root *root,
-				 struct uftrace_filter_setting *setting);
+			       struct rb_root *root,
+			       struct uftrace_filter_setting *setting);
```
namhyung, line 149: Unnecessary changes.

namhyung (commented):
Can you please split the filter implementation and its usage in replay in separate commits like you did for record?  The third patch looks too big and contains many changes.

utils/filter.h
ekyooo: no need to align?

honggyukim: That means it's not related to this change so can be fixed later in a separate patch.

libmcount/mcount.c
namhyung: Ah, I missed that.  So it's set by "UFTRACE_SRCLINE".  Thanks for the clarification.

namhyung:
> > Can you please split the filter implementation and its usage in replay in separate commits like you did for record? The third patch looks too big and contains many changes.
> 
> Except for the example usage of uftrace-replay.md, most of the changes are related to filters. Tests using the replay command can be split, but there is no test for filters after splitting. Do you agree? Thank you.

I'm ok having tests in a different commit.  But I think we can have filter changes without users first.  And then add `-L` option in replay command later.

ekyooo:
> > > Can you please split the filter implementation and its usage in replay in separate commits like you did for record? The third patch looks too big and contains many changes.
> > 
> > 
> > Except for the example usage of uftrace-replay.md, most of the changes are related to filters. Tests using the replay command can be split, but there is no test for filters after splitting. Do you agree? Thank you.
> 
> I'm ok having tests in a different commit. But I think we can have filter changes without users first. And then add `-L` option in replay command later.

Do you want me to leave the long option (--loc-filter) on the filter patch for testing? Or do you want to split it into a replay patch and then try to add filter unit tests to the filter patches (if possible)?

namhyung:
The option change should be in the replay patch and I think it's good to have the tests there too.

ekyooo:
It's updated. Please review it.
Thank you.

namhyung (approved):
LGTM, thanks a lot for your work!

namhyung (changes requested):
Oh, I have to take my LGTM back.  I've found a couple of problems during some manual tests.  Basically it needs to handle directories - or should not crash it least.  Also you need to check report output as well.

```
$ uftrace record -a -P . uftrace info -d my.data

# I want to check functions/files in the cmds directory
$ uftrace replay -L cmds
Segmentation fault

# retry with a trailing slash
$ uftrace replay -L cmds/
Usage: failed to set filter or trigger: cmds/

# files are ok
$ uftrace replay -L cmds/*
...
```

But even with that, report output shows some unrelated (and not symbolized) lines
```
$ uftrace report -L cmds/*
  925.676 ms   13.771 us           2  <0>
  915.330 ms   95.720 ms           1  <56173db8f5f0>
  819.609 ms   24.044 us           1  <559812621e95>
  817.904 ms   32.998 us           1  <55981268f685>
  817.680 ms    0.157 us           1  <559812639a45>
  817.680 ms   34.977 ms           1  <559812621600>
    1.207 ms    4.144 us           1  command_info
  558.870 us   11.269 us           1  process_uftrace_info
  123.719 us   10.083 us           1  read_uftrace_info
   26.658 us   12.786 us          28  print_info
   ...
```

uftrace graph and tui commands look fine.

ekyooo:
Oh, Thank you so much for finding that case. 
I found the cause of fault and I'll update it.
Thank you again.

ekyooo:
For the directory case, I'd like to clarify the spec before pushing.
In the current patch, the matching policy is:

- simple(PATT_SIMPLE): exact match except below.
> Basically I think it'd be better matching only with a basename unless there's a slash in the path.
https://github.com/namhyung/uftrace/issues/1378#issuecomment-960465802
- regex: handle it as a regex rule.
Directories can only be represented by regex.
```
uftrace replay -L .*cmds.*
``` 

After the change, the matching policy is:
- simple: partial match 
- regex: same as current

Could you let me know do you agree?
Thank you.

ekyooo:
1. segmentation fault is fixed for "cmds" argument.
2. directory support by additional patch :  https://github.com/namhyung/uftrace/pull/1395/commits/1f41c3aea03f2b7a84b9bd6eeb39efe5f6220625 
3. report output shows some unrelated (and not symbolized) lines: not yet fixed in my test.

Could you please let me know whether the additional patch contains all the requirements for the directory?
Thank you.

utils/filter.c
```diff
@@ -836,6 +838,16 @@ static void setup_trigger(char *filter_str, struct symtabs *symtabs,
 
 		/* use demangled name for triggers (some auto-args need it) */
 		name = demangle(name);
+
+		if (flags & TRIGGER_FL_LOC) {
+			if (!strpbrk(name, REGEX_CHARS) && is_directory(name)) {
+				re_suffix = name[strlen(name) - 1] == '/' ? ".*" : "/.*";
+				xasprintf(&tmp, "%s%s", name, re_suffix);
```
namhyung, line 845: Actually you can just add a slash since kernel can handle double slashes.  Also you need to check pattern type as well - it'd be '*' for glob patterns.

namhyung: Hmm.. we cannot simply check if it's a directory since the directory can be in a different location or does not exist at all.  Note that this is a name of directory at the compile time and we are checking it at runtime.

namhyung:
How about this?

```diff
diff --git a/utils/filter.c b/utils/filter.c
index 2799713a0a..1b8f2bc2a4 100644
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -406,6 +406,41 @@ bool match_filter_pattern(struct uftrace_pattern *p, char *name)
        }
 }
 
+bool match_location_filter(struct uftrace_pattern *p, struct debug_info *dinfo,
+                          int loc_idx)
+{
+       char *loc, *pos;
+
+       if (!dinfo || loc_idx >= dinfo->nr_locs || !dinfo->locs[loc_idx].file)
+               return false;
+
+       loc = dinfo->locs[loc_idx].file->name;
+
+       if (p->type != PATT_SIMPLE || strchr(p->patt, '/'))
+               return match_filter_pattern(p, loc);
+
+       /* try to match with a basename */
+       pos = strrchr(loc, '/');
+       if (pos)
+               loc = pos + 1;
+
+       if (!strcmp(p->patt, loc))
+               return true;
+
+       /* try to match with a directory name */
+       loc = dinfo->locs[loc_idx].file->name;
+       pos = strstr(loc, p->patt);
+       if (pos == NULL)
+               return false;
+
+       if (pos[strlen(p->patt)] != '/')
+               return false;
+       if (pos == loc || pos[-1] == '/')
+               return true;
+
+       return false;
+}
+
 void free_filter_pattern(struct uftrace_pattern *p)
 {
        free(p->patt);
@@ -740,25 +775,12 @@ static int add_trigger_entry(struct rb_root *root,
        struct sym *sym;
        unsigned i;
        int ret = 0;
-       int sym_idx;
-       char *pos;
 
        for (i = 0; i < symtab->nr_sym; i++) {
                sym = &symtab->sym[i];
 
                if (tr->flags == TRIGGER_FL_LOC) {
-                       sym_idx = sym - symtab->sym;
-
-                       if (!dinfo || sym_idx >= dinfo->nr_locs || !dinfo->locs[sym_idx].file)
-                               continue;
-
-                       if (patt->type == PATT_SIMPLE && strchr(patt->patt, '/') == NULL &&
-                           (pos = strrchr(dinfo->locs[sym_idx].file->name, '/')))
-                               pos = pos + 1;
-                       else
-                               pos = dinfo->locs[sym_idx].file->name;
-
-                       if (!match_filter_pattern(patt, pos))
+                       if (!match_location_filter(patt, dinfo, i))
                                continue;
                }
                else {
@@ -787,8 +809,6 @@ static void setup_trigger(char *filter_str, struct symtabs *symtabs,
 {
        struct strv filters = STRV_INIT;
        char *name;
-       char* tmp;
-       char* re_suffix;
        int j;
 
        if (filter_str == NULL)
@@ -839,15 +859,6 @@ static void setup_trigger(char *filter_str, struct symtabs *symtabs,
                /* use demangled name for triggers (some auto-args need it) */
                name = demangle(name);
 
-               if (flags & TRIGGER_FL_LOC) {
-                       if (!strpbrk(name, REGEX_CHARS) && is_directory(name)) {
-                               re_suffix = name[strlen(name) - 1] == '/' ? ".*" : "/.*";
-                               xasprintf(&tmp, "%s%s", name, re_suffix);
-                               free(name);
-                               name = tmp;
-                       }
-               }
-
                init_filter_pattern(setting->ptype, &patt, name);
                free(name);
 
```

namhyung:
I've merged some preparatory changes first.  Please rebase your change!

honggyukim:
> I've merged some preparatory changes first. Please rebase your change!

I've just rebase it and fixed the numbers of test scripts.

ekyooo:
I'm sorry for the late update and thank you for the suggestion. I will try to update this week.
Thank you.

ekyooo:
As a test result with the change, `cmds/*` and `cmds` work, but `cmds/` doesn't work as follows.
```
$ ./uftrace replay -L cmds/*
            [  7823] | } = 0; /* fork */

$ ./uftrace replay -L cmds
            [  7823] | } = 0; /* fork */

$ ./uftrace replay -L cmds/
Usage: failed to set filter or trigger: cmds/
 158.497 us [  7810] | getopt_long(4, 0x7ffc658426f8, "+aA:b:C:d:D:eE:f:F:hH:kK:lL:N:P:r:R:s:S:t:T:U:vVW:Z:") = -1;
```

As I simply checked, it returns `cmds/` in this condition. 
Could you let me know If you still want to support `cmds/` or have changed your mind?
Thank you.
```
 bool match_location_filter(
...
 419        if (p->type != PATT_SIMPLE || strchr(p->patt, '/'))
 420                return match_filter_pattern(p, loc);
```

namhyung:
Ok, let's ignore the `cmds/` case then.

namhyung:
I think you can squash the last commit into the first one.

honggyukim:
> As a test result with the change, cmds/* and cmds work, but cmds/ doesn't work as follows.

We can remove the trailing slash with this.
```diff
diff --git a/uftrace.c b/uftrace.c
index bde482de..791da05e 100644
--- a/uftrace.c
+++ b/uftrace.c
@@ -611,6 +611,14 @@ static int parse_option(struct opts *opts, int key, char *arg)
                        pr_warn("--libmcount-path option should be used to set libmcount path.\n");
                /* fall through */
        case OPT_loc_filter:
+               {
+                       /* remove trailing slash '/' at the end of input. */
+                       char *p = arg;
+                       while (*p++);
+                       p = p - 2;
+                       if (*p == '/')
+                               *p = '\0';;
+               }
                pos = strstr(arg, "@hide");
                if (!pos)
                        opts->loc_filter = opt_add_string(opts->loc_filter, arg);
```
With this change, `-L cmds/` and `-L cmds` are identical. However, we should be careful with `-L cmds/*` and it has to be `-L "cmds/*"` to avoid shell magic. For example, if there is a file `cmds/aaa`, then `-L cmds/*` will be changed to `-L cmds/aaa` by the shell. But we don't have a way to avoid it so let's be aware and leave it as is.

honggyukim:
To make it work better with `@hide`, this should be better.
```diff
diff --git a/uftrace.c b/uftrace.c
index bde482de..10d2a905 100644
--- a/uftrace.c
+++ b/uftrace.c
@@ -612,12 +612,20 @@ static int parse_option(struct opts *opts, int key, char *arg)
                /* fall through */
        case OPT_loc_filter:
                pos = strstr(arg, "@hide");
-               if (!pos)
-                       opts->loc_filter = opt_add_string(opts->loc_filter, arg);
-               else {
+               if (pos)
                        *pos = '\0';
-                       opts->loc_filter = opt_add_prefix_string(opts->loc_filter, "!", arg);
+               {
+                       /* remove trailing slash '/' at the end of input. */
+                       char *p = arg;
+                       while (*p++);
+                       p = p - 2;
+                       if (*p == '/')
+                               *p = '\0';;
                }
+               if (pos)
+                       opts->loc_filter = opt_add_prefix_string(opts->loc_filter, "!", arg);
+               else
+                       opts->loc_filter = opt_add_string(opts->loc_filter, arg);
                /*
                 * location filter focuses onto a given location,
                 * displaying sched event with it is annoying.
```

honggyukim:
@ekyooo And please rebase your patches again.

ekyooo:
I updated dffaaa7738a87ac77383a6c90db89e074ec73b6e.
If you agree with this, I''ll squash the last commit into the first one.
Thank you.

utils/filter.c
```diff
@@ -780,13 +780,34 @@ static int add_trigger_entry(struct rb_root *root,
...
+
+	/*
+	 * convert to a regex string to match all instances of including patt
+	 * in the path. exceptionally, if patt starts at the root, convert only
+	 * the right.
+	 * for example, if patt is 'cmds', it matches 'cmds', 'cmds/', 'a/cmds'
+	 * and 'a/cmds/b'. it matches 'abc.c', 'a/abc.c' if patt is 'abc.c'.
+	 */
+	xasprintf(&ret, "%s%s%s", (patt[0] == '/')? "" : "(^|(.*/))",
+		  patt, "($|(/.*))");
```
namhyung, line 799: We should support `--match=glob` case as well.  Also, shouldn't it start with `^` when it's a absolute pathname?

ekyooo: > when it's a absolute pathname?

To help understand, could you please give me an example of patt and loc for this case?

namhyung: Sorry for the late reply.  For glob matching, "cmds" should be converted to "cmds/*" (without the ".") and I was thinking about "/usr/bin/true" might match with "/opt/sysroot/usr/bin/true".

honggyukim:
@ekyooo Sorry but we have applied global reformatting so your PR currently shows many horrible conflicts. I have rebased your PR at https://github.com/namhyung/uftrace/compare/master...honggyukim:uftrace:clang-format/srcfilter3. So please get the patches and update your PR to make it conflict free again.

And please continue this work when you have time. Thanks!

utils/filter.c
```diff
@@ -1687,4 +1687,542 @@ TEST_CASE(trigger_setup_args)
...
+
+	sinfo->maps = map;
+	sinfo->exec_map = map;
+	sinfo->loaded = true;
+
+	sinfo->maps->next = map2;
+}
+
+/* Simple pattern arguments are treated as regex after conversion. */
+TEST_CASE(locfilter_setup_simple)
```
ekyooo, line 1787: I'll merge this with locfilter_setup_regex if you want.
Thank you.

utils/filter.c
```diff
@@ -1687,4 +1687,542 @@ TEST_CASE(trigger_setup_args)
...
+
+	pr_dbg("found 3 symbols. done\n");
+	uftrace_cleanup_filter(&root);
+	TEST_EQ(RB_EMPTY_ROOT(&root), true);
+
+	return TEST_OK;
+}
+
+/* Simple pattern arguments are treated as regex after conversion. */
+TEST_CASE(locfilter_setup_directory_simple)
```
ekyooo, line 1924: I'll merge this with locfilter_setup_directory_regex if you want.
Thank you.

ekyooo (commented):
Hi. Sorry for the late update.
Could you please review it?
After review, I will squash the last two commits with the previous one as requested.
This is the unit test result.

```
$ ./unittest locfilter
Running 7 test cases
======================
[001] locfilter_match : PASS
[002] locfilter_setup_directory_glob: PASS
[003] locfilter_setup_directory_regex: PASS
[004] locfilter_setup_directory_simple: PASS
[005] locfilter_setup_glob : PASS
[006] locfilter_setup_regex : PASS
[007] locfilter_setup_simple : PASS

unit test stats
====================
7 ran successfully
```
Thank you.

utils/filter.c
namhyung: I think it's good to leave as is.  But I prefer short names like 'locfilter_setup_dir_simple'.

namhyung (commented):
Thanks for working on this!  I just have a few nits.

utils/filter.c
```diff
@@ -780,14 +788,30 @@ static int add_trigger_entry(struct rb_root *root, struct uftrace_pattern *patt,
 	struct uftrace_symtab *symtab = &map->mod->symtab;
 	struct uftrace_dbg_info *dinfo = &map->mod->dinfo;
 	struct uftrace_symbol *sym;
-	unsigned i;
+	size_t i;
```
ekyooo, line 841: It's changed to also use for symtab->nr_sym.

ekyooo (commented):
I fixed as your comments. Thank you for the review.
Is it ok to squash the last two commits into the first one? Or is it too big?

namhyung:
I think it's good now.  Will check the code again later today.

honggyukim:
It's because I've got lots of reports about segfault in library tracing expecially in ARM/AArch64. So always my answer was asking to use `--no-libcall`. I would like to make default uftrace tracing as stable as possible.

@namhyung How do you think about it?

namhyung:
But I think the default library tracing is so valuable on x86.  Otherwise `--force` option would not be useful.  For segfaults on ARM/AArch64, I feel like we should find the root cause and fix the problem eventually instead of disabling it permanently.  The `--nest-libcall` is more dangerous, so better not to provided a short option.

namhyung (approved):
LGTM!!

ekyooo:
Sorry there was a typo, i fixed it.
Thank you.

PR #1396, opened 2021-12-14
python: Following python coding style in automated way with black

kangtegong:
python: Following python coding style in automated way with black

We can maintain python coding style using black[1].
If you install black and run

    $ black {path to pyproject.toml}
we can reformat python code as PEP 8 suggests. (e.g. black .)

In this PR, I included line-length as 100[2].

[1] https://black.readthedocs.io/en/stable/
[2]
    https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=bdc48fa11e46f867ea4d75fa59ee87a7f48be144

### DIFF
--- a/misc/gen-autoargs.py
+++ b/misc/gen-autoargs.py
@@ -22,27 +22,40 @@ argspec_file = "autoargs.h"
 storage_class_specifier = ["auto", "register", "static", "extern", "typedef"]
 type_qualifier = ["const", "volatile"]
 
-type_specifier = ["void", "char", "short", "int", "long", "float", "double", \
-                    "signed", "unsigned" ]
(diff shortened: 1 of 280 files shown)


kangtegong:
The reason why there are two blank lines between functions and classes is because it is [recommended by PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines)

Also, recommended line-length (number of character of each line) is 79, but I thought it is too short for uftrace, so I set to 84.

namhyung:
I'm not sure we really want this.  It generates too many changes for no real reason..

kangtegong:
ok, then I'll close this PR then..!

Issue #1387, opened 2021-11-21
arguments error in -fpatchable-function-entry compiled binaries


honggyukim:
There is a sample program `shared_ptr.cc`.
```cpp
#include <memory>

int main()
{
        std::shared_ptr<int> p(new int);
}
```
We can see the tracing output with `-pg` compiled binary as follows.
```
$ g++ -pg -g -o shared_ptr.pg shared_ptr.cc

$ uftrace --no-libcall -a shared_ptr.pg
# DURATION     TID     FUNCTION
            [ 49523] | main() {
            [ 49523] |   std::shared_ptr::shared_ptr(0x7fffb0edc590, 0xb3c9a0) {
            [ 49523] |     std::__shared_ptr::__shared_ptr(0x7fffb0edc590, 0xb3c9a0) {
            [ 49523] |       std::__shared_count::__shared_count(0x7fffb0edc598, 0xb3c9a0, integral_constant{...}) {
            [ 49523] |         std::__shared_count::__shared_count(0x7fffb0edc598, 0xb3c9a0) {
            [ 49523] |           std::_Sp_counted_ptr::_Sp_counted_ptr(0xb85ec0, 0xb3c9a0) {
   0.130 us [ 49523] |             std::_Sp_counted_base::_Sp_counted_base(0xb85ec0);
   1.068 us [ 49523] |           } /* std::_Sp_counted_ptr::_Sp_counted_ptr */
   9.487 us [ 49523] |         } /* std::__shared_count::__shared_count */
 116.611 us [ 49523] |       } /* std::__shared_count::__shared_count */
   0.146 us [ 49523] |       std::__shared_ptr::_M_enable_shared_from_this_with(0x7fffb0edc590, 0xb3c9a0);
 117.290 us [ 49523] |     } /* std::__shared_ptr::__shared_ptr */
 118.432 us [ 49523] |   } /* std::shared_ptr::shared_ptr */
            [ 49523] |   std::shared_ptr::~shared_ptr(0x7fffb0edc590) {
            [ 49523] |     std::__shared_ptr::~__shared_ptr(0x7fffb0edc590) {
            [ 49523] |       std::__shared_count::~__shared_count(0x7fffb0edc598) {
            [ 49523] |         std::_Sp_counted_base::_M_release(0xb85ec0) {
            [ 49523] |           __gnu_cxx::__exchange_and_add_dispatch() {
   0.078 us [ 49523] |             __gthread_active_p() = 0;
[...]

PR #1398, opened 2022-01-03
dwarf: Fix args reading error in -fpatchable-function-entry binaries

honggyukim:
Let's say there is a simple program 'shared_ptr.cc' as follows.
```cpp
  $ cat shared_ptr.cc
  #include <memory>

  int main()
  {
          std::shared_ptr<int> p(new int);
  }
```
If we compile this program with -fpatchable-function-entry, uftrace
record with arguments gets failed as follows.
```
  $ g++ -fpatchable-function-entry=5 -g -o shared_ptr.fpatchable shared_ptr.cc

  $ uftrace --no-libcall -a -P. shared_ptr.fpatchable
  fstack: /home/honggyu/work/uftrace/utils/fstack.c:1595:read_task_ustack
   ERROR: record missing argument info for std::shared_ptr::shared_ptr
```
This problem happens because dwarf_getsrc_die returns NULL with the
start address of functions and it's assigned to Dwarf_Line *line.
```
  line = dwarf_getsrc_die(&cudie, dwarf_addr);
```
-fpatchable-function-entry=5 adds 5 NOP instructions at the function
prologue and sym->addr has the address of the first NOP instruction.
However, we don't have dwarf info at the first NOP instruction.

For example, we can't find the debug info at the following function.
```asm
  0000000000400ebc <std::_Sp_counted_ptr<int*, (__gnu_cxx::_Lock_policy)2>::_M_get_deleter(std::type_info const&)>:
    400ebc:       90                      nop
    400ebd:       90                      nop
    400ebe:       90                      nop
    400ebf:       90                      nop
    400ec0:       90                      nop
    400ec1:       55                      push   %rbp
    400ec2:       48 89 e5                mov    %rsp,%rbp
    400ec5:       48 89 7d f8             mov    %rdi,-0x8(%rbp)
    400ec9:       48 89 75 f0             mov    %rsi,-0x10(%rbp)
    400ecd:       b8 00 00 00 00          mov    $0x0,%eax
    400ed2:       5d                      pop    %rbp
    400ed3:       c3                      retq
    400ed4:       66 2e 0f 1f 84 00 00    nopw   %cs:0x0(%rax,%rax,1)
    400edb:       00 00 00
    400ede:       66 90                   xchg   %ax,%ax
```
The same problem can be shown with addr2line and it can't find the
source location info as follows.
```
  $ addr2line -e shared_ptr.fpatchable 0x400ebc
  ??:?
```
The debug info can be found from the first actual instruction.
```
  $ addr2line -e shared_ptr.fpatchable 0x400ec1
  /usr/include/c++/9/bits/shared_ptr_base.h:384
```
To avoid this problem, we should adjust dwarf_addr to the first actual
instruction by skipping NOP instructions.

Fixed: #1387

### DIFF
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -13,6 +13,7 @@
 #define PR_DOMAIN  DBG_DWARF
 
 #include "uftrace.h"
+#include "mcount-arch.h"
 #include "utils/utils.h"
 #include "utils/dwarf.h"
 #include "utils/symbol.h"
@@ -1430,9 +1431,23 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
 		Dwarf_Die cudie;
 		Dwarf_Line *line;
 		unsigned long dwarf_addr = sym_to_dwarf_addr(dinfo, sym->addr);
+		unsigned long limit_addr = dwarf_addr + sym->size;
+		int search_limit = 10;
 
 		dwarf_diecu(die, &cudie, NULL, NULL);
-		line = dwarf_getsrc_die(&cudie, dwarf_addr);
+
+		/*
+		 * This loop is needed because gcc doesn't create dwarf info at NOP
+		 * instruction address so dwarf_getsrc_die() returns NULL.
+		 * To avoid this problem, we move forward until we see the first actual
+		 * instruction address of the function, which has a valid dwarf info.
+		 * The search is limited to the function size.
+		 */
+		do {
+			line = dwarf_getsrc_die(&cudie, dwarf_addr);
+			dwarf_addr += NOP_INSN_SIZE;
+		} while (line == NULL && --search_limit > 0 && dwarf_addr < limit_addr);
+
 		filename = dwarf_linesrc(line, NULL, NULL);
 		dfile = get_debug_file(dinfo, filename);
 		dwarf_lineno(line, &dline);
--- a/arch/aarch64/mcount-arch.h
+++ b/arch/aarch64/mcount-arch.h
@@ -38,6 +38,8 @@ struct mcount_disasm_engine;
 struct mcount_dynamic_info;
 struct mcount_disasm_info;
 
+#define NOP_INSN_SIZE   4
+
 int disasm_check_insns(struct mcount_disasm_engine *disasm,
 		       struct mcount_dynamic_info *mdi,
 		       struct mcount_disasm_info *info);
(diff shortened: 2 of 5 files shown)


utils/dwarf.c
```diff
@@ -1430,9 +1431,13 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
...
 		Dwarf_Line *line;
 		unsigned long dwarf_addr = sym_to_dwarf_addr(dinfo, sym->addr);
+		int search_limit = 10;
 
 		dwarf_diecu(die, &cudie, NULL, NULL);
-		line = dwarf_getsrc_die(&cudie, dwarf_addr);
+		do {
+			line = dwarf_getsrc_die(&cudie, dwarf_addr);
+			dwarf_addr += NOP_INSN_SIZE;
+		} while (line == NULL && --search_limit > 0);
```
namhyung, line 1440: Would be better if we can check the binary and it's NOP actually.  Also it needs to handle the failure case (`line == NULL`) properly so that just skip the function (and `ufftrace replay` would run without the error).

honggyukim: Hmm.. I didn't do that because we need load address to read the instruction so it requires map info. Is it possible to get the map info here?

honggyukim: Could you please let me know if you have a better idea? Otherwise, I can just add TODO comment in this commit. I think it's better to fix with a workaround than leaving it broken.

namhyung: Agreed, let's fix the problem first.  I think we should check the symbol size as well so that it can stop at the end of the function.

honggyukim: Right, I've handled it as you mentioned.

utils/dwarf.c
```diff
@@ -1430,9 +1431,22 @@ static void get_source_location(Dwarf_Die *die, struct build_data *bd,
...
+		 * This loop is needed because gcc doesn't create dwarf info at NOP
+		 * instruction address so dwarf_getsrc_die() returns NULL.
+		 * To avoid this problem, we move forward until we see the first actual
+		 * instruction address of the function, which has a valid dwarf info.
+		 * The search is limited to the function size.
+		 */
+		do {
+			line = dwarf_getsrc_die(&cudie, dwarf_addr);
+			dwarf_addr += NOP_INSN_SIZE;
+		} while (line == NULL && dwarf_addr < limit_addr);
```
namhyung, line 1448: Can we check both the function size and the count?  I'm afraid if there's a pathological case..

honggyukim: What could be the count? Do you mean we can set an arbitrary count limit just like `--search_limit > 0`?

namhyung: Yes, I meant that.

honggyukim: OK. Then I will set `int search_limit = 10` and check the count.

honggyukim: I just update it. Thanks.

namhyung (approved):
LGTM

Issue #1404, opened 2022-01-29
Is uftrace `-?` option an obsolete feature, or a bug?


kangtegong:
In README.md and doc/uftrace.md, there's an explation about `-?` option like below:

README.md

    You can use `-h`, `-?` or `--help` option to see available commands and options.

uftrace.md

    -?, \--help
        :   Print help message and list of options with description

But turns out, `-?` option doesn't work properly. (18.04.1-Ubuntu SMP x86_64)

```
$ uftrace -?
uftrace: invalid option -- '?'
```

Is this option is not available anymore? Or is this a bug?

If this feature is not available,  I'll erase this option from docs. And if this is a bug, can I fix this bug?

honggyukim:
You're right. It has become obsolete since we change the option from argp to `getopt_long` at 86f44a22bc6d87723f9a60344eecd3c1a04134cd.

Please send a PR to fix it. Thanks.

PR #1405, opened 2022-01-31
doc: Remove obsolete -? option from docs

kangtegong:
Since -? option is obsolete and currently not available,
this commit removes the feature from docs.

Fixed: #1404


honggyukim (approved):
Thanks for the fix.

kangtegong:
Thank you for reviewing this :)

PR #1416, opened 2022-04-04
uftrace: correct minor spelling errors

thepaul:
as found by the debian automated-checking tool Lintian. None of these
matter particularly, but it's a patch we apply to the Debian packaging,
so I'll pass it along.

Signed-off-by: paul cannon <pik@debian.org>


namhyung (approved):
LGTM

PR #1420, opened 2022-04-25
clock: Fix clock_source setup logic to handle perf events

honggyukim:
The --clock option currently doesn't change the clock id of perf events
and this makes the replay output incorrect when clock id is changed.

This patch applies clock id change to perf event as well and moves
clock_source from libmcount/mcount.c to utils/utils.c.


honggyukim:
@namhyung I need this fix this as early as possible so I will merge this fix. If you leave a review comment, I will make another PR to handle those comments.

PR #1421, opened 2022-04-25
symbol: Fix incorrect symbol file loading

honggyukim:
We also need to check if m->build_id[0] is not NULL just like we did for
build_id[0].

When m->build_id[0] is empty, it fails loading symbol files.


honggyukim:
@namhyung I need this fix as early as possible so I will merge this fix. If you leave a review comment, I will make another PR to handle those comments.

PR #1425, opened 2022-04-29
tests: Fix a regression in 248 dynamic_dlopen

honggyukim:
The test 248 dynamic_dlopen fails since the following patch.

  0096b3f3 dynamic: Do not skip weak symbol functions for dynamic tracing

It's because AAA:bar() is a weak symbol so it wasn't shown before, but
this patch makes it appear.
```
  $ nm -C libfoo.so
        ...
  0000000000000659 W AAA::bar(int)
```
This patch fixes the test by adding AAA:bar() in the expected test
result.


namhyung (approved):
LGTM

PR #1426, opened 2022-04-29
travis: Run system tests in CI

honggyukim:
It'd be more helpful to run our system tests in travis.

But there are some known test failures so we show the failed tests but
doesn't raise a red flag in the CI test result even if some tests fails.


honggyukim:
Our`runtest.py` fails in many tests when running with asan compiled `uftrace` so I will close this for now. Maybe we can enable this when all the tests pass even with asan enabled.

PR #1432, opened 2022-05-03
slide: Update dynamic tracing section

honggyukim:
We now have dynamic tracing for patchable function entry for both with a
compiler option and an attribute.

This usage is much better than the previous other dynamic tracing
options so it'd be better to replace old dynamic tracing ways to using
-fpatchable-function-entry=N and patchable_function_entry attribute in
the tutorial slide.


honggyukim:
The slide is updated as follows.
https://uftrace.github.io/slide/#dynamic-tracing

namhyung (approved):
LGTM, thanks for doing this!

PR #1437, opened 2022-06-25
arch: Fix missing renaming in arm and i386

honggyukim:
This patch is to fix missing renaming in arm and i386.
The original renaming change was introduced in the following commit.

  1b16f4ed uftrace: Rename 'symtabs' to 'uftrace_sym_info'


honggyukim:
@namhyung I will merge this as the current master is clearly broken in arm and i386 arches. I need to push other PR based on it.

namhyung:
Thanks for the fix!

PR #1438, opened 2022-06-26
A few fixes and enhancement for i386 testing

honggyukim:
This PR includes the following changes.
- afe48159 tests: Fix possibly incorrect libmcount.so loading
- 57b6b34e tests: Add -m32 option when the architecture is i386
- 31047961 tests: Add class Elf and move related code inside of runtest.py


namhyung (approved):
LGTM

PR #1447, opened 2022-07-01
session: correct the function name in the comment

qpakzk:
I found some typos in comments and corrected them.

### DIFF
--- a/utils/session.c
+++ b/utils/session.c
@@ -695,7 +695,7 @@ struct uftrace_symbol * task_find_sym(struct uftrace_session_link *sessions,
 }
 
 /**
- * task_find_sym - find a symbol that matches to @addr
+ * task_find_sym_addr - find a symbol that matches to @addr
  * @sessions: session link to manage sessions and tasks
  * @task: handle for functions in a task
  * @time: timestamp of the @addr


utils/fstack.c
```diff
@@ -503,7 +503,7 @@ int fstack_setup_filters(struct uftrace_opts *opts, struct uftrace_data *handle)
 
 /**
  * fstack_get - retrieve func_stack entry in a task
- * @task - tracee task
+ * @task - trace task
```
honggyukim, line 506: Ah. `tracee` might be correct and used on purpose. @namhyung Could you please confirm this?

namhyung: Yeah, it was intended.

utils/session.c
```diff
@@ -695,7 +695,7 @@ struct uftrace_symbol * task_find_sym(struct uftrace_session_link *sessions,
 }
 
 /**
- * task_find_sym - find a symbol that matches to @addr
+ * task_find_sym_addr - find a symbol that matches to @addr
```
honggyukim, line 698: As @namhyung mentioned, `tracee` is intentional, but we can leave this change. So the title of the commit is better to be changed.

utils/fstack.c
qpakzk: I understand that `tracee` is the opposite of `tracer`.

utils/session.c
qpakzk: @honggyukim I changed my commit. I appreciate it.

namhyung (approved):
LGTM

PR #1452, opened 2022-07-02
misc: Add an exception for distros that doesn't use /etc/os-release

kangtegong:
Although many linux distributions use /etc/os-release,
there are some systems that does not use that for OS identification.
(especially for some traditional non-systemd linux distros or macOS)

When these kinds on distros run /misc/install-deps.sh,
message is printed like below:
```
  $ sudo ./install-deps.sh
  grep: /etc/os-release: No such file or directory
  "" is not supported distro, so please install packages manually.
```
So this commit adds an exception for those does not use
/etc/os-release.

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -9,6 +9,12 @@ fi
 
 OPT="${@}"
 
+if [ ! -f /etc/os-release ]; then
+    echo "Your distribution is not supported, so please install pacakges manually."
+    echo
+    exit
+fi
+
 distro=$(grep "^ID=" /etc/os-release | cut -d\= -f2 | sed -e 's/"//g')
 
 case $distro in
@@ -34,5 +40,6 @@ case $distro in
         apk $OPT add luajit-dev || true
         apk $OPT add capstone-dev || true ;;
     *) # we can add more install command for each distros.
-        echo "\"$distro\" is not supported distro, so please install packages manually." ;;
+        echo "\"$distro\" is not supported distro, so please install packages manually."
+        echo ;;
 esac


misc/install-deps.sh
```diff
@@ -9,6 +9,12 @@ fi
 
 OPT="${@}"
 
+if [ ! -f /etc/os-release ]; then
+    echo "Your distribution is not supported, so please install pacakges manually."
+    echo
```
honggyukim, line 14: We don't need this empty `echo`. Instead, please add `\n` at the end of the previous message. The same thing can be applied to the above root permission check.

kangtegong: Thanks for your feedback, just amended the commit!

honggyukim (approved):
LGTM

misc/install-deps.sh
```diff
@@ -2,13 +2,17 @@
 
 if [ "x$(id -u)" != x0 ]; then
     echo "You might have to run it as root user."
-    echo "Please run it again with 'sudo'."
-    echo
+    echo "Please run it again with 'sudo'.\n"
```
namhyung, line 5: This requires `-e` option.

```
$ echo "abc\n"
abc\n

$ echo -e "abc\n"
abc

```

misc/install-deps.sh
```diff
@@ -2,13 +2,17 @@
...
     exit
 fi
 
 OPT="${@}"
 
+if [ ! -f /etc/os-release ]; then
+    echo "Your distribution is not supported, so please install pacakges manually.\n"
```
namhyung, line 12: Ditto.

misc/install-deps.sh
kangtegong: I think in order to use `echo -e` in shell script, 
shebang has to be `/bin/bash`, not (`/bin/sh`)

Otherwise, `-e` is also printed as follows:
(echo consider -e as a string)

```bash
#!/bin/sh

if [ "x$(id -u)" != x0 ]; then
    echo "You might have to run it as root user."
    echo -e "Please run it again with 'sudo'.\n"
    exit
fi
```

```
~/uftrace/misc$ ./install-deps.sh
You might have to run it as root user.
-e Please run it again with 'sudo'.
```

So I think these are the solution candidates for this problem:
1. keep shebang as /bin/sh && using empty echo instead of \n
2. chang shebang as /bin/bash && use `echo -e "...\n"`

reference:
[1] https://stackoverflow.com/questions/36531795/echo-e-is-not-working-in-my-script
[2] https://stackoverflow.com/questions/10267852/echo-e-working-in-terminal-but-not-in-bash-script

namhyung: Oh, I didn't know that it's a bash thing.  Then I think we should keep the empty echo line (if you really want a blank line).

honggyukim: OK. Let's keep using empty echo for newline again.

kangtegong: I just changed last newline as `echo`. Thank you for your feedbacks :)

namhyung (approved):
LGTM

Issue #1453, opened 2022-07-03
warning: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result


honggyukim:
There are some warnings when building `unittest` as follows.
```
$ make unittest
      ...
/home/honggyu/work/uftrace/utils/data-file.c: In function ‘create_data’:
/home/honggyu/work/uftrace/utils/data-file.c:666:2: warning: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result [-Wunused-result]
  666 |  asprintf(&filename, "%s/sid-%.*s.map",
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  667 |    opts->dirname, SESSION_ID_LEN, TEST_SESSION_ID);
      |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/honggyu/work/uftrace/utils/data-file.c:672:2: warning: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result [-Wunused-result]
  672 |  asprintf(&filename, "%s/%d.dat", opts->dirname, task_msg.pid);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      ...
```
We better fix those warnings.

SulRash:
Hello,

This issue seems easy to fix on the surface and I just wanted to ask if the bytes returned from asprintf are used? If not, I think I can push a fix pretty quickly.

honggyukim:
Hi @SulRash, sorry, the correct fix is to check the return value rather than ignoring it. @kangtegong already made the fix in #1489 and it's merged. So I will close this issue and your PR together.

PR #1454, opened 2022-07-04
fix: Issue #1453 Bug Fix

SulRash:
Fixed the issue by casting the return of asprintf to void to avoid unused variable warnings. Based on the expectation that the bytes returned from asprintf are unnecessary for the program's functionality.

Signed-off-by: Sultan AlRashed <sultan.m.rashed@gmail.com>

### DIFF
--- a/utils/data-file.c
+++ b/utils/data-file.c
@@ -663,13 +663,13 @@ static int create_data(struct uftrace_opts *opts)
 	write_dlopen_info(opts->dirname, &dlopen_msg, "libnothing.so");
 
 	/* create_session() requires a map file even if it's empty */
-	asprintf(&filename, "%s/sid-%.*s.map",
+	(void)asprintf(&filename, "%s/sid-%.*s.map",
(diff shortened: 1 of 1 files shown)


kangtegong:
Hi, nice patch :) But for your information,
prefix of commit message header shouldn't be just `fix: ...`.

It should be subject of uftrace (changed component of uftrace) like `utils: ...`.

Also, it would be nice if you clarify what change(s) did you made by your commit. For instance, rather than just
`fix: Issue #1453 Bug Fix ...`, `utils: casting return value of asprintf ...` would be better because other people can directly know what change(s) is made by your commit. 

If you put issue number in commit message body like
`Fixed: #<issue number>` as below, github automatically makes a link to the issue.

Screenshot_20220704-143707_Chrome

Thank you.

SulRash:
Will do! Thank you for the detailed response, I am still getting to grips with contributing to open source projects haha :)

Issue #1457, opened 2022-07-06
test 238 report_field2 failure


honggyukim:
The test 238 report_field2 fails as follows.
```
$ cd tests
$ ./runtest.py -vdp -O0 238
Start 1 tests with 1 worker
Test case                 pg
------------------------: O0
build command: gcc -o t-sort -fno-inline -fno-builtin -fno-ipa-cp -fno-omit-frame-pointer -D_FORTIFY_SOURCE=0  -pg -O0  s-sort.c
prerun command: /home/honggyu/work/uftrace/uftrace record --no-pager --no-event --libmcount-path=/home/parallels/work/uftrace  t-sort
test command: /home/honggyu/work/uftrace/uftrace report --no-pager --no-event --libmcount-path=/home/honggyu/work/uftrace -f total,total-max,self-min,call -s self-min,total-min t-sort
t238_report_field2: -pg -O0 returns 1

238 report_field2       : NZ
```
We can simply reproduce the problem as follows.
```
$ gcc -pg -o t-sort s-sort.c

$ uftrace record t-sort

$ uftrace report -f total,total-max,self-min,call -s self-min,total-min
uftrace: Unknown field name 'total-max'
uftrace:   Possible fields are: total total - avg total - min total - max self self - avg self - min self - max call
```
The problem is because of unknown filed name 'total-max' and we should fix this.

PR #1458, opened 2022-07-06
tests: Fix test 238 report_field2 failure

honggyukim:
clang-format mistakenly added whitespaces for input macro argument and
that makes the test 238 report_fields failure.
```
  $ ./runtest.py 238
  Start 1 tests with 1 worker
  238 report_field2       : NZ NZ NZ NZ NZ NZ NZ NZ NZ NZ
```
It can be reproduced as follows.
```
  $ gcc -pg -o t-sort s-sort.c
  $ uftrace record t-sort
  $ uftrace report -f total,total-max,self-min,call -s self-min,total-min
  uftrace: Unknown field name 'total-max'
  uftrace:   Possible fields are: total total - avg total - min total - max self self - avg self - min self - max call
```
It makes 'total-max' field unrecognizable, so this has to be fixed.

Fixed: #1457

### DIFF
--- a/utils/report.c
+++ b/utils/report.c
@@ -841,56 +841,52 @@ static void print_time_or_dash(uint64_t time_nsec)
 		pr_out("%10s", NODATA);
 }
 
+/* clang-format off */
 FIELD_TIME(REPORT_F_TOTAL_TIME, total, total.sum, total, "Total time");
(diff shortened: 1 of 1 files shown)


honggyukim:
@namhyung I will merge this as it's clearly a clang-format error.

Issue #1449, opened 2022-07-02
error: unknown type name ‘bool’ in utils/argspec.h


honggyukim:
We don't compile `utils/argspec.h` as a standalone, but it's obviously a problem so better to fix it. 
```
$ gcc -I. -c utils/argspec.h
utils/argspec.h:45:2: error: unknown type name ‘bool’
   45 |  bool   exact;
      |  ^~~~
```
We might have similar problems in other `.h` or `.c` files.

PR #1460, opened 2022-07-06
utils: Add missing header in argspec.h

kangtegong:
argspec.h lacks stdbool.h, and this causes compilation error
when argspec.h is compiled(as a standalone).

Although argspec.h doesn't usually compiled as a standalone,
this can cause potential error when other file(s) include
only argspec.h and nothing else.

Fixed: #1449


namhyung (approved):
LGTM

PR #1468, opened 2022-07-08
misc: Add Amazon Linux distro in install-deps.sh

kangtegong:
Although Amazon Linux uses yum as package manager[1],
current /misc/install-deps.sh cannot install packages because
ID of /etc/os-release is "amzn".
```
  $ cat /etc/os-release
  NAME="Amazon Linux"
  VERSION="2"
  ID="amzn"
  ID_LIKE="centos rhel fedora"
  VERSION_ID="2"
  PRETTY_NAME="Amazon Linux 2"
  ANSI_COLOR="0;33"
  CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
  HOME_URL="https://amazonlinux.com/"

  $ sudo ./install-deps.sh
  "amzn" is not supported distro, so please install packages manually.
```
So this commit adds Amazon Linux in /misc/install-deps.sh
so that Amazon Linux users can install uftrace dependency packages instantly.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-software.html

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -26,7 +26,7 @@ case $distro in
         dnf install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconf-pkg-config
         dnf install $OPT luajit-devel || true
         dnf install $OPT capstone-devel || true ;;
-    "rhel" | "centos")
+    "rhel" | "centos" | "amzn")
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
         yum install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconfig
         yum install $OPT luajit-devel || true


namhyung:
According to the output, it'd be nicer if we could parse ID_LIKE and try to match to one of them instead of adding one specific case.

kangtegong:
yes, I think that's better too. I'll work on it soon :)

kangtegong:
> According to the output, it'd be nicer if we could parse ID_LIKE and try to match to one of them instead of adding one specific case.

kangtegong:
> According to the output, it'd be nicer if we could parse ID_LIKE and try to match to one of them instead of adding one specific case.

I just checked other distros, and it turns out not all distros use ID_LIKE.

For instance, `/etc/os-release` of alpine linux is as follows.

```
/ # cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.16.0
PRETTY_NAME="Alpine Linux v3.16"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

And archlinux is as follows.

```
[root@ddfac8478da6 /]# cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux-logo
```

I think it's better to check ID_LIKE (if present) as well as ID (like uftrace uses now)

namhyung:
Right, I meant it as a fallback.  We can check ID first and then check ID_LIKE if any.

namhyung (commented):
This is not what we talked.  Please check ID first and try to match names in ID_LIKE then.

kangtegong:
Oh, yes. 
The forced-pushed branch above is just for rebase.
I'm currently working on it :)

Issue #1471, opened 2022-07-10
Incorrect help message when uftrace fails to trace kernel functions


shpark:
I tried to trace kernel functions with uftrace using the following command: `sudo uftrace record -k ./a.out`.

It failed to trace kernel function, saying that:
```
WARN: kernel tracing disabled due to an error
is CONFIG_FUNCTION_GRAPH_TRACER enabled in the kernel?
```

However, it seemed that the config was enabled for my setting (Debian 10 running inside a Docker container on MacOS Monterey 12.0.1 host on M1 pro chip):
```
vscode ➜ /com.docker.devenvironments.code/samples (master ✗) $ zcat /proc/config.gz | rg -i graph_tracer
CONFIG_HAVE_FUNCTION_GRAPH_TRACER=y
CONFIG_FUNCTION_GRAPH_TRACER=y
vscode ➜ /com.docker.devenvironments.code/samples (master ✗) $ uname -r
5.10.104-linuxkit
```

After looking at more verbose logs, I concluded that the root cause was prolly the absence of `tracing_on` file on proper path:
```
vscode ➜ /com.docker.devenvironments.code/samples (master ✗) $ sudo uftrace record -vv -k ./a.out
uftrace: running uftrace v0.12-43-g5f03a ( aarch64 dwarf python luajit tui perf sched dynamic )
uftrace: checking binary ./a.out
uftrace: removing uftrace.data.old directory
kernel: cannot open tracing file: tracing_on: No such file or directory
kernel: failed to reset tracing files
WARN: kernel tracing disabled due to an error
is CONFIG_FUNCTION_GRAPH_TRACER enabled in the kernel?
uftrace: skipping perf event due to error: Operation not permitted
uftrace: creating 2 thread(s) for recording
uftrace: start writer thread 1
uftrace: using /usr/local/lib/libmcount.so library for tracing
uftrace: start writer thread 0
mcount: initializing mcount library
```

Imho the error message would better be refined to provide better suggestions (e.g, mount debugfs, etc.).

Maybe related to #797

honggyukim:
Hi @shpark, thanks for the report. Could you please send a PR by detecting mount point of debugfs? There is a nice example at https://stackoverflow.com/questions/9280759/linux-function-to-get-mount-points.
```c
#include <stdio.h>
#include <stdlib.h>
#include <mntent.h>

int main(void)
{
  struct mntent *ent;
  FILE *aFile;

  aFile = setmntent("/proc/mounts", "r");
[...]

PR #1476, opened 2022-07-14
utils: detect tracefs or debugfs at runtime

### LINKED ISSUE #797: Fail to find tracing_on file for kernel tracing
It complains as follows:
```
uftrace record -vv --force -K 10 -- dd if=xxx of=/dev/mapper/??? bs=4096
uftrace: checking binary /bin/dd
WARN: Didn't you run 'make install' ?
uftrace: using libmcount.so library for tracing
uftrace: send UFTRACE_MSG_SEND_HDR
kernel: cannot open tracing file: tracing_on: No such file or directory
kernel: failed to reset tracing files
WARN: kernel tracing disabled due to an error
is CONFIG_FUNCTION_GRAPH_TRACER enabled in the kernel?
uftrace: start writer thread 0
uftrace: start writer thread 1
...
```
However, `tracing_on` file exists.
```
# cat /sys/kernel/tracing/tracing_on
1
```

shpark:
Currently, uftrace interacts with kernel tracing system through
a fixed TRACING_DIR; /sys/kernel/debug/tracing, therefore fails
to trace kernel when a tracefs is mounted at a different location.

This commit introduces runtime detection tracefs or debugfs from
/proc/self/mountinfo (which is newer than /proc</self>/mounts).

Fixes #797 

Signed-off-by: Seonghyun Park <shp4rk@gmail.com>

### DIFF
--- a/libmcount/misc.c
+++ b/libmcount/misc.c
@@ -13,8 +13,6 @@
 #include "utils/tracefs.h"
 #include "utils/utils.h"
 
-static char *TRACING_DIR = NULL;
-
 /* old kernel never updates pid filter for a forked child */
 void update_kernel_tid(int tid)
 {
@@ -23,16 +21,13 @@ void update_kernel_tid(int tid)
 	int fd;
 	ssize_t len;
 
-	if (!TRACING_DIR && !find_tracing_dir(&TRACING_DIR))
-		return;
-
 	if (!kernel_pid_update)
 		return;
 
 	/* update pid filter for function tracing */
-	xasprintf(&filename, "%s/set_ftrace_pid", TRACING_DIR);
+	filename = get_tracing_file("set_ftrace_pid");
 	fd = open(filename, O_WRONLY | O_APPEND);
-	free(filename);
+	put_tracing_file(filename);
 
 	if (fd < 0) {
 		pr_dbg("open kernel ftrace pid filter failed\n");
@@ -47,9 +42,9 @@ void update_kernel_tid(int tid)
 	close(fd);
 
 	/* update pid filter for event tracing */
-	xasprintf(&filename, "%s/set_event_pid", TRACING_DIR);
+	filename = get_tracing_file("set_event_pid");
 	fd = open(filename, O_WRONLY | O_APPEND);
-	free(filename);
+	put_tracing_file(filename);
 	if (fd < 0) {
 		pr_dbg("open kernel event pid filter failed\n");
 		return;
--- a/libmcount/misc.c
+++ b/libmcount/misc.c
@@ -10,16 +10,21 @@
 
 #include "libmcount/internal.h"
 #include "libmcount/mcount.h"
+#include "utils/tracefs.h"
 #include "utils/utils.h"
+
+static char *TRACING_DIR = NULL;
 
 /* old kernel never updates pid filter for a forked child */
 void update_kernel_tid(int tid)
 {
-	static const char TRACING_DIR[] = "/sys/kernel/debug/tracing";
 	char *filename = NULL;
 	char buf[8];
 	int fd;
 	ssize_t len;
+
+	if (!TRACING_DIR && !find_tracing_dir(&TRACING_DIR))
+		return;
 
 	if (!kernel_pid_update)
 		return;
--- a/libmcount/misc.c
+++ b/libmcount/misc.c
@@ -16,46 +16,20 @@
 /* old kernel never updates pid filter for a forked child */
 void update_kernel_tid(int tid)
 {
-	char *filename = NULL;
 	char buf[8];
-	int fd;
-	ssize_t len;
 
 	if (!kernel_pid_update)
 		return;
 
-	/* update pid filter for function tracing */
-	filename = get_tracing_file("set_ftrace_pid");
-	fd = open(filename, O_WRONLY | O_APPEND);
-	put_tracing_file(filename);
-
-	if (fd < 0) {
-		pr_dbg("open kernel ftrace pid filter failed\n");
-		return;
-	}
-
 	snprintf(buf, sizeof(buf), "%d", tid);
-	len = strlen(buf);
-	if (write(fd, buf, len) != len)
-		pr_dbg("update kernel ftrace pid filter failed\n");
 
-	close(fd);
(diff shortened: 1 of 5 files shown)


honggyukim:
@shpark Thanks for your patch, but is there a reason not to use `setmntent`/`getmntent` as I mentioned at https://github.com/namhyung/uftrace/issues/1471#issuecomment-1179689118? It will make the logic much shorter that the current implementation.

honggyukim:
>  I decided to use /proc/self/mountinfo as it was seemingly more modern.

If we can do the same thing then older interface is more portable than newer ones.

> Plus, I was unsure how universal mntent.h is, like you mentioned Android. I wanted to avoid having something like #ifdef HAVE_MNTENT_H.

I have checked `mntent.h` is also available on Android so I think we can use it.

To make it safer, we can check which one was introduced first between `mntent.h` and `/proc/self/mountinfo`.

@namhyung Do you have any preference on these approaches?

utils/mountinfo.c
```diff
@@ -0,0 +1,83 @@
...
+		xasprintf(tracing_dir, "%s", mnt);
+		pr_dbg2("tracefs found at %s\n", mnt);
+		pr_dbg2("TRACING_DIR: %s\n", *tracing_dir);
+		return true;
+	}
+
+	/*
+         * If tracefs is not found (e.g., older kernel), try to find debugfs
+         * and use its subdirectory `tracing` as `TRACING_DIR`.
+         */
```
namhyung, line 53: This doesn't follow the coding style.  Please do not indent the comment like this.

utils/mountinfo.h
```diff
@@ -0,0 +1,13 @@
+#ifndef UFTRACE_MOUNT_INFO_H
+#define UFTRACE_MOUNT_INFO_H
+
+#include <stdbool.h>
+
+/* TODO: Okay to use `self`? */
+#define PROC_MOUNTINFO "/proc/self/mountinfo"
```
namhyung, line 7: I think it's ok.

utils/mountinfo.c
```diff
@@ -0,0 +1,83 @@
...
+	size_t len;
+	int i;
+
+	fp = fopen(PROC_MOUNTINFO, "r");
+	if (fp == NULL)
+		return false;
+
+	/* First, try to find tracefs. */
+	while (getline(&line, &len, fp) > 0) {
+		if (!(ptr = strstr(line, "tracefs")))
```
namhyung, line 20: While this would work in most cases, it might not work properly if any other mount path contains the "tracefs".  You'd better use `sscanf()` to parse the mountinfo format properly - i.e. checking 9th position specifically.

namhyung (commented):
Thanks for the contribution.  I have no preference on it but /proc/PID/mountinfo is available from 2.6.26 kernel which is ancient so no worry for compatibility.

Also I think we can move more tracefs utilities to the new file like get/put/open_tracing_file and write/append_tracing_file.  It could be named as utils/tracefs.[ch].

namhyung (dismissed):
LGTM.

I suggest a follow up work to move `write_tracing_file()` (and `append_tracing_file()` too) to the utils/tracefs.c and use it in libmcount/misc.c.

namhyung:
It'd be nice if you could add it to this PR.

utils/tracefs.h
```diff
@@ -2,9 +2,30 @@
 #define UFTRACE_TRACEFS_H
 
 #include <stdbool.h>
+#include <stddef.h>
+#include <unistd.h>
 
 #define PROC_MOUNTINFO "/proc/self/mountinfo"
 
-bool find_tracing_dir(char **tracing_dir);
+bool find_tracing_dir(void);
```
namhyung, line 10: It doesn't need to be exposed, right?

libmcount/misc.c
```diff
@@ -13,54 +13,23 @@
...
-	}
-
 	snprintf(buf, sizeof(buf), "%d", tid);
-	len = strlen(buf);
-	if (write(fd, buf, len) != len)
-		pr_dbg("update kernel event pid filter failed\n");
-
-	close(fd);
+	if (append_tracing_file("set_event_pid", buf) < 0)
+		pr_dbg("write to kernel event pid filter failed\n");
```
namhyung, line 32: Can we just call `set_tracing_pid()` here?  But it could be a separate change.

namhyung (commented):
I think you are doing too many things in a single commit.  Let's break the second commit into parts:
* just move basic open/write
* convert libmcount/misc.c to use the functions
* introduce read_tracing_file() and use it

libmcount/misc.c
```diff
@@ -16,46 +16,20 @@
...
-	}
-
-	snprintf(buf, sizeof(buf), "%d", tid);
-	len = strlen(buf);
-	if (write(fd, buf, len) != len)
-		pr_dbg("update kernel event pid filter failed\n");
-
-	close(fd);
+	if (append_tracing_file("set_event_pid", buf) < 0)
+		pr_dbg("write to kernel ftrace pid filter failed\n");
```
namhyung, line 32: Can we simply call `set_tracing_pid()`?

utils/kernel.c
```diff
@@ -895,23 +895,23 @@ static int load_current_kernel(struct uftrace_kernel_reader *kernel)
...
+	fd = open_tracing_file("events/ftrace/funcgraph_entry/format", O_RDONLY);
 	if (fd < 0)
 		return -1;
 
 	len = read(fd, buf, sizeof(buf));
 	pevent_parse_event(pevent, buf, len, "ftrace");
 	close(fd);
 
-	fd = open_tracing_file("/events/ftrace/funcgraph_exit/format", O_RDONLY);
+	fd = open_tracing_file("events/ftrace/funcgraph_exit/format", O_RDONLY);
```
namhyung, line 914: I think these should be squashed to the first commit.

libmcount/misc.c
namhyung: Ah, ok.  Let's keep it for now. I think we can add `append_tracing_pid()`.

namhyung (approved):
LGTM, thanks a lot!

namhyung:
Could you please rebase your branch on to the current master?

Issue #1444, opened 2022-06-27
Split uftrace_info_bits to index and mask


namhyung:
Like in `enum uftrace_feat_bits`, it should have separate bit index and mask to reduce possible confustions.  Recent test failure was because of this confusion and it was painful to discover the root cause.

daeroro:
Can you explain more about the test failure due to this confusion?

honggyukim:
@namhyung Do you want to make the following change? If so, it looks it doesn't provide backward compatibility.
```diff
diff --git a/uftrace.h b/uftrace.h
index 5f016f11..5dc06e98 100644
--- a/uftrace.h
+++ b/uftrace.h
@@ -80,20 +80,39 @@ enum uftrace_feat_bits {
 };

 enum uftrace_info_bits {
-       EXE_NAME,
-       EXE_BUILD_ID,
-       EXIT_STATUS,
-       CMDLINE,
-       CPUINFO,
-       MEMINFO,
-       OSINFO,
-       TASKINFO,
-       USAGEINFO,
-       LOADINFO,
-       ARG_SPEC,
-       RECORD_DATE,
-       PATTERN_TYPE,
-       VERSION,
+       /* bit index */
+       EXE_NAME_BIT,
+       EXE_BUILD_ID_BIT,
+       EXIT_STATUS_BIT,
+       CMDLINE_BIT,
+       CPUINFO_BIT,
+       MEMINFO_BIT,
+       OSINFO_BIT,
+       TASKINFO_BIT,
+       USAGEINFO_BIT,
+       LOADINFO_BIT,
+       ARG_SPEC_BIT,
+       RECORD_DATE_BIT,
+       PATTERN_TYPE_BIT,
+       VERSION_BIT,
+
+       INFO_BIT_MAX,
+
+       /* bit mask */
+       EXE_NAME = (1U << EXE_NAME_BIT),
+       EXE_BUILD_ID = (1U << EXE_BUILD_ID_BIT),
+       EXIT_STATUS = (1U << EXIT_STATUS_BIT),
+       CMDLINE = (1U << CMDLINE_BIT),
+       CPUINFO = (1U << CPUINFO_BIT),
+       MEMINFO = (1U << MEMINFO_BIT),
+       OSINFO = (1U << OSINFO_BIT),
+       TASKINFO = (1U << TASKINFO_BIT),
+       USAGEINFO = (1U << USAGEINFO_BIT),
+       LOADINFO = (1U << LOADINFO_BIT),
[...]

PR #1478, opened 2022-07-15
dump: Add to print 'info' bit name

daeroro:
When executing 'dump' command, uftrace prints 'file header'.
And it prints 'feature mask' and its bit name.
(e.g. "PLTHOOK", "KERNEL", ..)

Similarly, it prints 'info mask' value. So, I added simple logic
to print 'info' bit name too.(e.g. "EXE_NAME", "CMDLINE", ...)

Signed-off-by: DaeRo Lee <skseofh@naver.com>

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -505,21 +505,18 @@ static void pr_event(struct uftrace_task_reader *task, unsigned evt_id)
 	free(evt_data);
 }
 
-static void get_feature_string(char *buf, size_t sz, uint64_t feature_mask)
+static void get_header_string(char *buf, size_t sz, uint64_t hdr_mask, const char **hdr_str,
+			      int hdr_max)
 {
 	int i;
 	size_t len;
 	bool first = true;
-	const char *feat_str[] = { "PLTHOOK",	 "TASK_SESSION", "KERNEL",     "ARGUMENT",
-				   "RETVAL",	 "SYM_REL_ADDR", "MAX_STACK",  "EVENT",
-				   "PERF_EVENT", "AUTO_ARGS",	 "DEBUG_INFO", "ESTIMATE_RETURN" };
 
-	/* feat_str should match to enum uftrace_feat_bits */
-	for (i = 0; i < FEAT_BIT_MAX; i++) {
-		if (!((1U << i) & feature_mask))
+	for (i = 0; i < hdr_max; i++) {
+		if (!((1U << i) & hdr_mask))
 			continue;
 
-		len = snprintf(buf, sz, "%s%s", first ? "" : " | ", feat_str[i]);
+		len = snprintf(buf, sz, "%s%s", first ? "" : " | ", hdr_str[i]);
 		buf += len;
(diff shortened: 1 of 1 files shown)


cmds/dump.c
```diff
@@ -544,10 +548,13 @@ static void dump_raw_header(struct uftrace_dump_ops *ops, struct uftrace_data *h
 	       handle->hdr.endian == 1 ? "little" : "big");
 	pr_out("uftrace file header: class         = %u (%s bit)\n", handle->hdr.elf_class,
 	       handle->hdr.elf_class == 2 ? "64" : "32");
-	get_feature_string(buf, sizeof(buf), handle->hdr.feat_mask);
+	get_header_string(buf, sizeof(buf), handle->hdr.feat_mask, feat_str, FEAT_BIT_MAX);
 	pr_out("uftrace file header: features      = %#" PRIx64 " (%s)\n", handle->hdr.feat_mask,
 	       buf);
-	pr_out("uftrace file header: info          = %#" PRIx64 "\n", handle->hdr.info_mask);
+	memset(buf, 0, sizeof(buf));
```
honggyukim, line 554: Do we need this `memset`?

honggyukim (commented):
Thanks for your patch! But I think we have to clarify #1444 first before this change. @namhyung Please have a look.

cmds/dump.c
daeroro: I thought that `buf` needs to be cleared if feature name length is shorter than info name length.
But `snprintf` includes the terminating null byte ('\0')) to `buf`, so `memset` doesn't need.
Thanks for your comment :)

namhyung:
Right, it'd be nice to include cleanup in #1444.

cmds/dump.c
```diff
@@ -533,6 +530,13 @@ static void dump_raw_header(struct uftrace_dump_ops *ops, struct uftrace_data *h
 	int i;
 	char buf[1024];
 	struct uftrace_raw_dump *raw = container_of(ops, typeof(*raw), ops);
+	const char *feat_str[] = { "PLTHOOK",	 "TASK_SESSION", "KERNEL",     "ARGUMENT",
+				   "RETVAL",	 "SYM_REL_ADDR", "MAX_STACK",  "EVENT",
+				   "PERF_EVENT", "AUTO_ARGS",	 "DEBUG_INFO", "ESTIMATE_RETURN" };
+	const char *info_str[] = { "EXE_NAME",	   "EXE_BUILD_ID", "EXIT_STATUS", "CMDLINE",
+				   "CPUINFO",	   "MEMINFO",	   "OSINFO",	  "TASKINFO",
+				   "USAGEINFO",	   "LOADINFO",	   "ARG_SPEC",	  "RECORD_DATE",
+				   "PATTERN_TYPE", "VERSION" };
```
namhyung, line 539: While it was not there originally, please put a comma after the last element so that it can format the code 'properly'.

namhyung (approved):
LGTM

Issue #1477, opened 2022-07-14
segfault when accessing string argument


honggyukim:
We can download hermes source code and build as follows.
```
$ git clone git@github.com:facebook/hermes.git && cd hermes

$ mkdir build && cd build

$ cmake ..

$ make -j
```
If we record it with `-a`, then it gets segfault as follows.
```
$ uftrace record -P. -a ./bin/hermes
WARN: Segmentation fault: invalid permission (addr: 0x29f1f1400000)
WARN:  if this happens only with uftrace, please consider -e/--estimate-return option.

WARN: Backtrace from uftrace v0.12-43-g5f03a ( x86_64 dwarf python luajit tui perf sched dynamic )
WARN: =====================================
WARN: [8] (std::__exchange[60adff] <= std::exchange[60abea])
WARN: [7] (std::exchange[60abc1] <= hermes::vm::_GLOBAL__N_1::ContiguousVAStorageProvider::newStorageImpl[60a723])
WARN: [6] (hermes::vm::_GLOBAL__N_1::ContiguousVAStorageProvider::newStorageImpl[60a679] <= hermes::vm::StorageProvider::newStorage[60a9b0])
WARN: [5] (hermes::vm::StorageProvider::newStorage[60a96f] <= hermes::vm::HeapRuntime::HeapRuntime[5d22d2])
WARN: [4] (hermes::vm::HeapRuntime::HeapRuntime[5d2277] <= hermes::vm::HeapRuntime::create[5cd08f])
WARN: [3] (hermes::vm::HeapRuntime::create[5cd038] <= hermes::vm::Runtime::create[5bf710])
WARN: [2] (hermes::vm::Runtime::create[5bf5e7] <= repl[417b10])
WARN: [1] (repl[417ad6] <= main[40bab8])
WARN: [0] (main[40b9ae] <= __libc_start_main[7f0132d9d0b3])

Please report this bug to https://github.com/namhyung/uftrace/issues.

WARN: child terminated by signal: 11: Segmentation fault
```
Here is a gdb output and I found it crashed when accessing `str[i]`.
```
$ gdb --args uftrace record -P. -a ./bin/hermes
(gdb) r
    ...
Thread 2.1 "hermes" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff3d3f840 (LWP 50272)]
[...]

PR #1480, opened 2022-07-16
args: Fix segfaults when reading string arguments

honggyukim:
In some cases, argument read for string could make the target program
get crashed.  The following sequence shows the problem.
```
  $ git clone git@github.com:facebook/hermes.git && cd hermes
  $ mkdir build && cd build
  $ cmake ..
  $ make -j

  $ uftrace record -P. -a ./bin/hermes
  WARN: Segmentation fault: invalid permission (addr: 0x29f1f1400000)
      ...
```
This crash happens becasue of string dereference for non readable memory
regions.

uftrace provides check_mem_region() for this purpose, but it doesn't
check whether each memory region has read permission or not.

In addition, some memory regions do have read permission but still make
the target program crashed.  To avoid this problem, we better skip
memory regions, which don't have names of them.

Fixed: #1477


libmcount/record.c
```diff
@@ -310,6 +310,22 @@ static void update_mem_regions(struct mcount_mem_regions *regions)
...
+		 * unnamed regions.
+		 * https://github.com/namhyung/uftrace/issues/1477
+		 *
+		 * 64-bit
+		 * 7ffff6bba000-7ffff6bdf000 r--p 00000000 fd:00 10229310                   /lib/x86_64-linux-gnu/libc-2.31.so
+		 * 32-bit
+		 * f7db5000-f7dd2000 r--p 00000000 fd:00 10701229                           /lib/i386-linux-gnu/libc-2.31.so
+		 */
+		if (strlen(buf) < 74)
+			continue;
```
namhyung, line 327: Is this really needed?  I think it's fragile and unnecessary as it also checks the permission.

honggyukim: Without checking this unnamed memory regions, I see hermes run still crashed because of dereferencing those regions.

namhyung: But I think unnamed (anon) regions can be legal and contain valid strings.

honggyukim: I initially checked only the read permission as follows.
```diff
diff --git a/libmcount/record.c b/libmcount/record.c
index 77a96742..292733f3 100644
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -318,6 +318,10 @@ static void update_mem_regions(struct mcount_mem_regions *regions)
                p = next + 1;
                end = strtoul(p, &next, 16);

+               p = next + 1;
+               if (*p != 'r')
+                       continue;
+
                if (strstr(next, "[heap]")) {
                        end = ROUND_UP(end, HEAP_REGION_UNIT);
                        if (end > regions->brk)
```
But I also saw another segfault.
```
$ uftrace record -P. -a ./bin/hermes
WARN: Segmentation fault: invalid permission (addr: 0x331e93800000)
WARN:  if this happens only with uftrace, please consider -e/--estimate-return option.

WARN: Backtrace from uftrace v0.12-53-g1ee2 ( x86_64 dwarf python luajit tui perf sched dynamic )
WARN: [14] (hermes::uninitializedCopy[6107db] <= hermes::vm::DynamicStringPrimitive::DynamicStringPrimitive[60cda1])
WARN: [13] (hermes::vm::DynamicStringPrimitive::DynamicStringPrimitive[60cd2f] <= hermes::vm::GCBase::constructCell[614324])
WARN: [12] (hermes::vm::GCBase::constructCell[6142c1] <= hermes::vm::HadesGC::makeA[6138a8])
WARN: [11] (hermes::vm::HadesGC::makeA[6137e9] <= hermes::vm::GCBase::makeA[612652])
WARN: [10] (hermes::vm::GCBase::makeA[6125c1] <= hermes::vm::GCBase::makeAVariable[61139d])
WARN: [9] (hermes::vm::GCBase::makeAVariable[611305] <= hermes::vm::Runtime::makeAVariable[610333])
WARN: [8] (hermes::vm::Runtime::makeAVariable[6102ed] <= hermes::vm::DynamicStringPrimitive::createLongLived[60cf1a])
WARN: [7] (hermes::vm::DynamicStringPrimitive::createLongLived[60ce8e] <= hermes::vm::StringPrimitive::createLongLived[5bef94])
WARN: [6] (hermes::vm::StringPrimitive::createLongLived[5bef2b] <= hermes::vm::Runtime::allocateCharacterString[5c92ca])
WARN: [5] (hermes::vm::Runtime::allocateCharacterString[5c9257] <= hermes::vm::Runtime::initCharacterStrings[5c91d1])
WARN: [4] (hermes::vm::Runtime::initCharacterStrings[5c90f7] <= hermes::vm::Runtime::Runtime[5c1486])
WARN: [3] (hermes::vm::Runtime::Runtime[5c019f] <= hermes::vm::Runtime::create[5bf77e])
WARN: [2] (hermes::vm::Runtime::create[5bf5e7] <= repl[417b10])
WARN: [1] (repl[417ad6] <= main[40bab8])
WARN: [0] (main[40b9ae] <= __libc_start_main[7f75cc3980b3])

```
With some debugging, I found that one of unnamed region make the crash as follows.
```
00400000-0102d000 r-xp 00000000 fd:02 1793899                            /home/honggyu/work/hermes/build/bin/hermes
0122c000-01235000 r--p 00c2c000 fd:02 1793899                            /home/honggyu/work/hermes/build/bin/hermes
01235000-01239000 rw-p 00c35000 fd:02 1793899                            /home/honggyu/work/hermes/build/bin/hermes
01239000-07d19000 rw-p 00000000 00:00 0                                  [heap]
52d5f000000-52d5f414000 rw-p 00000000 00:00 0
52d5f414000-52d5f415000 ---p 00000000 00:00 0
52d5f415000-52d5f814000 rw-p 00000000 00:00 0
52d5f814000-52d5f815000 ---p 00000000 00:00 0
52d5f815000-52d5fc00000 rw-p 00000000 00:00 0         <--- This region access makes the crash
52d5fc00000-52da0000000 ---p 00000000 00:00 0
```
I don't have a clear idea but the crashed region has a valid read permission but got crashed. I think the chances of accessing valid strings in unnamed regions are relatively low so wanted to avoid possible crashes.

namhyung: Unnamed regions can be (part of) heap so we cannot just skip.  It'd be  nice if we can debug what's the real issue.

In your above example, the address doesn't belong to the region.

namhyung: In my case, after checking the read permission I'm seeing another failure.

```
$ uftrace record -P. -a bin/hermes
hermes: hermes/lib/VM/gcs/HadesGC.cpp:3041: void hermes::vm::HadesGC::OldGen::setTargetSizeBytes(size_t): Assertion `!targetSizeBytes_ && "Should only initialise targetSizeBytes_ once."' failed.
WARN: child terminated by signal: 6: Aborted
```

namhyung: And checking the string length as your original patch doesn't work too.

honggyukim: Hmm.. That could be a different issue then.

namhyung: Ok, let's fix the permission check alone for now.  It's an obvious bug.

namhyung: Pushed aa24650f53.

bernhardkaindl:
@namhyung @honggyukim  Just sending a ping on this. It has a git push which has no comment (maybe this PR can be revived)

namhyung:
Note that the permission check is merged already.  The remaining part is the string length check which is arbitrary and I don't understand why it matters.

namhyung:
@honggyukim maybe we can close this for now if you don't work on it anymore.

honggyukim:
I have tested the same program, but it still gets crashed and can be fixed with this PR. This approach might be fragile but we better keep it open until we find the solid fix.

honggyukim:
I've just updated this PR only dropping already merged part.

PR #1481, opened 2022-07-17
trivial: Remove duplicated #include directive

JHH20:
Remove redundant `#include` in utils/report.c

There were a few other files with duplicated `#include` but I thought those should be kept. I'll list them below just for the record.

utils/kernel.c
- <sys/stat.h> L15, L1832
> The first is always included while the second is used for unit testing. Since the use case is different, being explicit seems better.

cmds/tui.c
- "uftrace.h" L14, L3033
- "utils/utils.h" L16, L3034
> The directives are specified once for `ifdef HAVE_LIBNCURSES` and once for `else`. Having dependencies contained for each condition would be better for future development in case the conditions become more complex.

misc/prototypes.h
- <sys/stat.h> L188, L496
> constant / function declarations are grouped by category, so keeping the directive serves as documentation for where the declarations originate from.

### DIFF
--- a/utils/report.c
+++ b/utils/report.c
@@ -1,7 +1,6 @@
 #include <stdio.h>
 #include <inttypes.h>
 #include <string.h>
-#include <inttypes.h>
 
 #include "uftrace.h"
 #include "utils/report.h"


honggyukim (approved):
Thanks. LGTM!

namhyung (approved):
LGTM.

Thanks for the analysis.  I think we can fix utils/kernel.c at least.  Please send a PR for that.

Issue #1448, opened 2022-07-01
add function size info to report field (graph, tui are optional)


honggyukim:
The function size info is not runtime information but it delivers useful ideas how big the function is.

We can add the function size info to report and graph field.

honggyukim:
@namhyung Please let me know if you don’t like the idea.

namhyung:
I Like the idea!

honggyukim:
The `uftrace_symbol` has `size` info so we can just print it in report and graph command output.
https://github.com/namhyung/uftrace/blob/79a7d940b484be95ded642f9c9a48925c2472281/utils/symbol.h#L50-L55

fistki:
I will try this issue!

fistki:
I add function size like this.
Values are just printing `size` of `uftrace_symbol`.

```
$ ~/func-size/uftrace report -d hello.data -f +func-size
  Total time   Self time       Calls   Func size  Function
  ==========  ==========  ==========  ==========  ====================
   48.661 us    0.320 us           1          45  _GLOBAL__sub_I_main
   48.341 us    1.433 us           1          92  __static_initialization_and_destruction_0
   46.797 us   46.797 us           1          16  std::ios_base::Init::Init
    7.755 us    0.211 us           1          46  main
    7.544 us    7.544 us           1          16  std::operator<<
    0.111 us    0.111 us           1          16  __cxa_atexit
```

Does it fit for purpose? @honggyukim

honggyukim:
@fistki Yes, that's what I meant. Could you upload a PR?

fistki:
> @fistki Yes, that's what I meant. Could you upload a PR?

i uploaded it #1495

PR #1495, opened 2022-07-27
report: Add "size" field to display function size

fistki:
Signed-off-by: Joonho Seo <tjwnsgh1110@gmail.com>


honggyukim:
Looks good, but could you make it work with `-s`/`--sort` option? You can have a look at https://github.com/namhyung/uftrace/blob/v0.12/utils/report.c#L164-L172.

And please add more description in the commit message. You better read other commit messages with `git log` for reference.

kangtegong:
One more thing please, if this commit resolved issue https://github.com/namhyung/uftrace/issues/1448,

please add `Closes: #issue number` at the bottom of your commit message like below.

```
commit 69ce2c186d8d5af6be6659c341637c2747f5c764
Author: Kang Minchul <tegongkang@gmail.com>
Date:   Tue Jan 25 16:08:48 2022 +0900

    uftrace: Better default message for uftrace

    Currently, when uftrace is used soely,
    there are no explanations for COMMANDS like below:

      COMMAND: record|replay|live|report|graph|info|dump|recv|script|tui"

    But I think it's better to put explanation for each commands like perf, git, etc
    for users who don't know uftrace well.

      ubuntu@ubuntu:~$ perf

       usage: perf [--version] [--help] [OPTIONS] COMMAND [ARGS]

       The most commonly used perf commands are:
         annotate        Read perf.data (created by perf record) and display annotated code
         archive         Create archive with object files with build-ids found in perf.data file
         bench           General framework for benchmark suites
         buildid-cache   Manage build-id cache.
         buildid-list    List the buildids in a perf.data file
         ...

    So this commit adds better explanation for commands when uftrace is used solely.

    Closes: #1348

```
 
This will automatically close the issue. Thanks for your contribution.

honggyukim:
@fistki Thanks for the update. The sorting works fine, but I have one more request. `uftrace report` also has a diff mode for two different records. We better have the diff mode for function size as well. Please have a look the following links.
- https://github.com/namhyung/uftrace/blob/v0.12/utils/report.c#L409-L417
- https://github.com/namhyung/uftrace/blob/v0.12/utils/report.c#L895-L923
- https://github.com/namhyung/uftrace/blob/master/doc/uftrace-report.md

utils/report.c
```diff
@@ -175,10 +178,10 @@ static struct sort_key sort_func = {
...
 };
 
-static struct sort_key *all_sort_keys[] = {
-	&sort_total,	&sort_total_avg, &sort_total_min, &sort_total_max, &sort_self,
-	&sort_self_avg, &sort_self_min,	 &sort_self_max,  &sort_call,	   &sort_func,
-};
+static struct sort_key *all_sort_keys[] = { &sort_total,     &sort_total_avg, &sort_total_min,
+					    &sort_total_max, &sort_self,      &sort_self_avg,
+					    &sort_self_min,  &sort_self_max,  &sort_call,
+					    &sort_func,	     &sort_func_size };
```
namhyung, line 184: Please check the codying style with pre-commit.

utils/report.c
```diff
@@ -761,6 +764,14 @@ void report_sort_tasks(struct uftrace_data *handle, struct rb_root *name_root,
...
 	FIELD_STRUCT(_id, _name, _func, _header, 10)
 
+#define FIELD_FUNC_SIZE(_id, _name, _field, _func, _header)                                        \
+	static void print_##_func(struct field_data *fd)                                           \
+	{                                                                                          \
+		struct uftrace_report_node *node = fd->arg;                                        \
+		pr_out("%10" PRIu64 "", node->_field);                                             \
+	}                                                                                          \
+	FIELD_STRUCT(_id, _name, _func, _header, 10)
+
```
namhyung, line 774: Can we reuse the existing `FIELD_CALL`?  Maybe you want to rename it to `FIELD_UINT` or something.

cmds/report.c
```diff
@@ -43,15 +43,19 @@ static void find_insert_node(struct rb_root *root, struct uftrace_task_reader *t
...
 
 	sym = task_find_sym_addr(&task->h->sessions, task, timestamp, addr);
 	if (needs_srcline)
 		loc = task_find_loc_addr(&task->h->sessions, task, timestamp, addr);
 
+	origin_sym = task->func;
+	task->func = sym;
 	symname = symbol_getname(sym, addr);
 	insert_node(root, task, symname, loc);
+	task->func = origin_sym;
```
namhyung, line 58: I don't understand why this is necessary.  Could you please add comments?

honggyukim:
I have one more thing, as we all know the information is for each function so how about drop `func-` prefix and use this field name as `size` and fix the related names in the code?

utils/report.c
fistki: Ok, I will try it. Thanks for advice!

cmds/report.c
fistki: That's because I had no idea of which cause side effect.
But now, I think that is unnecessary too.

utils/report.c
fistki: Fixed commit is updated. Please check it.

utils/report.c
```diff
@@ -175,10 +178,10 @@ static struct sort_key sort_func = {
...
 };
 
-static struct sort_key *all_sort_keys[] = {
-	&sort_total,	&sort_total_avg, &sort_total_min, &sort_total_max, &sort_self,
-	&sort_self_avg, &sort_self_min,	 &sort_self_max,  &sort_call,	   &sort_func,
-};
+static struct sort_key *all_sort_keys[] = { &sort_total,     &sort_total_avg, &sort_total_min,
+					    &sort_total_max, &sort_self,      &sort_self_avg,
+					    &sort_self_min,  &sort_self_max,  &sort_call,
+					    &sort_func,	     &sort_size };
```
namhyung, line 184: Please follow the existing coding style like
```
static struct sort_key *all_sort_keys[] = {
   ...
};
```

The open and closing braces are on separate lines.

namhyung (commented):
Thanks for working on this.

utils/report.c
fistki: pre-commit automatically modify separate lines to one line.

```
Check Yaml...........................................(no files to check)Skipped
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
isort................................................(no files to check)Skipped
clang-format.............................................................Failed
- hook id: clang-format
- files were modified by this hook
fistki@uftrace ~/func-size(func-size)$ git diff
diff --git a/utils/report.c b/utils/report.c
index 62a8b94d..834cec22 100644
--- a/utils/report.c
+++ b/utils/report.c
@@ -181,8 +181,7 @@ static struct sort_key sort_func = {
 static struct sort_key *all_sort_keys[] = { &sort_total,     &sort_total_avg, &sort_total_min,
                                            &sort_total_max, &sort_self,      &sort_self_avg,
                                            &sort_self_min,  &sort_self_max,  &sort_call,
-                                           &sort_func,      &sort_size
-};
+                                           &sort_func,      &sort_size };
```
What should I do?

honggyukim: Let's keep it as is. We should find if there is a such rule in clang-format configs.
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

I can't find any so far.

namhyung: Hmm.. I'm a bit surprised that they don't have 'Linux' base style yet.

namhyung: It looks like they don't have a specific option for this.  But I also found that trailing comma (after `&sort_size` in this case) could do the trick.

https://stackoverflow.com/questions/47409797/how-to-wrap-struct-initializer-in-clang-format

fistki: It works. Thanks!

utils/report.c
```diff
@@ -407,7 +412,7 @@ static struct diff_key sort_diff_func = {
 static struct diff_key *all_diff_keys[] = {
 	&sort_diff_total, &sort_diff_total_avg, &sort_diff_total_min, &sort_diff_total_max,
 	&sort_diff_self,  &sort_diff_self_avg,	&sort_diff_self_min,  &sort_diff_self_max,
-	&sort_diff_call,  &sort_diff_func,
+	&sort_diff_call,  &sort_diff_func,	&sort_diff_size
```
namhyung, line 415: I think it also needs a trailing comma for clang-format.

utils/report.c
```diff
@@ -753,7 +758,7 @@ void report_sort_tasks(struct uftrace_data *handle, struct rb_root *name_root,
 	}                                                                                          \
 	FIELD_STRUCT(_id, _name, _func, _header, 10)
 
-#define FIELD_CALL(_id, _name, _field, _func, _header)                                             \
+#define FIELD_UNIT(_id, _name, _field, _func, _header)                                             \
```
namhyung, line 761: It's not about 'unit' as there's no such thing here.  What I suggested was 'uint' (yeah, it's confusing) which means 'unsigned int'.

fistki: I was confused. I updated it.

namhyung (approved):
LGTM

PR #1496, opened 2022-07-27
misc: update description about libunwind support

JSYoo5B:
On building uftrace from the scratch, `./configure` shows dependent
library install status. `misc/install-deps.sh` will automatically
install those libraries with each distribution's package manager.

But, current `misc/install-deps.sh` doesn't install libunwind.
This is especially optional package for debugging. (though other
packages are also optional packages.) Some users may want to build
uftrace with full feature support. So add `--full` option to install
all libraries which checked by `./configure`.
Additionally, add usage message to show about `--full` options.

Also, change description about `libunwind`. It supports backtracing
information, which is optional for debugging.

Reported-by: Lee Dong Wook <sh95119@gmail.com>
Signed-off-by: JaeSang Yoo <jsyoo5b@gmail.com>

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -85,7 +85,7 @@ It will show the prefix directory and detected features like:
     ...     perf_event: [ on  ] - perf (PMU) event support
     ...       schedule: [ on  ] - scheduler event support
     ...       capstone: [ on  ] - full dynamic tracing support
-    ...      libunwind: [ OFF ] - libunwind support
+    ...      libunwind: [ OFF ] - stacktrace support (optional for debugging)
 
 Then you can run `make` to build the source.
 
--- a/configure
+++ b/configure
@@ -260,7 +260,7 @@ print_feature "cxa_demangle" "cxa_demangle" "full demangler support with libstdc
 print_feature "perf_event" "perf_clockid" "perf (PMU) event support"
 print_feature "schedule" "perf_context_switch" "scheduler event support"
 print_feature "capstone" "have_libcapstone" "full dynamic tracing support"
-print_feature "libunwind" "have_libunwind" "libunwind support"
+print_feature "libunwind" "have_libunwind" "stacktrace support (optional for debugging)"
 
 cat >$output <<EOF
 # this file is generated automatically


JSYoo5B:
I didn't realized `$OPT` were passed into package manager.

I'll fix and re-submit it.

kangtegong:
Thanks for your contribution. While you amend your commit, can you also write `--full` option in docs (both English, Korean) as well?

JSYoo5B:
> Thanks for your contribution. While you amend your commit, can you also write `--full` option in docs (both English, Korean) as well?

I'm not sure which file to write this `--full` options (especially in Korean)

I've seen [./INSTALL.md](https://github.com/namhyung/uftrace/blob/master/INSTALL.md) is the install documentation in english. But I cannot find Korean version.

Also, `./configure` [example output](https://github.com/namhyung/uftrace/blob/master/INSTALL.md) need to be updated because I changed libunwind's description message.

JSYoo5B:
By the way, I think additional github action for other distribution may required.

As previous [github action](https://github.com/namhyung/uftrace/actions/runs/2747966061) failed, it may occur in other distros too.

kangtegong:
I initially thought writing 

```
Also, If you want optional package for debugging, you can use `--full` option like below:

    $ sudo misc/install-deps.sh --full
```

in these locations, but I think we need more opinion about this..

https://github.com/namhyung/uftrace/blob/07332935deae59231e3305743be2cf3554fe1a32/README.md?plain=1#L50

https://github.com/namhyung/uftrace/blob/07332935deae59231e3305743be2cf3554fe1a32/doc/ko/README.md?plain=1#L49

And I also think output of `./configure` should be updated in INSTALL.md, like below too. Can you please change the doc?

```
...         libelf: [ on  ] - more flexible ELF data handling
...          libdw: [ on  ] - DWARF debug info support
...      libpython: [ on  ] - python scripting support
...      libluajit: [ on  ] - luajit scripting support
...    libncursesw: [ on  ] - TUI support
...   cxa_demangle: [ on  ] - full demangler support with libstdc++
...     perf_event: [ on  ] - perf (PMU) event support
...       schedule: [ on  ] - scheduler event support
...       capstone: [ on  ] - full dynamic tracing support
...      libunwind: [ on  ] - backtrace support (optional for debugging)
```

And I also saw github action failing. I think that should be changed as soon as possible.

honggyukim:
We don’t have to add this option in the doc because it’s just for specifics developers only.

JSYoo5B:
> We don’t have to add this option in the doc because it’s just for specifics developers only.

Then, I'll just update `./configure` output example with updated version.

honggyukim:
@JSYoo5B Sorry but I think it's unlikely to have many more packages that are used only for debugging so I think this change is not needed. Besides that, could you please keep the description change for `libunwind`, which is from `libunwind support` to `backtrace support (optional for debugging)`?.

JSYoo5B:
@honggyukim PR title & commit are updated. :)

honggyukim:
Ah. sorry, could you please change `backtrace` to `stacktrace`. We have a function name called `stacktrace()` at https://github.com/namhyung/uftrace/blob/fb495a8c83ff646577ecd83cd3058553ddc07b63/utils/utils.c#L927.

honggyukim (approved):
Thanks for the update. LGTM.

PR #1501, opened 2022-07-30
info: Split uftrace_info_bits to index and mask

honggyukim:
Like in enum uftrace_feat_bits, it should have separate bit index and
mask to reduce possible confustions.

Recent test failure was because of this confusion and it was painful to
discover the root cause, but fixed in the following commit anyway.

  41d08aed test: Use info bit mask properly in the data_basic test

Fixed: #1444


namhyung (approved):
LGTM

PR #1506, opened 2022-08-02
dump: Do not print perf-cpuXX and kernel-cpuXX if empty

honggyukim:
If there are many cores in the system, most of the cpus stays empty when
recording perf and kernel data.

If the files are empty, do not print "reading {perf,kernel}-cpuXX.dat"
messages in uftrace dump output.


namhyung (approved):
LGTM

Issue #1078, opened 2019-12-29
support non multiprocessing runtest.py


honggyukim:
It seems that the test speed is a lot faster with multiprocessing pool, but it makes the result unreliable sometime. I've seen many random failures as I posted it in #1063.

In addition to having `runtest.py` with multiprocessing pool, we better have a way to run test without it when we really have to see what tests are exactly failed.

honggyukim:
@DanielTimLee Could you please help us if you can find some time to work on this?

namhyung:
Isn't `runtest.py -j 1` enough?

honggyukim:
That doesn't seem to fix random failures.

honggyukim:
Please see #1063. They are all tested with a single worker.

DanielTimLee:
I'll look into it and get back to #1063.

DanielTimLee:
```bash
for i in {1..10}; do ./runtest.py 226 > $(mktemp /tmp/uftrace-test.XXXX); done;
A=( $(ls /tmp/uftrace-test*) )
for i in ${A[@]}; do diff ${A[0]} $i; done
```

honggyukim:
Sometimes the test gets stuck and does not make progress, so I really need this non-multiprocessing execution is needed.

DanielTimLee:
Sorry, I'm in the middle of something, so I can't get to the issue right now.

namhyung:
Feel free to do it anytime you feel comfortable. :)

namhyung:
You can specify the test as `serial=True` like in test 22 (filter kernel).

Issue #1063, opened 2019-12-14
random failure in tests


honggyukim:
It seems very weird that some tests are randomly failed. For example, 226 test result is different for each execution.
```
$ ./runtest.py 226
Start 1 tests with 1 worker
Test case                 pg             finstrument-fu
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
226 default_opts        : OK OK OK OK OK OK NG OK OK OK
...

$ ./runtest.py 226
Start 1 tests with 1 worker
Test case                 pg             finstrument-fu
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
226 default_opts        : NG OK OK OK NG OK NG NG OK NG
...

$ ./runtest.py 226
Start 1 tests with 1 worker
Test case                 pg             finstrument-fu
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
226 default_opts        : OK OK OK OK OK OK OK NG NG OK
...

$ ./runtest.py 226
Start 1 tests with 1 worker
Test case                 pg             finstrument-fu
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os
226 default_opts        : OK NG NG OK NG OK OK OK OK OK
...
```
@DanielTimLee It seems that this problem has been seen since #570 was merged. Could you please take a look at this problem?

honggyukim:
Its result is different even when executing only a single test.
```
$ ./runtest.py 226 -p -O1
Start 1 tests with 1 worker
Test case                 pg
------------------------: O1
226 default_opts        : OK
...

$ ./runtest.py 226 -p -O1
Start 1 tests with 1 worker
Test case                 pg
------------------------: O1
226 default_opts        : NG
```

honggyukim:
The failed result is as follows:
```diff
$ ./runtest.py 226 -vdp -O1
...
===========  result  ===========
main() {
   usleep();
   foo() {
     usleep();
   } /* foo */
 } /* main */
=========== expected ===========
main() {
   foo() {
     usleep();
   } /* foo */
 } /* main */
t226_default_opts: diff result of -pg -O1
--- expect      2019-12-14 09:17:04.977697525 +0900
+++ result      2019-12-14 09:17:04.977697525 +0900
@@ -1,2 +1,3 @@
 main() {
+   usleep();
    foo() {

226 default_opts        : NG
```

DanielTimLee:
[...]

PR #1510, opened 2022-08-05
test: Support non-multiprocessing runtest.py

JHH20:
Disable multiprocessing pool explicitly with -j option
- Not including -j option will behave as before, even if the actual `arg.worker` value is set to 1 later
- Mode reflected in console message as `"with %d worker" -> "without worker pool"`

dotdict class (L849-L853) from https://stackoverflow.com/a/23689767
- To minimize code changes that access the `shared` dict
- Mentioning reference just in case of licensing complications

Addresses #1078 but did not solve the problem raised in that issue (#1063)

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -822,6 +822,9 @@ if __name__ == "__main__":
 
     arg = parse_argument()
 
+    # Arg is checked before being bounded to be explicitly toggleable
+    use_pool = arg.worker > 1
+
     if arg.case == 'all':
         testcases = glob.glob('t???_*.py')
     else:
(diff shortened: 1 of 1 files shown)


kangtegong:
Thanks for your patch,
but whenever your patch is ran with 1 worker like below,

`$ python3 ./runtests.py -j1`

output of the result is not printed during the test 
but printed after the whole test is ran...

JHH20:
Oh, I did not fully consider the case when running multiple testcases.
I tried to avoid changing too much of the test code, so all the tests are started (and completed since serial) before any printing begins, but it looks like fixing this side effect shouldn't require too much change.

I will push another commit which prints after each testcase completes. If this new version is preferred, I will squash the commits to reflect the newer change.

namhyung (approved):
LGTM

Issue #1499, opened 2022-07-30
create a single root node for graphviz dump output


honggyukim:
The current `--graphviz` output for `t-abc` is as follows.
```dot
$ uftrace record t-abc

$ uftrace dump --graphviz
      ...
digraph "t-abc" {
    "__monstartup" [xlabel = "1"]
    "__cxa_atexit" [xlabel = "1"]
    "main" [xlabel = "1"]
    "main" -> "a" [xlabel = "1"]
    "a" -> "b" [xlabel = "1"]
    "b" -> "c" [xlabel = "1"]
    "c" -> "getpid" [xlabel = "1"]
}
```
It has 3 root nodes in total - `__monstartup`, `__cxa_atexit`, and `main` and the output image is as follows.
<img width="452" alt="image" src="https://user-images.githubusercontent.com/19642328/181993170-7ad84281-9e67-43c3-9d52-3b06ff475ffe.png">

But rather than creating multiple root nodes, we better create a single root node with its `basename(exename)`. i.e. `t-abc` in this example. So the output can be rewritten as follows.
```dot
digraph G {
    "t-abc"
    "t-abc" -> "__monstartup" [xlabel = "1"]
    "t-abc" -> "__cxa_atexit" [xlabel = "1"]
    "t-abc" -> "main" [xlabel = "1"]
    "main" -> "a" [xlabel = "1"]
    "a" -> "b" [xlabel = "1"]
    "b" -> "c" [xlabel = "1"]
    "c" -> "getpid" [xlabel = "1"]
}
```
Then this new output image can be shown as follows.
<img width="442" alt="image" src="https://user-images.githubusercontent.com/19642328/181994441-335d1871-6f5b-4166-a268-9167c5540dcb.png">

Similarly, `uftrace graph` command also creates a fake root node `t-abc` as follows.
```
$ uftrace graph
# Function Call Graph for 't-abc' (session: 721637281f8729ec)
========== FUNCTION CALL GRAPH ==========
# TOTAL TIME   FUNCTION
    7.616 us : (1) t-abc
    1.656 us :  +-(1) __monstartup
             :  |
    1.032 us :  +-(1) __cxa_atexit
             :  |
    4.928 us :  +-(1) main
    4.413 us :    (1) a
    3.895 us :    (1) b
    3.236 us :    (1) c
    2.023 us :    (1) getpid
```

honggyukim:
The name of the root node is the `basename(exename)` as follows.
[...]

PR #1511, opened 2022-08-06
graph: Add the option to show function call graph as mermaid diagram

wendyisdream:
This option shows the function call graph as a mermaid flowchart
for visualization.

- mermaid page : https://mermaid-js.github.io/mermaid/

The usage for this option is add --mermaid in graph command.

The result can be rendered in the browser so it is formatted as HTML.
The mermaid syntax itself can be rendered in other services
including a mermaid plugin like github or draw.io.

The function name cannot be used as a node id because some suffix
like 'end', 'self' is indicated to special usage in mermaid.
So 'id' is added to identify the node to 'struct uftrace_graph_node'.

Signed-off-by: Jia Ha <wendyisdream1004@gmail.com>

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -82,6 +82,10 @@ struct uftrace_graphviz_dump {
 	struct uftrace_dump_ops ops;
 };
 
+struct uftrace_mermaid_dump {
+	struct uftrace_dump_ops ops;
+};
+
 static const char *rstack_type(struct uftrace_record *frs)
 {
 	return frs->type == UFTRACE_EXIT  ? "exit " :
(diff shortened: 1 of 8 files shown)


honggyukim:
It looks nice. Thanks very much for doing this work!
image

I have two comments on this work.
1. I understand this work is more related to `graph` than `dump`. If this is implemented in `graph` command, then `--graphviz` is also better to be moved in `graph` command. Moreover, I would like to hear how @namhyung thinks about this.

2. `uftrace graph` has a partial graph mode. The usage of `graph` is as follows.
```
    uftrace graph [options] [FUNCTION]
```
Let's say there is an example program and it can be compiled and recorded as follows.
```
$ gcc -pg -o t-racycount tests/s-racycount.c -pthread

$ uftrace record t-racycount
```
The `uftrace graph` shows its call graph from the top.
```
$ uftrace graph
# Function Call Graph for 'a.out' (session: 240a6ef673ec8bcf)
# TOTAL TIME   FUNCTION
    1.864 ms : (1) a.out
    2.130 us :  +-(1) __monstartup
             :  |
    1.163 us :  +-(1) __cxa_atexit
             :  |
    1.134 ms :  +-(1) main
    1.273 us :  |  +-(1) pthread_barrier_init
             :  |  |
  358.416 us :  |  +-(2) pthread_create
             :  |  |
  639.737 us :  |  +-(1) racy_count
  306.615 us :  |  | (1) pthread_barrier_wait
  294.710 us :  |  | (1) linux:schedule
             :  |  |
  128.560 us :  |  +-(2) pthread_join
  115.741 us :  |    (1) linux:schedule
             :  |
  726.979 us :  +-(2) thread_fn
  720.392 us :    (2) racy_count
   94.733 us :    (1) linux:schedule
```
If `FUNCTION` is given as `racy_count`, it builds a call graph from the `FUNCTION` and it also collects all the backtrace that reaches to the `FUNCTION`.
```
$ uftrace graph racy_count
# Function Call Graph for 'racy_count' (session: 240a6ef673ec8bcf)
=============== BACKTRACE ===============
 backtrace #0: hit 2, time 720.392 us
   [0] thread_fn (0x4009a5)
   [1] racy_count (0x40095f)

 backtrace #1: hit 1, time 639.737 us
   [0] main (0x4009c9)
   [1] racy_count (0x40095f)

# TOTAL TIME   FUNCTION
    1.360 ms : (3) racy_count
  306.615 us :  +-(1) pthread_barrier_wait
  294.710 us :  | (1) linux:schedule
             :  |
   94.733 us :  +-(1) linux:schedule
```
This case has to be covered in `--mermaid` option if it's used in `graph` command.

namhyung:
I think it's better to have it in dump command as it's for external tools.

wendyisdream:
@honggyukim, @namhyung Thank you for the review. 
Okay, I will move this option to dump command part. 
Oh! uftrace have the partial graph mode!
I will try to include it in this option.

wendyisdream:
I updated the code to move the mermaid option to the dump command.
I tried to add this option following the existing dump option's format to remove this PR's other effect.
I added the code to be shown as a single root to graphviz and mermaid which is raised in #1499.

In dump command, adding a partial graph mode and handling multiple sessions, and refactoring duplicate code between commands is needed but I think it will be better to handle in each separate PR.

namhyung:
In dump command, we don't support the partial graph.  But we can use filters to limit the output.

honggyukim:
> I added the code to be shown as a single root to graphviz and mermaid which is raised in https://github.com/namhyung/uftrace/issues/1499.

Thanks for the update, but it's not a good idea to handle two different tasks in a single commit. Please separate them in different patches.

wendyisdream:
I deleted the update in graphviz to separate from #1499. :)
Thank you for the review!

cmds/dump.c
```diff
@@ -1244,6 +1248,126 @@ static void dump_graphviz_footer(struct uftrace_dump_ops *ops, struct uftrace_da
...
+	if (opts->caller)
+		pr_out(" /caller: %s", opts->caller);
+	if (opts->hide)
+		pr_out(" /hide: %s", opts->hide);
+	if (opts->tid)
+		pr_out(" /tid: %s", opts->tid);
+	if (opts->depth != OPT_DEPTH_DEFAULT)
+		pr_out(" /depth: %d", opts->depth);
+	if (opts->threshold) {
+		pr_out(" /threshold:");
```
namhyung, line 1345: Please print it as "time-filter" as it's the name of the option.

utils/graph.c
```diff
@@ -87,6 +88,7 @@ static int add_graph_entry(struct uftrace_task_graph *tg, char *name, size_t nod
 
 		node = xzalloc(node_size);
 
+		node->id = next_id++;
```
namhyung, line 91: Although this is a small change, I think it's better to split it as a separate commit because it's a general change.  Please add a comment to `graph_add_node()` function that it's not thread-safe due to the static id.

wendyisdream:
Thank you for the review!
I modified it as commented and separated this PR into three commits.

b518c3c2 dump: Add a single root node as the executed file in graphviz
9bc8bc38 dump: Add the option to show function call graph as mermaid diagram
493ea934 graph: Add id to uftrace_graph_node to identify graph node

cmds/dump.c
```diff
@@ -1250,6 +1254,126 @@ static void dump_graphviz_footer(struct uftrace_dump_ops *ops, struct uftrace_da
...
+static void dump_mermaid_header(struct uftrace_dump_ops *ops, struct uftrace_data *handle,
+				struct uftrace_opts *opts)
+{
+	graph_init_callbacks(NULL, NULL, NULL, ops);
+	mermaid_graph.root.name = basename(handle->info.exename);
+
+	pr_out("<html>\n");
+	pr_out("  <body>\n");
+	pr_out("    <script src=\"https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js\"></script>\n");
+	pr_out("    <script>\n");
```
honggyukim, line 1312: We don't need to add whitespaces for indentation for html tags. Please remove them here and rest of the other places.

I'm okay with having spaces for javascript code.

cmds/dump.c
```diff
@@ -1250,6 +1254,126 @@ static void dump_graphviz_footer(struct uftrace_dump_ops *ops, struct uftrace_da
...
+}
+static void dump_mermaid_footer(struct uftrace_dump_ops *ops, struct uftrace_data *handle,
+				struct uftrace_opts *opts)
+{
+	/* skip empty graph */
+	if (mermaid_graph.root.time == 0 && mermaid_graph.root.nr_edges == 0)
+		return;
+
+	if (mermaid_graph.root.time || mermaid_graph.root.nr_edges) {
+		pr_out("    <h2>Function Call Graph for <span style=\"color:blue\">%s</span>",
```
honggyukim, line 1366: Please remove spaces before `<h2`>.

honggyukim (commented):
Thanks for the update. It looks good in general, but I have a few more requests.

utils/graph.c
wendyisdream: Thank you for the review!
I added the comment for it to line 210 because graph_add_node is a public interface for others.
Do you think it is better to move it here?
Except it, I updated the code.

wendyisdream: > > Please add a comment to graph_add_node() function that it's not thread-safe due to the static id.
> 
> Please address @namhyung's comment above.

Thank you for the review!
I added the comment for utils/graph.c to line 210 because graph_add_node is a public interface for others.
Do you think it is better to move it here?
Except for it, I updated the code.

honggyukim:
Thanks for the update. It's almost done, but could you add a test in `tests/t272_dump_mermaid`. You can make it similar as `tests/t101_dump_chrome.py`.

wendyisdream:
Testcase 272 is updated to 5aad7f4 commit. Thank you for informing of it!

cmds/dump.c
```diff
@@ -1250,6 +1254,104 @@ static void dump_graphviz_footer(struct uftrace_dump_ops *ops, struct uftrace_da
...
+	pr_out("</script>\n");
+}
+
+static void print_graph_node_mermaid(struct uftrace_graph *graph, struct uftrace_graph_node *node)
+{
+	char *symname = node->name;
+	struct uftrace_graph_node *child;
+
+	if (node->addr == EVENT_ID_PERF_SCHED_IN)
+		symname = "linux:schedule";
```
namhyung, line 1328: I don't think we need this.  It should be handled by `graph_add_node()` already.

namhyung (commented):
Mostly looks good.  A nit below..

cmds/dump.c
wendyisdream: I applied the code to delete this :)

namhyung (commented):
Looks good.  Could you please add an example output (and the comment line) in the second commit message even if it's in the test case?  I think you can just copy the same output of the test case.  Thanks!

wendyisdream:
I added the runtest.py result to the commit message. :)

wendyisdream:
I deleted unnecessary brackets in a python pattern expression.

namhyung (approved):
LGTM

PR #1270, opened 2021-05-29
tui: Customize fields in report mode interactively

qpakzk:
Hello @namhyung @honggyukim.

I implemented fields menu for report mode in tui.
Review this commit please.

This commit allows users to customize fields for report mode in interactive mode.
All mechanisms are the same as graph mode as described in commit 3d65ed4.

The report command has 9 fields: total time, total avg, total max, total min,
self time, self avg, self max, self min, call.
So users can customize 9 fields.
The default fields are total time, self time, and call.

Depending on the session where you are, you can use different fields menu when you press `f` key.
If you're in the graph session, graph fields menu appears.
If you're in the report session, report fields menu appears.
Except two sessions, you cannot see any menu windows even though you press `f` key.

PR #1513, opened 2022-08-07
tui: add avg, min, max fields to tui graph

dongwooklee96:
add total-avg, total-min, total-max, self-avg, self-min, self-max fields
with tui -f option

Signed-off-by: Lee Dong Wook <sh95119@gmail.com>

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1083,7 +1083,7 @@ static void print_flame_graph(struct uftrace_dump_ops *ops, struct uftrace_graph
 	unsigned long sample = node->nr_calls;
 
 	if (sample && flame->sample_time)
-		sample = (node->time - node->child_time) / flame->sample_time;
+		sample = (node->total_time.sum - node->child_time) / flame->sample_time;
 
 	if (sample) {
 		struct uftrace_graph_node *parent = node;
@@ -1336,7 +1336,7 @@ static void dump_mermaid_footer(struct uftrace_dump_ops *ops, struct uftrace_dat
 	const char *interactive_option_js;
 
 	/* skip empty graph */
-	if (mermaid_graph.root.time == 0 && mermaid_graph.root.nr_edges == 0)
+	if (mermaid_graph.root.total_time.sum == 0 && mermaid_graph.root.nr_edges == 0)
 		return;
 
 	pr_out("<h2>Function Call Graph for <span style=\"color:blue\" id=\"baseName\">%s</span>",
--- a/cmds/graph.c
+++ b/cmds/graph.c
@@ -45,7 +45,7 @@ static void print_total_time(struct field_data *fd)
 	struct uftrace_graph_node *node = fd->arg;
 	uint64_t d;
 
-	d = node->time;
+	d = node->total_time.sum;
 
 	print_time_unit(d);
 }
@@ -55,7 +55,7 @@ static void print_self_time(struct field_data *fd)
 	struct uftrace_graph_node *node = fd->arg;
 	uint64_t d;
 
-	d = node->time - node->child_time;
+	d = node->total_time.sum - node->child_time;
 
 	print_time_unit(d);
 }
@@ -489,7 +489,8 @@ static int print_graph(struct session_graph *graph, struct uftrace_opts *opts)
 	bool *indent_mask;
 
 	/* skip empty graph */
-	if (list_empty(&graph->bt_list) && graph->ug.root.time == 0 && graph->ug.root.nr_edges == 0)
+	if (list_empty(&graph->bt_list) && graph->ug.root.total_time.sum == 0 &&
+	    graph->ug.root.nr_edges == 0)
 		return 0;
 
 	pr_out("# Function Call Graph for '%s' (session: %.16s)\n", graph->func,
@@ -503,7 +504,7 @@ static int print_graph(struct session_graph *graph, struct uftrace_opts *opts)
 	setup_field(&output_fields, opts, &setup_default_field, field_table,
 		    ARRAY_SIZE(field_table));
 
-	if (graph->ug.root.time || graph->ug.root.nr_edges) {
+	if (graph->ug.root.total_time.sum || graph->ug.root.nr_edges) {
 		pr_out("========== FUNCTION CALL GRAPH ==========\n");
 		print_header(&output_fields, "# ", "FUNCTION", 2, false);
 		if (!list_empty(&output_fields)) {
@@ -695,8 +696,8 @@ static void build_graph(struct uftrace_opts *opts, struct uftrace_data *handle, 
 		struct uftrace_graph_node *node;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
-			graph->ug.root.time += node->time;
-			graph->ug.root.child_time += node->time;
+			graph->ug.root.total_time.sum += node->total_time.sum;
+			graph->ug.root.child_time += node->total_time.sum;
 		}
 
 		graph = graph->next;
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -266,7 +266,7 @@ static void print_graph_total(struct field_data *fd)
 	struct uftrace_graph_node *node = fd->arg;
 	uint64_t d;
 
-	d = node->time;
+	d = node->total_time.sum;
 
 	print_time(d);
 }
@@ -276,7 +276,7 @@ static void print_graph_self(struct field_data *fd)
 	struct uftrace_graph_node *node = fd->arg;
 	uint64_t d;
 
-	d = node->time - node->child_time;
+	d = node->total_time.sum - node->child_time;
 
 	print_time(d);
 }
@@ -687,7 +687,7 @@ static void copy_graph_node(struct uftrace_graph_node *dst, struct uftrace_graph
 		}
 
 		node->n.addr = child->addr;
-		node->n.time += child->time;
+		node->n.total_time.sum += child->total_time.sum;
 		node->n.child_time += child->child_time;
 		node->n.nr_calls += child->nr_calls;
 
@@ -736,8 +736,8 @@ static struct tui_graph *tui_graph_init(struct uftrace_opts *opts)
 		top->nr_calls = 1;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
-			top->time += node->time;
-			top->child_time += node->time;
+			top->total_time.sum += node->total_time.sum;
+			top->child_time += node->total_time.sum;
 		}
 
 		tui_window_init(&graph->win, &graph_ops);
@@ -796,7 +796,7 @@ static void build_partial_graph(struct tui_report_node *root_node, struct tui_gr
 	root->n.name = str;
 	root->n.parent = NULL;
 
-	root->n.time = 0;
+	root->n.total_time.sum = 0;
 	root->n.child_time = 0;
 	root->n.nr_calls = 0;
 
@@ -817,7 +817,7 @@ static void build_partial_graph(struct tui_report_node *root_node, struct tui_gr
 			tmp = append_graph_node(&tmp->n, target, parent->n.name);
 
 			tmp->n.addr = parent->n.addr;
-			tmp->n.time = node->n.time;
+			tmp->n.total_time.sum = node->n.total_time.sum;
 			tmp->n.child_time = node->n.child_time;
 			tmp->n.nr_calls = node->n.nr_calls;
 
@@ -843,7 +843,7 @@ static void build_partial_graph(struct tui_report_node *root_node, struct tui_gr
 			continue;
 
 		root->n.addr = node->n.addr;
-		root->n.time += node->n.time;
+		root->n.total_time.sum += node->n.total_time.sum;
 		root->n.child_time += node->n.child_time;
 		root->n.nr_calls += node->n.nr_calls;
 
@@ -1352,8 +1352,8 @@ static bool win_longest_child_graph(struct tui_window *win, void *node)
 
 	list_for_each_entry(child, &curr->n.head, n.list) {
 		fold_graph_node(child, true, false, 0);
-		if (longest_child_time < child->n.time) {
-			longest_child_time = child->n.time;
+		if (longest_child_time < child->n.total_time.sum) {
+			longest_child_time = child->n.total_time.sum;
 			longest_child = child;
 		}
 	}
--- a/utils/graph.c
+++ b/utils/graph.c
@@ -172,7 +172,7 @@ static int add_graph_exit(struct uftrace_task_graph *tg)
 	}
 
 out:
-	node->time += fstack->total_time;
+	node->total_time.sum += fstack->total_time;
 	node->child_time += fstack->child_time;
 
 	if (exit_cb)
@@ -386,7 +386,7 @@ TEST_CASE(graph_basic)
 	node = graph_find_node(&graph.root, data[1].addr);
 	TEST_NE(node, NULL);
 	TEST_STREQ(node->name, data[1].name);
-	TEST_EQ(node->time, data[1].total_time);
+	TEST_EQ(node->total_time.sum, data[1].total_time);
 	TEST_EQ(node->child_time, data[1].child_time);
 	TEST_EQ(node->nr_calls, 1);
 
@@ -394,7 +394,7 @@ TEST_CASE(graph_basic)
 	node = graph_find_node(&graph.root, data[5].addr);
 	TEST_NE(node, NULL);
 	TEST_STREQ(node->name, data[5].name);
-	TEST_EQ(node->time, data[5].total_time);
+	TEST_EQ(node->total_time.sum, data[5].total_time);
 	TEST_EQ(node->child_time, data[5].child_time);
 	TEST_EQ(node->nr_calls, 1);
 
@@ -402,7 +402,7 @@ TEST_CASE(graph_basic)
 	node = graph_find_node(node, data[4].addr);
 	TEST_NE(node, NULL);
 	TEST_STREQ(node->name, data[4].name);
-	TEST_EQ(node->time, data[4].total_time);
+	TEST_EQ(node->total_time.sum, data[4].total_time);
 	TEST_EQ(node->child_time, data[4].child_time);
 	TEST_EQ(node->nr_calls, 1);
 
--- a/utils/graph.h
+++ b/utils/graph.h
@@ -9,12 +9,16 @@
 #include "utils/list.h"
 #include "utils/rbtree.h"
 
+struct graph_time_stat {
+	uint64_t sum;
+};
+
 struct uftrace_graph_node {
 	uint64_t addr;
 	char *name;
 	int nr_edges;
 	int nr_calls;
-	uint64_t time;
+	struct graph_time_stat total_time;
 	uint64_t child_time;
 	uint32_t id;
 	struct list_head head;


honggyukim:
Please fix pre-commit and unit test errors.

cmds/dump.c
```diff
@@ -1058,8 +1058,8 @@ static void adjust_fg_time(struct uftrace_task_graph *tg, void *arg)
 	 */
 	accounted_time = (curr_time / sample_time) * sample_time;
 
-	node->parent->child_time -= curr_time;
-	node->parent->child_time += accounted_time;
+	node->parent->self_time.sum -= curr_time;
+	node->parent->self_time.sum += accounted_time;
```
namhyung, line 1062: This looks suspicious... why did you replace child time with self time?

cmds/tui.c
```diff
@@ -287,6 +344,36 @@ static struct display_field graph_field_total = {
 	.list = LIST_HEAD_INIT(graph_field_total.list),
 };
 
+static struct display_field graph_field_total_avg = {
+	.id = GRAPH_F_TOTAL_AVG,
+	.name = "total-avg",
+	.alias = "tot-avg",
```
namhyung, line 350: Aliases can be short like 'tavg'.

namhyung (commented):
This does many things in a single commit.  Please separate it like
* add graph_time_stat and set the fields properly
* add output fields for the new stats

cmds/dump.c
dongwooklee96: I changed the name to match the term with `uftrace_report_node`.
```
struct uftrace_report_node {
    char *name;
    struct report_time_stat total;
    struct report_time_stat self;
    struct uftrace_dbg_loc *loc;
    uint64_t call;
    struct rb_node name_link;
    struct rb_node sort_link;

    /* used by diff */
    struct uftrace_report_node *pair;
};
```
https://github.com/namhyung/uftrace/blob/master/utils/report.h#L25-L36

cmds/tui.c
dongwooklee96: I will apply other fields with the same rules.

dongwooklee96:
@namhyung I separated my commit

honggyukim:
@dongwooklee96 Your two patches have the same commit message title, but does different tasks. Please update it so that we can understand what the change does.

cmds/dump.c
dongwooklee96: Okay, I'll change it to that.

utils/graph.h
```diff
@@ -9,13 +9,20 @@
...
+
 struct uftrace_graph_node {
 	uint64_t addr;
 	char *name;
 	int nr_edges;
 	int nr_calls;
-	uint64_t time;
-	uint64_t child_time;
+	struct graph_time_stat total_time;
+	struct graph_time_stat child_time;
```
namhyung, line 25: I don't think we need a stat for child time.  The current code is just to get the sum of self time easily.  But if we want to have min/max the logic has to change.

cmds/tui.c
```diff
@@ -267,6 +294,36 @@ static void print_graph_self(struct field_data *fd)
...
+
+	print_time(d);
+}
+
+static void print_graph_self_min(struct field_data *fd)
+{
+	struct uftrace_graph_node *node = fd->arg;
+	uint64_t d;
+
+	d = node->total_time.min - node->child_time.min;
```
namhyung, line 312: This won't work correctly.  The min (and max) of self time should be calculalted each time it adds a graph node like report does.

cmds/tui.c
```diff
@@ -267,6 +294,36 @@ static void print_graph_self(struct field_data *fd)
...
+
+	print_time(d);
+}
+
+static void print_graph_self_max(struct field_data *fd)
+{
+	struct uftrace_graph_node *node = fd->arg;
+	uint64_t d;
+
+	d = node->total_time.max - node->child_time.max;
```
namhyung, line 322: Ditto.

utils/graph.c
```diff
@@ -173,6 +189,12 @@ static int add_graph_exit(struct uftrace_task_graph *tg)
 	node->total_time.sum += fstack->total_time;
 	node->child_time.sum += fstack->child_time;
 
+	update_graph_time_stat(&node->total_time, fstack->total_time);
+	update_graph_time_stat(&node->child_time, fstack->child_time);
```
namhyung, line 193: So you need to calculate the min/max of SELF time here.

namhyung (commented):
IIUC this work adds min/max column for graph command not only for tui, right?  Then I think you can:
* convert graph code to keep self time instead of child time
* introduce graph_time_stat
* add min and max fields to graph
* add min and max fields to tui

utils/graph.h
dongwooklee96: @namhyung 
stat for child time Isn't it useful to store the average?

dongwooklee96:
@namhyung @honggyukim 

I don't know if I understood it well.
I change the fields and logic of the calculation as requested.

utils/graph.c
```diff
@@ -172,6 +188,13 @@ static int add_graph_exit(struct uftrace_task_graph *tg)
 out:
 	node->total_time.sum += fstack->total_time;
 	node->child_time += fstack->child_time;
+	node->self_time.sum += fstack->total_time - fstack->child_time;
+
+	update_graph_time_stat(&node->total_time, fstack->total_time);
+	update_graph_time_stat(&node->self_time, fstack->total_time - fstack->child_time);
+
+	node->total_time.avg = (node->total_time.sum) / node->nr_calls;
+	node->self_time.avg = (node->self_time.sum) / node->nr_calls;
```
honggyukim, line 197: I hit divide by zero error here because `node->nr_calls` comes with `0` in this case. This can be reproduced as follows.
```
$ gcc -pg -o t-fork tests/s-fork.c
$ uftrace record t-fork
$ uftrace tui -f +total-avg,total-min,total-max,self-avg,self-min,self-max
```
If you add the following line above, you will hit the assertion failure easily.
```c
ASSERT(node->nr_calls > 0);
```

honggyukim (commented):
I couldn't read all the code, but I hit an error. Please fix it.

honggyukim:
@dongwooklee96 In addition, the current commit message doesn't explain what the patches do. You can think that the commit message is a letter to the code reviewer and future readers.

Rather than explaining many things in the comment here, you better explain what the patches do in the commit message so that reviewers do not have to ask. Please read other commit messages and get reference from them.

utils/graph.c
namhyung: For those cases, ASSERT will crash the uftrace and I don't think we want it.  Maybe we can drop the node or set nr_calls to 1.

utils/graph.c
```diff
@@ -10,6 +10,19 @@ static void *cb_arg;
 
 static struct rb_root task_graph_root = RB_ROOT;
 
+static void init_time_stat(struct graph_time_stat *ts)
+{
+	ts->min = -1ULL;
```
namhyung, line 15: Please init other fields too.

cmds/tui.c
```diff
@@ -781,6 +906,7 @@ static void build_partial_graph(struct tui_report_node *root_node, struct tui_gr
 
 	root->n.total_time.sum = 0;
 	root->n.child_time = 0;
+	root->n.self_time.sum = 0;
```
namhyung, line 909: You'd better to reset other fields like min, max too.

honggyukim:
The current PR still gets crashed when running the following command because `node->nr_calls` being `0`.
```
$ gcc -pg -o t-fork tests/s-fork.c
$ uftrace record t-fork
$ uftrace tui -f all
```
It will be more difficult to debug the problem in tui mode. This work is not only for tui graph mode, but also can be done for `graph` command. I would suggest you to apply the same work to `graph` command. If you feel it's too much work, then I think this is better to be added to `graph` command first, then apply the same work to `tui` command after that. I feel that some code can be shared between those two commands.

In addition, `t-fork` executable binary shouldn't be included in the commit. Please remove it.

utils/graph.c
dongwooklee96: @namhyung 
> Maybe we can drop the node or set nr_calls to 1.

I don't know if this is what you mean.

```
if (node->nr_calls > 0) {
		node->total_time.avg = (node->total_time.sum) / node->nr_calls;
		node->self_time.avg = (node->self_time.sum) / node->nr_calls;
}
```

dongwooklee96:
> @dongwooklee96 In addition, the current commit message doesn't explain what the patches do. You can think that the commit message is a letter to code reviewers and future readers. (or even to yourself)
> 
> Rather than explaining many things in the comment here, you better explain what the patches do in the commit message as much as possible so that reviewers do not have to ask. Please read other commit messages and get reference from them.

First, I'll fix the bug and add commit message at the end.

utils/graph.c
dongwooklee96: @namhyung @honggyukim

```
static void init_time_stat(struct report_time_stat *ts)
{
	ts->min = -1ULL;
}
``` 
ref: https://github.com/namhyung/uftrace/blob/master/utils/report.c#L11-L14

`init_time_state` function in `utils/report.c` file it didn't init too except min.
does it necessary?

honggyukim:
Thanks for the update. It looks fine now, but please add all to fields pane so that we can select with check box.

image

It currently has `TOTAL TIME`, `SELF TIME`, `ADDRESS` only.

honggyukim:
You can have a look at #1270 to customize fields in tui mode.

cmds/graph.c
```diff
@@ -695,8 +696,8 @@ static void build_graph(struct uftrace_opts *opts, struct uftrace_data *handle,
 		struct uftrace_graph_node *node;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
-			graph->ug.root.time += node->time;
-			graph->ug.root.child_time += node->time;
+			graph->ug.root.total_time.sum += node->total_time.sum;
+			graph->ug.root.child_time += node->child_time;
```
namhyung, line 700: This is a behavior change.  Since they are top-level nodes, It should use total_time.sum for root's child time.

cmds/tui.c
```diff
@@ -719,8 +719,8 @@ static struct tui_graph *tui_graph_init(struct uftrace_opts *opts)
 		top->nr_calls = 1;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
-			top->time += node->time;
-			top->child_time += node->time;
+			top->total_time.sum += node->total_time.sum;
+			top->child_time += node->child_time;
```
namhyung: Same here.

utils/graph.h
```diff
@@ -9,12 +9,20 @@
...
+};
+
 struct uftrace_graph_node {
 	uint64_t addr;
 	char *name;
 	int nr_edges;
 	int nr_calls;
-	uint64_t time;
+	struct graph_time_stat total_time;
+	struct graph_time_stat self_time;
```
namhyung, line 25: I think you can add self_time in the next commit.

cmds/tui.c
```diff
@@ -717,12 +834,20 @@ static struct tui_graph *tui_graph_init(struct uftrace_opts *opts)
 		top = &graph->ug.root;
 		top->name = basename(graph->ug.sess->exename);
 		top->nr_calls = 1;
+		top->total_time.min = -1ULL;
+		top->self_time.min = -1ULL;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
 			top->total_time.sum += node->total_time.sum;
-			top->child_time += node->child_time;
+			top->self_time.sum += node->self_time.sum;
```
namhyung, line 842: Top level node is a pseudo node and should not have self time.

utils/graph.c
namhyung: Then you can use it when building partial graphs.

cmds/tui.c
```diff
@@ -314,9 +431,9 @@ static struct display_field graph_field_addr = {
 
 /* index of this table should be matched to display_field_id */
 static struct display_field *graph_field_table[] = {
-	&graph_field_total,
-	&graph_field_self,
-	&graph_field_addr,
+	&graph_field_total,	&graph_field_total_avg, &graph_field_total_min,
+	&graph_field_total_max, &graph_field_self,	&graph_field_self_avg,
+	&graph_field_self_min,	&graph_field_self_max,	&graph_field_addr,
```
namhyung, line 436: I think it's better to split this part into another commit and to focus on self time calculation here.

utils/graph.h
dongwooklee96: @namhyung 
Do you mean I should split into each commit total_time and self_time?

cmds/tui.c
```diff
@@ -162,10 +162,11 @@ static const char *help[] = {
 	"q             Quit",
 };
 
-#define NUM_GRAPH_FIELD 6
+#define NUM_GRAPH_FIELD 9
 
 static const char *graph_field_names[NUM_GRAPH_FIELD] = { "TOTAL TIME", "TOTAL AVG", "TOTAL MIN",
-							  "TOTAL MAX",	"SELF TIME", "ADDRESS" };
+							  "TOTAL MAX",	"SELF TIME", "SELF AVG",
+							  "SELF MIN",	"SELF MAX",  "ADDRESS" };
```
namhyung, line 169: Please fix the indentation by adding a comma after "ADDRESS".  It's not your fault but I'd like to change it here as well.

utils/graph.c
```diff
@@ -10,6 +10,22 @@ static void *cb_arg;
...
+	ts->max = 0;
+}
+
+void update_graph_time_stat(struct graph_time_stat *ts, uint64_t time_ns)
+{
+	if (ts->min > time_ns)
+		ts->min = time_ns;
+	if (ts->max < time_ns)
+		ts->max = time_ns;
```
namhyung, line 26: I think it's better to update ts->sum too.

utils/graph.c
```diff
@@ -170,9 +188,14 @@ static int add_graph_exit(struct uftrace_task_graph *tg)
 	}
 
 out:
-	node->time += fstack->total_time;
+	node->total_time.sum += fstack->total_time;
 	node->child_time += fstack->child_time;
 
+	update_graph_time_stat(&node->total_time, fstack->total_time);
+	if (node->nr_calls > 0) {
+		node->total_time.avg = (node->total_time.sum) / node->nr_calls;
```
namhyung, line 200: This also can be done in the `update_graph_time_stat` but I think we can just drop 'avg' from the stat as it can be easily calculated from sum and calls when it's printed.

cmds/tui.c
```diff
@@ -252,7 +249,34 @@ static void print_graph_total(struct field_data *fd)
...
+	d = node->total_time.sum;
+
+	print_time(d);
+}
+
+static void print_graph_total_avg(struct field_data *fd)
+{
+	struct uftrace_graph_node *node = fd->arg;
+	uint64_t d;
+	d = node->total_time.avg;
```
namhyung, line 262: Please have a blank line after the variable declaration (here and in other functions).

utils/graph.h
```diff
@@ -57,6 +64,9 @@ void graph_init(struct uftrace_graph *graph, struct uftrace_session *s);
 void graph_init_callbacks(graph_fn entry, graph_fn exit, graph_fn event, void *arg);
 void graph_destroy(struct uftrace_graph *graph);
 
+extern void update_graph_time_stat(struct graph_time_stat *ts, uint64_t time_n);
+extern void init_time_stat(struct graph_time_stat *ts);
```
namhyung, line 69: For external functions, it'd be good to have a common prefix like graph.  Please rename to `graph_update_time_stat` and `graph_init_time_stat`.

cmds/tui.c
```diff
@@ -771,14 +833,19 @@ static struct tui_graph *tui_graph_init(struct uftrace_opts *opts)
...
 		top->nr_calls = 1;
 		top->total_time.min = -1ULL;
+		top->self_time.min = -1ULL;
 
 		list_for_each_entry(node, &graph->ug.root.head, list) {
 			top->total_time.sum += node->total_time.sum;
 			top->child_time += node->total_time.sum;
+
 			update_graph_time_stat(&top->total_time, node->total_time.sum);
+			update_graph_time_stat(&top->self_time, node->self_time.sum);
```
namhyung, line 843: At this point, I'm not sure whether this is correct.  At least each node should have `nr_calls == 1`.  Could you please add `ASSERT` to check the condition before calling the function?

namhyung (commented):
Thanks for your work!

cmds/tui.c
```diff
@@ -722,10 +778,11 @@ static struct tui_graph *tui_graph_init(struct uftrace_opts *opts)
 		top = &graph->ug.root;
 		top->name = basename(graph->ug.sess->exename);
 		top->nr_calls = 1;
+		top->total_time.min = -1ULL;
```
namhyung, line 844: `graph_init_time_stat()`?

cmds/tui.c
```diff
@@ -162,13 +162,10 @@ static const char *help[] = {
...
-#define NUM_GRAPH_FIELD 3
+#define NUM_GRAPH_FIELD 6
 
-static const char *graph_field_names[NUM_GRAPH_FIELD] = {
-	"TOTAL TIME",
-	"SELF TIME",
-	"ADDRESS",
-};
+static const char *graph_field_names[NUM_GRAPH_FIELD] = { "TOTAL TIME", "TOTAL AVG", "TOTAL MIN",
+							  "TOTAL MAX",	"SELF TIME", "ADDRESS" };
```
namhyung: A comma after "ADDRESS" please.

cmds/tui.c
```diff
@@ -291,7 +293,37 @@ static void print_graph_self(struct field_data *fd)
...
+
+	print_time(d);
+}
+
+static void print_graph_self_avg(struct field_data *fd)
+{
+	struct uftrace_graph_node *node = fd->arg;
+	uint64_t d;
+
+	d = node->self_time.sum / node->nr_calls;
```
namhyung, line 306: You need to check if `node->nr_calls` is 0.

utils/graph.h
```diff
@@ -21,6 +21,7 @@ struct uftrace_graph_node {
 	int nr_edges;
 	int nr_calls;
 	struct graph_time_stat total_time;
+	struct graph_time_stat self_time;
 	uint64_t child_time;
```
namhyung, line 25: I think you can remove `child_time` now.

cmds/tui.c
namhyung: Hmm.. you didn't do that.

Anyway, please consider updating the documentation too.

dongwooklee96: @namhyung 
I was told to add a function corresponding to each commit.
 Wasn't it added to the two commits?

utils/graph.h
dongwooklee96: Is `child_time` not required to get the `self_time`?

utils/graph.c
namhyung: I thought it's an error if `node->nr_calls != 1`.

cmds/tui.c
namhyung: I expect we would have 4 commits at the end.

utils/graph.h
namhyung: It should come from fstack.

cmds/tui.c
namhyung: Please see my earlier comment.

> I think your work contains the following changes.
> 
> 1.  convert node->total_time to struct graph_time_stat
> 2.  change node->child_time to node->self_time (and make it struct graph_time_stat)
> 3.  add avg/max/min to uftrace graph command
> 4.  add avg/max/min to uftract tui command
>
> Each change can be in a different commit.

utils/graph.h
dongwooklee96: @namhyung I have no idea can you explain more?

cmds/tui.c
namhyung: No, nr_calls can be any number here.  You should not add ASSERT here.  Just check if it's zero to prevent a divide-by-zero.

utils/graph.h
namhyung: In `add_graph_exit()` you already used `fstack->total_time - fstack->child_time` to calculate the self time.  So `node->child_time` is not needed anymore.

cmds/tui.c
dongwooklee96: Oh, that'sure

utils/graph.h
dongwooklee96: @namhyung @honggyukim 

How should I modify the code below when I remove `child_time` from the `utrace_graph_node` structure?

```
...
if (sample && flame->sample_time)
		sample = (node->time - node->child_time) / flame->sample_time;
...
```
https://github.com/namhyung/uftrace/blob/master/cmds/dump.c#L1075-L1084

cmds/tui.c
dongwooklee96: @namhyung 

Oh, I'm sorry that I missed it.

honggyukim:
@dongwooklee96 Sorry for the late reply. I've looked into your work but it's not easy to review because many different changes are mixed into a single commit. As @namhyung asked, the patches have to be split for easier review.

But I've tried to split your work into separate patches and pushed it into https://github.com/namhyung/uftrace/compare/master...honggyukim:uftrace:check/add_field_with_tuimode.

It'd be good if you find it useful to take it and improve it.

As @namhyung asked at https://github.com/namhyung/uftrace/pull/1513#pullrequestreview-1072117803, this work can be shared for both `graph` and `tui` commands. But start working on `tui` first is not a good idea because it's more difficult when it comes to debugging. I suggest you to implement it for `graph` command first.

One last comment is having `avg`/`min`/`max for self time is tricky because it always requires calculation unlike total time. I would like to ask you to make this work simpler and complete partially.

To sum up, my suggestion is as follows.
- do not add `avg`/`min`/`max` to self time (for now)
- implement `avg`/`min`/`max` for `graph` command first (if you feel comfortable)

You can write commits for `tui` in the next PR. In the `tui` PR, please follow this.
- take my patches from https://github.com/namhyung/uftrace/compare/master...honggyukim:uftrace:check/add_field_with_tuimode and improve based on it.
- The current implementation crashes when pressing `g` button because it didn't handle partial graph properly. So please fix it.

To help your debugging, I pushed a commit https://github.com/namhyung/uftrace/commit/125beace26c072fc08bf913901cff3259348ae95 that supports `TUI_ASSERT()` because backtrace output is broken in tui mode. Please use it when you add assertion.

dongwooklee96:
@honggyukim 
I didn't mean to, but my pull request ended up closing.

dongwooklee96:
I didn't mean to, but my pull request ended up closing.

namhyung (approved):
LGTM

honggyukim:
@namhyung I think the work needs to be improved. As I mentioned, we need to fix crash when pressing ‘g’ button and this makes the screen update too slow when moving cursors.

Could you please check it again.

dongwooklee96:
> @namhyung I think the work needs to be improved. As I mentioned, we need to fix crash when pressing ‘g’ button and this makes the screen update too slow when moving cursors.
> 
> 
> 
> Could you please check it again?
@namhyung 
yeah this is not final version of work. I'm working on it

namhyung:
Ok, let me know once you're done.

bernhardkaindl:
@dongwooklee96: I just found this PR and that [namhyung](https://github.com/namhyung) asked you to notify him when done. Is it finished/working now?

dongwooklee96:
> @dongwooklee96: I just found this PR and that [namhyung](https://github.com/namhyung) asked you to notify him when done. Is it finished/working now?

@bernhardkaindl 
I'm not working these days, so if you want to, you can proceed.

PR #1514, opened 2022-08-07
misc: Make 'uftrace --version' show python version info

### LINKED ISSUE #1464: make 'uftrace --version' show python version info
Currently, `uftrace --version` shows whether [python scripting](https://github.com/namhyung/uftrace/blob/v0.12/doc/uftrace-script.md#script-execution) is supported, but it doesn't inform whether python2.7 or python3 is supported.

We better distinguish the python version info by modifying the following code.
https://github.com/namhyung/uftrace/blob/v0.12/misc/version.sh#L37-L40

So the expected output of `uftrace --version` is
```
$ uftrace --version
uftrace v0.12 ( x86_64 dwarf python3 ... )
```
or
```
$ uftrace --version
uftrace v0.12 ( x86_64 dwarf python2.7 ... )
```

ghojeong:
Closes #1464 

Show python3 and python2.7 version,
when user command 'uftrace --version'.

Signed-off-by: JeongWan Gho <rhwjddhks@gmail.com>

### DIFF
--- a/misc/version.sh
+++ b/misc/version.sh
@@ -34,9 +34,10 @@ DEPS=" ${ARCH}"
 if test -f ${SRCDIR}/check-deps/have_libdw; then
     DEPS="${DEPS} dwarf"
 fi
-if test -f ${SRCDIR}/check-deps/have_libpython2.7 ||
-   test -f ${SRCDIR}/check-deps/have_libpython3; then
-    DEPS="${DEPS} python"
+if test -f ${SRCDIR}/check-deps/have_libpython3; then
+    DEPS="${DEPS} python3"
+elif test -f ${SRCDIR}/check-deps/have_libpython2.7; then
+    DEPS="${DEPS} python2"
 fi
 if test -f ${SRCDIR}/check-deps/have_libluajit; then
     DEPS="${DEPS} luajit"


misc/version.sh
```diff
@@ -34,9 +34,11 @@ DEPS=" ${ARCH}"
...
 fi
-if test -f ${SRCDIR}/check-deps/have_libpython2.7 ||
-   test -f ${SRCDIR}/check-deps/have_libpython3; then
-    DEPS="${DEPS} python"
+if test -f ${SRCDIR}/check-deps/have_libpython2.7; then
+    DEPS="${DEPS} python2.7"
+fi
+if test -f ${SRCDIR}/check-deps/have_libpython3; then
+    DEPS="${DEPS} python3"
 fi
```
honggyukim, line 42: uftrace uses python3 only when it has both versions so this should be as follows.
```suggestion
if test -f ${SRCDIR}/check-deps/have_libpython3; then
    DEPS="${DEPS} python3"
elif test -f ${SRCDIR}/check-deps/have_libpython2.7; then
    DEPS="${DEPS} python2.7"
fi
```

honggyukim (commented):
This makes uftrace version show both python2.7 and python3.
```
$ uftrace --version
uftrace v0.12-77-g393ec ( aarch64 dwarf python2.7 python3 luajit tui perf sched dynamic )
```

misc/version.sh
ghojeong: I will accept the suggestion.
Thank you!

namhyung: As we all know it's python 2.7, I'd rather have it just like "python2" instead of "python2.7".

ghojeong: I will change to "python2" !

honggyukim:
The pre-commit failure is not related to this change. I've just fixed it and merged so please rebase it onto the current master again.
```
codespell................................................................Failed
- hook id: codespell
- exit code: 65

utils/graph.c:406: destory ==> destroy
```

honggyukim (approved):
LGTM

Issue #1515, opened 2022-08-07
Install libunwind on install-dep script and add support Fedora WSL2


paranlee:
Add install libunwind library on `misc/install-dep.sh` script.

And I want to add the install-deps.sh script to be available to Fedora WSL2 users.

There are additional postfix on `ID=fedoraremixforwsl`

```
[kim@DESKTOP-S4LJL03 uftrace]$ cat /etc/os-release
NAME="Fedora Remix for WSL"
VERSION="36"
ID=fedoraremixforwsl
ID_LIKE=fedora
VERSION_ID=36
PLATFORM_ID="platform:f36"
```
Is wildcard appropriate?

```
case $distro in

    "fedora"*)
        dnf install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconf-pkg-config
        dnf install $OPT luajit-devel || true
        dnf install $OPT capstone-devel || true
        dnf install $OPT libunwind-devel || true ;;
```

paranlee:
#1382 : libunwind-dev is only required in debug build

PR #1382, opened 2021-11-06
misc: add libunwind-dev to install-deps.sh

DanielTimLee:
Recently, uftrace started to make use of libunwind library.
commit df1f47075a3d ("Merge pull request #1301 from honggyukim/check/use-libunwind")

This commit adds libunwind-dev to install-deps.sh.
Since libunwind-dev ubuntu package only exists in upper than 18.04,
directive ' || true' has been added to suppress error.
(Same for Fedora, CentOS)

Signed-off-by: Daniel T. Lee <danieltimlee@gmail.com>

--- 

Ubuntu libunwind-dev search result
https://packages.ubuntu.com/search?keywords=libunwind-dev

Tested on vagrant 'generic/ubuntu1804', 'generic/ubuntu2110', 'generic/centos7', 'generic/fedora33', 'generic/arch'

PR #1516, opened 2022-08-07
misc: Add install libunwind lib and WSL2 Fedora support in install-deps.sh

paranlee:
Add install libunwind library and WSL2 Fedora support in issue #1515

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -21,23 +21,28 @@ case $distro in
     "ubuntu" | "debian" | "raspbian" | "kali" | "linuxmint")
         apt-get $OPT install pandoc libdw-dev python3-dev libncursesw5-dev pkg-config
         apt-get $OPT install libluajit-5.1-dev || true
+        [ $DEBUG ] && apt-get $OPT install libunwind-dev || true
         apt-get $OPT install libcapstone-dev || true ;;
-    "fedora")
+    "fedora"*)
         dnf install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconf-pkg-config
         dnf install $OPT luajit-devel || true
+        [ $DEBUG ] && dnf install $OPT libunwind-devel || true
         dnf install $OPT capstone-devel || true ;;
     "rhel" | "centos")
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
         yum install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconfig
         yum install $OPT luajit-devel || true
+        [ $DEBUG ] && yum install $OPT libunwind-devel || true
         yum install $OPT capstone-devel || true ;;
     "arch" | "manjaro")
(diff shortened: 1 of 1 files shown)


honggyukim:
Thanks @paranlee, but please have a look at #1382.

paranlee:
@honggyukim , Thanks you for review! 

Would it be better to create a separate issue for WSL2 fedora support only?

honggyukim:
> WSL2 fedora

What do you mean by WSL2 fedora support? Isn't it same as normal fedora?

paranlee:
@honggyukim Updated PR. 

When DEBUG=1, because libunwind-dev is only required in debug build.

` [ $DEBUG ] && apt-get $OPT install libunwind-dev `

#1515 : There are additional postfix on ID=fedoraremixforwsl

[kim@DESKTOP-S4LJL03 uftrace]$ cat /etc/os-release
NAME="Fedora Remix for WSL"
VERSION="36"
ID=fedoraremixforwsl
ID_LIKE=fedora
VERSION_ID=36
PLATFORM_ID="platform:f36"

misc/install-deps.sh
```diff
@@ -21,23 +21,28 @@ case $distro in
     "ubuntu" | "debian" | "raspbian" | "kali" | "linuxmint")
         apt-get $OPT install pandoc libdw-dev python3-dev libncursesw5-dev pkg-config
         apt-get $OPT install libluajit-5.1-dev || true
+        [ $DEBUG ] && apt-get $OPT install libunwind-dev || true
         apt-get $OPT install libcapstone-dev || true ;;
-    "fedora")
+    "fedora"*)
```
namhyung, line 26: No wildcard plesae.  I'd like to fix it with ID_LIKE as in #1468.

namhyung (commented):
I don't have a strong opinion on the libunwind installment. @honggyukim ?

misc/install-deps.sh
```diff
@@ -21,23 +21,28 @@ case $distro in
     "ubuntu" | "debian" | "raspbian" | "kali" | "linuxmint")
         apt-get $OPT install pandoc libdw-dev python3-dev libncursesw5-dev pkg-config
         apt-get $OPT install libluajit-5.1-dev || true
+        [ $DEBUG ] && apt-get $OPT install libunwind-dev || true
```
honggyukim, line 24: We use `DEBUG=1` only during `make` stage as of now. I think it's better to provide a way to install libunwind as follows.
```
$ ./misc/install-deps.sh libunwind
```

honggyukim (commented):
It looks this PR does libunwind installation and WSL2 Fedora support, but I think they have to be in separate patches.

misc/install-deps.sh
paranlee: Nice! :)
I'm going to isolate the issue.

misc/install-deps.sh
paranlee: Thanks you for code review. :)

PR #1517, opened 2022-08-07
utils: Archive old record data (WIP)

JHH20:
Currently, only the previous record data is saved (uftrace.data.old) and any older data are discarded.
This complicates debugging `runtest.py` since most record data are lost even with `-k` option.

This commit changes how old data is archived so that the last `n` record data are saved (where `n` is configurable).
Currently that number is hardcoded to 10 (`utils/util.c` #L364) because I have yet to implement an interface with `struct uftrace_opts`.

Current method of archiving appends `.old` to the directory name, but this change will append `.old#` where `#` is the smallest positive index that has not yet been taken. When the limit `n` is exceeded, the smaller indices are discarded and the remaining are renamed such that their index is shifted to lie within the range [1, n].

tests/.gitignore was also updated to exclude `failed-tests.txt.old` and `test_*` which get created from running `runtest.py -k`

### DIFF
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -2,3 +2,5 @@
 __pycache__/
 unittest
 failed-tests.txt
+failed-tests.txt.old
+test_*/
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -290,23 +290,84 @@ out:
 	return ret;
 }
 
+int archive_directory(const char *dirname, int limit)
+{
+	// Rename dirname -> dirname.old# (1 = oldest, bigger # = newer)
+	// Return -1 for shift failure, 0 for success, 1 for rename failure
+	int ret = -1;
+	char name_fmt[] = "%s.old%d";
+
+	// Rename dirname to newest id
+	int id = 0;
+	while (ret < 0) {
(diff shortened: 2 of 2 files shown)


namhyung:
I thought each test runs in a separate directory.  Can we just keep the directory?

JHH20:
If you mean the temp directory created by `runtest.py` then that is preserved with `-k` option already. However, the directory is shared for the same testcase built with different compile flags. So if more than 2 flags are tested, the record data from previous runs are lost.

namhyung:
Then, how about separating them?  Maybe we can add the flags to the directory name.

JHH20:
Currently, the temp directory is prefixed with the name of the binary only, so if the same binary is tested for different testcases, then it becomes difficult to tell them apart. How about the following prefix: `test_{testcase num}_{binary}_{flag}_` ?

honggyukim:
I have tried it but didn’t complete making all the tests passed. You can have a look at https://github.com/honggyukim/uftrace/commit/83b6451926a86aa66e0fcb745e638a32e59c6a4f and improve it.

PR #1520, opened 2022-08-09
tests: Separate test directory for runtest.py

JHH20:
Follow-up from PR #1517 but changing runtest.py instead of uftrace.

- Change directory structure to "test_{case num}/{flag}_{opt}"
- Fix improper call to debug logging
- Add test directory to .gitignore


tests/runtest.py
```diff
@@ -685,9 +682,14 @@ def run_single_case(case, flags, opts, arg):
     timeout = int(arg.timeout)
 
     origdir = os.getcwd()
+    case_num = case.split('_')[0][1:]
+    case_dir = os.path.abspath('test_%s' % case_num)
```
namhyung, line 686: I'd like to see the test name instead of the number.

namhyung (commented):
Please add your sign-off after Honggyu's in the first commit.  That's how we maintain the chain of authorship in a collaboration.

honggyukim:
My patch is just in work-in-progress so you can squash my patch into your one and add my Signed-off-by. I was testing with all the tests, but saw some failures because of the change. The change in [t220_trace_script.py](https://github.com/namhyung/uftrace/pull/1520/commits/788ec29582e2bcf228539261a6f824bead58d970#diff-b29cd4a99bee2f7aa107eb983cc85b37aaac46dbb64f03c3eba381da59a59be9) is one of the fix.

Please rerun all the tests and check and fix if some other tests fail.

tests/runtest.py
namhyung: Hmm.. now I think that we can have the number.  Sorry for the noise and please keep the current code.

namhyung:
Test 261 fails like this:

```
  File "/home/namhyung/project/uftrace/tests/./runtest.py", line 941, in <module>
    results = run_single_case(name, flags, opts.split(), arg, compilers)
  File "/home/namhyung/project/uftrace/tests/./runtest.py", line 704, in run_single_case
    ret, dif = tc.run(case, cflags, arg.diff, timeout)
  File "/home/namhyung/project/uftrace/tests/runtest.py", line 538, in run
    self.setup()
  File "/home/namhyung/project/uftrace/tests/t261_info_note.py", line 54, in setup
    with open('uftrace.data/note.txt', 'w+') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'uftrace.data/note.txt'
```

JHH20:
> Test 261 fails like this:
> 
> ```
> Traceback (most recent call last):
>   File "/home/namhyung/project/uftrace/tests/./runtest.py", line 941, in <module>
>     results = run_single_case(name, flags, opts.split(), arg, compilers)
>   File "/home/namhyung/project/uftrace/tests/./runtest.py", line 704, in run_single_case
>     ret, dif = tc.run(case, cflags, arg.diff, timeout)
>   File "/home/namhyung/project/uftrace/tests/runtest.py", line 538, in run
>     self.setup()
>   File "/home/namhyung/project/uftrace/tests/t261_info_note.py", line 54, in setup
>     with open('uftrace.data/note.txt', 'w+') as f:
> FileNotFoundError: [Errno 2] No such file or directory: 'uftrace.data/note.txt'
> ```

Found the reason.
https://github.com/namhyung/uftrace/blob/44a73f01e8f279870a4c245e033d0ac21266f977/tests/t261_info_note.py#L42-L57

`self.exearg` is not set correctly which results in the runcmd in `prerun()` missing the program name.
The default value is the prog name `'t-' + self.name` which is appropriate for `uftrace live/record` and this is set in `__init__()` as well as `prerun()` in the `TestBase` class. `setup()` is called from `run()` and if several uftrace subcommands are used in a single test iteration, the `self.exearg` needs to be updated appropriately.

In test 261, `prerun()` does not account for `'t-' + self.name` being cleared out from a previous test iteration's `setup()`, which caused the record command to fail. As a result, the record command becomes `uftrace record {uftrace opts}  notes.txt` instead of the intended `uftrace record {uftrace opts} t-hello notes.txt`

I think it would be good practice to always be explicit in full rather than using `+=` even when a test only uses one type of uftrace subcommand. The following testcases use `+=` which I've updated:
- t133
- t178
- t179
- t180
- t183 (actually causes problem)
- t190
- t205
- t261 (actually causes problem)

Issue #1414, opened 2022-03-21
[suggest] feature for tracing to already running process


ParkHanbum:
The feature to trace running processes is useful. In particular, it is attractive that it can skip the initialization stage of the program and track only the functions user want. 

I think the function of tracing dynamically executed programs should be developed separately as follows.

1.Injection of shared object file responsible for tracing  function 
feature to load shared objects that enable tracing into processes that are already running (like, libmcount.so).

2. create channel to communicate between shared object to be injected
If it inject shared objects, it is impossible to deliver environmental variables differently than before, so it will have to take a differently. Previously, there was a function to save options in the default.opts file, so I think you should come up with a way to communicate that path.

3. Ability to properly control running threads and perform dynamic patches 
In my opinion, the injection of breakpoint is essential to implement this function. It seems that you need to insert breakpoint into the function you want to patch first, then perform code patching, then release breakpoint and rerun suspended threads.

honggyukim:
Hi @ParkHanbum, it's good to see you back here :)

I can summary your words as follows.
1. Injecting `libmcount.so` to an already running process.
2. Create a communication channel between a target process and uftrace.
3. Modify the target process for full dynamic tracing.

First of all, we already have experimental work for 2 and 3 done by @gpollo and @clementguidi. In #1269, creating a communication channel is already done and applying dynamic patching is done by #1274. It's worth you take a look at those works.

Regarding injecting `libmcount.so`, this attach feature is definitely needed for sure. But I think we better have a more robust interim solution to achieve this long term goal.

What I would like to do is as follows.
1. Provide a way to make an explicit link to `libmcount.so` when a target program is built. (to avoid injection)
2. Ask users to compile the target program with `-fpatchable-function-entry=5` as we support this dynamic tracing as described at https://github.com/namhyung/uftrace/blob/master/doc/uftrace-record.md#patchable-function-entry.
3. When running the program, control the target process with already loaded `libmcount.so` and apply dynamic patching to the patchable entries filled with `NOP` instructions.

The downside of this approach is that it requires users to have special build for this, but it's harmless when running the program in normal scenarios even without uftrace attached because loading `libmcount.so` and having `NOP`s do not affect the normal execution of the target program.

By doing this, we can have a robust and actionable working model. How do you think about this approach?

ParkHanbum:
thank you for welcom @honggyukim  :)
I'm glad to hear that! looks many changes.

I have no objection to your approach. it is good approach if user can be support.
[...]

PR #1526, opened 2022-08-13
record: Do not initialize when libmcount.so is explictly linked

honggyukim:
Currently, libmcount.so library cannot be explictly linked to a binary
because it always initializes many things for uftrace record expecting
uftrace.data directoy is already created by uftrace.

Because of this assumption, it shows an error when a binary has an
explicit link to libmcount.so as follows.
```
  # make sure there is existing uftrace.data directory.
  $ rm -rf uftrace.data

  $ gcc -pg -g -o t-hello tests/s-hello.c -L./libmcount -lmcount
  $ ldd t-hello | grep libmcount
          libmcount.so => /home/honkim01/usr/lib/libmcount.so (0x00007fad4b99e000)

  $ ./t-hello
  mcount: /home/honggyu/work/uftrace/libmcount/record.c:1127:record_proc_maps
   ERROR: cannot open for writing maps file: No such file or directory
```
This patch is to avoid this error by adding UFTRACE_TRACING_ON env
variable to check whether libmcount.so init is invoked with uftrace or
not.  Having this patch, it doesn't show an error and the target binary
can still be traced by uftrace.
```
  $ ./t-hello
  Hello world

  $ uftrace ./t-hello
  Hello world
              [ 40432] | main() {
    13.835 us [ 40432] |   printf();
    17.606 us [ 40432] | } /* main */
```


honggyukim:
This can be a very first step of making uftrace as a tracing library. In the long term, the target program can be recorded without running `uftrace`, but just by linking `libmcount.so`.

namhyung:
I think we need to think how can we enable tracing it later especially with dynamic tracing.

honggyukim:
@namhyung Hmm.. I don’t get what you mean. Could you please explain more about your comment?

namhyung:
Well.. by linking the libmcount.so users might want to use full features of uftrace including dynamic tracing.

honggyukim:
Hmm.. It works find with dynamic tracing with this explicit linking.
```
$ gcc tests/s-abc.c -Wl,--no-as-needed -L./libmcount -lmcount
$ ldd a.out | grep libmcount
        libmcount.so => /home/honggyu/usr/lib/libmcount.so (0x00007fa352746000)

$ uftrace a.out
uftrace: /home/honggyu/work/uftrace/cmds/record.c:1692:check_binary
 ERROR: Can't find 'mcount' symbol in the 'a.out'.
        It seems not to be compiled with -pg or -finstrument-functions flag.
        You can rebuild your program with it or use -P option for dynamic tracing.

$ uftrace -P. -v a.out
      ...
uftrace: libmcount.so is already linked so do not preload it.
      ...
            [ 39937] | main() {
            [ 39937] |   a() {
            [ 39937] |     b() {
            [ 39937] |       c() {
   1.875 us [ 39937] |         getpid();
   5.417 us [ 39937] |       } /* c */
   5.942 us [ 39937] |     } /* b */
   6.378 us [ 39937] |   } /* a */
   7.787 us [ 39937] | } /* main */
```
Do you mean something else? What I meant to make libmcount.so self contained is a future work, not in this PR.

namhyung:
Oh I thought you want to run the program and record data without uftrace.  Actually I don't see why the explicit link is needed in this use case.

honggyukim:
This is a preparation patch for #1414. I mentioned as follows at https://github.com/namhyung/uftrace/issues/1414#issuecomment-1081313032.

> 1. Provide a way to make an explicit link to libmcount.so when a target program is built. (to avoid injection)

This PR makes attaching to a running process much simpler.

namhyung:
I see.  Then it'd work with #1543.

honggyukim:
@namhyung I've just rebased it. Could you please have a look again?

honggyukim:
It shows that when linking `libmcount.so` explicitly with `-pg` option, then it shows the `libmcount.so` in the link list.
```
$ gcc -pg -g -o t-hello tests/s-hello.c -L./libmcount -lmcount
$ ldd t-hello | grep libmcount
        libmcount.so => /home/honggyu/usr/lib/uftrace/libmcount.so (0x00007f228212a000)
```
However, when compiling with `-fpatchable-function-entry` then `libmcount.so` is not shown even if `-lmcount` is added during compilation.
```
$ gcc -fpatchable-function-entry=5 -g -o t-hello tests/s-hello.c -L./libmcount -lmcount
$ ldd t-hello | grep libmcount
(... no libmcount.so found ...)
```
This is because the compiler drops the library by detecting no function is used inside the library in this case. We might have to find a way to force linking `libmcount.so`.

namhyung:
GNU `ld` has `--as-needed` option, and I guess it'd do the job.

namhyung:
I understand it's an easier solution than injecting a library but still don't think it's a better way.  Anyway maybe I can take this later if injection is not feasible.  But it's not an urgent issue and no real use case yet.

PR #1543, opened 2022-08-15
Introduce an agent in libmcount

clementguidi:
Hi,

This is a rework of PR #1269 from DORSAL. I moved the code to a new repository and made modifications according to old discussions. Sorry for the inconvenience.

This PR is the first step to interact with uftrace targets at runtime. It separates the CLI from the libmcount runtime, by spawning a daemon in libmcount, that is controlled by an separate and independent uftrace instance.

I introduce two new associated parameters:
- `--daemon`/`-m` to spawn the daemon in the target (off by default)
- `--pid`/`-p`to set uftrace to "client mode" and forward its parameters to a running target

Note that the daemon code is compiled conditionally in libmcount. It can be disabled using the `--without-daemon` flag in the configure script.

Follow-up steps:
- implement parameter tweaking at runtime (e.g. filters, depth) - PR to come
- implement dynamic patching and unpatching (from #1274)

Please let me know what you think about it!

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -3,12 +3,14 @@
 #include <signal.h>
 #include <stdio.h>
 #include <sys/stat.h>
+#include <sys/un.h>
 #include <unistd.h>
 
 #include "libmcount/mcount.h"
 #include "uftrace.h"
 #include "utils/fstack.h"
 #include "utils/kernel.h"
+#include "utils/socket.h"
 #include "utils/utils.h"
 
 static char *tmp_dirname;
@@ -95,6 +97,37 @@ static void setup_child_environ(struct uftrace_opts *opts)
 	free(libpath);
 }
 
+/* Forward all client options to the agent */
+static int forward_options(struct uftrace_opts *opts)
+{
+	int sfd;
+	struct sockaddr_un addr;
+	int ret = 0;
+
+	sfd = socket_create(&addr, opts->pid);
+	if (sfd == -1)
+		return -1;
+
+	if (socket_connect(sfd, &addr) == -1) {
+		ret = -1;
+		goto socket_error;
+	}
+
+	if (socket_send_option(sfd, UFTRACE_DOPT_CLOSE, NULL, 0) == -1) {
+		pr_warn("cannot terminate agent connection\n");
+		ret = -1;
+	}
+	else {
+		enum uftrace_dopt ack;
+		if (read(sfd, &ack, sizeof(enum uftrace_dopt)) < 0 || ack != UFTRACE_DOPT_CLOSE)
+			ret = -1;
+	}
+
+socket_error:
+	close(sfd);
+	return ret;
+}
+
 int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 {
 	char template[32] = "/tmp/uftrace-live-XXXXXX";
@@ -144,6 +177,9 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 		}
 		return 0;
 	}
+
+	if (opts->pid)
+		return forward_options(opts);
 
 	ret = command_record(argc, argv, opts);
 	if (!can_skip_replay(opts, ret)) {
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -14,6 +14,7 @@
 #include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/uio.h>
+#include <sys/un.h>
 #include <unistd.h>
 
 /* This should be defined before #include "utils.h" */
@@ -26,6 +27,7 @@
 #include "mcount-arch.h"
 #include "utils/filter.h"
 #include "utils/script.h"
+#include "utils/socket.h"
 #include "utils/symbol.h"
 #include "utils/utils.h"
 #include "version.h"
@@ -91,6 +93,12 @@ unsigned long mcount_return_fn;
 
 /* do not hook return address and inject EXIT record between functions */
 bool mcount_estimate_return;
+
+/* agent thread */
+static pthread_t agent;
+
+/* state flag for the agent */
+static bool agent_run = false;
 
 __weak void dynamic_return(void)
 {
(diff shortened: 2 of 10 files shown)


honggyukim:
Hi @clementguidi, thanks for coming back. I understand you want to make runtime dynamic patching. Is it correct?

One of our long term plan is to support attaching to a running process. It can be done together with runtime dynamic patching. I've summarized the approach at https://github.com/namhyung/uftrace/issues/1414#issuecomment-1081313032. You could have a look at it.

I've also uploaded making `libmcount.so` linked to a target binary at #1526. I think we can ask users to build the target program with explicitly linking `libmcount.so` and control whether it keeps socket open or not with an environmental variable such as `UFTRACE_TRACING_ON`.

This work is not same as your approach, but I think we can take advantage of both approaches. Thanks.

namhyung:
@clementguidi thanks for working on this!

Actually I was skeptical about the daemon work at first, but now I think it might be good both for changing tracing configurations at run time and for attaching to a task as an interim step.  A few thoughts:

* It's not a daemon actually, can we have a better name?
* most of socket messaging code should be in a separate file under utils directory
* please split the -m and -p option handling into separate commits
* I don't like the idea of optional argument, please make it required
* I don't think it needs a config option to disable it as it's already optional

clementguidi:
I made changes according to your comments first @namhyung.

As a new name for the daemon I suggest using "agent", what do you think? Other choices might be "controller" or "session", but the latter can be ambiguous in uftrace.

namhyung:
'agent' looks good.  Will review the patch soon.

utils/socket.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include <sys/socket.h>
+
+#include "uftrace.h"
+
+#define MCOUNT_AGENT_SOCKET_DIR "/tmp/uftrace"
+
+int create_socket(struct sockaddr_un *addr, pid_t pid);
+int listen_socket(int sfd, struct sockaddr_un *addr);
+int send_options(struct uftrace_opts *opts);
```
namhyung, line 12: For these external functions, I prefer having a common prefix.  How about `socket_create`, `socket_listen`, `socket_send_options`?  It'd be nice if we could have some comments for them.

utils/socket.h
```diff
@@ -0,0 +1,14 @@
+#ifndef UFTRACE_SOCKET_H
+#define UFTRACE_SOCKET_H
+
+#include <sys/socket.h>
+
+#include "uftrace.h"
```
namhyung, line 6: Please don't include this unnecesarily.  If it's only for `struct uftrace_opts`, we can have a forward declaration.

utils/socket.c
```diff
@@ -0,0 +1,65 @@
...
+	return sfd;
+}
+
+int send_options(struct uftrace_opts *opts)
+{
+	int sfd; /* Socket file descriptor, to communicate with the agent */
+	char *channel = NULL;
+	struct sockaddr_un addr;
+
+	sfd = create_socket(&addr, opts->pid);
```
namhyung, line 53: Please check the return value.

libmcount/mcount.c
```diff
@@ -1716,6 +1723,133 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+}
+
+void agent_fini(int sfd)
+{
+	if (sfd != -1)
+		close(sfd);
+}
+
+/* Agent routine, applying instructions from the CLI. */
+void *apply_client_commands(void *arg)
```
namhyung, line 1761: It'd be nice we could have `agent_` prefix.  How about `agent_apply_commands()`?

namhyung (commented):
Could you please add a test case?  You could add a test command to check if it's communicating properly.

clementguidi:
Thank you very much for all your comments, I made the relevant changes.

As for the test case, I'm not sure how to do it. I've been trying to add a unit test in `uftrace.c` but I can't configure `libmcount` so it starts the agent when running that test.

uftrace.c
```diff
@@ -703,6 +705,11 @@ static int parse_option(struct uftrace_opts *opts, int key, char *arg)
 	case 'h':
 		return -3;
 
+	case 'p':
+		opts->pid = strtol(arg, NULL, 0);
+		opts->exename = "";
```
namhyung, line 710: Could you explain why this is needed?

namhyung: Maybe we can add this functionality to `live` command and skip recording/replaying.

namhyung (commented):
You can add a runtime test as similar as your use case.  You might want to add `UFTRACE_DOPT_TEST` to verify agent behavior in some way.

uftrace.c
namhyung: Ok, we don't need to use `record` for this.  I think `live` is a better fit since it'll only talk to a running process.  It can handle this similar to `--list-event`.  Wdyt?

clementguidi: I agree, and it also means it can be invoked without explicitly using the `live` command.

As for the syntax, should I keep the `-p` option to trigger client behavior? I was thinking maybe use the target program as the PID value when it is numeric. For example
```
$ uftrace --some-option target # executes and traces the "target" binary
$ uftrace --some-option 1234   # client mode, forwards options to agent in process with PID 1234
```
Would that be relevant?

namhyung: I think it's better to use `-p` explicitly to trigger the client mode.

namhyung (commented):
The first commit contains changes for later commits.  Please split the first and squash them to the second and third so that each option comes together with the actual implementation.

clementguidi:
I added a test case to check if the agent is responsive. I'm using a structure similar to the `recv` tests, where you spawn a background uftrace instance and terminate it at the end of the test.

I split the commits as you requested.

namhyung (commented):
The test 272 was taken by mermaid recently, please update.  Also we might want to add more tests for agent later.  Please rename it to something like '273_agent_basic'.  Otherwise, looks good! :)

clementguidi:
Great! I'm fixing this and rebasing.

libmcount/mcount.c
```diff
@@ -1716,6 +1724,145 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
 	strv_free(&info.cmds);
 }
 
+/* unlink socket file if it exists */
+static void unlink_socket_maybe(struct sockaddr_un *addr)
```
namhyung, line 1728: `socket_unlink` and move it to socket.c?

libmcount/mcount.c
```diff
@@ -1716,6 +1724,145 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+	}
+}
+
+static int agent_init(struct sockaddr_un *addr)
+{
+	int sfd;
+
+	if (mkdir(MCOUNT_AGENT_SOCKET_DIR, 0775) == -1) {
+		if (errno != EEXIST) {
+			pr_dbg("error creating run directory %s\n", MCOUNT_AGENT_SOCKET_DIR);
```
namhyung, line 1743: Isn't it a critical error?  I wonder if it's a part of `socket_create`.

libmcount/mcount.c
```diff
@@ -1716,6 +1724,145 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+		}
+	}
+
+	sfd = socket_create(addr, getpid());
+	if (sfd == -1)
+		return sfd;
+
+	unlink_socket_maybe(addr);
```
namhyung, line 1751: Why do you want to remove it?

libmcount/mcount.c
```diff
@@ -1716,6 +1724,145 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+ * connection . */
+static void agent_kill()
+{
+	int sfd;
+	struct sockaddr_un addr;
+
+	if (agent_run)
+		agent_run = false;
+	else
+		return;
```
namhyung, line 1840: I prefer having return condition separately like

```c
if (!agent_run)
    return;

agent_run = false;
```

libmcount/mcount.c
```diff
@@ -1716,6 +1724,145 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+
+	close(sfd);
+
+	switch (pthread_join(agent, NULL)) {
+	case 0:
+	case EINVAL:
+		break;
+	default:
+		pr_dbg("agent left in unknown state\n");
+	}
```
namhyung, line 1862: Why do you handle EINVAL specially?  I think we can just add `pr_dbg()` for the any error return.

utils/socket.c
```diff
@@ -0,0 +1,67 @@
...
+	return 0;
+}
+
+int socket_connect(int fd, struct sockaddr_un *addr)
+{
+	if (connect(fd, (struct sockaddr *)addr, sizeof(struct sockaddr_un)) == -1) {
+		pr_warn("cannot connect to socket '%s': %s\n", addr->sun_path, strerror(errno));
+		return -1;
+	}
+	else
```
namhyung, line 60: You can remove the else.

cmds/live.c
```diff
@@ -95,6 +97,39 @@ static void setup_child_environ(struct uftrace_opts *opts)
...
+	if (socket_send_option(sfd, UFTRACE_DOPT_CLOSE, NULL, 0) == -1) {
+		pr_warn("cannot terminate agent connection\n");
+		ret = -1;
+	}
+	else {
+		enum uftrace_dopt ack;
+		if (read(sfd, &ack, sizeof(enum uftrace_dopt)) == -1)
+			ret = -1;
+		else
+			ret = -!(ack == UFTRACE_DOPT_CLOSE);
```
namhyung, line 125: Please change it to

```c
if (read(...) < 0 ||
    ack != UFTRACE_DOPT_CLOSE)
        ret = -1;
```

tests/t273_agent_basic.py
```diff
@@ -0,0 +1,37 @@
...
+        sleep(.01)              # time for the agent to start
+        self.subcmd = 'live'
+        self.option = '-p %d' % self.record_p.pid
+        self.exearg = ''
+
+    def postrun(self, result):
+        self.record_p.communicate(b"^D") # target waits for a char to end
+        self.record_p.wait()
+        if result == TestBase.TEST_DIFF_RESULT: # the expected result is empty
+            result = TestBase.TEST_SUCCESS
```
namhyung, line 36: This is wierd... What's the actual output?

namhyung (commented):
Nit: small style changes.

libmcount/mcount.c
clementguidi: It is a critical error, so I'm making it return.

However `socket_create()` doesn't bind to the socket, so it doesn't interact with the filesystem.
`socket_listen` does (I gather `bind()` and `listen()` calls there), but I feel like verifying that the directory exists prior to attempting socket creation/binding.

libmcount/mcount.c
clementguidi: Socket binding fails when the file already exists. The only case I see it happening is if a PID is used twice by a uftrace target, and isn't cleared by the first one.

tests/t273_agent_basic.py
clementguidi: I only want to test the exit value of `uftrace live --pid`, as it produces no trace output. That's the only way I could think of using the test setup: starting a target with an agent in the background, and check if the client (`uftrace live --pid`) can communicate with the agent.
The test doesn't expect an output from the `live` command, so `TEST_DIFF_RESULT` is returned. But I assume the test is successful as long as the exit code from `live` is 0.

libmcount/mcount.c
clementguidi: It was in case the agent thread terminated by itself, but this shouldn't happen.

libmcount/mcount.c
namhyung: Ok, then I think we should warn instead if the file exists.

libmcount/mcount.c
namhyung: I'm ok with checking EEXIST.  Other error code means it has a more serious error so it should fail.

tests/t273_agent_basic.py
namhyung: I guess it can get the usual output with the agent and the expected output is something like below.  Then, why not just set the output and check the return code?

```
            [266409] | main() {
   2.530 us [266409] |   getchar();
   3.037 us [266409] | } /* main */
```

tests/t273_agent_basic.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+EXEARG = 't-agent'
```
namhyung, line 8: I don't think we need this because it'd be set by default.

tests/t273_agent_basic.py
clementguidi: Right, I wasn't sure how to use the test structure to start multiple processes. Now I check the client return code in the `prerun()` and use the `replay` command in the `run()` so I get a trace output.

tests/t273_agent_basic.py
```diff
@@ -0,0 +1,43 @@
...
+        self.option = '--keep-pid -g'
+        self.exearg = 't-' + self.name
+        record_cmd  = self.runcmd()
+        self.pr_debug("prerun command: " + record_cmd)
+        record_p = sp.Popen(record_cmd.split(), stdin=sp.PIPE, stdout=sp.PIPE, stderr=sp.PIPE)
+
+        sleep(.05)              # time for the agent to start
+        self.subcmd = 'live'
+        self.option = '-p %d' % record_p.pid
+        self.exearg = ''
```
namhyung, line 28: It'd nice if this command actually doing something to change the output.  I think we can change this test after we add some new features/operations later.

namhyung (approved):
LGTM

tests/t273_agent_basic.py
clementguidi: Sure, I'll add tests in other pull requests.

clementguidi:
Great, thank you for your guidance!

PR #1545, opened 2022-08-17
utils: Add TRAP() for adding explicit breakpoints

honggyukim:
Sometimes it's useful to add explicit breakpoints in the source code.
This patch adds a portable way of adding breakpoints with TRAP() macro.

### DIFF
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -396,12 +396,18 @@ char *uftrace_strerror(int errnum, char *buf, size_t buflen);
 
 void stacktrace(void);
 
+#if defined(__x86_64__) || defined(__i386__)
+#define TRAP() asm volatile("int $3")
+#else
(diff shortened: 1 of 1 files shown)


utils/utils.h
```diff
@@ -413,6 +413,12 @@ void stacktrace(void);
 		raise(SIGTRAP);                                                                    \
 	}
 
+#if defined(__x86_64__) || defined(__i386__)
+#define TRAP() asm volatile("int $3")
+#else
+#define TRAP() raise(SIGTRAP)
+#endif
```
namhyung, line 420: Maybe `ASSERT` and `DASSERT` can use it?

honggyukim: Yes, right. I've just updated it.

namhyung (approved):
LGTM

Issue #1509, opened 2022-08-05
Random failure in ./runtest 222


JHH20:
Reproduced random failure for `./runtest -vdik 222` on x64 WSL2 Ubuntu 20.04
The info below is from one instance of failure where the external data comment is missing

`uftrace replay -f time` output
```
#     TIMESTAMP       FUNCTION
   426883.003490100 | main() {
   426883.003490500 |   a() {
   426883.003490600 |     b() {
   426883.003490700 |       c() {
   426883.003491400 |         getpid();
   426883.003493200 |       } /* c */
   426883.003493500 |     } /* b */
   426883.003493600 |   } /* a */
   426883.003493700 | } /* main */
```
Expected output
```
#     TIMESTAMP       FUNCTION
   426883.003490100 | main() {
   426883.003490101 | /* external-data: user message */
   426883.003490500 |   a() {
   426883.003490600 |     b() {
   426883.003490700 |       c() {
   426883.003491400 |         getpid();
   426883.003493200 |       } /* c */
   426883.003493500 |     } /* b */
   426883.003493600 |   } /* a */
   426883.003493700 | } /* main */
```
`extern.dat` content: `426883.3490101 user message`
- Editing the file to add 2 leading zeros after the decimal does not solve the problem

Above failure might be due to parsing error when timestamp has leading zero (unverified claim)

_Originally posted by @JHH20 in https://github.com/namhyung/uftrace/issues/1063#issuecomment-1206474208_

So far, this test seems to fail in 2 different ways:
- The external-data message is missing entirely
- The message is misplaced between 2 functions rather than directly below main()

I originally thought it might be related to buffering differences between tty and non-tty, but passing a pseudo-tty in ./runtest.py via pty.openpty did not solve the issue.

JHH20:
Adding output when external-data message is misplaced between 2 functions.

Test ran as `./runtest.py -dk 222` and the following was the 4th test
```
#     TIMESTAMP       FUNCTION
    35775.262034200 | __monstartup();
    35775.262035200 | __cxa_atexit();
    35775.262035800 | main() {
    35775.262035800 |   a() {
    35775.262035801 | /* external-data: user message */
    35775.262035900 |     b() {
    35775.262035900 |       c() {
    35775.262036000 |         getpid();
    35775.262036400 |       } /* c */
    35775.262036500 |     } /* b */
    35775.262036600 |   } /* a */
    35775.262036600 | } /* main */
```

[...]

PR #1549, opened 2022-08-19
tests: Fix int conversion issue where zeros were discarded

jonghyeon-an:
This is a fix for the test case found in #1509.

There was a issue where if there was a zero after the decimal point
like 8490.080467500, the zero was deleted when converting to int.

```
  $ uftrace replay -f +time
              [  9250]     8490.080467500 | main() {
              [  9250]     8490.080467800 |   a() {
              [  9250]     8490.080467800 |     b() {
              [  9250]     8490.080467900 |       c() {
     1.200 us [  9250]     8490.080468700 |         getpid();
     2.700 us [  9250]     8490.080470600 |       } /* c */
     2.900 us [  9250]     8490.080470700 |     } /* b */
     3.000 us [  9250]     8490.080470800 |   } /* a */
     3.400 us [  9250]     8490.080470900 | } /* main */
```

After the fix, it converts to 8490.080467501 instead of 8490.80467501.

### Current
```
$ cat uftrace.data/extern.dat 
8490.80467501 user message
```

### After changes
```
$ cat uftrace.data/extern.dat 
8490.080467501 user message
```

### Test result
```
$ ./runtest.py -dik 222
Compiler                  gcc
Test case                 finstrument-fu
------------------------: O0 O1 O2 O3 Os
222 external_data       : OK OK OK OK OK

$ ./runtest.py -dik 222
Compiler                  gcc
Test case                 finstrument-fu
------------------------: O0 O1 O2 O3 Os
222 external_data       : OK OK OK OK OK

$ ./runtest.py -dik 222
Compiler                  gcc
Test case                 finstrument-fu
------------------------: O0 O1 O2 O3 Os
222 external_data       : OK OK OK OK OK

$ ./runtest.py -dik 222
Compiler                  gcc
Test case                 finstrument-fu
------------------------: O0 O1 O2 O3 Os
222 external_data       : OK OK OK OK OK

$ ./runtest.py -dik 222
Compiler                  gcc
Test case                 finstrument-fu
------------------------: O0 O1 O2 O3 Os
222 external_data       : OK OK OK OK OK
```

### DIFF
--- a/tests/t222_external_data.py
+++ b/tests/t222_external_data.py
@@ -50,7 +50,7 @@ class TestCase(TestBase):
             nsec = int(t[point+1:point+10])
 
             # add the external data right after the first line
-            msg = '%s.%d %s\n' % (t[:point], nsec + 1, 'user message')
+            msg = '%s.%09d %s\n' % (t[:point], nsec + 1, 'user message')
 
             data_file = open(os.path.join('uftrace.data', 'extern.dat'), 'w')
             data_file.write(msg)


JHH20:
How about `str.rjust(9, '0')` for python 2.4+ or `str.zfill(9)` for python 2.2.2+?

jonghyeon-an:
Thanks for the comment!
Isn't there a case where the decimal point of timestamp is output as less than 9?
When str.rjust(9, '0') or str.zfill(9) is used, it seems to be a problem when it comes output like 8490.467501.

JHH20:
The width could be set as `len(nsec)` but since uftrace output uses fixed width for the nanosecond portion, I think it's better to go for consistency. Anyways, I just wanted to suggest str method over string concatenation for better legibility.

jonghyeon-an:
If uftrace uses a fixed-point length, I also think it is better to change it for readability. 
Thanks for the confirmation, I'll change to use zfill.

honggyukim:
Thanks for the PR, but this is not correct. We should fix `parse_timestamp` logic instead of changing the test passed.

jonghyeon-an:
@honggyukim, Thanks for the comment!
In issue #1509, not only the parse_timestamp function but also the test case should be changed.
In this issue, the problem that occurs in the parse_timestamp function can be solved by changing the base of the strtoull function, but the other problem occurs in a case such as 8490.080467500 in test222.

honggyukim:
@jonghyeon-an You're right. The leading zeros are needed to compare the result, but how about adding zeros with this?
```diff
diff --git a/tests/t222_external_data.py b/tests/t222_external_data.py
index 48db8ce3..e78b0f97 100644
--- a/tests/t222_external_data.py
+++ b/tests/t222_external_data.py
@@ -50,7 +50,7 @@ class TestCase(TestBase):
             nsec = int(t[point+1:point+10])

             # add the external data right after the first line
-            msg = '%s.%d %s\n' % (t[:point], nsec + 1, 'user message')
+            msg = '%s.%09d %s\n' % (t[:point], nsec + 1, 'user message')

             data_file = open(os.path.join('uftrace.data', 'extern.dat'), 'w')
             data_file.write(msg)
```

jonghyeon-an:
@honggyukim Oh, this way is better! I'll change it.
Thanks for the suggestion :+1:

honggyukim (approved):
LGTM

namhyung (approved):
LGTM

PR #1550, opened 2022-08-20
utils: Fix the problem of handling numbers after the decimal point

jonghyeon-an:
This is a fix for an issue found in #1509.

In the parse_timestamp function, when a string such as 8959832.082682731
was input as an argument value, there was a problem that it was not parsed
correctly in the number of decimal places.

```
  $ make TESTARG="-v parse_times" unittest
    TEST     test_unit
  Running 1 test cases
  ======================
  Testing parse_timestamp...
  test failed at /home/user/uftrace/utils/utils.c:1121: time == tc[i].val
    time             = 8959832000000000
    tc[i].val        = 8959832082682731
  [001] parse_timestamp               : FAIL
  -------------
```

This was a parsing issue in the strtoull function caused by setting the base
to 0, and I solved the problem by specifying an explicit decimal base.
The parse_time function that uses strtoull was also changed.

Additionally, in the parse_time function, according to the limited_digits
condition, the number of decimal places was not converted correctly,
so the problem was solved through an additional condition.

```
  $ make TESTARG="-v parse_time" unittest
    TEST     test_unit
  Running 1 test cases
  ======================
  Testing parse_time...
  test failed at /home/user/uftrace/utils/utils.c:1094: time == tc[i].val
    time             = 148765
    tc[i].val        = 50987
  [001] parse_time                    : FAIL
  -------------

  $ make TESTARG="-v parse_time" unittest
    TEST     test_unit
  Running 1 test cases
  ======================
  Testing parse_time...
  test failed at /home/user/uftrace/utils/utils.c:1094: time == tc[i].val
    time             = 8107641850123
    tc[i].val        = 8107006418501
  [001] parse_time                    : FAIL
  -------------
```

Added unit test cases related to parse_time and parse_stamp so that the
same problem can be detected in advance.

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -475,13 +475,13 @@ uint64_t parse_time(char *arg, int limited_digits)
 	char *unit, *pos;
 	int i, decimal_places = 0, exp = 0;
 	uint64_t limited, decimal = 0;
-	uint64_t val = strtoull(arg, &unit, 0);
+	uint64_t val = strtoull(arg, &unit, 10);
 
 	pos = strchr(arg, '.');
 	if (pos != NULL) {
 		while (*(++pos) == '0')
 			decimal_places++;
-		decimal = strtoull(pos, &unit, 0);
+		decimal = strtoull(pos, &unit, 10);
 	}
 
 	limited = 10;
@@ -532,12 +532,12 @@ uint64_t parse_timestamp(char *arg)
 	uint64_t ts, tmp;
 	int len;
 
-	tmp = strtoull(arg, &sep, 0);
+	tmp = strtoull(arg, &sep, 10);
 	ts = tmp * NSEC_PER_SEC;
 
 	if (*sep == '.') {
 		arg = sep + 1;
-		tmp = strtoull(arg, &sep, 0);
+		tmp = strtoull(arg, &sep, 10);
 
 		len = 0;
 		while (isdigit(*arg)) {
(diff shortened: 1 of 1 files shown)


jonghyeon-an:
Is it better to prefix the test case name of utils with utils_?

namhyung (commented):
> Additionally, in the parse_time function, according to the limited_digits
condition, the number of decimal places was not converted correctly,
so the problem was solved through an additional condition.

This part is not clear.  Please be more specific - how it's not converted
correctly and what's the desired result, and so on.

namhyung:
> Is it better to prefix the test case name of utils with utils_?

Yeah, that's convention. :)

utils/utils.c
```diff
@@ -521,6 +521,8 @@ uint64_t parse_time(char *arg, int limited_digits)
 
 		for (i = decimal_places; i < exp; i++)
 			decimal *= 10;
+		while (decimal_places-- > exp)
+			decimal /= 10;
```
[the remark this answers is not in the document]
jonghyeon-an: OK, I'll write specifically when I reflect on another commit.

jonghyeon-an:
Only the part where the default argument value of the strtoull function is changed to 10 has been committed.

```
$ make TESTARG="-v utils_parse_time" unittest
  TEST     test_unit
Running 2 test cases
======================
Testing utils_parse_time...
[001] utils_parse_time              : PASS
-------------
Testing utils_parse_timestamp...
[002] utils_parse_timestamp         : PASS
-------------

unit test stats
====================
  2 ran successfully
  0 failed
  0 skipped
  0 signal caught
  0 unknown result
```

utils/utils.c
```diff
@@ -1065,4 +1065,57 @@ TEST_CASE(utils_strv)
...
+		{ "17.1m", 5, 1026000000000 },
+		{ "8960070.725168293s", 9, 8960070725168293 },
+		{ "8959832.082682731s", 9, 8959832082682731 },
+		{ "426883.003490100s", 9, 426883003490100 },
+		{ "8107.0641850s", 9, 8107064185000 },
+		{ "8107.006418501s", 9, 8107006418501 },
+	};
+	uint64_t time;
+	int i;
+
```
namhyung, line 1089: Could you add a `pr_dbg()` to briefly describe the test like "parsing time with a unit"?

utils/utils.c
```diff
@@ -1065,4 +1065,57 @@ TEST_CASE(utils_strv)
...
+	} tc[] = {
+		{ "8960070.725168293", 8960070725168293 },
+		{ "8959832.082682731", 8959832082682731 },
+		{ "426883.003490100", 426883003490100 },
+		{ "8107.0641850", 8107064185000 },
+		{ "8107.006418501234567890", 8107006418501 },
+	};
+	uint64_t time;
+	int i;
+
```
namhyung, line 1112: Ditto.  "parsing timestamp in various precisions with leading zeros".

utils/utils.c
jonghyeon-an: Sure. I'll add a description.

utils/utils.c
jonghyeon-an: I'll add it here too.

jonghyeon-an:
When I looked at the previous test cases, all the TC description were added.
In the future, when adding a test case, I'll add a description through pr_dbg.

```
$ make TESTARG="-v utils_parse_time" unittest
  TEST     test_unit
Running 2 test cases
======================
Testing utils_parse_time...
uftrace: parsing time with a unit
[001] utils_parse_time              : PASS
-------------
Testing utils_parse_timestamp...
uftrace: parsing timestamp in various precisions with leading zeros
[002] utils_parse_timestamp         : PASS
-------------

unit test stats
====================
  2 ran successfully
  0 failed
  0 skipped
  0 signal caught
  0 unknown result
```

namhyung (approved):
LGTM

honggyukim:
@jonghyeon-an This patch is partially based on @JHH20's report and analysis at #1509. He also suggested an additional fix at https://github.com/namhyung/uftrace/issues/1509#issuecomment-1220791077 so you better add his signed-off-by. So I would suggest you to update the commit message as follows.
```
tests: Fix int conversion issue where zeros were discarded
...
After the fix, it converts to 8490.080467501 instead of 8490.80467501.

Fixed: #1509
Signed-off-by: Jonghyeon An <markajh@gmail.com>
```

jonghyeon-an:
> @jonghyeon-an This patch is partially based on @JHH20's report and analysis at #1509. He also suggested an additional fix at [#1509 (comment)](https://github.com/namhyung/uftrace/issues/1509#issuecomment-1220791077) so you better add his signed-off-by. So I would suggest you to update the commit message as follows.
> 
> ```
> tests: Fix int conversion issue where zeros were discarded
> ...
> After the fix, it converts to 8490.080467501 instead of 8490.80467501.
> 
> Fixed: #1509
> Signed-off-by: JongHyeon Hwang <jhh.20@icloud.com>
> Signed-off-by: Jonghyeon An <markajh@gmail.com>
> ```

Thanks for the good suggestion!

honggyukim (approved):
LGTM

PR #1559, opened 2022-08-23
tests: Allow runtest.py --color to explicitly enable color

honggyukim:
This patch changes the usage of `--color` option in runtest.py from
boolean to string type, which can accept one of 'auto', 'on', or 'off'.

Having this option, we can always enable color in github action nightly.


honggyukim:
An example output in github action can be seen at https://github.com/honggyukim/uftrace/runs/7980904268?check_suite_focus=true. The test result shows colored output.

namhyung (approved):
LGTM

PR #1562, opened 2022-08-30
tests: Fix failed-tests.txt output alignment

honggyukim:
This patch fixes the following incorrect alignment in failed-tests.txt.
```
  $ cat failed-tests.txt
  Compiler                  gcc                            clang
  Test case                 pg             finstrument-fu  pg             finstrument-fu
  ------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os  O0 O1 O2 O3 Os O0 O1 O2 O3 Os
  203 arg_dwarf3          : OK OK NG NG NG SK SK SK SK SK OK OK OK OK OK SK SK SK SK SK
```
It also makes the failed-tests.txt output with color because there is
another way to print the plain output with '--color off' option.


namhyung (approved):
LGTM

Issue #1553, opened 2022-08-21
Code static checks infered by Infer(x86_64)


paranlee:
There are infering to improve the uftrace project codes from Facebook Infer.
- https://fbinfer.com/docs/analyzing-apps-or-projects/#using-a-compilation-database

**Install and static analyze process**
```
sudo apt -y install ocaml sqlite

git clone https://github.com/facebook/infer
cd infer
./build-infer.sh clang
## clang compile -> y
sudo make install

cd ~/uftrace
infer run --force-integration make
```

**Result**
```
Found 36 issues
          Issue Type(ISSUED_TYPE_ID): #
              Dead Store(DEAD_STORE): 27
  Null Dereference(NULL_DEREFERENCE): 5
        Resource Leak(RESOURCE_LEAK): 4
```

**List**
```
#0
cmds/dump.c:254: error: Dead Store
  The value written to `&end` (type `char*`) is never used.
  252. 				goto out;
  253. 			}
  254. 			*end++ = '\0';
           ^
  255. 
  256. 			pr_out("%s  session of task %d: %.*s (%s)\n", timestamp, tid, 16, sid,

#1
cmds/dump.c:735: error: Null Dereference
  pointer `event` last assigned on line 731 could be null and is dereferenced at line 735, column 2.
  733. 
  734. 	pr_time(frs->time);
  735. 	pr_out("%5d: [%s] %s:%s(%ld) %.*s\n", tid, rstack_type(frs), event->system, event->name,
        ^
  736. 	       frs->addr, size, event_data);
  737. 

#2
cmds/record.c:2081: error: Resource Leak
  resource acquired by call to `open()` at line 2072, column 14 is not released after line 2081, column 2.
  2079. 		pr_out("uftrace: install signal handlers to task %d\n", pid);
  2080. 
  2081. 	setup_writers(&wd, opts);
         ^
  2082. 	start_tracing(&wd, opts, ready);
  2083. 	close(ready);

#3
cmds/recv.c:266: error: Dead Store
  The value written to `&namelen` (type `int`) is never used.
  264. 		pr_err("map read failed");
  265. 
  266. 	namelen = htonl(namelen);
        ^
  267. 
  268. 	pr_dbg2("send UFTRACE_MSG_SEND_META_DATA: %s\n", filename);

#4
cmds/tui.c:1842: error: Dead Store
  The value written to `&len` (type `int`) is never used.
  1840. 	switch ((long)s) {
[...]

PR #1567, opened 2022-09-04
utils: Null dereference amend

paranlee:
issue #1553 related Null dereference amend works.

### DIFF
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -127,6 +127,9 @@ bool check_kernel_pid_filter(void)
 	bool ret = true;
 	char *filename = get_tracing_file("options/function-fork");
 
+	if (filename == NULL)
+		return false;
+
 	if (!access(filename, F_OK))
 		ret = false;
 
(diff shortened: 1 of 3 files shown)


utils/kernel.c
```diff
@@ -127,6 +127,9 @@ bool check_kernel_pid_filter(void)
 	bool ret = true;
 	char *filename = get_tracing_file("options/function-fork");
 
+	if (filename == NULL)
+		return false;
```
namhyung, line 131: I'm not sure if it could return NULL.

utils/script.c
```diff
@@ -41,6 +41,9 @@ enum script_type_t get_script_type(const char *str)
 {
 	char *ext = strrchr(str, '.');
 
+	if (ext == NULL)
+		return;
```
namhyung, line 45: Return `SCRIPT_UNKNOWN`

paranlee: ok, thank you!

utils/kernel.c
paranlee: Would it be better to remove that null check?

namhyung: No, I think it's possible.  Please keep it.

paranlee: I got it. ;)

paranlee:
@namhyung Thank you for code review! my patch updated.

PR #1571, opened 2022-09-04
scripts: Correct dump scripts timestamp representation

JSYoo5B:
As the timestamp represents in nanosecond unit, dump script wants to
show in second unit. But, in modular, the digits may not filled enough

ex.
  _time = 1001002003
  _time / unit == 1
  _time % unit == 1002003

expected:
  1.001002003

actual:
  1.1002003
  // may translated as 1s, 100ms, 200us, 300ns

So, need to change formatting text to start with leading zero, fixed
nine digits.

Signed-off-by: JaeSang Yoo <jsyoo5b@gmail.com>

### DIFF
--- a/scripts/dump.lua
+++ b/scripts/dump.lua
@@ -20,7 +20,7 @@ function uftrace_entry(ctx)
     local _name       = ctx['name']
 
     local unit = 1000000000
-    print(string.format('%d.%d %6d: [entry] %s(%x) depth: %d',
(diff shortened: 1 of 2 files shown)


honggyukim:
Thanks for fixing the potential bug. But please change the commit title by reading other commit messages. I would suggest this.
```
scripts: Correct dump scripts timestamp representation
```

namhyung (approved):
LGTM

PR #1572, opened 2022-09-06
dump: Add an interaction option to the mermaid result.

wendyisdream:
This adds an interaction option to mermaid results. It allows users to
 check the call graph step by step with one uftrace result.
It supports the following graph following the user's input.
  a) Show the graph until specific call depth level.
     When the function has sub call flow, clicking the function node
      makes it attach the remaining call flow.
     It allows to toggle show/hide sub-graph.
  b) Show the partial graph starting from a specific function.
     This partial graph also support the a) depth option

For this feature, the following is added.
  1. Add the call depth level to mermaid node id as "depth_id" format.
  2. Add html and javascript to make the option. This code is included
      in the header file as raw string.
     This header file is converted incorrectly in pre-commit
      clang-format hook. So it is excluded in pre-commit.

Signed-off-by: Jia Ha <wendyisdream1004@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,9 @@ endif
 
 COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
 
+C_STR_TARGET = utils/mermaid.js utils/mermaid.html
+C_STR_EXTENSION = cstr
+
 #
 # Note that the plain CFLAGS and LDFLAGS can be changed
 # by config/Makefile later but *_*FLAGS can not.
@@ -281,7 +284,7 @@ $(UFTRACE_ARCH_OBJS): $(wildcard $(srcdir)/arch/$(ARCH)/*.[cS]) $(COMMON_DEPS)
 $(objdir)/libtraceevent/libtraceevent.a: $(wildcard $(srcdir)/libtraceevent/*.[ch]) $(objdir)/.config
 	@$(MAKE) -C $(srcdir)/libtraceevent BUILD_SRC=$(srcdir)/libtraceevent BUILD_OUTPUT=$(objdir)/libtraceevent CONFIG_FLAGS="$(TRACEEVENT_CFLAGS)"
 
-$(objdir)/uftrace.o: $(srcdir)/uftrace.c $(objdir)/version.h $(COMMON_DEPS)
+$(objdir)/uftrace.o: $(srcdir)/uftrace.c $(objdir)/version.h c-str-conversion $(COMMON_DEPS)
 	$(QUIET_CC)$(CC) $(UFTRACE_CFLAGS) -c -o $@ $<
 
 $(objdir)/misc/demangler.o: $(srcdir)/misc/demangler.c $(objdir)/version.h $(COMMON_DEPS)
@@ -383,6 +386,7 @@ clean:
 	$(Q)$(RM) $(objdir)/uftrace-*.tar.gz $(objdir)/version.h
 	$(Q)find -name "*\.gcda" -o -name "*\.gcno" | xargs $(RM)
 	$(Q)$(RM) coverage.info
+	$(Q)$(RM) $(objdir)/utils/*.$(C_STR_EXTENSION)
 	@$(MAKE) -sC $(srcdir)/arch/$(ARCH) clean
 	@$(MAKE) -sC $(srcdir)/tests ARCH=$(ARCH) clean
 	@$(MAKE) -sC $(docdir) clean
@@ -395,5 +399,10 @@ reset-coverage:
 ctags:
 	@find . -name "*\.[chS]" -o -path ./tests -prune -o -path ./check-deps -prune \
 		| xargs ctags --regex-asm='/^(GLOBAL|ENTRY|END)\(([^)]*)\).*/\2/'
+
+c-str-conversion:
+	@for i in ${C_STR_TARGET}; do \
+		sed -e 's#\\#\\\\#g;s#\"#\\"#g;s#$$#\\n\"#;s#^#\"#' $${i} > $${i}.$(C_STR_EXTENSION); \
+	done
 
 .PHONY: all config clean test dist doc ctags PHONY
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1311,45 +1311,65 @@ static void dump_mermaid_header(struct uftrace_dump_ops *ops, struct uftrace_dat
 
 	pr_out("<html>\n");
 	pr_out("<body>\n");
-	pr_out("<script src=\"https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js\"></script>\n");
-	pr_out("<script>\n");
-	pr_out("  var config = {\n");
-	pr_out("  startOnLoad: true,\n");
-	pr_out("  maxTextSize: 99999999,\n");
-	pr_out("  flowchart: { useMaxWidth: false }\n");
-	pr_out("  };\n");
-	pr_out("  mermaid.initialize(config);\n");
-	pr_out("</script>\n");
 }
 
 static void print_graph_node_mermaid(struct uftrace_graph *graph, struct uftrace_graph_node *node)
 {
 	char *symname = node->name;
 	struct uftrace_graph_node *child;
-
+	static int depth = -1;
+	depth++;
 	list_for_each_entry(child, &node->head, list) {
-		pr_out("  %d[\"%s\"] -->|%d| %d[\"%s\"];\n", node->id, symname, child->nr_calls,
-		       child->id, child->name);
+		pr_out("  %d_%d[\"%s\"] -->|%d| %d_%d[\"%s\"];\n", depth, node->id, symname,
+		       child->nr_calls, depth + 1, child->id, child->name);
 		print_graph_node_mermaid(graph, child);
+	}
+	if (list_no_entry(child, &node->head, list)) {
+		depth--;
 	}
 }
 
 static void dump_mermaid_footer(struct uftrace_dump_ops *ops, struct uftrace_data *handle,
 				struct uftrace_opts *opts)
 {
+	const char *interactive_option_html;
+	const char *interactive_option_js;
+
 	/* skip empty graph */
 	if (mermaid_graph.root.time == 0 && mermaid_graph.root.nr_edges == 0)
 		return;
 
-	if (mermaid_graph.root.time || mermaid_graph.root.nr_edges) {
-		pr_out("<h2>Function Call Graph for <span style=\"color:blue\">%s</span>",
-		       mermaid_graph.root.name);
-		pr_out("</h2>\n");
-		pr_out("<div class=\"mermaid\">\n");
-		pr_out("flowchart TB\n");
-		print_graph_node_mermaid(&mermaid_graph, &mermaid_graph.root);
-		pr_out("</div>\n");
-	}
+	pr_out("<h2>Function Call Graph for <span style=\"color:blue\" id=\"baseName\">%s</span>",
+	       mermaid_graph.root.name);
+	pr_out("</h2>\n");
+
+	interactive_option_html =
+#include "utils/mermaid.html.cstr" /* This file is a converted of mermaid.html to one string literal in build-time */
+		;
+
+	pr_out("%s", interactive_option_html);
+
+	pr_out("<div class=\"mermaid\" id=\"meraid_graph\">\n");
+	pr_out("flowchart TB\n");
+	print_graph_node_mermaid(&mermaid_graph, &mermaid_graph.root);
+	pr_out("</div>\n");
+	pr_out("<script src=\"https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js\"></script>\n");
+	pr_out("<script>\n");
+	pr_out("var config = {\n");
+	pr_out("	startOnLoad: true,\n");
+	pr_out("	maxTextSize: 99999999,\n");
+	pr_out("	flowchart: { useMaxWidth: false, curve : \'basis\', htmlLabels: \'false\'},\n");
+	pr_out("	securityLevel:\'loose\'\n");
+	pr_out("};\n");
+	pr_out("mermaid.initialize(config);\n");
+
+	interactive_option_js =
+#include "utils/mermaid.js.cstr" /* This file is a converted of mermaid.html to one string literal in build-time */
+		;
+
+	pr_out("%s", interactive_option_js);
+
+	pr_out("</script>\n");
 	pr_out("</body>\n");
 	pr_out("</html>\n");
 }
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -449,7 +449,7 @@ class TestBase:
         start_mermaid = False
 
         for ln in output.split('\n'):
-            if ln.find('<div class=\"mermaid\">') >= 0:
+            if ln.find('<div class=\"mermaid\"') >= 0:
                 start_mermaid = True
                 continue
             if start_mermaid == False:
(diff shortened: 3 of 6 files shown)


wendyisdream:
Below is this option's operation example.
example_mermaid_option

namhyung:
Thanks for this work.  Looks very useful!  I'll take a look later.

cmds/dump.c
```diff
@@ -7,6 +7,7 @@
 
 #include "libtraceevent/event-parse.h"
 #include "libtraceevent/kbuffer.h"
+#include "mermaid.h"
```
namhyung, line 10: I'd like to move the header in the utils directory.

tests/runtest.py
```diff
@@ -449,17 +449,17 @@ def mermaid_sort(self, output, ignore_children=False):
...
+            if ln.find('<div class=\"mermaid\"') >= 0:
                 start_mermaid = True
                 continue
             if start_mermaid == False:
                 continue
             if ln.find('</div>') >= 0:
                 break
 
-            m = re.match( r'\s+(?P<start_id>\d+)\[\"(?P<start_name>\S+)\"\]\s+-->\|(?P<call_num>\d+)\|\s+(?P<end_id>\d+)\[\"(?P<end_name>\S+)\"\];', ln)
+            m = re.match( r'\s+(?P<start_depth>\d+)_(?P<start_id>\d+)\[\"(?P<start_name>\S+)\"\]\s+-->\|(?P<call_num>\d+)\|\s+(?P<end_depth>\d+)_(?P<end_id>\d+)\[\"(?P<end_name>\S+)\"\];', ln)
```
namhyung, line 460: Could you please this line shorter?  You can break it in multiple lines if possible.

cmds/mermaid.h
```diff
@@ -0,0 +1,256 @@
+const char *INTERACTIVE_OPTION_HTML = R""(
```
honggyukim, line 1: I think this header is not needed. Instead, the contents of this variable can be in a separate file such as `mermaid.html` and it can be written as follows and `cmds/dump.c` can directly use it.

In `cmds/dump.c`,
```c
const char *interactive_option_html =
#include "utils/mermaid.html"
;
```
In `utils/mermaid.html`,
```html
R""(
<style>
#options {
  border-collapse: collapse;
  width: 100%;
}
    ... (put all the remaining contents html related) ...
</table>
<br>
</details>
)""
```

cmds/mermaid.h
```diff
@@ -0,0 +1,256 @@
...
+          <option value="7">7</option>
+        </select>
+    </td>
+  </tr>
+</table>
+<br>
+</details>
+)"";
+
+const char *INTERACTIVE_OPTION_JS = R""(
```
honggyukim, line 61: This part can also be written similarly as follows.

In `cmds/dump.c`,
```c
const char *interactive_option_js = 
#include "utils/mermaid.js"
;
```
In `utils/mermaid.js`,
```js
R""(
const GRAPH_TITLE = new String('graph TB\n');
const FULL_DEPTH = 64;
    ... (put all the remaining contents html related) ...
)""
```

cmds/dump.c
honggyukim: I think it doesn't have to be a header. Please see my comments below.

honggyukim (commented):
Thanks for doing this work! I have some requests as follows.

honggyukim:
By the way, I didn't know R string with `R""(...)""` is allowed in C. Is this a standard C? If so, from which version?

Otherwise, this could make build errors in various other environments.

wendyisdream:
Thank you for the review!
I applied the code as you suggested.
For raw string literals,  I can find it was added to C++11 standards. 
But on C standard side, it seems to be just a kind of extension.
I can see gcc support it but clang for C doesn't.
So I changed this option string to C style including escape character for it. (I wanted to avoid this format but I couldn't find any other way in C standard yet)

cmds/dump.c
```diff
@@ -1308,45 +1308,63 @@ static void dump_mermaid_header(struct uftrace_dump_ops *ops, struct uftrace_dat
...
-		pr_out("</div>\n");
-	}
+	pr_out("<h2>Function Call Graph for <span style=\"color:blue\" id=\"baseName\">%s</span>",
+	       mermaid_graph.root.name);
+	pr_out("</h2>\n");
+
+	interactive_option_html =
+#include "utils/mermaid.html"
+		;
+	pr_out("%s", interactive_option_html);
```
namhyung, line 1346: Maybe you can make it a normal html file then copy the file directly (using `fwrite(..., outfp)`?)

cmds/dump.c
```diff
@@ -1308,45 +1308,63 @@ static void dump_mermaid_header(struct uftrace_dump_ops *ops, struct uftrace_dat
...
+	pr_out("	maxTextSize: 99999999,\n");
+	pr_out("	flowchart: { useMaxWidth: false, curve : \'basis\', htmlLabels: \'false\'},\n");
+	pr_out("	securityLevel:\'loose\'\n");
+	pr_out("};\n");
+	pr_out("mermaid.initialize(config);\n");
+
+	interactive_option_js =
+#include "utils/mermaid.js"
+		;
+	pr_out("%s", interactive_option_js);
```
namhyung, line 1365: Ditto.

cmds/dump.c
honggyukim: It makes uftrace binary depend on having `mermaid.{html,js}` when running so we chose to do this way.

wendyisdream: At first, I also considered using js/html file itself.
But in that case, it is additionally required for uftrace binary to know the path of js/html file because uftrace can be installed everywhere. 
So this PR includes js/html file to uftrace as Hongkki mentioned.

namhyung: I see.   Then how about keeping the file as a normal file and convert it to a C file at build time so that we can include it like here?  That'd be better than manually adding escape sequences.  Something like https://stackoverflow.com/a/5500077.

wendyisdream:
Thank you for this suggestion!

I updated the code to keep this js/html and convert it in build time and use the converted file itself.
I also considered converting the js/html file in place.
But I think using the converted file is simple for managing because it doesn't have any effect on the original in case of exception cases like build failure.

Makefile
```diff
@@ -396,4 +400,9 @@ ctags:
 	@find . -name "*\.[chS]" -o -path ./tests -prune -o -path ./check-deps -prune \
 		| xargs ctags --regex-asm='/^(GLOBAL|ENTRY|END)\(([^)]*)\).*/\2/'
 
+c-str-conversion:
+	@for i in ${C_STR_TARGET}; do \
+		sed -e 's/\\/\\\\/g;s/\"/\\"/g;s/$$/\\n\"/;s/^/\"/' $${i} > $${i}.$(C_STR_EXTENSION);\
```
namhyung, line 405: Does `s/^/\"/` part really necessary?

wendyisdream: (s/^/\"/) is for adding __"__ at the beginning of the line as the pair of __"__  of (s/$/\\n\"/) at the end of the line like *"          <option value=\"1\">1</option>\n"* for string literal concatenation.


Another way add the backslash at the end of the line and add " for the first and last only.
It will work in the sed command
's/\\/\\\\/g;s/\"/\\"/g;s/$/\\n\\/;1s/^/\"/;$ a \"'.

The current commit applied the string literal concatenation.

namhyung: Ah, I see.  Thanks for the explanation.  Looks good now :)

utils/mermaid.html
```diff
@@ -0,0 +1,57 @@
+<style>
+#options {
+  border-collapse: collapse;
+  width: 100%;
+}
+#options td, #options th {
+  border: 1px solid #ddd;
```
namhyung, line 7: Is it same as #0d0d0d or #000ddd?

utils/mermaid.html
```diff
@@ -0,0 +1,57 @@
...
+    <tr>
+      <td>
+          <button class="button" onclick="drawBaseGraph(0)">Full-Graph</button>
+          <button class="button" onclick="drawSelectGraph()">Sub-Graph Selection</button>
+          <span id="description" style="color:blue" > </span>
+      </td>
+      <td>
+        Depth:
+        <select id="depth-select">
+          <option value="64" selected>Full</option>
```
namhyung, line 44: 64 ?

utils/mermaid.js
```diff
@@ -0,0 +1,192 @@
+const GRAPH_TITLE = new String("graph TB\n");
+const FULL_DEPTH = 64;
```
namhyung, line 2: Where does this come from?

cmds/dump.c
```diff
@@ -1308,45 +1308,65 @@ static void dump_mermaid_header(struct uftrace_dump_ops *ops, struct uftrace_dat
...
+	pr_out("</h2>\n");
+
+	interactive_option_html =
+#include "utils/mermaid.html.cstr" /* This file is a converted of mermaid.html to one string literal in build-time */
+		;
+
+	pr_out("%s", interactive_option_html);
+
+	pr_out("<div class=\"mermaid\" id=\"meraid_graph\">\n");
+	pr_out("flowchart TB\n");
```
namhyung, line 1350: What is TB?

utils/mermaid.js
```diff
@@ -0,0 +1,192 @@
...
+		callFlow[i] = callFlow[i].replace(COMMENT, "");
+		if (Number(startDepth[i]) >= depth)
+			callFlow[i] = (COMMENT + callFlow[i]);
+		if (Number(startDepth[i]) == depth) {
+			id = startDepth[i] + "_" + startId[i];
+			if (nodeList.includes(id) == false) {
+				nodeList.push(id);
+				showHideDesc = showHideDesc.concat(
+					"  style " + id +
+					" fill:#f9f,stroke:#333,stroke-width:px;\n");
```
namhyung, line 91: Can we move this style change to a separate place?

utils/mermaid.js
```diff
@@ -0,0 +1,192 @@
...
+	let selectionDesc = new String();
+	let nodeList = new Array();
+	if (callFlow.length == 0)
+		drawBaseGraph(0);
+	for (let i = 1; i < callFlow.length; i++) {
+		let id = startDepth[i] + "_" + startId[i];
+		if (nodeList.includes(id) == false) {
+			nodeList.push(id);
+			selectionDesc = selectionDesc.concat(
+				"  style " + id + " fill:#CAB8FF,stroke:#333,stroke-width:px;\n");
```
namhyung, line 110: Ditto.

utils/mermaid.js
```diff
@@ -0,0 +1,192 @@
...
+	mermaid.initialize(config);
+	initializeBaseGraph();
+	let start = new Boolean(false);
+	maxDepth = 0;
+	var result = mermaidTxt.split(/\r?\n/);
+	for (let i = 0; i < result.length; i++) {
+		if (result[i].length == 0)
+			continue;
+		let oneCallFlow = result[i].match(
+			/\s+(?<start_depth>\d+)_(?<start_id>\d+)\[\"(?<start_name>\S+)\"\]\s+--\S+\|(?<call_num>\d+)\|\s+(?<end_depth>\d+)_(?<end_id>\d+)\[\"(?<end_name>\S+)\"];/);
```
namhyung, line 129: It's a bit hard to follow.  Could you please add a comment with examples for each case?

utils/mermaid.js
```diff
@@ -0,0 +1,192 @@
...
+	baseGraph = "".concat(GRAPH_TITLE, callFlow.join("\n"));
+	updateDepthList();
+	renderGraph(baseGraph);
+};
+
+var updateDepthList =
+	function() {
+	var select = document.getElementById("depth-select");
+	select.options.length = 0;
+	select.options[0] = new Option("Full", 64);
```
namhyung, line 161: FULL_DEPTH ?

namhyung (commented):
Sorry for the late review.

utils/mermaid.html
wendyisdream: It means #dddddd.
Please refer to the below link's [3 Digit HEX Value] part.
https://www.w3schools.com/css/css_colors_hex.asp

utils/mermaid.html
wendyisdream: 64 is the value I set to the initial call depth limit.
I just think there is no normal case to be over 64 call depth.

utils/mermaid.js
wendyisdream: I updated to use the value in Javascript as HTML's initial value for "full".

cmds/dump.c
wendyisdream: TB means the flowchart is oriented from Top to Bottom. '
(The other option is LR(Left to Right)])
Please refer to the below link.
https://mermaid-js.github.io/mermaid/#/flowchart

utils/mermaid.js
wendyisdream: I updated it to use the separate funtion.

utils/mermaid.js
wendyisdream: I updated it to use the separate function.

utils/mermaid.js
wendyisdream: I added the comment of an example.

utils/mermaid.js
wendyisdream: Yes. I updated it to use the defined constant.
Thank you for the review.

utils/mermaid.html
namhyung: Ok, thanks!

utils/mermaid.html
namhyung: You don't know how much it would use.  The default max call depth in uftrace is 1024 as C++ programs tend to have many small functions.  But practically we won't allow that many in the display..

utils/mermaid.js
```diff
@@ -0,0 +1,202 @@
...
+	for (let i = 0; i < callFlow.length; i++) {
+		callFlow[i] = callFlow[i].replace(COMMENT, "");
+		if (Number(startDepth[i]) >= depth)
+			callFlow[i] = (COMMENT + callFlow[i]);
+		if (Number(startDepth[i]) == depth) {
+			id = startDepth[i] + "_" + startId[i];
+			if (nodeList.includes(id) == false) {
+				nodeList.push(id);
+				showHideDesc =
+					showHideDesc.concat(getNodeStyle(id, "#f9f", "#333"), "\n");
```
namhyung, line 95: It'd be nice if we have a symbolic name for this constant so that one can customize it easily.

utils/mermaid.js
```diff
@@ -0,0 +1,202 @@
...
+	let selectionDesc = new String();
+	let nodeList = new Array();
+	if (callFlow.length == 0)
+		drawBaseGraph(0);
+	for (let i = 1; i < callFlow.length; i++) {
+		let id = startDepth[i] + "_" + startId[i];
+		if (nodeList.includes(id) == false) {
+			nodeList.push(id);
+			selectionDesc =
+				selectionDesc.concat(getNodeStyle(id, "#cab8ff", "#333"), "\n");
```
namhyung, line 115: Ditto.  Please use symbolic names for the hard coded values.

utils/mermaid.js
namhyung: Thanks!

namhyung (commented):
Thanks for the update!

utils/mermaid.js
wendyisdream: Thank you for the information!
I updated it to 1024

And I changed this color value to use a constant definition. It looks cleaner. Thank you.

namhyung (approved):
LGTM

Makefile
honggyukim: I would like to ask a small cosmetic change request as follows.
```suggestion
		sed -e 's#\\#\\\\#g;s#\"#\\"#g;s#$$#\\n\"#;s#^#\"#' $${i} > $${i}.$(C_STR_EXTENSION); \
```
It's to make less confusing by using `#` as a separator instead of `/` because using `/` and `\` together looks confusing.

honggyukim (commented):
Sorry, I couldn't take a look for a while but left a little nitpick change request.

wendyisdream:
> Sorry, I couldn't take a look for a while but left a little nitpick change request.

I updated the code as you suggested. Thank you!

honggyukim (approved):
Thanks for the update. LGTM.

Issue #1552, opened 2022-08-21
More Docker File, more Linux Distribution build support test 


paranlee:
Add Docker files based on the from Supported LTS to Latest LTS Linux distribution.

Adds SUSE, Rocky Linux Docker build Supports.

paranlee:
Related PR #1582

PR #1582, opened 2022-09-12
misc: More Docker File to distribution build support

paranlee:
Add suse and Latest Distribution Dockerfile. `docker` build and run `uftrace` (#1552 )


honggyukim:
Thanks for the work, but I feel like "Add dockerfiles for xxx" patches can be squashed into a single patch.

In addition, could you write the commit message more in details? I've made the same request at https://github.com/namhyung/uftrace/pull/1513#issuecomment-1218463369 so please read it and update the commit message again.
> the current commit message doesn't explain what the patches do. You can think that the commit message is a letter to code reviewers and future readers. (or even to yourself)
> 
> Rather than explaining many things in the comment here, you better explain what the patches do in the commit message as much as possible so that reviewers do not have to ask. Please read other commit messages and get reference from them.

honggyukim:
It'd be much more useful if you write the steps to run uftrace under those docker images in the commit message.

namhyung:
> It'd be much more useful if you write the steps to run uftrace under those docker images in the commit message.

Agreed.  Please add something like `misc/docker/run.sh` or something to do the build test for all (or some selected) docker images.  Optionally it can give a shell prompt.

yskelg:
Thank you @namhyung for the review. I will write a script that can build and run all or each of the Docker images. Additionally, I will add Docker files to match the latest distribution.

PR #1588, opened 2022-10-05
mcount: Update depth and time filters at runtime

clementguidi:
Hi,

This pull request implements a dynamic mechanism for `-D` and `-t` options. Both can be updated at runtime, using the agent.

~~It is constructed on top of #1543 (not accepted yet at the time of writing). The first four commits are from #1543. The last two commits actually implement the dynamic features.~~

Given a running `uftrace` instance with its agent enabled:
```
$ uftrace -g <some-target>
```

One can use the following commands to set new depth/time filter value:
```
$ uftrace -p `pidof some-target` -D <depth>
$ uftrace -p `pidof some-target` -t <threshold>
```

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -122,6 +122,16 @@ static int forward_options(struct uftrace_opts *opts)
 			pr_dbg3("sent option depth = %d\n", opts->depth);
 	}
 
+	if (opts->threshold) {
+		if (socket_send_option(sfd, UFTRACE_DOPT_THRESHOLD, &opts->threshold,
+				       sizeof(uint64_t)) == -1) {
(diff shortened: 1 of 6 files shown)


namhyung:
Could you please finish #1543 first?  I think there were remaining comments to be addressed and it still has some conflicts with the current master.

namhyung:
Rebases are fine, and it's actually recommended. :)

It'd be nice to modify the agent basic test, but you need to come up with an idea how to verify its behavior.  Feel free to change the target program (tests/s-agent.c) if needed.

clementguidi:
Superseded by #1644

PR #1600, opened 2022-11-13
make -Z/--size-filter work with non dynamic tracing

fistki:
-Z/--size-filter option supports non-dynamic tracing mode.

Here are examples.

```
./uftrace report -f size --no-pager
          Size  Function
    ==========  ====================
            72  main
            16  __monstartup
            22  a
           356  d
            20  b
            16 __cxa_atexit
            17  c
```

You can see functions under 20 bytes are filtered.
```
./uftrace record -Z 20 ~/a.out
./uftrace report -f size --no-pager
          Size  Function
    ==========  ====================
            72  main
            22  a
           356  d
```

You can also set a trigger for specific functions.
```
./uftrace -T main@size=20 ~/a.out
   7.500 us [ 14201] | __monstartup();
   0.900 us [ 14201] | __cxa_atexit();
            [ 14201] | main() {
   0.800 us [ 14201] |   a();
   0.800 us [ 14201] |   d();
   7.600 us [ 14201] | } /* main */
```


fistki:
I found that this patch works wrong with --estimate-return option.
If there are other errors, failed test cases or anything wrong, please tell me.

libmcount/mcount.c
```diff
@@ -1360,6 +1364,8 @@ static unsigned long __mcount_exit(long *retval)
 
 	rstack = &mtdp->rstack[mtdp->idx - 1];
 
+	sym = find_symtabs(&mcount_sym_info, rstack->child_ip);
+	rstack->size = sym->size;
```
namhyung, line 1378: This will add more overhead in the fast path.  I think it should call the function only if the size filter is specified.

libmcount/internal.h
```diff
@@ -53,6 +53,8 @@ struct filter_control {
 	uint16_t saved_depth;
 	uint64_t time;
 	uint64_t saved_time;
+	unsigned size;
+	unsigned saved_size;
```
namhyung, line 57: Once we check it in the entry path, I think we don't need to save it here.

namhyung (commented):
Thanks for working on this!

namhyung:
I wonder what would be the output when you run the program with size filter of 300.

fistki:
> I wonder what would be the output when you run the program with size filter of 300.

```
uftrace -Z 300 a.out
                [ 12226] | main() {
       0.800 us [ 12226] |   d();
     414.900 us [ 12226] | } /* main */
```
You can see the parent function remains even though size of the function is less than 300.
Trigger makes the same result.

```
uftrace -T main@size=300 a.out
  11.800 us [ 12682] | __monstartup();
   1.300 us [ 12682] | __cxa_atexit();
            [ 12682] | main() {
   1.200 us [ 12682] |   d();
  10.800 us [ 12682] | } /* main */
 ```

File: libmcount/internal.h (hunk 2 above)
Comment (fistki) [reply libmcount/internal.h]: Thanks for comments. I update what you advised, so please check them.

File: libmcount/mcount.c [hunk 3]
```diff
@@ -1196,7 +1196,8 @@ void mcount_exit_filter_record(struct mcount_thread_data *mtdp, struct mcount_re
 {
 	mtdp->record_idx--;
 
-	if (rstack->end_time - rstack->start_time > mcount_threshold ||
+	if (((rstack->end_time - rstack->start_time > mcount_threshold) &&
+	     (mcount_min_size == 0 || rstack->size > mcount_min_size)) ||
```
namhyung, line 1212: It seems we can remove `mcount_min_size == 0` part.

libmcount/mcount.c
```diff
@@ -1366,6 +1368,7 @@ static unsigned long __mcount_exit(long *retval)
 	unsigned long retaddr;
 
 	mtdp = get_thread_data();
+
```
namhyung, line 1383: Unnecessary change.

namhyung (commented):
I was thinking that `-Z 300` should not have `main`.  Also you need to change the title of the first commit since it didn't enable -Z option for non-dynamic cases at the moment.

libmcount/mcount.c
namhyung: Also I said the logic should be moved to `mcount_entry_filter_check()`.

libmcount/internal.h
fistki: Don't you need to save it when triggering?

libmcount/mcount.h
```diff
@@ -52,6 +52,7 @@ struct mcount_ret_stack {
 	int tid;
 	unsigned dyn_idx;
 	uint64_t filter_time;
+	unsigned size;
```
namhyung, line 55: I don't think we need this now.

utils/filter.c
```diff
@@ -94,6 +94,8 @@ static void print_trigger(struct uftrace_trigger *tr)
 		pr_dbg("\ttrigger: time filter %" PRIu64 "\n", tr->time);
 	if (tr->flags & TRIGGER_FL_CALLER)
 		pr_dbg("\ttrigger: caller filter\n");
+	if (tr->flags & TRIGGER_FL_SIZE_FILTER)
+		pr_dbg("\ttrigger: size filter %d\n", tr->size);
```
namhyung, line 98: Shouldn't it be `%u` as tr->size is unsigned?

libmcount/internal.h
namhyung: You're right, a trigger can change the size filter setting so it needs to remember old filter data.  That means we cannot simply check the size in the entry path to drop them early. Hmm..

libmcount/mcount.c
```diff
@@ -1182,6 +1182,9 @@ enum filter_result mcount_entry_filter_check(struct mcount_thread_data *mtdp, un
 	if (mcount_check_rstack(mtdp))
 		return FILTER_RSTACK;
 
+	if (mcount_min_size > 0 && mcount_getsize(&mcount_sym_info, child) < mcount_min_size)
+		return FILTER_OUT;
+
```
namhyung, line 1202: This is only used when DISABLE_MCOUNT_FILTER is defined which means it's used without other filter/trigger options.  But I see you added the trigger action in the later patch.

libmcount/mcount.c
```diff
@@ -106,6 +106,9 @@ static pthread_t agent;
 /* state flag for the agent */
 static bool agent_run = false;
 
+/* whether size filter is activated */
+static bool mcount_size_enabled;
```
namhyung, line 110: Maybe we can get rid of this.

libmcount/mcount.c
```diff
@@ -908,10 +914,16 @@ enum filter_result mcount_entry_filter_check(struct mcount_thread_data *mtdp, un
...
 			mtdp->filter.time = tr->time;
+
+		if (tr->flags & TRIGGER_FL_SIZE_FILTER)
+			mtdp->filter.size = tr->size;
 	}
 
 #undef FLAGS_TO_CHECK
 
+	if (mtdp->filter.size > 0)
+		mcount_size_enabled = true;
```
namhyung, line 925: This would hurt performance since it keeps writing to a shared variable.

libmcount/mcount.c
```diff
@@ -995,11 +1007,13 @@ void mcount_entry_filter_record(struct mcount_thread_data *mtdp, struct mcount_r
 				struct uftrace_trigger *tr, struct mcount_regs *regs)
 {
 	if (mtdp->filter.out_count > 0 ||
-	    (mtdp->filter.in_count == 0 && mcount_filter_mode == FILTER_MODE_IN))
+	    (mtdp->filter.in_count == 0 && mcount_filter_mode == FILTER_MODE_IN) ||
+	    rstack->size < mtdp->filter.size)
```
namhyung, line 1011: can we do like this?

```c
(mtdp->filter.size > 0 && mcount_get_size(rstack->child_ip) < mtdp->filter.size)
```

Then we don't need `rstack->size`.

libmcount/mcount.c
```diff
@@ -1339,6 +1354,9 @@ static int __mcount_entry(unsigned long *parent_loc, unsigned long child, struct
 	rstack->nr_events = 0;
 	rstack->event_idx = ARGBUF_SIZE;
 
+	if (mcount_size_enabled)
+		rstack->size = mcount_getsize(&mcount_sym_info, child);
```
namhyung, line 1358: Also we have `__cygprof_entry()` and `_xray_entry()`.  That why it's better to do it in `mcount_entry_filter_record()`.

namhyung (commented):
Mostly looks good.  Could you please add some tests?

fistki:
@namhyung Thanks for reviews! I updated what you said, but testing may takes more time.

fistki (commented):
This commit makes same problem that parent function still remains even though it has small size than size filter.
`get_task_ustack()` returns `rstack_list` before it reach parent function's exit phase.
I don't know how to deal with this issue...

libmcount/mcount.c
```diff
@@ -992,11 +1007,13 @@ void mcount_entry_filter_record(struct mcount_thread_data *mtdp, struct mcount_r
 				struct uftrace_trigger *tr, struct mcount_regs *regs)
 {
 	if (mtdp->filter.out_count > 0 ||
-	    (mtdp->filter.in_count == 0 && mcount_filter_mode == FILTER_MODE_IN))
+	    (mtdp->filter.in_count == 0 && mcount_filter_mode == FILTER_MODE_IN) ||
+	    rstack->size < mtdp->filter.size)
 		rstack->flags |= MCOUNT_FL_NORECORD;
```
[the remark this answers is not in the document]
honggyukim: You can check if your logic works with -finstrument-functions compiled binary.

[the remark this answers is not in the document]
namhyung: The following test case didn't work.. please check.

```
$ uftrace tests/t-abc
   2.566 us [844343] | __monstartup();
   1.051 us [844343] | __cxa_atexit();
            [844343] | main() {
            [844343] |   a() {
            [844343] |     b() {
            [844343] |       c() {
   2.291 us [844343] |         getpid();
   3.133 us [844343] |       } /* c */
   3.575 us [844343] |     } /* b */
   3.963 us [844343] |   } /* a */
   4.289 us [844343] | } /* main */

$ readelf -s tests/t-abc | grep -e main -e ' [abc]$'
    30: 0000000000001210    49 FUNC    LOCAL  DEFAULT   14 c
    31: 0000000000001250    20 FUNC    LOCAL  DEFAULT   14 b
    32: 0000000000001270    20 FUNC    LOCAL  DEFAULT   14 a
    69: 0000000000001080    53 FUNC    GLOBAL DEFAULT   14 main

$ uftrace -T b@size=100 abc-pg-full
   1.199 us [890714] | __monstartup();
   0.447 us [890714] | __cxa_atexit();
            [890714] | main() {
   2.523 us [890714] |   a();
   2.964 us [890714] | } /* main */

$ uftrace -Z 100 -T b@read=10 tests/t-abc

```

For the last case, I expect to see `b`, `c` and `getpid` functions.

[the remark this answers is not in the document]
namhyung: It needs to clear the size filter for live command.

```diff
diff --git a/cmds/live.c b/cmds/live.c
index a9cc362726..74362c0ddc 100644
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -36,6 +36,7 @@ static void reset_live_opts(struct uftrace_opts *opts)
        opts->filter = NULL;
        free(opts->caller);
        opts->caller = NULL;
+       opts->size_filter = 0;
 
        /*
         * Likewise, most trigger options are ignored, but color settings and
```

honggyukim:
@fistki Thanks for the update. I have tested it as follows.
```
$ uftrace record t-abc

$ uftrace report -f +size
  Total time   Self time       Calls        Size  Function
  ==========  ==========  ==========  ==========  ====================
    0.875 us    0.125 us           1         120  main
    0.750 us    0.125 us           1          44  a
    0.625 us    0.125 us           1          44  b
    0.500 us    0.208 us           1          88  c
    0.333 us    0.333 us           1          16  __monstartup
    0.292 us    0.292 us           1          16  getpid
    0.250 us    0.250 us           1          16  __cxa_atexit
```
The following result looks correct.
```
$ uftrace replay -Z 100
   0.875 us [1007515] | main();
```
However, this result looks unexpected because it shows `a` and `b` although their size are both 44, which is less than 50.
```
$ uftrace replay -Z 50
            [1007515] | main() {
            [1007515] |   a() {
            [1007515] |     b() {
   0.500 us [1007515] |       c();
   0.625 us [1007515] |     } /* b */
   0.750 us [1007515] |   } /* a */
   0.875 us [1007515] | } /* main */
```
This case works fine when running in record or live command.
```
$ uftrace -Z 50 t-abc
            [1008150] | main() {
   1.750 us [1008150] |   c();
   2.751 us [1008150] | } /* main */

$ uftrace record -Z 50 t-abc
$ uftrace replay
            [1008755] | main() {
   0.458 us [1008755] |   c();
   0.917 us [1008755] | } /* main */
```
We also need to cover replay time usage as well.

fistki:
@honggyukim The failure is caused by https://github.com/namhyung/uftrace/pull/1600/commits/a130752da0473470cb40528262087a088465a068.
record and live command filter function size by `mcount_entry_filter_record()`, but replay and other commands are filtering by `get_task_ustack()`.
`get_task_ustack()` returns `rstack_list` before it reach parent function's exit phase, so the failure is ocurring.
When time filter is applied, it is not a problem because parent function always takes much time then child, but when size filter is applied, size of parent function can be smaller then child.

At this time, I don't know how to deal it. I will try something else and get back to you. Thanks!

namhyung:
I think the current code works fine and hides small parents.  But the problem is that uftrace doesn't keep the function size precisely.  It only saves the start address and the size will be calculated as a difference between two adjacent function addresses.  This leads inaccurate function size if function start address is aligned (it adds a padding at the end of function).  So the size filter at replay will show different results than record (which uses accurate function size from ELF symbol table).

fistki:
```
$ uftrace report -f +size 
 Total time   Self time       Calls        Size  Function
  ==========  ==========  ==========  ==========  ====================
   27.922 us   27.922 us           1          16  __monstartup
    6.101 us    1.322 us           1          94  main
    4.779 us    1.092 us           1          24  a
    3.687 us    1.172 us           1          24  b
    2.515 us    1.393 us           1          57  c
    1.293 us    1.293 us           1          16  __cxa_atexit
    1.122 us    1.122 us           1          16  getpid
```
```
$ readelf -s tests/t-abc | grep -e main -e ' [abc]$'
    13: 0000000000001229    24 FUNC    LOCAL  DEFAULT   16 a
    14: 0000000000001241    24 FUNC    LOCAL  DEFAULT   16 b
    15: 0000000000001259    57 FUNC    LOCAL  DEFAULT   16 c
    41: 0000000000001292    80 FUNC    GLOBAL DEFAULT   16 main
```

@namhyung Do you say difference of these cases?

namhyung:
It looks like your compiler doesn't align the function start address, but I still noticed a difference in the `main` function.

fistki:
> @namhyung In that case, Would it good to save size information to the symbol file?

I updated PR about it. Now, It seems that it has same func size with ELF sym table.

```
  Total time   Self time       Calls        Size  Function
  ==========  ==========  ==========  ==========  ====================
    1.222 us    0.220 us           1          80  main
    1.002 us    0.170 us           1          24  a
    0.832 us    0.211 us           1          24  b
    0.621 us    0.450 us           1          57  c
    0.310 us    0.310 us           1          16  __monstartup
    0.171 us    0.171 us           1          16  getpid
    0.120 us    0.120 us           1          16  __cxa_atexit
```

But, There's still problem that function is printed although `a`, `b` is smaller than size filter `50`.
```
            [  1689] | main() {
            [  1689] |   a() {
            [  1689] |     b() {
   0.621 us [  1689] |       c();
   0.832 us [  1689] |     } /* b */
   1.002 us [  1689] |   } /* a */
   1.222 us [  1689] | } /* main */

```

honggyukim:
> @namhyung In that case, Would it good to save size information to the symbol file?

Your change looks good but it breaks the backward compatibility, which means previously recorded `uftrace.data` cannot be properly replayed after your change is committed because one more column is added to `.sym` file to store the size information.

@namhyung I think having the size information is useful, but do you have any idea if we can keep it while preserving the backward compatibility? I feel like it's not easy because we don't have versioning schemes in `.sym` files.

fistki:
@honggyukim It makes a fault when old uftrace reads the new `.sym` file.
```
  Total time   Self time       Calls        Size  Function
  ==========  ==========  ==========  ==========  ====================
    1.283 us    0.221 us           1          94  00000050 main
    1.062 us    0.171 us           1          24  00000018 a
    0.891 us    0.190 us           1          24  00000018 b
    0.701 us    0.540 us           1          57  00000039 c
    0.261 us    0.261 us           1          16  00000010 __monstartup
    0.161 us    0.161 us           1          16  00000010 getpid
    0.130 us    0.130 us           1          16  00000010 __cxa_atexit
```
I can't find any error cases which make problems when changed uftrace reads the old `uftrace.data`. But I also think it needs to keep compatibility too.

namhyung:
Please drop the symbol size changes.  I think we should handle that in a separate change.

Anyway keeping backward compatibility is important as we'd like to support old data.

namhyung:
For the replay, I think it should have a different approach than the time filter because time filter is ok just to check the last function.  But size filter needs to handle each function separately.  I think it should check it at both ENTRY and EXIT time.

It seems the [following change](https://github.com/namhyung/uftrace/files/10334095/replay-size-filter.patch.txt) works for me:
```diff
diff --git a/utils/fstack.c b/utils/fstack.c
index abc21dd5fc..712eea6454 100644
--- a/utils/fstack.c
+++ b/utils/fstack.c
@@ -1581,30 +1581,41 @@ static struct uftrace_record *get_task_ustack(struct uftrace_data *handle, int i
 		if (sess && (curr->type == UFTRACE_ENTRY || curr->type == UFTRACE_EXIT))
 			uftrace_match_filter(curr->addr, &sess->filters, &tr);
 
-		if (task->filter.time)
-			time_filter = task->filter.time->threshold;
+		if (task->filter.stack) {
+			time_filter = task->filter.stack->threshold;
+			size_filter = task->filter.stack->size;
+		}
 
-		if (task->filter.size)
-			size_filter = task->filter.size;
+		if (tr.flags & TRIGGER_FL_TIME_FILTER)
+			time_filter = tr.time;
+		if (tr.flags & TRIGGER_FL_SIZE_FILTER)
+			size_filter = tr.size;
 
  		if (curr->type == UFTRACE_ENTRY) {
-			/* it needs to wait until matching exit found */
-			add_to_rstack_list(rstack_list, curr, &task->args);
+			if (size_filter) {
+				struct uftrace_symbol *sym;
+
+				sym = find_symtabs(&sess->sym_info, curr->addr);
+				if (sym && sym->size >= size_filter)
+					add_to_rstack_list(rstack_list, curr, &task->args);
+			}
+			else {
+				/* it needs to wait until matching exit found */
+				add_to_rstack_list(rstack_list, curr, &task->args);
+			}
 
-			if (tr.flags & TRIGGER_FL_TIME_FILTER) {
-				struct uftrace_time_filter_stack *tfs;
+			if (tr.flags & (TRIGGER_FL_TIME_FILTER | TRIGGER_FL_SIZE_FILTER)) {
+				struct uftrace_task_filter_stack *tfs;
 
 				tfs = xmalloc(sizeof(*tfs));
-				tfs->next = task->filter.time;
+				tfs->next = task->filter.stack;
 				tfs->depth = curr->depth;
 				tfs->context = FSTACK_CTX_USER;
-				tfs->threshold = tr.time;
+				tfs->threshold = time_filter;
+				tfs->size = size_filter;
 
-				task->filter.time = tfs;
+				task->filter.stack = tfs;
 			}
-
-			if (tr.flags & TRIGGER_FL_SIZE_FILTER)
-				task->filter.size = tr.size;
 		}
 		else if (curr->type == UFTRACE_EXIT) {
 			struct uftrace_rstack_list_node *last;
@@ -1612,17 +1623,25 @@ static struct uftrace_record *get_task_ustack(struct uftrace_data *handle, int i
 			int last_type;
 			bool filtered = false;
 
-			if (task->filter.time) {
-				struct uftrace_time_filter_stack *tfs;
+			if (task->filter.stack) {
+				struct uftrace_task_filter_stack *tfs;
 
-				tfs = task->filter.time;
+				tfs = task->filter.stack;
 				if (tfs->depth == curr->depth && tfs->context == FSTACK_CTX_USER) {
 					/* discard stale filter */
-					task->filter.time = tfs->next;
+					task->filter.stack = tfs->next;
 					free(tfs);
 				}
 			}
 
+			if (size_filter) {
+				struct uftrace_symbol *sym;
+				sym = find_symtabs(&sess->sym_info, curr->addr);
+
+				if (sym && sym->size < size_filter)
+					continue;
+			}
+
 			list_for_each_entry_reverse(last, &rstack_list->read, list) {
 				if (last->rstack.type == UFTRACE_ENTRY)
 					break;
@@ -1644,14 +1663,6 @@ static struct uftrace_record *get_task_ustack(struct uftrace_data *handle, int i
 			if (handle->caller_filter)
 				filtered |= !(tr.flags & TRIGGER_FL_CALLER);
 
-			if (size_filter) {
-				struct uftrace_symbol *sym;
-				sym = find_symtabs(&sess->sym_info, curr->addr);
-
-				if (sym != NULL)
-					filtered |= (sym->size < size_filter);
-			}
-
 			if (filtered) {
 				/*
 				 * it might set TRACE trigger, which shows
diff --git a/utils/fstack.h b/utils/fstack.h
index 97b3e68335..7904a15f18 100644
--- a/utils/fstack.h
+++ b/utils/fstack.h
@@ -24,10 +24,11 @@ enum uftrace_fstack_context {
 	FSTACK_CTX_KERNEL = 2,
 };
 
-struct uftrace_time_filter_stack {
-	struct uftrace_time_filter_stack *next;
+struct uftrace_task_filter_stack {
+	struct uftrace_task_filter_stack *next;
 	uint64_t threshold;
 	int depth;
+	unsigned size;
 	enum uftrace_fstack_context context;
 };
 
@@ -70,8 +71,7 @@ struct uftrace_task_reader {
 		int in_count;
 		int out_count;
 		int depth;
-		unsigned size;
-		struct uftrace_time_filter_stack *time;
+		struct uftrace_task_filter_stack *stack;
 	} filter;
 	struct uftrace_fstack {
 		uint64_t addr;
```

honggyukim:
@namhyung Thanks. Your change works fine for me, but it can't be compiled because of `utils/kernel.c` not being changed. It needs more changes as follows.
```diff
diff --git a/utils/kernel.c b/utils/kernel.c
index bfebbf5e..bcc9b51d 100644
--- a/utils/kernel.c
+++ b/utils/kernel.c
@@ -1432,8 +1432,8 @@ static int read_kernel_cpu(struct uftrace_data *handle, int cpu)
 		if (prev_tid == -1)
 			prev_tid = tid;
 
-		if (task->filter.time)
-			time_filter = task->filter.time->threshold;
+		if (task->filter.stack)
+			time_filter = task->filter.stack->threshold;
 
 		/* filter match needs full (64-bit) address */
 		real_addr = get_kernel_address(&sess->sym_info, curr->addr);
@@ -1450,15 +1450,15 @@ static int read_kernel_cpu(struct uftrace_data *handle, int cpu)
 			add_kfunc_addr(&kfunc_tree, real_addr);
 
 			if (tr.flags & TRIGGER_FL_TIME_FILTER) {
-				struct uftrace_time_filter_stack *tfs;
+				struct uftrace_task_filter_stack *tfs;
 
 				tfs = xmalloc(sizeof(*tfs));
-				tfs->next = task->filter.time;
+				tfs->next = task->filter.stack;
 				tfs->depth = curr->depth;
 				tfs->context = FSTACK_CTX_KERNEL;
 				tfs->threshold = tr.time;
 
-				task->filter.time = tfs;
+				task->filter.stack = tfs;
 			}
 
 			/* XXX: handle scheduled task properly */
@@ -1474,14 +1474,14 @@ static int read_kernel_cpu(struct uftrace_data *handle, int cpu)
 			if (!find_kfunc_addr(&kfunc_tree, real_addr))
 				continue;
 
-			if (task->filter.time) {
-				struct uftrace_time_filter_stack *tfs;
+			if (task->filter.stack) {
+				struct uftrace_task_filter_stack *tfs;
 
-				tfs = task->filter.time;
+				tfs = task->filter.stack;
 				if (tfs->depth == curr->depth &&
 				    tfs->context == FSTACK_CTX_KERNEL) {
 					/* discard stale filter */
-					task->filter.time = tfs->next;
+					task->filter.stack = tfs->next;
 					free(tfs);
 				}
 			}
```
But the above change didn't properly use size filter so it needs more changes to make it work for kernel functions as well.

namhyung:
Oops, sorry.  I forgot to upload the full version.  Please try [this patch](https://github.com/namhyung/uftrace/files/10341338/replay-size-filter.patch.txt).

namhyung:
Please apply my patch to enable the size filter at replay.

fistki:
@namhyung Thanks for the patch!

honggyukim:
@fistki Thanks for the update. It works as expected for both record and replay commands. We also need to update the man pages for related commands so please update or add the description for size filter in `doc/uftrace-*.md` and `doc/ko/uftrace-*.md` files.

namhyung (approved):
LGTM.  Please update the documentation.

fistki:
I just updated it. Please check

doc/ko/uftrace-record.md
```diff
@@ -148,6 +148,11 @@ RECORD 옵션
     으로 '추적' trigger가 적용된 경우, 그 함수는 실행 시간과 상관없이 항상
     출력된다. 필터에 대한 설명은 *FILTERS* 를 참고한다.
 
+-Z *SIZE*, \--size-filter=*SIZE*
+:   SIZE 바이트보다 큰 함수들을 표시하지 않게 한다. 만약 어떤 함수가 명시적으로
+    '추적' trigger가 적용된 경우, 그 함수는 함수 크기와 상관없이 항상 출력된다.
```
namhyung, line 153: Please do not translate `'trace'` as it's the name of the trigger action and needs to be specified in the command line as is.

doc/ko/uftrace-graph.md
```diff
@@ -70,6 +70,10 @@ GRAPH 옵션
     명시적으로 'trace' 트리거가 적용된 경우, 그 함수는 실행 시간과 상관없이 항상
     출력된다.
 
+-Z *SIZE*, \--size-filter=*SIZE*
+:   SIZE 바이트보다 큰 함수들을 표시하지 않게 한다. 만약 어떤 함수가 명시적으로
```
namhyung, line 74: s/큰/작은/

doc/ko/uftrace-replay.md
```diff
@@ -83,6 +83,11 @@ REPLAY 옵션
     명시적으로 'trace' 트리거가 적용된 경우, 그 함수는 실행 시간과 상관없이 항상
     출력된다. 필터에 대한 설명은 *FILTERS* 를 참고한다.
 
+-Z *SIZE*, \--size-filter=*SIZE*
+:   SIZE 바이트보다 큰 함수들을 표시하지 않게 한다. 만약 어떤 함수가 명시적으로
+    '추적' trigger가 적용된 경우, 그 함수는 함수 크기와 상관없이 항상 출력된다.
```
namhyung, line 88: Ditto for all others.

doc/uftrace-graph.md
```diff
@@ -68,6 +68,11 @@ COMMON OPTIONS
     explicitly have the 'trace' trigger applied, those are always traced
     regardless of execution time.
 
+-Z *SIZE*, \--size-filter=*SIZE*
+:   Do not show functions bigger than size bytes.  If some functions explicitly
```
namhyung, line 72: s/bigger/smaller/

doc/uftrace-record.md
```diff
@@ -40,7 +40,7 @@ RECORD OPTIONS
     See *DYNAMIC TRACING*.
 
 -Z *SIZE*, \--size-filter=*SIZE*
-:   Patch functions bigger than SIZE bytes dynamically.  See *DYNAMIC TRACING*.
+:   patch functions bigger than size bytes dynamically.  See *dynamic tracing*.
```
namhyung, line 43: Why did you change this?  It uses the upper case letters intentionally.

Issue #433, opened 2018-08-06
Support build on Android (AOSP)


gigony:
Thank you for developing a great tool!
It would be very nice if makefile of uftrace is applicable for Android platform.

Current uftrace source code heavily depends on argp.
Unfortunately, Android's [bionic](https://en.wikipedia.org/wiki/Bionic_(software)) doesn't implement argp so uftrace is not buildable for arm-based Android, showing argp-related error messages (attached build error messages: [build_failure.txt](https://github.com/namhyung/uftrace/files/2261830/build_failure.txt)).

https://android.googlesource.com/platform/external/elfutils/+/master/bionic-fixup/argp.h

```
export TOP=/build/gbae/dev-other-next
# cp $TOP/external/elfutils/bionic-fixup/argp.h $TOP/prebuilts/ndk/r11/platforms/android-21/arch-arm/usr/include/    <== used for fixing include-related error but argp implementation is missing..
export CFLAGS="-march=armv7-a --sysroot=$TOP/prebuilts/ndk/r11/platforms/android-21/arch-arm"
./configure --prefix=`pwd`/build --with-elfutils=$TOP/external/elfutils
make
```

I wonder if you have any idea on making uftrace available on Android.

namhyung:
@gigony thanks for kind words.  Currently uftrace is tested on glibc-based systems only so I'm not sure it works well on Android too.  Unfortunately I'm not familiar with the Android code base..

For argp errors, it looks like there's a standalone argp library.  Is it possible for you to try to build with it?

gigony:
@namhyung Thanks for the reply!

It seems that there were similar issues with v4l-utils.
   https://patchwork.kernel.org/patch/7117581/

I think I can try to build library with https://github.com/jahrome/argp-standalone on Android.
I am not sure if it would work without significant modifications because Android is using Clang as a default compiler nowadays.
Would let you know when I am able to build it.

ParkHanbum:
google offer the strace that has compiled statically by trace tool. 
because Android OS environment could be customized by OS vendor. 
if there is the way to build by statically, could helpful for these cases.

honggyukim:
In termux project, they maintain some local patches for Android porting. Maybe we can apply some of their local patches into uftrace.

One of the major obstacles for porting uftrace to Android is to provide `shm_open`/`shm_unlink`.

[...]

PR #1603, opened 2022-12-14
Draft port to Android platform

yugr:
This is an initial take on porting uftrace to Android platform. It works for simple command-line tools in adb shell but please consider it as a base for discussion rather than a complete PR.

Currently patch contains too many ifdefs which is caused by the fact that Android and Linux use different shmem APIs (Ashmem+`sendmsg`/`recvmsg` VS SysV shmem respectively). To get rid of this we could
- unify Android and Linux by switching to `memfd_create`+`sendmsg`/`recvmsg` on Linux instead
- re-implement SysV shmem APIs (`shm_open`, etc.) on top of Ashmem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,12 @@ COMMON_CFLAGS := -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
 COMMON_CFLAGS += -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
 COMMON_CFLAGS += -Wdeclaration-after-statement
 #CFLAGS-DEBUG = -g -D_GNU_SOURCE $(CFLAGS_$@)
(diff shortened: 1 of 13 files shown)


namhyung:
Thanks for working on this!

I remember we had some discussion in #433.  Please take a look at the [termux](https://github.com/termux/termux-packages/tree/master/packages/uftrace) project as they provide a package for uftrace.  I think it's not a pure Android environment but I still think you can get the idea and hopefully collaborate with them.

When it comes to the shmem APIs, I think we need an abstraction layer that hides the implementation detail.  And then we can add a new implementation easily.  I think the termux folks re-implemented shmem and it'd be easier to do so.  But feel free to follow the memfd path if you think it's better.

Finally, please send a PR for the first commit separately as it's unrelated.  Thanks!

yugr:
Superseded by #1605

PR #1605, opened 2022-12-16
uftrace: Port to Android x86_64/AArch64.

yugr:
This is an updated port of uftrace to Android platform. I have got rid of dedicated Ashmem-based code and instead followed the [termux approach](https://github.com/termux/termux-packages/blob/master/packages/uftrace/cmds-record.c.patch) which mimics the `shm_open`/`shm_unlink` APIs using temporary files.

With these changes uftrace works for simple command-line tools in x86_64/AArch64 adb shell (compiled with `-fpatchable-function-entry=5`/`=2` respectively), both on emulator and on real device.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -168,6 +168,22 @@ This assumes you already installed the cross-built `libelf` on the sysroot
 directory.  Otherwise, you can also build it from source (please see below) or
 use it on a different path using `--with-elfutils=<PATH>`.
 
+To compile for Android 9+, export the CC environment variable and disable the not
+yet implemented python and libstdc++ support. E.g. to configure for Android
+AArch64 do
+
+    $ export CC=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android33-clang
+    $ export LD=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/ld.lld
+    $ ./configure --arch=aarch64 --cross-compile=aarch64-linux-gnu- --without-libpython --without-libstdc++
+
+It's recommended to compile instrumented program with
+`-fpatchable-function-entry` on Android. You could also use
+`-finstrument-functions` or `-pg` but in that case you also need to link program
+with `-Wl,-z,undefs` because Android runtime does not include
+`__cyg_profile_func_enter` or `mcount`.
+
+Note that Android has been tested on AArch64 and x86\_64 so far.
+
 
 BUILD WITH ELFUTILS (libelf)
 ============================
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -252,7 +252,7 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 			/* can't reuse first template because it was trashed by mkstemp */
 			strcpy(template, LIVE_NAME);
 
-			if (errno != EPERM)
+			if (errno != EPERM && errno != ENOENT)
 				pr_err("cannot access to /tmp");
 
 			fd = mkstemp(template);
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -300,6 +300,7 @@ static int find_dynamic_module(struct dl_phdr_info *info, size_t sz, void *data)
 	struct find_module_data *fmd = data;
 	struct uftrace_sym_info *sym_info = fmd->sinfo;
 	struct uftrace_mmap *map;
+	bool is_executable = mcount_is_main_executable(info->dlpi_name, sym_info->filename);
 
 	mdi = create_mdi(info);
 
@@ -315,7 +316,7 @@ static int find_dynamic_module(struct dl_phdr_info *info, size_t sz, void *data)
 		free(mdi);
 	}
 
-	return !fmd->needs_modules;
+	return !fmd->needs_modules && is_executable;
 }
 
 static void prepare_dynamic_update(struct uftrace_sym_info *sinfo, bool needs_modules)
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -2147,6 +2147,21 @@ void __visible_default __cyg_profile_func_exit(void *child, void *parent)
 }
 UFTRACE_ALIAS(__cyg_profile_func_exit);
 
+bool mcount_is_main_executable(const char *filename, const char *exename)
+{
+	/* on Linux main executable has empty name
+	   whereas on Android we need to compare with exename */
+	char filename_canonized[PATH_MAX];
+	char exename_canonized[PATH_MAX];
+
+	if (!*filename)
+		return true;
+	if (realpath(filename, filename_canonized) && realpath(exename, exename_canonized)) {
+		return strcmp(filename_canonized, exename_canonized) == 0;
+	}
+	return false;
+}
+
 #ifndef UNIT_TEST
 /*
  * Initializer and Finalizer
--- a/utils/shmem.c
+++ b/utils/shmem.c
@@ -1,0 +1,96 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "utils/shmem.h"
+
+#ifdef __ANDROID__
+
+const char *uftrace_shmem_root(void)
+{
+	static char uftrace_dir[PATH_MAX] = "";
+
+	if (uftrace_dir[0] == 0) {
+		const char *tmpdir;
+		tmpdir = getenv("TMPDIR");
+		if (!tmpdir)
+			tmpdir = "/tmp";
+
+		snprintf(uftrace_dir, sizeof(uftrace_dir), "%s/uftrace", tmpdir);
+	}
+
+	return uftrace_dir;
+}
+
+int uftrace_shmem_open(const char *name, int oflag, mode_t mode)
+{
+	const char *uftrace_dir;
+	char *fname;
+	int fd;
+	int status;
+
+	uftrace_dir = uftrace_shmem_root();
+
+	status = mkdir(uftrace_dir, mode);
+	if (status < 0 && errno != EEXIST)
+		return -1;
+
+	if (asprintf(&fname, "%s/%s", uftrace_dir, name) < 0)
+		return -1;
+
+	fd = open(fname, oflag, mode);
+	if (fd >= 0) {
+		int flags = fcntl(fd, F_GETFD, 0);
+		flags |= FD_CLOEXEC;
+		if (fcntl(fd, F_SETFD, flags) < 0) {
+			int saved_errno = errno;
+			close(fd);
+			fd = -1;
+			errno = saved_errno;
+		}
+	}
+
+	free(fname);
+
+	return fd;
+}
+
+int uftrace_shmem_unlink(const char *name)
+{
+	const char *uftrace_dir;
+	char *fname;
+	int status;
+
+	uftrace_dir = uftrace_shmem_root();
+
+	if (asprintf(&fname, "%s/%s", uftrace_dir, name))
+		return -1;
+	status = unlink(fname);
+	free(fname);
+
+	return status;
+}
+
+#else /* ! __ANDROID__ */
+
+#include <sys/mman.h>
+
+const char *uftrace_shmem_root(void)
+{
+	return "/dev/shm";
+}
+
+int uftrace_shmem_open(const char *name, int oflag, mode_t mode)
+{
+	return shm_open(name, oflag, mode);
+}
+
+int uftrace_shmem_unlink(const char *name)
+{
+	return shm_unlink(name);
+}
+
+#endif
--- a/utils/symbol.c
+++ b/utils/symbol.c
@@ -825,6 +825,10 @@ enum uftrace_trace_type check_trace_functions(const char *filename)
 
 		/* undefined function is ok here */
 		if (elf_symbol_type(&iter.sym) != STT_FUNC &&
+#ifdef __ANDROID__
+		    // Profiling functions are undefined on Android
+		    elf_symbol_type(&iter.sym) != STT_NOTYPE &&
+#endif
 		    elf_symbol_type(&iter.sym) != STT_GNU_IFUNC)
 			continue;
 
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,7 @@ LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/script.c $(srcdir)/utils/script-python.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/auto-args.c $(srcdir)/utils/dwarf.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/hashmap.c $(srcdir)/utils/argspec.c
 LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/tracefs.c $(srcdir)/utils/socket.c
+LIBMCOUNT_UTILS_SRCS += $(srcdir)/utils/shmem.c
 LIBMCOUNT_UTILS_SRCS += $(wildcard $(srcdir)/utils/symbol*.c)
 LIBMCOUNT_UTILS_OBJS := $(patsubst $(srcdir)/utils/%.c,$(objdir)/libmcount/%.op,$(LIBMCOUNT_UTILS_SRCS))
 
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,12 @@ COMMON_CFLAGS := -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
 COMMON_CFLAGS += -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
 COMMON_CFLAGS += -Wdeclaration-after-statement
 #CFLAGS-DEBUG = -g -D_GNU_SOURCE $(CFLAGS_$@)
-COMMON_LDFLAGS := -lrt -ldl -pthread -Wl,-z,noexecstack $(LDFLAGS)
+COMMON_LDFLAGS := -ldl -pthread -Wl,-z,noexecstack $(LDFLAGS)
+ifeq ($(ANDROID),)
+COMMON_LDFLAGS += -lrt
+else
+COMMON_LDFLAGS += -landroid
+endif
 ifneq ($(elfdir),)
   COMMON_CFLAGS  += -I$(elfdir)/include
   COMMON_LDFLAGS += -L$(elfdir)/lib
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ now.
 
 Limitations
 ===========
-- It can trace a native C/C++ application on Linux.
+- It can trace a native C/C++ application on Linux and Android.
 - It *cannot* trace already running process.
 - It *cannot* be used for system-wide tracing.
 - It supports x86 (32 and 64 bit), ARM (v6 or later) and AArch64 for now.
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -25,6 +25,7 @@
 #include "utils/kernel.h"
 #include "utils/list.h"
 #include "utils/perf.h"
+#include "utils/shmem.h"
 #include "utils/symbol.h"
 #include "utils/utils.h"
 
@@ -826,7 +827,7 @@ static void record_mmap_file(const char *dirname, char *sess_id, int bufsize)
 	struct mcount_shmem_buffer *shmem_buf;
 
 	/* write (append) it to disk */
-	fd = shm_open(sess_id, O_RDWR, 0600);
+	fd = uftrace_shmem_open(sess_id, O_RDWR, 0600);
 	if (fd < 0) {
 		pr_dbg("open shmem buffer failed: %s: %m\n", sess_id);
 		return;
@@ -935,12 +936,12 @@ static void unlink_shmem_list(void)
 		sscanf(sl->id, "/uftrace-%[^-]-%*d-%*d", shmem_session);
 		pr_dbg2("unlink for session: %s\n", shmem_session);
 
-		num = scandir("/dev/shm/", &shmem_bufs, filter_shmem, alphasort);
+		num = scandir(uftrace_shmem_root(), &shmem_bufs, filter_shmem, alphasort);
 		for (i = 0; i < num; i++) {
 			sid[0] = '/';
 			memcpy(&sid[1], shmem_bufs[i]->d_name, MSG_ID_SIZE);
 			pr_dbg3("unlink %s\n", sid);
-			shm_unlink(sid);
+			uftrace_shmem_unlink(sid);
 			free(shmem_bufs[i]);
 		}
 
--- a/configure
+++ b/configure
@@ -151,6 +151,10 @@ fi
 CC=${CC:-${CROSS_COMPILE}gcc}
 LD=${LD:-${CROSS_COMPILE}ld}
 
+if $CC --version | grep -q Android; then
+    ANDROID=1
+fi
+
 # objdir can be changed, reset output
 objdir=$(readlink -f ${objdir})
 output=${output:-${objdir}/.config}
@@ -282,6 +286,7 @@ override CC     := $CC
 override LD     := $LD
 override CFLAGS  = $CFLAGS
 override LDFLAGS = $LDFLAGS
+override ANDROID = $ANDROID
 
 override srcdir := $srcdir
 override objdir := $objdir
--- a/libmcount/internal.h
+++ b/libmcount/internal.h
@@ -443,4 +443,6 @@ int read_pmu_event(struct mcount_thread_data *mtdp, enum uftrace_event_id id, vo
 void release_pmu_event(struct mcount_thread_data *mtdp, enum uftrace_event_id id);
 void finish_pmu_event(struct mcount_thread_data *mtdp);
 
+bool mcount_is_main_executable(const char *filename, const char *exename);
+
 #endif /* UFTRACE_MCOUNT_INTERNAL_H */
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -481,6 +481,11 @@ static int setup_mod_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
 		"ld-linux-*.so.*",
 		"libdl.so.2",
 		"libdl-2.*.so",
+#ifdef __ANDROID__
+		"linker64",
+		"libc.so",
+		"libm.so",
+#endif
 	};
 	size_t k;
 	static bool exe_once = true;
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -511,6 +511,9 @@ static int setup_exe_plthook_data(struct dl_phdr_info *info, size_t sz, void *ar
 	const char *exename = arg;
 	unsigned long offset = info->dlpi_addr;
 
+	if (!mcount_is_main_executable(info->dlpi_name, exename))
+		return 0;
+
 	pr_dbg2("setup plthook data for %s (offset: %lx)\n", exename, offset);
 
 	hook_pltgot(exename, offset);
--- a/libmcount/record.c
+++ b/libmcount/record.c
@@ -17,6 +17,7 @@
 #include "mcount-arch.h"
 #include "utils/event.h"
 #include "utils/filter.h"
+#include "utils/shmem.h"
 #include "utils/symbol.h"
 #include "utils/utils.h"
 
@@ -33,7 +34,7 @@ static struct mcount_shmem_buffer *allocate_shmem_buffer(char *sess_id, size_t s
 
 	snprintf(sess_id, size, SHMEM_SESSION_FMT, mcount_session_name(), tid, idx);
 
-	fd = shm_open(sess_id, O_RDWR | O_CREAT | O_TRUNC, 0600);
+	fd = uftrace_shmem_open(sess_id, O_RDWR | O_CREAT | O_TRUNC, 0600);
 	if (fd < 0) {
 		saved_errno = errno;
 		pr_dbg("failed to open shmem buffer: %s\n", sess_id);
--- a/utils/shmem.h
+++ b/utils/shmem.h
@@ -1,0 +1,8 @@
+#ifndef UFTRACE_SHMEM_H
+#define UFTRACE_SHMEM_H
+
+int uftrace_shmem_open(const char *name, int oflag, mode_t mode);
+int uftrace_shmem_unlink(const char *name);
+const char *uftrace_shmem_root(void);
+
+#endif /* UFTRACE_SHMEM_H */


arch/aarch64/mcount-dynamic.c
```diff
@@ -84,10 +87,21 @@ static void read_patchable_loc(struct mcount_dynamic_info *mdi, struct uftrace_e
...
+	unsigned long *patchable_loc;
 
 	mdi->nr_patch_target = shdr->sh_size / sizeof(long);
 	mdi->patch_target = xmalloc(shdr->sh_size);
-
+	patchable_loc = mdi->patch_target;
+
+#ifdef __ANDROID__
+	/* symbol addresses on Android are stored in relocation records
+	   so read them from relocated section */
```
yugr, line 99: On Android the relative offset is stored not in `__patchable_function_entries` but in `.rela.dyn` section.

libmcount/dynamic.c
```diff
@@ -315,7 +316,7 @@ static int find_dynamic_module(struct dl_phdr_info *info, size_t sz, void *data)
 		free(mdi);
 	}
 
-	return !fmd->needs_modules;
+	return !fmd->needs_modules && is_executable;
```
yugr, line 319: On Android linker64 precedes the executable in list of modules so we can not bail out after the first module.

utils/symbol.c
```diff
@@ -824,6 +824,10 @@ enum uftrace_trace_type check_trace_functions(const char *filename)
 
 		/* undefined function is ok here */
 		if (elf_symbol_type(&iter.sym) != STT_FUNC &&
+#ifdef __ANDROID__
+		    // Profiling functions are undefined on Android
+		    elf_symbol_type(&iter.sym) != STT_NOTYPE &&
```
yugr, line 829: Executables linked with `-finstrument-functions`/`-pg` need to be linked with `-Wl,-z,undefs` because corresponding callbacks are missing from Android runtime library and only come from `libmcount`.

libmcount/mcount.c
```diff
@@ -1874,7 +1887,7 @@ static void agent_kill()
 	return;
 
 error:
-	pthread_cancel(agent);
+	pthread_kill(agent, SIGUSR1);
```
yugr, line 1890: No `pthread_cancel` in Bionic.

namhyung (commented):
Please separate each change to different commits for easier review.  We can have a shmem change, patch location fix, pthread_cancel, PLT hooking commits separately for example.  Also please update the INSTALL.md document for Android build.

yugr:
> Please separate each change to different commits for easier review.

Ok, makes sense. Done.

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
...
+{
+#ifdef __ANDROID__
+	const char *tmpdir;
+	static char uftrace_dir[PATH_MAX];
+
+	tmpdir = getenv("TMPDIR");
+	if (!tmpdir)
+		tmpdir = "/tmp";
+
+	sprintf(uftrace_dir, "%s/uftrace", tmpdir);
```
namhyung, line 19: `sprintf()` considered harmful, please use `snprintf()` instead.

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+
+char *get_shm_root()
```
namhyung, line 9: Should be `const char * get_shm_root(void)`.

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
```
namhyung, line 4: No need for a blank line here.

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
...
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+
+char *get_shm_root()
+{
+#ifdef __ANDROID__
+	const char *tmpdir;
+	static char uftrace_dir[PATH_MAX];
```
namhyung, line 13: Please do not use a static buffer this way as it's not thread safe.  As it'd be written only and read only after it, we can do something like this

```c
{
    static char shmem_root[PATH_MAX] = "";

    if (shmem_root[0] == '\0') {
        // ...
    }
    return shmem_root;
}

File: utils/shmem.c [hunk 9]
```diff
@@ -0,0 +1,74 @@
...
+
+	uftrace_dir = get_shm_root();
+
+	status = mkdir(uftrace_dir, mode);
+	if (status < 0 && errno != EEXIST)
+		return -1;
+
+	namelen = strlen(name);
+	fname = alloca(strlen(uftrace_dir) + 1 + namelen + 1);
+	sprintf(fname, "%s/%s", uftrace_dir, name);
```
namhyung, line 44: Ditto.

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
...
+	uftrace_dir = get_shm_root();
+
+	namelen = strlen(name);
+	fname = alloca(strlen(uftrace_dir) + 1 + namelen + 1);
+	sprintf(fname, "%s/%s", uftrace_dir, name);
+
+	return unlink(fname);
+}
+
+#endif
```
namhyung, line 74: I prefer having the macro conditionals out of functions.  Let's do this.

```c
#ifdef __ANDROID__
const char * uftrace_shmem_root(void)
{
    ...
}

int uftrace_shmem_open(const char *name, int oflag, mode_t mode)
{
    ...
}

int uftrace_shmem_unlink(const char *name)
{
    ...
}

#else /* __ANDROID__ */

const char * uftrace_shmem_root(void)
{
    return "/dev/shm";
}

int uftrace_shmem_open(const char *name, int oflag, mode_t mode)
{
    return shm_open(name, oflag, mode);
}

int uftrace_shmem_unlink(const char *name)
{
    return shm_unlink(name);
}
#endif
```

utils/shmem.c
```diff
@@ -0,0 +1,74 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+
```
namhyung, line 8: And you'll need `#include "utils/shmem.h"` here.

libmcount/mcount.c
```diff
@@ -2139,6 +2139,20 @@ void __visible_default __cyg_profile_func_exit(void *child, void *parent)
 }
 UFTRACE_ALIAS(__cyg_profile_func_exit);
 
+int mcount_is_main_executable(const char *filename, const char *exename)
```
namhyung, line 2142: Why not return `bool` type?

libmcount/mcount.c
```diff
@@ -2139,6 +2139,20 @@ void __visible_default __cyg_profile_func_exit(void *child, void *parent)
...
+{
+	/* on Linux main executable has empty name
+	   whereas on Android we need to compare with exename */
+	char filename_canonized[PATH_MAX];
+	char exename_canonized[PATH_MAX];
```
namhyung, line 2147: Please put a blank line after the declaration.

namhyung (commented):
Thanks for the quick update!

arch/aarch64/mcount-dynamic.c
yugr: Firstly let me first clarify what's going on here. At runtime entries in `__patchable_function_entries` section contain absolute addresses of patched functions. Statically, in ELF file these entries may contain relative offsets (on Linux) or just zeroes (on Android). At startup, runtime linker will patch them by adding the load address of module.

Existing code reads static contents of `__patchable_function_entries` from ELF file but it's useless on Android as it's all zeroes. Instead I reconstruct relative function offsets by subtracting the module start address from already relocated contents of `__patchable_function_entries`.

I suspect that new code should also work on Linux but haven't tested it (`make test` does not seem to run tests for patched functions, only `-finstrument-functions` and `-pg`, perhaps that should be fixed?).

utils/shmem.c
yugr: Well, it won't be thread-safe with suggested change either. We could switch to dynamic memory but then it would be ugly and slow...

arch/aarch64/mcount-dynamic.c
yugr: > I suspect that new code should also work on Linux but haven't tested it

Ok, I tested it and it worked at least with s-abc.c. I'd be cautious to introduce this change on Linux without larger tests though.

BTW note that `patchable_gcc_nop` (and probly other patterns too) need to be updated because modern compilers prefix code with endbr64:
```
0000000000001169 <a>:
    1169:	f3 0f 1e fa          	endbr64 
    116d:	90                   	nop
    116e:	90                   	nop
    116f:	90                   	nop
    1170:	90                   	nop
    1171:	90                   	nop
    ...
```

cmds/live.c
```diff
@@ -218,14 +219,17 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 		umask(022);
 		fd = mkstemp(template);
 		if (fd < 0) {
+			/* can't reuse first template because it was trashed by mkstemp */
+			strcpy(template, LIVE_NAME);
+
 			if (errno != EPERM)
-				pr_err("cannot access to /tmp");
+				pr_dbg("cannot access to /tmp\n");
```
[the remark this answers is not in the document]
yugr: Hm, you are right, seems to be redundant.

libmcount/mcount.c
yugr: This means `sfd` and `addr` variables should be made global?

arch/aarch64/mcount-dynamic.c
namhyung: > Existing code reads static contents of __patchable_function_entries from ELF file but it's useless on Android as it's all zeroes.

Ok, good to know.  Yeah, if it contains absolute addresses at runtime we can read from the memory.  I was confused that you wanted to read from the `.rela.dyn` section.

namhyung: > `make test` does not seem to run tests for patched functions, only `-finstrument-functions` and `-pg`, perhaps that should be fixed?

That'd be nice :).

namhyung: > BTW note that patchable_gcc_nop (and probly other patterns too) need to be updated because modern compilers prefix code with endbr64.

Thanks for pointing that out.  We should update CET support.

libmcount/mcount.c
yugr: Please check if updated PR is ok.

yugr:
@namhyung unfortunately I'll not be able to work on this PR after Dec 21 due to job change. I hope it will soon be uptaken by another member of our team though.

namhyung:
@yugr sorry to hear that.  You did a great job in such a short period of time.  Hope you find an interesting job.  Keep up the good work!

yugr:
> @yugr sorry to hear that. You did a great job in such a short period of time. Hope you find an interesting job. Keep up the good work!

Note that if you post more comments I'll have time to fix them tomorrow (as a courtesy for whoever picks up this PR later).

utils/shmem.c
```diff
@@ -0,0 +1,92 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "shmem.h"
```
namhyung, line 8: Nit: I keep the include path from the root directory even if it's in the same directory.  I think it's better for maintenance.  So please change it to `"utils/shmem.h"`.

utils/shmem.c
```diff
@@ -0,0 +1,92 @@
...
+
+	uftrace_dir = uftrace_shmem_root();
+
+	status = mkdir(uftrace_dir, mode);
+	if (status < 0 && errno != EEXIST)
+		return -1;
+
+	fnamelen = strlen(uftrace_dir) + 1 + strlen(name) + 1;
+	fname = alloca(fnamelen);
+	snprintf(fname, fnamelen, "%s/%s", uftrace_dir, name);
```
namhyung, line 28: I think we can simply call `xasprintf()` and free the filename after `open()`.

utils/shmem.c
```diff
@@ -0,0 +1,92 @@
...
+	char *fname;
+	size_t fnamelen;
+
+	uftrace_dir = uftrace_shmem_root();
+
+	fnamelen = strlen(uftrace_dir) + 1 + strlen(name) + 1;
+	fname = alloca(fnamelen);
+	snprintf(fname, fnamelen, "%s/%s", uftrace_dir, name);
+
+	return unlink(fname);
```
namhyung, line 56: Ditto.  Please use `xasprintf()` and `free()`.

namhyung (commented):
I think it's good now.  But I don't have a test environment.  Also it'd be great if anybody in the termux team can take a look.  @sylirre and/or @xtkoba ?

namhyung:
@gigony might be interested too. :)

namhyung:
I've cherry-picked some commits in this PR.  Could you please rebase?

yugr:
> I've cherry-picked some commits in this PR. Could you please rebase?

Sure, done.

kangtegong:
Really nice patch :) I think it would be better if you can remove these warnings while make.

```
  CC       utils/shmem.o
/home/minchul/workspace/uftrace/utils/shmem.c: In function ‘uftrace_shmem_open’:
/home/minchul/workspace/uftrace/utils/shmem.c:81:16: warning: implicit declaration of function ‘shm_open’ [-Wimplicit-function-declaration]
   81 |         return shm_open(name, oflag, mode);
      |                ^~~~~~~~
/home/minchul/workspace/uftrace/utils/shmem.c: In function ‘uftrace_shmem_unlink’:
/home/minchul/workspace/uftrace/utils/shmem.c:86:16: warning: implicit declaration of function ‘shm_unlink’; did you mean ‘unlink’? [-Wimplicit-function-declaration]
   86 |         return shm_unlink(name);
      |                ^~~~~~~~~~
      |                unlink

  CC FPIC  libmcount/shmem.op
/home/minchul/workspace/uftrace/utils/shmem.c: In function ‘uftrace_shmem_open’:
/home/minchul/workspace/uftrace/utils/shmem.c:81:16: warning: implicit declaration of function ‘shm_open’ [-Wimplicit-function-declaration]
   81 |         return shm_open(name, oflag, mode);
      |                ^~~~~~~~
/home/minchul/workspace/uftrace/utils/shmem.c: In function ‘uftrace_shmem_unlink’:
/home/minchul/workspace/uftrace/utils/shmem.c:86:16: warning: implicit declaration of function ‘shm_unlink’; did you mean ‘unlink’? [-Wimplicit-function-declaration]
   86 |         return shm_unlink(name);
      |                ^~~~~~~~~~
      |                unlink

  CC FPIC  arch/aarch64/mcount-dynamic.op
/home/minchul/workspace/uftrace/arch/aarch64/mcount-dynamic.c: In function ‘read_patchable_loc’:
/home/minchul/workspace/uftrace/arch/aarch64/mcount-dynamic.c:91:24: warning: variable ‘patchable_loc’ set but not used [-Wunused-but-set-variable]
   91 |         unsigned long *patchable_loc;
      |                        ^~~~~~~~~~~~~
/home/minchul/workspace/uftrace/arch/aarch64/mcount-dynamic.c:90:18: warning: unused variable ‘i’ [-Wunused-variable]
   90 |         unsigned i;
      |                  ^
```

kangtegong:
Just for your information, I've tried this patch in Android 8, which uses API level below 27[1] and I couldn't build/run uftrace:

```
ld: error: undefined symbol: glob
```

```
# ./uftrace
CANNOT LINK EXECUTABLE "./uftrace": cannot locate symbol "glob" referenced by "/data/local/tmp/uftrace"...
Aborted
```

Turns out, `glob.h` is not included in earlier than Android version 8[2], so this patch is relevant to Android version 9 and above. I share this information because I think it would be helpful to know about it.

[1] https://source.android.com/docs/setup/about/build-numbers 
[2]

```
Author:     Elliott Hughes <enh@google.com>
AuthorDate: Tue Sep 26 17:09:07 2017 -0700
Commit:     Elliott Hughes <enh@google.com>
CommitDate: Tue Sep 26 18:46:09 2017 -0700

    Add <glob.h>.
    
    This is the FreeBSD implementation, plus some tests. The FreeBSD
    implementation includes the GNU extensions and seems to be what
    iOS is using too, which should provide bug compatibility for app
    developers.
    
    The code unfortunately uses a lot of stack, and uses FreeBSD locale
    implementation that we don't have, but it does seem better maintained
    than the other BSDs.
    
    Bug: http://b/29251134
    Test: ran tests
    Change-Id: Ie477b45e42a6df1319b25712098519d2b33adf67
```

yugr:
> Really nice patch :) I think it would be better if you can remove these warnings while make.

Oh, crumbs! Thanks for spotting.

yugr:
> Just for your information, I've tried this patch in Android 8, which uses API level below 27[1] and I couldn't build/run uftrace:

Right, I think I tested on 9.

namhyung:
> > Just for your information, I've tried this patch in Android 8, which uses API level below 27[1] and I couldn't build/run uftrace:
> 
> Right, I think I tested on 9.

Ok, we can only support Android 9 or above since Android 8 will go away (it's released 5 years ago).  Please add this in the documentation.

yugr:
> > > Just for your information, I've tried this patch in Android 8, which uses API level below 27[1] and I couldn't build/run uftrace:
> > 
> > 
> > Right, I think I tested on 9.
> 
> Ok, we can only support Android 9 or above since Android 8 will go away (it's released 5 years ago). Please add this in the documentation.

Right, fixed.

arch/aarch64/mcount-dynamic.c
yugr: > Thanks for pointing that out. We should update CET support.

Added #1613

yugr: > > `make test` does not seem to run tests for patched functions, only `-finstrument-functions` and `-pg`, perhaps that should be fixed?
> 
> That'd be nice :).

Added #1614

yugr: > I suspect that new code should also work on Linux but haven't tested it (`make test` does not seem to run tests for patched functions, only `-finstrument-functions` and `-pg`, perhaps that should be fixed?).

Added in #1623

yugr:
> But I don't have a test environment.

FYI I successfully ran this in Android emulator. I can provide detailed instructions if needed.

I've also updated my PR to support x86_64.

cmds/live.c
```diff
@@ -252,7 +252,7 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 			/* can't reuse first template because it was trashed by mkstemp */
 			strcpy(template, LIVE_NAME);
 
-			if (errno != EPERM)
+			if (errno != EPERM && errno != ENOENT)
```
namhyung, line 255: Not related to the change, but please add your Signed-off-by in the commit message.

honggyukim:
> FYI I successfully ran this in Android emulator. I can provide detailed instructions if needed.

@yugr Yeah, please share the instruction details. I will try to test.

cmds/live.c
yugr: I keep forgetting...

yugr:
> @yugr Yeah, please share the instruction details. I will try to test.

I'm not an Android pro but this worked for me:
1) Download Android Studio and run it (it may suggest installing additional components at startup, do it)
2) Run SDK manager (Tools -> SDK manager) and in "SDK Tools" tab make sure that SDK build tools, NDK, command-line tools, Android emulator, platform tools are installed (install if missing)
3) In Device Manager create x86_64 device with the latest image that does NOT include the "Google API" in its name
4) In Device Manager click "Play" and wait for phone screen to appear
5) Root the device:
```
$ adb root
$ adb remount
```
6) Build uftrace:
```
$ export CC=$HOME/install/android-ndk-r25b/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android33-clang
$ ./configure --arch=x86_64 --cross-compile=x86_64-linux-gnu- --without-libpython --without-libstdc++
$ make
$ $CC tests/s-abc.c -o s-abc -fpatchable-function-entry=5
```
7) Copy uftrace files to device
```
$ adb push uftrace libmcount*.so s-abc /data
```
8) Run test:
```
$ adb shell 'cd /data/yugr/uftrace && LD_LIBRARY_PATH=. ./uftrace -P . a.out'
            [  3838] | __libc_init() {
            [  3838] |   main() {
            [  3838] |     a() {
            [  3838] |       b() {
            [  3838] |         c() {
  19.000 us [  3838] |           getpid();
  56.800 us [  3838] |         } /* c */
  96.000 us [  3838] |       } /* b */
 130.400 us [  3838] |     } /* a */
 162.400 us [  3838] |   } /* main */
            [  3838] |   /* linux:task-exit */
```

honggyukim:
> I'm not an Android pro but this worked for me:

@yugr Thanks a lot for sharing the steps. But I see my ubuntu gets freeze when running android emulator so couldn't test it much. Could you please let me know which specific virtual device and android version you used?

INSTALL.md
```diff
@@ -168,6 +168,21 @@ This assumes you already installed the cross-built `libelf` on the sysroot
 directory.  Otherwise, you can also build it from source (please see below) or
 use it on a different path using `--with-elfutils=<PATH>`.
 
+To compile for Android 9+, export the CC environment variable and disable the not
+yet implemented python and libstdc++ support. E.g. to configure for Android
+AArch64 do
+
+    $ export CC=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android33-clang
+    $ ./configure --arch=aarch64 --cross-compile=aarch64-linux-gnu- --without-libpython --without-libstdc++
```
honggyukim, line 176: I see an error as follows.
```
$ make
  CC       uftrace.o
  CC       arch/aarch64/cpuinfo.o
  LINK     arch/aarch64/uftrace.o
/bin/sh: 1: aarch64-linux-gnu-ld: not found
make[1]: *** [Makefile:22: /home/honggyu/work/uftrace/arch/aarch64/uftrace.o] Error 127
make: *** [Makefile:293: /home/honggyu/work/uftrace/arch/aarch64/uftrace.o] Error 2
```
Adding one more line for `LD` makes it successful.
```
$ export CC=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android33-clang
$ export LD=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/ld.lld
$ ./configure --arch=aarch64 --cross-compile=aarch64-linux-gnu- --without-libpython --without-libstdc++
```

honggyukim (commented):
I tested it on one of my android phone, which isn't rooted and it fails with some reasons.

1. inaccessible `/tmp`
```
1|beryllium:/data/local/tmp $ ./uftrace --libmcount-path=. --force pwd
uftrace: /home/honggyu/work/uftrace/cmds/live.c:255:command_live
 ERROR: cannot access to /tmp: No such file or directory
```
2. `mkfifo` fails due to a permission issue
```
1|beryllium:/data/local/tmp $ ./uftrace record --libmcount-path=. --force pwd
uftrace: /home/honggyu/work/uftrace/cmds/record.c:2207:command_record
 ERROR: cannot create a communication channel: Permission denied
```
Both of the reasons can be avoided when having root permission, but especailly `mkfifo` cannot be executed without having root as described in the link below.
https://stackoverflow.com/questions/2740321/how-to-create-named-pipe-mkfifo-in-android/2758038#2758038

yugr:
> > I'm not an Android pro but this worked for me:
> 
> @yugr Thanks a lot for sharing the steps. But I see my ubuntu gets freeze when running android emulator so couldn't test it much. Could you please let me know which specific virtual device and android version you used?

I used Pixel 2 (default) and Android 12.0 (without "Google API"). Do you have enough RAM? I think emulator needs several gigabytes... I used Windows version but in theory it shouldn't matter.

INSTALL.md
yugr: That's interesting, I can't repro. BTW isn't `LD=$CC` enough in your case?

yugr:
> I tested it on one of my android phone, which isn't rooted and it fails with some reasons.

That's correct, I tested on rooted device.

>     1. inaccessible `/tmp`

I think we need to teach uftrace to respect TMPDIR.

>     2. `mkfifo` fails due to a permission issue

Ok, I'll try to take a look over weekend.

honggyukim:
> I used Pixel 2 (default) and Android 12.0 (without "Google API"). Do you have enough RAM? I think emulator needs several gigabytes... I used Windows version but in theory it shouldn't matter.

Yes, my system has 16GB so I think it should be enough. The emulator is terminated with an error message
```
2023-01-31 20:18:50,400 [  77968] SEVERE - Emulator: Pixel 2 API 31 - Emulator terminated with exit code 139
java.lang.Throwable: Emulator terminated with exit code 139
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:182)
	at com.android.tools.idea.avdmanager.EmulatorProcessHandler$ConsoleListener.onTextAvailable(EmulatorProcessHandler.kt:89)
	at jdk.internal.reflect.GeneratedMethodAccessor27.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.intellij.execution.process.ProcessHandler$2.invoke(ProcessHandler.java:248)
	at com.sun.proxy.$Proxy25.onTextAvailable(Unknown Source)
	at com.intellij.execution.process.ProcessHandler.notifyTextAvailable(ProcessHandler.java:222)
	at com.intellij.execution.process.ProcessTerminatedListener.processTerminated(ProcessTerminatedListener.java:54)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
...
```
I use ubuntu 22.04 and it might be related to ubuntu issue as someone said in https://stackoverflow.com/questions/65148279/android-the-emulator-process-for-avd-was-killed.

INSTALL.md
honggyukim: With the current instruction guide, `.config.` file becomes as follows.
```
$ cat .config 
    ...
override ARCH   := aarch64
override CC     := /home/honggyu/toolchains/android-ndk-r25b/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android33-clang
override LD     := aarch64-linux-gnu-ld
    ...
```
But there is no `aarch64-linux-gnu-ld` file in android NDK. It might be the case that your system already has `aarch64-linux-gnu-ld` somewhere else outside of android NDK. I think we also need
```
export LD=$NDK/toolchains/llvm/prebuilt/linux-x86_64/bin/ld.lld
```
along with `export CC`.

honggyukim: This problem is shown in @kangtegong's environment and having `export LD=.../ld.lld` fixed his problem as well.

honggyukim:
> I think we need to teach uftrace to respect TMPDIR.

`TMPDIR` is not defined by default in every environments so I think we should first check whether `getenv("TMPDIR")` isn't `NULL` then use it. Otherwise, use `/tmp` as usual.

honggyukim:
The emulator runs fine randomly and I got a chance to run it without being crashed. The test result looks good.
```
emulator64_x86_64:/data # ./uftrace -P. -a --libmcount-path=. t-abc
            [  3481] | __libc_init() {
            [  3481] |   main() {
            [  3481] |     a() {
            [  3481] |       b() {
            [  3481] |         c() {
   0.228 us [  3481] |           getpid() = 3481;
   8.284 us [  3481] |         } /* c */
   8.492 us [  3481] |       } /* b */
   8.638 us [  3481] |     } /* a */
   8.906 us [  3481] |   } /* main */
  24.500 us [  3481] |   /* linux:schedule (pre-empted) */
   4.235 ms [  3481] |   /* linux:schedule (pre-empted) */
   5.017 ms [  3481] |   /* linux:schedule (pre-empted) */
   6.980 us [  3481] |   /* linux:schedule (pre-empted) */
            [  3481] |   /* linux:task-exit */

task: 3481
[0] __libc_init
```
```
emulator64_x86_64:/data # ./uftrace --libmcount-path=. -la -t 1us pwd
/data
            [  4999] | __libc_init() {
   4.581 us [  4999] |   signal(SIGPIPE, 0) = 0;
            [  4999] |   basename("pwd") {
   8.008 ms [  4999] |     /* linux:schedule (pre-empted) */
   8.952 ms [  4999] |   } = "pwd"; /* basename */
   1.611 us [  4999] |   __umask_chk();
   7.166 us [  4999] |   getcwd(0, 0) = "/data";
            [  4999] |   puts("/data") {
  12.227 us [  4999] |     /* linux:schedule (pre-empted) */
   2.583 us [  4999] |     /* linux:schedule (pre-empted) */
  31.574 us [  4999] |   } = 0; /* puts */
            [  4999] |   /* linux:task-exit */

task: 4999
[0] __libc_init
```

yugr:
> > I think we need to teach uftrace to respect TMPDIR.
> 
> `TMPDIR` is not defined by default in every environments so I think we should first check whether `getenv("TMPDIR")` isn't `NULL` then use it. Otherwise, use `/tmp` as usual.

Yes, that's what I had in mind.

INSTALL.md
yugr: Makes sense, I had AArch64 binutils installed so couldn't repro. I updated INSTALL.md.

honggyukim:
Thanks for the update! I will need to test more this if it doesn't break the existing Linux environment. I will do it during later weekend. Thanks again.

honggyukim:
I manually checked the result of running `runtest.py` in Linux environment and it doesn't show difference. I think it's safe to be merged unless @namhyung has some comments.

I actually want to do a little bit more refactoring, but I can do that myself. I don't want to hold @yugr for long because of some nitpicks. We can improve based on this work later.

@kangtegong If you could test it on your rooted Android device, then that might be perfect and @namhyung Please have another look.

namhyung:
> (update) After testing a bit more, my system got freeze so I had to reboot my ubuntu machine. It’s weird that I didn’t run android emulator with root permission but I don’t know why it stops my entire system.

Maybe because it consumed a lot of shared memory space?  Sometimes it didn't cleanup the shared memory after record, probably due to some bugs.

utils/shmem.c
```diff
@@ -0,0 +1,96 @@
...
+	fd = open(fname, oflag, mode);
+	if (fd >= 0) {
+		int flags = fcntl(fd, F_GETFD, 0);
+		flags |= FD_CLOEXEC;
+		if (fcntl(fd, F_SETFD, flags) < 0) {
+			int saved_errno = errno;
+			close(fd);
+			fd = -1;
+			errno = saved_errno;
+		}
```
namhyung, line 37: I think we can do this for non-Android as well.  But it should be a separate change.

Also I guess it can be a source of system-wide memory leak if it's not collected by uftrace after exec.  We need to make sure it doesn't leak the shared memory.

utils/shmem.c
```diff
@@ -0,0 +1,96 @@
...
+		const char *tmpdir;
+		tmpdir = getenv("TMPDIR");
+		if (!tmpdir)
+			tmpdir = "/tmp";
+
+		snprintf(uftrace_dir, sizeof(uftrace_dir), "%s/uftrace", tmpdir);
+	}
+
+	return uftrace_dir;
+}
```
namhyung, line 75: It'd be natural to place this function at the beginning since others call it.

namhyung (approved):
LGTM.

And I noticed we have a wiki entry for Android.  Maybe we can update it once this work is merged.

https://github.com/namhyung/uftrace/wiki/uftrace-on-Android

yugr:
> And I noticed we have a wiki entry for Android. Maybe we can update it once this work is merged.

Wow, someone really invested quite some time in it.

namhyung:
Merged. Thanks for your great work, Yury!

honggyukim:
@yugr Thanks very much for your effort. It will be really useful!

yugr:
> @yugr Thanks very much for your effort. It will be really useful!

Thanks for the great tool!

PR #1606, opened 2022-12-18
mcount: Change TLS type to initial-exec

ParkHanbum:
-. Benefit
changed the model of the TLS variable mtd to initial-exec. The initial-exec model doesn't ocurred problem because it is not make any calls to refer the TLS. the way to refer to variables in the intial-exec is adding or subtracting offsets from static TLS blocks as follows.

mov 0x2d6a6 (% rip),% rax # 35fc8 <.got + 0x38>
mov% fs: (% rax),% rbx

as a rough bench result, the initial-exec type TLS variable can be expected to increase the reference speed by about x2 times or more than the TLS variable using the POSIX pthread.

In addition, the initial-exec type TLS variable could be expected to improve the reference speed by about 20% compared to the dynamic type TLS variable created when using a general __thread.

-. Limitation
the initial-exec TLS model have limitation. TLS will allocated by loader and the initial-exec TLS block allocate statically. after process already running and the initial-exec allocated, it can be that there is not enough space in initial-exec static TLS block to allocate mtd. typical execution environment no need to worry about.

--. Limitation mitigation
but to prepare future update, there is need to reduce size of `tls_mtd` to minimized. added new two TLS variable pointer `mtd_tls` and bool `mcount_recursion_marker`.

allocate `struct mcount_thread_data` to heap. and manage it by `*tls_mtd` which declared as TLS variable. and `mcount_recursion_marker` has same role with `mcount_recursion_marker` that inside of `struct mcount_thread_data`.

-.. reference
For more information on this, please see the following link: https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt

reference document written in Korean :
https://die4taoam.tistory.com/37

스크린샷 2022-12-19 오전 5 10 24

233 items excluding standard deviations from approximately 240 samples.
- Reduced execution time by 8.98 ms on average.
- About 7.2% performance improvement

the above benchmark was conducted with the code that exists in review/bench.

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -53,7 +53,8 @@ unsigned long mcount_global_flags = MCOUNT_GFL_SETUP;
 pthread_key_t mtd_key = (pthread_key_t)-1;
 
 /* thread local data to trace function execution */
-TLS struct mcount_thread_data mtd;
+TLS struct mcount_thread_data *tls_mtd TLS_ATTR;
+TLS bool mcount_recursion_marker TLS_ATTR;
 
 /* pipe file descriptor to communite to uftrace */
 int pfd = -1;
@@ -592,15 +593,15 @@ void mtd_dtor(void *arg)
 	struct mcount_thread_data *mtdp = arg;
 	struct uftrace_msg_task tmsg;
 
-	if (mtdp->dead)
+	if (mcount_need_dead(mtdp))
 		return;
 
 	if (mcount_should_stop())
 		mcount_trace_finish(true);
 
 	/* this thread is done, do not enter anymore */
-	mtdp->recursion_marker = true;
-	mtdp->dead = true;
+	__mcount_guard_recursion();
+	mcount_should_dead(get_thread_data());
 
 	if (mcount_estimate_return)
 		mcount_rstack_estimate_finish(mtdp);
@@ -617,6 +618,7 @@ void mtd_dtor(void *arg)
 	mcount_watch_release(mtdp);
 	finish_mem_region(&mtdp->mem_regions);
 	shmem_finish(mtdp);
+	free(mtdp);
 
 	tmsg.pid = getpid();
 	tmsg.tid = mcount_gettid(mtdp);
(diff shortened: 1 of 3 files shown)


namhyung (commented):
But in my benchmark, it doesn't give a noticeable performance benefit.

Base:
```
# uftrace bench
    Self avg    Self min  Function
  ==========  ==========  ====================
    0.089 us    0.072 us  bar
   45.807 us    3.584 us  linux:schedule (pre-empted)
    0.030 us    0.025 us  baz
    0.030 us    0.026 us  foo
```

Your change:
```
# uftrace bench
    Self avg    Self min  Function
  ==========  ==========  ====================
    0.091 us    0.073 us  bar
   46.826 us    3.639 us  linux:schedule (pre-empted)
    0.029 us    0.025 us  foo
    0.029 us    0.024 us  baz
```

namhyung:
I've just merged the bench mark framework.  Please use `make bench` on a real machine (hopefully in idle state) as it would take care of things like cpufreq to get more reliable results.

ParkHanbum:
ok I'll check. 

@namhyung 
have you checked the number of records in the test results? I don't know why, but the number of records was recorded differently, so I had to run a lot.

If you test about 2000 times, less than 300 will be recorded as a normal number. Otherwise, the record count is missing. Please confirm.

ParkHanbum:
and weird result

```
root@f039c186f545:/uftrace# ./uftrace record --libmcount-path=. misc/bench;./uftrace report;
  155.624 ms    0.167 us           1  main
  155.624 ms   59.497 ms           1  bench
   76.973 ms   57.184 ms      500000  bar
   19.789 ms   19.789 ms      441893  baz
   19.153 ms   19.153 ms      420017  foo
    0.459 us    0.459 us           1  __monstartup
    0.042 us    0.042 us           1  __cxa_atexit
root@f039c186f545:/uftrace# vi misc/bench.c 
root@f039c186f545:/uftrace# ./uftrace record --libmcount-path=. misc/bench;./uftrace report;
  160.097 ms    0.166 us           1  main
  160.097 ms   63.551 ms           1  bench
   77.389 ms   57.533 ms      500000  bar
   19.856 ms   19.856 ms      435272  baz
   19.156 ms   19.156 ms      415368  foo
    0.416 us    0.416 us           1  __monstartup
    0.083 us    0.083 us           1  __cxa_atexit

```

some as you can see, some counting is missed while calling bar and foo.

namhyung:
Hmm.. strange.  But I don't see it on my machine.

ParkHanbum:
New test results notified!

First of all, the loss of counts didn't happen when I tested it on a real machine, but I'm still not sure why.

The new test was tested on ubuntu 22.04 on a real machine.
Here are the test results.

스크린샷 2022-12-20 오후 5 44 15


351 items within 1 std of standard deviation from approximately 1000 samples.
-    Reduced execution time by 5.04 ms on average.
-    About 3.3% performance improvement 

Points to note. As you can see in the image, the difference between when the standard deviation is applied and when it is not is large.

namhyung:
Did you run it with `misc/bench.sh` (in other words, did you change the cpufreq governor to `performance`?)  Please share what the script shows otherwise I have no idea what the number means.

libmcount/mcount.c
```diff
@@ -783,15 +800,16 @@ struct mcount_thread_data *mcount_prepare(void)
...
+
 	mcount_filter_setup(mtdp);
 	mcount_watch_setup(mtdp);
-	mtdp->rstack = xmalloc(mcount_rstack_max * sizeof(*mtd.rstack));
+	mtdp->rstack = xmalloc(mcount_rstack_max * sizeof(*mtdp->rstack));
 
 	pthread_once(&once_control, mcount_init_file);
 	prepare_shmem_buffer(mtdp);
 
-	pthread_setspecific(mtd_key, mtdp);
```
namhyung, line 793: With this change, `mtd_dtor()` won't get called when a thread exits.  We rely on this to synchronize thread status with uftrace process and release the resources.  So even if we don't use the TSD APIs anymore, we should keep it for this reason.

namhyung (commented):
I'm really sorry for the late review.

You need to split the change as it contains a lot of changes.  Regarding the limitation, I think we can start from the simple implementation and then add complexities gradually.

PR #1609, opened 2023-01-01
bench: Fix multiple problems and issues in bench

honggyukim:
There are multiple issues in bench compilation as follows.

First, the baz does not return a value although its return type is int so set it to void.
```
  uftrace/misc/bench.c:12:1: warning: non-void function does not return a value [-Wreturn-type]
```
Second, the bench.sh script uses bash syntax, but its shebang is /bin/sh so it makes the following error in some environments.
```
  $ make bench
      ...
  ./bench.sh: 11: Syntax error: "(" unexpected
  make: *** [Makefile:390: bench] Error 2
```
Lastly, the bench executable requires it to be compiled with -pg, but android build doesn't support -pg build so it shows the following error.
```
  clang-14: warning: argument unused during compilation: '-pg' [-Wunused-command-line-argument]
  ld: error: undefined symbol: _mcount
```
It seems there is no reason to build bench image in default build, so it'd be better to opt it out and build it only in 'make bench'.

This patch fixes all the problems and issues above.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ LIBMCOUNT_TARGETS += libmcount/libmcount-single.so libmcount/libmcount-fast-sing
 
 _TARGETS := uftrace libtraceevent/libtraceevent.a
 _TARGETS += $(LIBMCOUNT_TARGETS) libmcount/libmcount-nop.so
-_TARGETS += misc/demangler misc/symbols misc/dbginfo misc/bench
+_TARGETS += misc/demangler misc/symbols misc/dbginfo
 TARGETS  := $(patsubst %,$(objdir)/%,$(_TARGETS))
 
 UFTRACE_SRCS := $(srcdir)/uftrace.c $(wildcard $(srcdir)/cmds/*.c $(srcdir)/utils/*.c)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -380,7 +380,7 @@ unittest: all
 runtest: all
 	@$(MAKE) -C $(srcdir)/tests TESTARG="$(TESTARG)" RUNTESTARG="$(RUNTESTARG)" test_run
 
-bench: all
+bench: all $(objdir)/misc/bench
```
namhyung, line 383: Ok, then it fails when we run `make bench`, right?

misc/bench.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
namhyung, line 1: Hmm.. but I think keeping POSIX shell behavior would be better for a this kind of small script.  What about removing 'function' keyword, is that all it complains?

honggyukim: I have tried it but there are more bash syntaxes so it still complains.

Makefile
honggyukim: Yes, right.

misc/bench.sh
namhyung: Ah, ok.  Let's change it to `bash` then.

namhyung (approved):
LGTM

PR #1611, opened 2023-01-06
misc: Check ID_LIKE along with ID in /etc/os-release

kangtegong:
Since ID in /etc/os-release cannot contain distribution in /misc/install-deps.sh,

/etc/os-release can also have ID_LIKE,
which can contain distribution in /misc/install-deps.sh like below:

```
  $ cat /etc/os-release
  NAME="Amazon Linux"
  VERSION="2"
  ID="amzn"
  ID_LIKE="centos rhel fedora"
  VERSION_ID="2"
  PRETTY_NAME="Amazon Linux 2"
  ANSI_COLOR="0;33"
  CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
  HOME_URL="https://amazonlinux.com/"
```

So this commit adds parsing logic in /misc/install-deps.sh to check ID first, and ID_LIKE also.

Closes: #1468

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -1,5 +1,41 @@
 #!/bin/sh
 
+install_packages() {
+    case $1 in
+        "ubuntu" | "debian")
+            apt-get install $OPT pandoc libdw-dev python3-dev libncursesw5-dev pkg-config
+            apt-get install $OPT libluajit-5.1-dev || true
+            apt-get install $OPT libcapstone-dev || true
+            exit
+            ;;
+        "fedora")
+            dnf install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconf-pkg-config
+            dnf install $OPT luajit-devel || true
+            dnf install $OPT capstone-devel || true
+            exit
+            ;;
+        "rhel" | "centos")
+            rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+            yum install $OPT pandoc elfutils-devel python3-devel ncurses-devel pkgconfig
+            yum install $OPT luajit-devel || true
+            yum install $OPT capstone-devel || true
+            exit
+            ;;
+        "arch" | "manjaro")
+            pacman $OPT -S pandoc libelf python3 ncurses pkgconf
+            pacman $OPT -S luajit || true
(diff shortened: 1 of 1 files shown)


kangtegong:
This PR is tested on amazonlinux2.

misc/install-deps.sh
```diff
@@ -1,5 +1,41 @@
 #!/bin/sh
 
+install_packages() {
+    case $1 in
+        "ubuntu" | "debian" | "raspbian" | "kali" | "linuxmint")
```
namhyung, line 5: I think we can remove "raspbian", "kali", "linuxmint" as they are variants of former two and have ID_LIKE field.

namhyung (commented):
Thanks for the update!

misc/install-deps.sh
kangtegong: Okay, I just amended the commit :)

misc/install-deps.sh
```diff
@@ -16,30 +52,14 @@ if [ ! -f /etc/os-release ]; then
...
-        pacman $OPT -S capstone || true ;;
-    "alpine")
-        apk $OPT add elfutils-dev python3-dev ncurses-dev pkgconf
-        apk $OPT add luajit-dev || true
-        apk $OPT add capstone-dev || true ;;
-    *) # we can add more install command for each distros.
-        echo "\"$distro\" is not supported distro, so please install packages manually."
-        echo ;;
-esac
+echo "\"$distro\" is not supported distro, so please install packages manually."
```
namhyung, line 63: I think the `$distro` should have the value from the original `ID` field and not sure if the for loop above keeps it.  How about using a different variable for `ID_LIKE`?

kangtegong: Sure. I just changed the variable name so `$distro` should only have value from the original `ID`.

misc/install-deps.sh
```diff
@@ -16,30 +52,14 @@ if [ ! -f /etc/os-release ]; then
...
+echo
+exit
```
namhyung, line 65: I don't think we need the `exit` at the end.

kangtegong: Just removed `exit`. Thanks :)

namhyung (approved):
LGTM

PR #1614, opened 2023-01-14
Enable testing of -fpatchable-function-entry instrumentation.

yugr:
This PR adds testing configuration for `-fpatchable-function-entry` to `runtests.py`. It mostly fails on GCC due to #1613 but works well with clang (955 tests passed VS 1006 with `-pg`).

PR #1623, opened 2023-01-21
Read patchable entries from memory instead of disk

yugr:
This change simplifies the code and is also compatible with Android.

It shows no regressions on x86_64 when tested on top of #1614

### DIFF
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -145,21 +145,21 @@ static void read_patchable_loc(struct mcount_dynamic_info *mdi, struct uftrace_e
 			       struct uftrace_elf_iter *iter, unsigned long offset)
 {
 	typeof(iter->shdr) *shdr = &iter->shdr;
+	unsigned i;
+	unsigned long *patchable_loc;
+	unsigned long sh_addr;
 
 	mdi->nr_patch_target = shdr->sh_size / sizeof(long);
 	mdi->patch_target = xmalloc(shdr->sh_size);
+	patchable_loc = mdi->patch_target;
 
-	elf_get_secdata(elf, iter);
-	elf_read_secdata(elf, iter, 0, mdi->patch_target, shdr->sh_size);
+	sh_addr = shdr->sh_addr;
+	if (elf->ehdr.e_type == ET_DYN)
+		sh_addr += offset;
 
-	/* symbol has relative address, fix it to match each other */
-	if (elf->ehdr.e_type == ET_EXEC) {
-		unsigned long *patchable_loc = mdi->patch_target;
-		unsigned i;
-
-		for (i = 0; i < mdi->nr_patch_target; i++) {
-			patchable_loc[i] -= offset;
-		}
+	for (i = 0; i < mdi->nr_patch_target; i++) {
+		unsigned long *entry = (unsigned long *)sh_addr + i;
+		patchable_loc[i] = *entry - offset;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


arch/x86_64/mcount-dynamic.c
```diff
@@ -145,21 +145,16 @@ static void read_patchable_loc(struct mcount_dynamic_info *mdi, struct uftrace_e
...
-		unsigned long *patchable_loc = mdi->patch_target;
-		unsigned i;
-
-		for (i = 0; i < mdi->nr_patch_target; i++) {
-			patchable_loc[i] -= offset;
-		}
+	for (i = 0; i < mdi->nr_patch_target; i++) {
+		unsigned long *entry = (unsigned long *)(offset + shdr->sh_addr) + i;
+		patchable_loc[i] = *entry - offset;
 	}
```
honggyukim, line 158: I tested it and it works fine in default build, which is `ET_DYN` as follows.
```
$ gcc -o t-abc.dyn -fpatchable-function-entry=5 tests/s-abc.c 

$ readelf -h t-abc.dyn | grep Type
  Type:                              DYN (Position-Independent Executable file)

$ uftrace -P. --debug-domain dynamic:4 t-abc.dyn 
dynamic: dynamic patch type: t-abc.dyn: 5 (fpatchable)
dynamic: update 0x56351b77616d for 'a' function dynamically to call __fentry__
dynamic: update 0x56351b776184 for 'b' function dynamically to call __fentry__
dynamic: update 0x56351b77619b for 'c' function dynamically to call __fentry__
dynamic: update 0x56351b7761d3 for 'main' function dynamically to call __fentry__
dynamic: patched all (4) functions in 't-abc.dyn'
            [ 22119] | main() {
            [ 22119] |   a() {
            [ 22119] |     b() {
            [ 22119] |       c() {
   0.886 us [ 22119] |         getpid();
   2.081 us [ 22119] |       } /* c */
   2.306 us [ 22119] |     } /* b */
   2.529 us [ 22119] |   } /* a */
   3.016 us [ 22119] | } /* main */
```
But having `-no-pie` option makes the executable type `ET_EXEC` then it gets segfaulted as follows.
```
$ gcc -o t-abc.exec -no-pie -fpatchable-function-entry=5 tests/s-abc.c 

$ readelf -h t-abc.exec | grep Type
  Type:                              EXEC (Executable file)

$ uftrace -P. --debug-domain dynamic:4 t-abc.exec 
WARN: cannot open record data: /tmp/uftrace-live-VpBStd: No data available
```
Please check if `shdr->sh_addr` is correct in this case.

yugr: > But having `-no-pie` option makes the executable type `ET_EXEC` then it gets segfaulted as follows.
...
> Please check if `shdr->sh_addr` is correct in this case.

Yeah, looks like for ET_EXECs the addresses in section are absolute so we do not need the `offset` addend...

honggyukim: Thanks for the update! I've just tested and it works fine on both x86_64 and aarch64.

honggyukim: @namhyung Please have a final look.

honggyukim: @yugr It'd be better if you could rebase it targeting to the current master.

yugr: Done!

configure
```diff
@@ -177,6 +177,7 @@ check_command() {
 
 check_command make
 check_command ${CC}
+check_command pkg-config
```
namhyung, line 180: I don't think this change belongs to the PR.  Let's move it to #1605.

namhyung (commented):
Looks good.  But I think the change log for the first commit is somewhat misleading.  It does nothing with `.rela.dyn` section, instead please describe that `__patchable_function_entries` has `SHF_ALLOC` and it has the actual addresses of NOP instructions in memory.  So we can read from the memory instead of the file.

configure
yugr: Ok, removed.

namhyung (approved):
LGTM

Issue #1620, opened 2023-01-21
Rust symbol demangling for Traits


namhyung:
When I run uftrace with a rust program, I found that some symbols are not demangled properly.

```
$ cargo new hello_cargo
$ cd hello_cargo
$ cargo build

$ uftrace -P. -F hello_cargo::main target/debug/hello_cargo
Hello, world!
# DURATION     TID     FUNCTION
            [952910] | hello_cargo::main() {
   0.263 us [952910] |   core::fmt::Arguments::new_v1();
            [952910] |   std::io::stdio::_print() {
   0.203 us [952910] |     std::io::stdio::print_to_buffer_if_capture_used();
            [952910] |     std::sync::once_lock::OnceLock<T>::initialize() {
            [952910] |       std::sys_common::once::futex::Once::call() {
   0.093 us [952910] |         __rdl_alloc();
   0.557 us [952910] |       } /* std::sys_common::once::futex::Once::call */
   1.240 us [952910] |     } /* std::sync::once_lock::OnceLock<T>::initialize */
            [952910] |     _<&std..io..stdio..Stdout as std..io..Write>::write_fmt() {
            [952910] |       core::fmt::write() {
            [952910] |         _<std..io..Write..write_fmt..Adapter<T> as core..fmt..Write>::write_str() {
            [952910] |           _<std..io..stdio..StdoutLock as std..io..Write>::write_all() {
   0.128 us [952910] |             std::io::stdio::handle_ebadf();
   5.620 us [952910] |           } /* _<std..io..stdio..StdoutLock as std..io..Write>::write_all */
   5.795 us [952910] |         } /* _<std..io..Write..write_fmt..Adapter<T> as core..fmt..Write>::write_str */
   6.061 us [952910] |       } /* core::fmt::write */
   6.270 us [952910] |     } /* _<&std..io..stdio..Stdout as std..io..Write>::write_fmt */
   8.438 us [952910] |   } /* std::io::stdio::_print */
  10.112 us [952910] | } /* hello_cargo::main */
```

The symbols start with `_<` seems to be for traits.  I think the `".."` part should be `"::"` as it's the standard separator.  Also I think we'd better skip the `"as TRAIT"` part for simplicity.

For example,
```
_<&std..io..stdio..Stdout as std..io..Write>::write_fmt()
```
should be
```
_<&std::io::stdio::Stdout>::write_fmt()
```

PR #1625, opened 2023-01-29
demangle: More elaborated demangling for Rust

ChoKyuWon:
This commit fixes following things:
1. Ignore traits (See #1620)
2. Demangle '..' to '::'
3. Consume extra '_' at starting point of string
4. Add placeholder to support Rust v0 mangling rule (start with '_R')

However, because of 2, it replace the '..' to '::', it need a write permission for the source string. It makes segfault for the demangle test because the source string is const.

Signed-off-by: ChoKyuWon <kyuwoncho18@gmail.com>

### DIFF
--- a/utils/demangle.c
+++ b/utils/demangle.c
@@ -1426,9 +1426,18 @@ static int dd_source_name(struct demangle_data *dd)
 	/* check special symbol mappings (e.g. '$LT$') for Rust */
 	while (dollar != NULL && dollar < end) {
 		bool found = false;
+		char *separator = p;
 
 		num = dollar - p;
-		dd_append_len(dd, p, num);
+		while (true) {
+			char *update = strstr(separator, "..");
+			if (!update || update > dollar)
+				break;
+			dd_append_len(dd, separator, update - separator);
+			dd_append_separator(dd, "::");
+			separator = update + 2;
+		}
+		dd_append_len(dd, separator, dollar - separator);
 
 		for (i = 0; i < ARRAY_SIZE(rust_mappings); i++) {
 			if (strncmp(rust_mappings[i].code, dollar + 1,
@@ -1946,6 +1955,9 @@ TEST_CASE(demangle_rust1)
 		      "$u20$as$u20$foo..Bar$LT$Test$GT$$GT$3barE",
 		      "_<Test + 'static>::bar");
 	DEMANGLE_TEST("_ZN3foo3bar17h05af221e174051e9E", "foo::bar");
+	DEMANGLE_TEST(
+		"_ZN61_$LT$$RF$std..io..stdio..Stdout$u20$as$u20$std..io..Write$GT$9write_fmt17h75c561f414a62159E",
+		"_<&std::io::stdio::Stdout>::write_fmt");
 
 	return TEST_OK;
 }
--- a/utils/demangle.c
+++ b/utils/demangle.c
@@ -1436,9 +1436,17 @@ static int dd_source_name(struct demangle_data *dd)
 				continue;
 
 			dd_add_debug(dd);
-			dd_append(dd, rust_mappings[i].punc);
-			num += strlen(rust_mappings[i].code) + 2;
-			__dd_consume_n(dd, num, NULL);
+			/* skip "as TRAIT" */
+			if (strncmp(dollar, "$u20$as$u20$", 12) == 0) {
+				dd_append(dd, ">");
+				num += (end - dollar);
+				__dd_consume_n(dd, num, NULL);
+			}
+			else {
+				dd_append(dd, rust_mappings[i].punc);
+				num += strlen(rust_mappings[i].code) + 2;
+				__dd_consume_n(dd, num, NULL);
+			}
 
 			p += num;
 			found = true;
@@ -1691,7 +1699,13 @@ static char *demangle_simple(char *str)
 		dd.len -= 15;
 	}
 
-	if (dd.old[0] != '_' || dd.old[1] != 'Z')
+	if (dd.old[0] != '_') {
+		if (dd.old[1] != 'Z' || dd.old[1] != 'R')
+			return xstrdup(str);
+	}
+
+	// TODO: implement demangling Rust v0 mangling
+	if (dd.old[1] == 'R')
 		return xstrdup(str);
 
 	if (dd_encoding(&dd) < 0 || dd.level != 0) {
@@ -1930,7 +1944,7 @@ TEST_CASE(demangle_rust1)
 	DEMANGLE_TEST("_ZN35Bar$LT$$u5b$u32$u3b$$u20$4$u5d$$GT$E", "Bar<[u32; 4]>");
 	DEMANGLE_TEST("_ZN71_$LT$Test$u20$$u2b$$u20$$u27$static"
 		      "$u20$as$u20$foo..Bar$LT$Test$GT$$GT$3barE",
-		      "_<Test + 'static as foo..Bar<Test>>::bar");
+		      "_<Test + 'static>::bar");
 	DEMANGLE_TEST("_ZN3foo3bar17h05af221e174051e9E", "foo::bar");
 
 	return TEST_OK;


utils/demangle.c
```diff
@@ -1415,6 +1415,15 @@ static int dd_source_name(struct demangle_data *dd)
 	dd_append_separator(dd, "::");
 
 	p = dd->old + dd->pos;
+	/* replace ".." to standard separator */
+	for (int i = 0; i < num - 1; ++i) {
+		if (p[i] == '.' && p[i + 1] == '.') {
+			// FIXME: This cause segfault in test because test array is constant.
```
namhyung, line 1421: You cannot change the old string.  Please do it for new string.

namhyung (commented):
Thanks for the PR.  Can you please split the ".." part and trait handling?

utils/demangle.c
```diff
@@ -1426,19 +1427,30 @@ static int dd_source_name(struct demangle_data *dd)
 	/* check special symbol mappings (e.g. '$LT$') for Rust */
 	while (dollar != NULL && dollar < end) {
 		bool found = false;
-
 		num = dollar - p;
-		dd_append_len(dd, p, num);
+		/* consume extra '_' before '$' */
```
namhyung, line 1432: I'm not sure about this.  How can we make sure if it's not a valid identifier.  For example it might have `my_mod_` as a module name, then any dollar sign after that would delete the trailing underscore, right?

utils/demangle.c
```diff
@@ -1691,7 +1703,13 @@ static char *demangle_simple(char *str)
 		dd.len -= 15;
 	}
 
-	if (dd.old[0] != '_' || dd.old[1] != 'Z')
+	if (dd.old[0] != '_') {
+		if (dd.old[1] != 'Z' || dd.old[1] != 'R')
+			return xstrdup(str);
+	}
+
+	// TODO: implement demangling Rust v0 mangling
```
namhyung, line 1717: Thanks for doing this.  Could you please create an issue with this?  It'd be great if it can have some examples.

utils/demangle.c
```diff
@@ -1469,6 +1470,11 @@ static int dd_source_name(struct demangle_data *dd)
 	if (add_name)
 		dd_append_len(dd, p, num);
 	__dd_consume_n(dd, num, NULL);
+	separator = strstr(dd->new, "..");
```
namhyung, line 1473: I'm afraid it might have multiple `".."`.  Does this handle it properly?

namhyung (commented):
Can you rewrite the header line of the second commit message?  The same message confuses readers and I'd prefer being more specific like 'demangle: convert ".." to "::" for Rust'.

utils/demangle.c
ChoKyuWon: I'm not sure why, but in my test case, it seems to work properly. But to make sure, I'll add a loop for that.

utils/demangle.c
ChoKyuWon: That's a good point. I'll revert this code. However, for clearer context, the first underscore would be deleted IMO.

utils/demangle.c
namhyung: I think it worked by luck.  It just converts a single instance, but later `dd_source_name()` call would convert it again.  If the original name contains more than two ".." and it has a `dd_source_name()` after that, not all ".." would be converted.

utils/demangle.c
```diff
@@ -1464,6 +1465,15 @@ static int dd_source_name(struct demangle_data *dd)
...
 		dd_append_len(dd, p, num);
 	__dd_consume_n(dd, num, NULL);
+	while (true) {
+		separator = strstr(dd->new, "..");
+		if (separator) {
+			separator[0] = ':';
+			separator[1] = ':';
+		}
+		else
+			break;
```
namhyung, line 1475: In general, I prefer reducing the nesting level by checking the break condition early.  Something liike below..

```c
while (true) {
    separator = strstr(dd->new, "..");
    if (separator == NULL)
        break;
    separator[0] = ':';
    separator[1] = ':';
}
```

utils/demangle.c
```diff
@@ -1382,6 +1382,7 @@ static int dd_source_name(struct demangle_data *dd)
 {
 	int num = dd_number(dd);
 	char *dollar;
+	char *separator;
```
namhyung, line 1385: You may put it inside the loop below.

utils/demangle.c
ChoKyuWon: You're right. Thanks for the review.

utils/demangle.c
```diff
@@ -1464,6 +1465,15 @@ static int dd_source_name(struct demangle_data *dd)
...
 	__dd_consume_n(dd, num, NULL);
+	while (true) {
+		separator = strstr(dd->new, "..");
+		if (separator) {
+			separator[0] = ':';
+			separator[1] = ':';
+		}
+		else
+			break;
+	}
```
namhyung, line 1476: Another concern is that it's suboptimal.  It checks ".." whenever it calls `dd_source_name()`.  I think we can do it only when the dollar sign was found.  Also it doesn't need to search the whole (new) string.  Could we check the newly added region only?

namhyung (commented):
Please add some explanation in the second commit message.  Like when it needs to be converted or just include an example.

Also I think we should add a test for more complex case like in the #1620.

utils/demangle.c
ChoKyuWon: I'm not sure that the ".." is always inside the <>. I'll try to find a counterexample, and if there's nothing, I'll fix this commit.

ChoKyuWon: I collect the LLVM IR of large-scale projects and search the pattern like this:
```
$ cat *.ll | grep "define[^w]*\.\." | grep -v "$"
```
and it returns nothing.
So ".." only generated with a dollar sign is quite a safe guess. I'll work on the optimal solution.

utils/demangle.c
```diff
@@ -1426,7 +1426,6 @@ static int dd_source_name(struct demangle_data *dd)
 	/* check special symbol mappings (e.g. '$LT$') for Rust */
 	while (dollar != NULL && dollar < end) {
 		bool found = false;
-
```
namhyung, line 1429: Unnecessary change.  Please leave it.

utils/demangle.c
```diff
@@ -1930,8 +1943,11 @@ TEST_CASE(demangle_rust1)
 	DEMANGLE_TEST("_ZN35Bar$LT$$u5b$u32$u3b$$u20$4$u5d$$GT$E", "Bar<[u32; 4]>");
 	DEMANGLE_TEST("_ZN71_$LT$Test$u20$$u2b$$u20$$u27$static"
 		      "$u20$as$u20$foo..Bar$LT$Test$GT$$GT$3barE",
-		      "_<Test + 'static as foo..Bar<Test>>::bar");
+		      "_<Test + 'static>::bar");
 	DEMANGLE_TEST("_ZN3foo3bar17h05af221e174051e9E", "foo::bar");
+	DEMANGLE_TEST(
+		"_ZN61_$LT$$RF$std..io..stdio..Stdout$u20$as$u20$std..io..Write$GT$9write_fmt17h75c561f414a62159E",
+		"_<&std::io::stdio::Stdout>::write_fmt");
```
namhyung, line 1959: This should be moved to the next commit.

utils/demangle.c
```diff
@@ -1426,8 +1426,17 @@ static int dd_source_name(struct demangle_data *dd)
 	/* check special symbol mappings (e.g. '$LT$') for Rust */
 	while (dollar != NULL && dollar < end) {
 		bool found = false;
+		char *separator = p;
```
namhyung, line 1429: The general guide is to have a blank line between declaration and body.

namhyung (commented):
Thanks for your work.  I only have a few nitpicks.

namhyung (approved):
LGTM. Can you please rebase it onto the current master?

namhyung:
You need to add a blank line after the commit header to separate subject and body.  Please take a look at other commit messages.

Issue #1631, opened 2023-02-16
incorrect floating point argument/return values


honggyukim:
The `t083_arg_float.py` fails as follows.
```diff
$ ./runtest.py -vdp -c gcc -O0 083
      ...
t083_arg_float: diff result of gcc -pg -O0
--- expect      2023-02-16 19:28:09.750616419 +0900
+++ result      2023-02-16 19:28:09.750616419 +0900
@@ -4,3 +4,3 @@
    float_mul(300.000000, 400.000000) = 120000.000000;
-   float_div(40000000000.000000, -0.020000) = -2000000000000.000000;
+   float_div(40000000000.000000, -0.020000) = 40000000000.000000;
  } /* main */

083 arg_float           : NG
```
This can simply be reproduced as follows.
```
$ gcc -pg -g tests/s-exp-float.c

$ uftrace -a -F float_div t-exp-float
# DURATION     TID     FUNCTION
   0.851 us [1775190] | float_div(40000000000.000000, -0.020000) = 40000000000.000000;
```

honggyukim:
This problem disappears when uftrace itself is compiled with `-O0` and it shows only when `libmcount.so` is optimized.

The root cause is that `mcount_memcpy4` overwrites `$xmm0` when it's optimized as follows.
```c
274│ static inline void mcount_memcpy4(void *restrict dst, const void *restrict src, int len)
275│ {
276│         unsigned int *restrict p = dst;
277│         const unsigned int *restrict q = src;
278│         int len4 = len / 4;
279│
280│         while (len4-- > 0)
281├───────────────> *p++ = *q++;
282│ }
```
The above line actually calls `memmove@plt` inside.
```asm
208│    0x00007ffff7f8238c <+284>:   call   0x7ffff7f78960 <memmove@plt>
```
Inside of `memmove` implementation, there are assembly instructions that overwrites `$xmm0` as follows.
```asm
   1│ Dump of assembler code for function __memmove_avx_unaligned_erms:
   2│ ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:
   3│ 307     in ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S
   4│    0x00007ffff7ed9940 <+0>:     endbr64
   5│
   6│ 308     in ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S
   7│    0x00007ffff7ed9944 <+4>:     mov    rax,rdi
   8│
   ...
[...]

PR #1632, opened 2023-02-17
libmcount: Turn off unexpected vectorization in libmcount

honggyukim:
Sometimes compiler inserts some builtin functions such as memcpy, memmove even if the source code doesn't have them at all.

It's internally used because the compiler thinks that using those calls is faster when optimized.

However, this can overwrite vector registers such as xmm registers in x86 or VFP registers in arm and that makes uftrace record floating point arguments or return values incorrectly.

One of the example can be found as follows.
```diff
$ ./runtest.py -vdp -c gcc -O0 083
      ...
t083_arg_float: diff result of gcc -pg -O0
--- expect      2023-02-16 19:28:09.750616419 +0900
+++ result      2023-02-16 19:28:09.750616419 +0900
@@ -4,3 +4,3 @@
    float_mul(300.000000, 400.000000) = 120000.000000;
-   float_div(40000000000.000000, -0.020000) = -2000000000000.000000;
+   float_div(40000000000.000000, -0.020000) = 40000000000.000000;
  } /* main */

083 arg_float           : NG
```
To avoid this problem, libmcount should be compiled with '-fno-builtin' and '-fno-tree-vectorize' options.

The problem is fixed by having those options.

Fixed: #1631


namhyung (approved):
LGTM

Issue #1618, opened 2023-01-19
Change the library install location


namhyung:
As uftrace has a number of libraries to install and it has more to come, I think it'd be better to create a sub-directory in the `lib` and then copy the libraries there.  Currently the default value of `libdir` is `${prefix}/lib` and I'd like to change it to `${prefix}/lib/uftrace`.  Users can change the value with `--libdir=<PATH>` option when running `configure` script.

Then it should work like below:
```
$ cd uftrace
$ ./configure

$ cat .config
# this file is generated automatically
override prefix := /usr/local
override bindir := /usr/local/bin
override libdir := /usr/local/lib/uftrace
override mandir := /usr/local/share/man
override etcdir := /usr/local/etc

override ARCH   := x86_64
override CC     := gcc
override LD     := ld
override CFLAGS  = 
override LDFLAGS = 

override srcdir := /home/namhyung/project/uftrace
override objdir := /home/namhyung/project/uftrace
```

After the install, you can see the libmcount libraries in the uftrace directory.
```
$ sudo make install

$ ls /usr/local/lib/uftrace
libmcount-fast-single.so  libmcount-fast.so  libmcount-nop.so  libmcount-single.so  libmcount.so
```

namhyung:
Something like below should work:

```diff
diff --git a/INSTALL.md b/INSTALL.md
index 580eb54fe2..009fe59263 100644
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -116,7 +116,7 @@ directories or build directory with this script.
       --help                print this message
       --prefix=<DIR>        set install root dir as <DIR>        (default: /usr/local)
       --bindir=<DIR>        set executable install dir as <DIR>  (default: ${prefix}/bin)
-      --libdir=<DIR>        set library install dir as <DIR>     (default: ${prefix}/lib)
+      --libdir=<DIR>        set library install dir as <DIR>     (default: ${prefix}/lib/uftrace)
       --mandir=<DIR>        set manual doc install dir as <DIR>  (default: ${prefix}/share/man)
       --objdir=<DIR>        set build dir as <DIR>               (default: ${PWD})
       --sysconfdir=<DIR>    override the etc dir as <DIR>
diff --git a/Makefile b/Makefile
index 70dc12f0c9..047f99655b 100644
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ endif
 
 prefix ?= /usr/local
 bindir = $(prefix)/bin
[...]

PR #1633, opened 2023-02-17
uftrace: Modify install location for uftrace libraries

kangtegong:
This commit changes default uftrace library location from `${prefix}/lib` to `${prefix}/lib/uftrace`.

Also users can set library location with `--libdir=<DIR>` option of configure script like below:
```
  $ ./configure --libdir=/home/minchul/libuftrace
  $ cat .config
  # this file is generated automatically
  override prefix := /usr/local
  override bindir := /usr/local/bin
  override libdir := /home/minchul/libuftrace
  ...
  $ make && sudo make install
  ...
  $ ls /home/minchul/libuftrace/
  libmcount-fast-single.so  libmcount-fast.so  libmcount-nop.so  libmcount-single.so  libmcount.so
```
Closes: #1618

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -116,7 +116,7 @@ directories or build directory with this script.
       --help                print this message
       --prefix=<DIR>        set install root dir as <DIR>        (default: /usr/local)
       --bindir=<DIR>        set executable install dir as <DIR>  (default: ${prefix}/bin)
-      --libdir=<DIR>        set library install dir as <DIR>     (default: ${prefix}/lib)
+      --libdir=<DIR>        set library install dir as <DIR>     (default: ${prefix}/lib/uftrace)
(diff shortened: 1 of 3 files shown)


kangtegong:
This commit passes all unittests and runtests.py and had no problems in various usages.

Also, just for your information, this commit once had [pre-commit failure ](https://github.com/namhyung/uftrace/actions/runs/4206723227/jobs/7300589500#step:12:134)
but the failure was attributed to a recent isort bug[1][2].

When I re-ran the CI, all tests passed successfully.

[1] https://stackoverflow.com/questions/75281731/pre-commit-fails-to-install-isort-5-10-1-with-error-runtimeerror-the-poetry-co
[2] https://github.com/PyCQA/isort/issues/2077

namhyung (approved):
LGTM

Issue #1630, opened 2023-02-13
s-nested.c with -fpatchable-function-entry gets crashed


honggyukim:
The `s-nested.c` can be compiled as follows.
```
$ gcc -fpatchable-function-entry=5 s-nested.c
```
But this gets crashed when recording with `-P`.
```
$ uftrace -P. a.out 
WARN: Segmentation fault: invalid permission (addr: 0x7ffc7c59a080)
WARN:  if this happens only with uftrace, please consider -e/--estimate-return option.

WARN: Backtrace from uftrace v0.13.1-24-g55da ( x86_64 dwarf python3 luajit tui perf sched dynamic )
WARN: =====================================
WARN: [2] (foo_internal.0[5641d530e18e] <= foo[5641d530e1de])
WARN: [1] (foo[5641d530e1ab] <= main[5641d530e342])
WARN: [0] (main[5641d530e2ff] <= <7fbe80e56d90>[7fbe80e56d90])

Please report this bug to https://github.com/namhyung/uftrace/issues.

WARN: child terminated by signal: 11: Segmentation fault
# DURATION     TID     FUNCTION
            [125510] | main() {
            [125510] |   foo() {
            [125510] |     foo_internal.0() {

uftrace stopped tracing with remaining functions
================================================
task: 125510
[2] foo_internal.0
[1] foo
[0] main
```
This can simply be reproduced with `-P foo_internal.0`.
```
$ uftrace record -P foo_internal.0 --debug-domain dynamic:3 a.out 
dynamic: dynamic patch type: a.out: 5 (fpatchable)
dynamic: update 0x560ebe2c8189 for 'foo_internal.0' function dynamically to call __fentry__
dynamic: patched all (1) functions in 'a.out'
WARN: Segmentation fault: invalid permission (addr: 0x7fff263b3080)
WARN:  if this happens only with uftrace, please consider -e/--estimate-return option.

WARN: Backtrace from uftrace v0.13.1-24-g55da ( x86_64 dwarf python3 luajit tui perf sched dynamic )
WARN: =====================================
WARN: [0] (foo_internal.0[560ebe2c818e] <= foo[560ebe2c81de])

Please report this bug to https://github.com/namhyung/uftrace/issues.

WARN: child terminated by signal: 11: Segmentation fault
```
It looks fine in `foo_internal.0` with `objdump -d`.
```asm
0000000000001189 <foo_internal.0>:
    1189:       90                      nop
    118a:       90                      nop
[...]

PR #1637, opened 2023-02-25
x86_64: Fix an error in s-nested.c with -fpatchable-function-entry

honggyukim:
The current s-nested.c gets crashed when built with -fpatchable-function-entry as follows.
```
  $ gcc -fpatchable-function-entry=5 s-nested.c
  $ uftrace record -P foo_internal.0 a.out
  WARN: Segmentation fault: invalid permission (addr: 0x7fff263b3080)
```
The crash happens because the inner function 'foo_internal.0' reads r10 register to access variables in the outter function.

However, our '__fentry__' implementation doesn't push r10 register for later usage so r10 is changed unexpectedly.

The x86-64-ABI document[1] describes r10 register as follows.
```
  r10 is used for passing a function's static chain pointer.
```
This patch pushes r10 register inside '__fentry__' and it also pushes r11 register together just like '__dentry__' does[2].

[1] https://gitlab.com/x86-psABIs/x86-64-ABI/-/commit/adc986909f2c5aad7aeedbcc05b5f449a2eabfbf
[2] 68104739 arch/x86_64: Fix crash on (full) dynamic tracing

Fixed: #1630

### DIFF
--- a/arch/x86_64/fentry.S
+++ b/arch/x86_64/fentry.S
@@ -56,8 +56,14 @@ GLOBAL(__fentry__)
 	/* save rax (implicit argument for variadic functions) */
 	push %rax
 
+	/* save scratch registers due to -fipa-ra */
+	push %r10
+	push %r11
+
 	call mcount_entry
 
+	pop  %r11
+	pop  %r10
 	pop  %rax
 
 	/* restore original stack pointer */


namhyung (approved):
LGTM

PR #1642, opened 2023-03-03
unittest: Fix c99 compiler error

clementguidi:
Hello Namhyung and Honggyu!

I'm back with this very short fix to begin with, because some unit tests can't compile with old gcc.

After having figured out our upstreaming process at Ciena, I'm ready to send you propositions for the agent in upcoming pull requests. I suggest you label my work as `ciena` from now if you want.

### DIFF
--- a/misc/bench.c
+++ b/misc/bench.c
@@ -20,8 +20,9 @@ int bar(volatile int *ptr)
 int bench(int count)
 {
 	volatile int result = 0;
+	int i;
 
-	for (int i = 0; i < count; i++) {
+	for (i = 0; i < count; i++) {
 		if (i % 2 == 0)
 			foo(&result);
 		else
@@ -33,6 +34,7 @@ int bench(int count)
 int main(int argc, char *argv[])
 {
 	int n = 1;
+	int i;
 	int loop = 1000000;
 	int result = 0;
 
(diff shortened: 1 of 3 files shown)


namhyung:
Thanks Clement, good to see you working on it again.

I remember there was a conversation to increase the minimal supported C version to 99.  But I don't think we updated the Makefile actually.  Can you please add `--std=c99` to COMMON_CFLAGS instead?

clementguidi:
Sorry I cannot find this discussion. I tried looking up `c99` in issues but can't find it.

If we want to upgrade to `--std=c99` we need some refactoring as the code won't compile. For example we need to use `__typeof__` instead of `typeof`:

```
uftrace/cmds/recv.c: In function ‘find_client’:
uftrace/cmds/recv.c:327:2: warning: implicit declaration of function ‘typeof’ [-Wimplicit-function-declaration]
```

Could `--std=gnu99` do the trick?

honggyukim:
@clementguidi Thanks for coming back.

> I remember there was a conversation to increase the minimal supported C version to 99. But I don't think we updated the Makefile actually. Can you please add --std=c99 to COMMON_CFLAGS instead?

How about using C11 instead? I know that it requires some refactoring to conform C standard as @clementguidi mentioned. So maybe we can use the same version as Linux kernel, which is `-std=gnu11`.

If possible, I think we can consider using `thread_local` keyword instead of using `pthread_setspecific`/`pthread_getspecific`.

namhyung (approved):
LGTM

PR #1643, opened 2023-03-03
mcount: Toggle tracing at runtime

clementguidi:
Hi Namhyung and Honggyu,

This PR makes it possible to pause and resume tracing from the client, in a similar fashion to `trace_on` and `trace_off` triggers. When tracing is stopped, the instrumentation is simply skipped on function entry or exit, and event collection doesn't take place.

This mechanism can be used in conjunction with function or signal trigger actions.
```
$ uftrace record --agent --trace=off ./prog
$ uftrace --pid $(pidof prog) --trace=on     # toggle tracing on
$ uftrace --pid $(pidof prog) --trace=off    # toggle tracing back off
```

I'm refactoring the `--disable` option to `--trace=on|off` for clarity.
I'm simplifying agent internals to make it easier to integrate other options.

~Based on: #1642~
~Based on: #1665~

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -56,7 +56,7 @@ static bool match_replay_triggers(const char *trigger)
 static void reset_live_opts(struct uftrace_opts *opts)
 {
 	/* this is needed to set display_depth at replay */
-	live_disabled = opts->disabled;
+	live_disabled = (opts->trace == TRACE_STATE_OFF);
 
 	/*
 	 * These options are handled in record and no need to do it in
@@ -144,7 +144,7 @@ static void reset_live_opts(struct uftrace_opts *opts)
 
 others:
 	opts->depth = MCOUNT_DEFAULT_DEPTH;
-	opts->disabled = false;
+	opts->trace = TRACE_STATE_ON;
 	opts->no_event = false;
 	opts->no_sched = false;
 }
@@ -461,7 +461,7 @@ TEST_CASE(live_reset_options)
 {
 	struct uftrace_opts o = {
 		.depth = 3,
-		.disabled = true,
+		.trace = TRACE_STATE_OFF,
 		.no_event = true,
 		.no_sched = true,
 	};
@@ -482,7 +482,7 @@ TEST_CASE(live_reset_options)
 	TEST_EQ(o.caller, NULL);
 	/* it should only have the color trigger */
 	TEST_STREQ(o.trigger, "bar@time=1us,color=red;baz@backtrace,trace");
-	TEST_EQ(o.disabled, false);
+	TEST_EQ(o.trace, TRACE_STATE_ON);
 	TEST_EQ(o.no_event, false);
 	TEST_EQ(o.no_sched, false);
 
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -340,9 +340,13 @@ static int forward_options(struct uftrace_opts *opts)
 	if (capabilities < 0)
 		goto close;
 
-	/* FIXME Forward user options and set status */
-	if (0)
-		status = forward_option(sfd, capabilities, 0, NULL, 0);
+	if (opts->trace != TRACE_STATE_NONE) {
+		int trace = (opts->trace == TRACE_STATE_ON);
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_TRACE, &trace,
+					sizeof(trace));
+		if (status < 0)
+			goto close;
+	}
 
 close:
 	status_close = agent_message_send(sfd, UFTRACE_MSG_AGENT_CLOSE, NULL, 0);
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -106,7 +106,7 @@ static pthread_t agent;
 /* state flag for the agent */
 static volatile bool agent_run = false;
 
-#define MCOUNT_AGENT_CAPABILITIES 0
+#define MCOUNT_AGENT_CAPABILITIES (UFTRACE_AGENT_OPT_TRACE)
 
 __weak void dynamic_return(void)
 {
@@ -1865,8 +1865,17 @@ static int agent_read_option(int fd, int *opt, void **value, size_t read_size)
 static int agent_apply_option(int opt, void *value, size_t size, struct rb_root *triggers)
 {
 	int ret = 0;
+	int trace;
 
 	switch (opt) {
+	case UFTRACE_AGENT_OPT_TRACE:
+		trace = *((int *)value);
+		if (mcount_enabled != trace) {
+			mcount_enabled = trace;
+			pr_dbg("turn trace %s\n", mcount_enabled ? "on" : "off");
+		}
+		break;
+
 	default:
 		ret = -1;
 	}
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -445,7 +445,7 @@ static void mcount_filter_init(enum uftrace_pattern_type ptype, bool force)
 	if (getenv("UFTRACE_DEPTH"))
 		mcount_depth = strtol(getenv("UFTRACE_DEPTH"), NULL, 0);
 
-	if (getenv("UFTRACE_DISABLED"))
+	if (getenv("UFTRACE_TRACE_OFF"))
 		mcount_enabled = false;
 }
 
--- a/tests/t281_agent_trace_toggle.py
+++ b/tests/t281_agent_trace_toggle.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    """Run the target with the agent activated. Tracing is disabled by the
+    client before triggering the second loop in the target, and restored just
+    after.
+
+    That way, func() gets executed 3 times in the target, but is only recorded 2
+    times by uftrace."""
+    def __init__(self):
+        TestBase.__init__(self, 'agent', """
+# DURATION     TID     FUNCTION
+            [  5650] | main() {
+            [  5650] |   func() {
+            [  5650] |     a() {
+            [  5650] |       b() {
+   0.113 us [  5650] |         c();
+   1.618 us [  5650] |       } /* b */
+   1.910 us [  5650] |     } /* a */
+   2.301 us [  5650] |   } /* func */
+            [  5650] |   func() {
+            [  5650] |     a() {
+            [  5650] |       b() {
+   0.430 us [  5650] |         c();
+   3.898 us [  5650] |       } /* b */
+   4.993 us [  5650] |     } /* a */
+   6.843 us [  5650] |   } /* func */
+  37.310 ms [  5650] | } /* main */
+""")
+
+    def client_send_command(self, pid, option):
+        self.subcmd = 'live'
+        self.option = '-p %d %s' % (pid, option)
+        self.exearg = ''
+        client_cmd = self.runcmd()
+        self.pr_debug('prerun command: ' + client_cmd)
+        client_p = sp.run(client_cmd.split())
+        return client_p.returncode
+
+    def prerun(self, timeout):
+        self.subcmd = 'record'
+        self.option  = '--agent'
+        self.option += ' --keep-pid'
+        self.option += ' --no-libcall'
+        self.exearg = 't-' + self.name
+        record_cmd  = self.runcmd()
+        self.pr_debug("prerun command: " + record_cmd)
+        record_p = sp.Popen(record_cmd.split(), stdin=sp.PIPE, stderr=sp.PIPE, bufsize=0)
+
+        sleep(.05)              # time for the agent to start
+        if self.client_send_command(record_p.pid, '--trace=off') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        if self.client_send_command(record_p.pid, '--trace=on') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        record_p.stdin.close()
+        record_p.wait()
+
+        return TestBase.TEST_SUCCESS
+
+    def setup(self):
+        self.subcmd = 'replay'
+        self.option = ''
+        self.exearg = ''
--- a/uftrace.c
+++ b/uftrace.c
@@ -59,6 +59,7 @@ enum uftrace_short_options {
 	OPT_avg_self,
 	OPT_color,
 	OPT_disabled,
+	OPT_trace,
 	OPT_demangle,
 	OPT_dbg_domain,
 	OPT_report,
@@ -149,7 +150,7 @@ __used static const char uftrace_help[] =
 "      --diff=DATA            Report differences\n"
 "      --diff-policy=POLICY   Control diff report policy\n"
 "                             (default: 'abs,compact,no-percent')\n"
-"      --disable              Start with tracing disabled\n"
+"      --disable              Start with tracing disabled (deprecated)\n"
 "  -D, --depth=DEPTH          Trace functions within DEPTH\n"
 "  -e, --estimate-return      Use only entry record type for safety\n"
 "      --event-full           Show all events outside of function\n"
@@ -221,6 +222,7 @@ __used static const char uftrace_help[] =
 "      --task-newline         Interleave a newline when task is changed\n"
 "      --tid=TID[,TID,...]    Only replay those tasks\n"
 "      --time                 Print time information\n"
+"      --trace=STATE          Set the recording state: on, off (default: on)\n"
 "  -T, --trigger=FUNC@act[,act,...]\n"
 "                             Trigger action on those FUNCs\n"
 "  -U, --unpatch=FUNC         Don't apply dynamic patching for FUNCs\n"
@@ -284,6 +286,7 @@ static const struct option uftrace_options[] = {
 	NO_ARG(avg-self, OPT_avg_self),
 	REQ_ARG(color, OPT_color),
 	NO_ARG(disable, OPT_disabled),
+	REQ_ARG(trace, OPT_trace),
 	REQ_ARG(demangle, OPT_demangle),
 	NO_ARG(record, OPT_record),
 	NO_ARG(report, OPT_report),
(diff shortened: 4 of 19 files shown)


cmds/live.c
```diff
@@ -203,12 +203,33 @@ static void setup_child_environ(struct uftrace_opts *opts)
...
+ * @size - size of the option data load
+ * @return - status: < 0 on error, >= on success
+ */
+static int forward_option(int fd, enum agent_opt opt, void *value, size_t size)
+{
+	int ret = 0;
+
+	ret = socket_send_option(fd, opt, value, size);
+	if (ret >= 0)
+		pr_dbg("sent option %s\n", stringify(opt));
```
namhyung, line 220: This `stringify` won't work as intended and it'd show "opt".

utils/socket.h
```diff
@@ -5,13 +5,13 @@
 
 #define MCOUNT_AGENT_SOCKET_DIR "/tmp/uftrace"
 
-enum uftrace_dopt;
+enum agent_opt;
```
namhyung, line 8: Please keep `uftrace` prefix for non-static functions or data structures.  Also I think 'message' would be a better term than 'option'.  How about renaming it to `enum uftrace_agent_msg`?

uftrace.c
```diff
@@ -857,8 +857,13 @@ static int parse_option(struct uftrace_opts *opts, int key, char *arg)
...
 		break;
 
-	case OPT_disabled:
-		opts->disabled = true;
+	case OPT_trace:
+		if (!strcmp(arg, "on"))
+			opts->trace = TRACE_STATE_ON;
+		else if (!strcmp(arg, "off"))
+			opts->trace = TRACE_STATE_OFF;
+		else if (strcmp("arg", "on"))
```
namhyung, line 865: Wouldn't it be simply `else`?

uftrace.c
```diff
@@ -1385,6 +1390,7 @@ int main(int argc, char *argv[])
 		.patt_type = PATT_REGEX,
 		.show_args = true,
 		.clock = "mono",
+		.trace = TRACE_STATE_NONE,
```
namhyung, line 1393: We can skip this part as long as its value is 0.

uftrace.c
```diff
@@ -1484,6 +1490,9 @@ int main(int argc, char *argv[])
 	if (opts.mode == UFTRACE_MODE_TUI)
 		opts.srcline = true;
 
+	if (!opts.pid) /* Keep uninitialized values in client mode */
```
namhyung, line 1493: Do we really need to keep the value?  It feels like the sensible default value is ON and we can change the state for uncommon case like client mode or explicit trace-off.

cmds/live.c
```diff
@@ -240,6 +240,13 @@ static int forward_options(struct uftrace_opts *opts)
 		goto socket_error;
 	}
 
+	if (opts->trace != TRACE_STATE_NONE) {
+		bool trace = opts->trace == TRACE_STATE_ON;
+		if ((ret = forward_option(sfd, AGENT_TRACE, &trace, sizeof(bool))) < 0)
```
namhyung, line 245: Please break the statement and don't assign a variable in the condition.  It's just a matter of preference and I like simple things. :)

And for I/O, I think we'd better using a sized int (like uint8_t or int32_t) rather than bool.

libmcount/mcount.c
```diff
@@ -1838,6 +1839,14 @@ void *agent_apply_commands(void *arg)
 			}
 
 			switch (opt) {
+			case AGENT_TRACE:
+				read(cfd, &trace, sizeof(bool));
```
namhyung, line 1843: Please check the return value.

uftrace.h
```diff
@@ -422,6 +422,8 @@ enum uftrace_msg_type {
 /* Options sent by the client to the agent */
 enum agent_opt {
 	AGENT_CLOSE = (1U << 0), /* Close the connection with the client */
+
+	AGENT_TRACE = (1U << 5), /* Tracing state (on/off) */
```
namhyung, line 426: Do you intend to have a complex option with bitwise-OR?  Otherwise we can just use natural numbers.

namhyung (commented):
Please do not include many changes in a single commit unless it's trivial.  I prefer small commits divided by logical units.  Refactoring can be done separately.

utils/socket.h
clementguidi: Sounds good. Should I use a prefix for the enum members too? Like `UFTRACE_AGENT_MSG_CLOSE`? I'm afraid the names are getting too long.

cmds/live.c
```diff
@@ -203,12 +203,33 @@ static void setup_child_environ(struct uftrace_opts *opts)
...
 
+/**
+ * forward_option - forward a given option and its data to the agent
+ * @fd - agent socket file descriptor
+ * @opt - option to send
+ * @data - option data load
+ * @size - size of the option data load
+ * @return - status: < 0 on error, >= on success
+ */
+static int forward_option(int fd, enum agent_opt opt, void *value, size_t size)
```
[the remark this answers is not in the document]
clementguidi: I wanted to centralize debug output. And later add a capability feature, to check if the agent supports a given feature. This would fit in this function.

uftrace.h
clementguidi: Yes I want to introduce a capability system, so the client can request supported features from the agent. For example changing filters could require locks, and thus be disabled on certain libmcount builds.

uftrace.c
clementguidi: Yes I messed up when rebasing changes, sorry.

uftrace.c
clementguidi: When in client mode, we only want to forward the trace option to the agent went the user specifies `--trace=on|off`. Otherwise, when tracing is off, the default `trace=on` applies, so the user needs to use `--trace=off` explicitly so the tracing state doesn't change by itself.
The `TRACE_STATE_NONE` value means the user hasn't specified a new tracing state. I need to have it as a default value to know whether the `TRACE_STATE_ON` value comes from user action or is a default value. Do you see another solution?

clementguidi:
Force-push: fixed the conversations marked as resolved.

cmds/live.c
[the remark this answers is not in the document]
namhyung: Ok, then it'd be fine.

utils/socket.h
namhyung: What about `UAM_CLOSE`?  UAM for uftrace agent message. :)

uftrace.c
namhyung: No, I think we're good.

uftrace.h
namhyung: I think the client should receive a response from the agent after sending the message to check if it's supported.  But I'm not sure if it needs OR-ing the messages like `UAM_CLOSE | UAM_TRACE` or something.

utils/socket.h
clementguidi: I'm good with it.

uftrace.h
clementguidi: Ok, the way I intended it was to ask for all the supported features, and then only forwarding options that are supported. What do you think?
You can check the relevant commit on #1645: b6710b2.

clementguidi:
- Rebase on master
- Change `uftrace_agent_msg` to `UAM_`
- Fix client debug messages

uftrace.h
namhyung: Yeah, I mean it too.  Then we can define the enum simply like:

```c
enum uftrace_agent_message {
    UAM_CLOSE = 0,
    UAM_TRACE = 1,
    ...
};
```

namhyung: Also, I think agent need to send 'ACK' back to client after receiving a message so that it can confirm whether it's handled successfully.

utils/socket.c
```diff
@@ -52,13 +52,20 @@ int socket_listen(int fd, struct sockaddr_un *addr)
...
-int socket_send_option(int fd, enum uftrace_dopt opt, void *value, size_t size)
+/**
+ * socket_send_option - send a single option to the agent through its socket
+ * @fd - agent socket file descriptor
+ * @opt - option to send
+ * @data - data load for the option
+ * @size - data size
+ * @return - status code, negative for error
+ */
+int socket_send_option(int fd, enum uftrace_agent_msg opt, void *data, size_t size)
```
namhyung, line 63: Maybe we can call it `socket_send_message()` or `uftrace_socket_send_message()`.

uftrace.h
```diff
@@ -424,6 +424,8 @@ enum uftrace_msg_type {
  */
 enum uftrace_agent_msg {
 	UAM_CLOSE = (1U << 0), /* Close the connection with the client */
+
+	UAM_TRACE = (1U << 5), /* Tracing state (on/off) */
```
namhyung, line 428: Please just use the next integer rather than make a hole.

clementguidi: I'm doing it. Just to clarify, and wanted to separate internal messages - such as `close` or later `data size` - from real options like `trace=on|off`. I wanted to fit the internal messages in later commits in the hole, rather than reorder options each time I need to add one.

uftrace.h
clementguidi: I'm OK with having simple integers in the enum. I'm going to send `ACK` from the agent for every message from the client.

> I think the client should receive a response from the agent after sending the message to check if it's supported.

Something like a handshake message prior to sending other messages?

utils/socket.c
clementguidi: OK, same with all other socket utils that are used in libmcount/live?

clementguidi:
I changed the message enum and added ack messages for every option sent.

cmds/live.c
```diff
@@ -203,6 +203,79 @@ static void setup_child_environ(struct uftrace_opts *opts)
...
+
+	switch (ack) {
+	case UFTRACE_AGENT_MSG_OK:
+		pr_dbg3("sent message '%s'\n", name);
+		break;
+
+	case UFTRACE_AGENT_MSG_EFAIL:
+		ret = -1;
+		pr_dbg2("agent operation failed: '%s'\n", name);
+
```
honggyukim, line 264: You missed adding `break` here.

libmcount/mcount.c
```diff
@@ -1890,6 +1927,8 @@ static void agent_kill()
 	return;
 
 error:
+	if (status < 0)
```
honggyukim, line 1942: My compiler complains here as follows.
```
/home/honggyu/work/uftrace/libmcount/mcount.c:1942:12: warning: ‘status’ may be used uninitialized [-Wmaybe-uninitialized]
 1942 |         if (status < 0)
      |            ^
/home/honggyu/work/uftrace/libmcount/mcount.c:1905:13: note: ‘status’ was declared here
 1905 |         int status;
      |             ^~~~~~
```

uftrace.c
```diff
@@ -858,7 +861,17 @@ static int parse_option(struct uftrace_opts *opts, int key, char *arg)
 		break;
 
 	case OPT_disabled:
-		opts->disabled = true;
+		pr_use("'--disable' is deprecated, use --trace=on|off instead.\n");
```
honggyukim, line 864: I think the message is better to be `use --trace=off instead` because `--trace=on` is not related to `--disable`.

uftrace.c
```diff
@@ -858,7 +861,17 @@ static int parse_option(struct uftrace_opts *opts, int key, char *arg)
...
+		opts->trace = TRACE_STATE_OFF;
+		break;
+
+	case OPT_trace:
+		if (!strcmp(arg, "on"))
+			opts->trace = TRACE_STATE_ON;
+		else if (!strcmp(arg, "off"))
+			opts->trace = TRACE_STATE_OFF;
+		else
+			pr_use("unknown tracing state: %s (ignoring..)\n", arg);
```
honggyukim, line 874: It doesn't have to be included in this PR, but I think we can extend this option for later usage to control trace on/off with time duration. For example, we can support following usages.
```
# trace target program only for 1ms from the beginning.
$ uftrace record --trace=~1ms ./prog

# trace target program since 1ms elapsed from the beginning.
$ uftrace record --trace=1ms~ ./prog

# trace target program only from 1ms elapsed from the beginning to 3ms elapsed, then stop tracing.
# so target program is recorded during 2ms in total.
$ uftrace record --trace=1ms~3ms ./prog
```
Otherwise, we can support `-r`/`--time-range` option for this purpose instead for record time.

honggyukim (commented):
Thanks for doing this work! I have tested it a bit, but it seems it records nothing when I followed the sequence you provided. I tested it with `sleep 30` with `--trace=off` and `-K 5` for kernel tracing at the record time.

Then I ran `--trace=on` and `--trace=off` for the `sleep` process. But `uftrace replay` shows nothing. If I ran only `--trace=on` then `uftrace replay` shows something.

honggyukim:
One more minor request is that we better print something in debug level 1 when agent receives a request whether it's `--trace` on or off.

uftrace.c
clementguidi: Ok. Should I remove `--disable` from the documentation? Or mark it as deprecated there too?

clementguidi:
> Then I ran `--trace=on` and `--trace=off` for the `sleep` process. But `uftrace replay` shows nothing. If I ran only `--trace=on` then `uftrace replay` shows something.

From what I understand, this is because there is no userspace data recorded. See the debug output:
```
kernel: found kernel ftrace data for 8 cpus
[...]
fstack: cannot open task data file: /tmp/uftrace-live-DHm4TZ/2782.dat: No such file or directory
```

uftrace doesn't expect to only have kernel data. Should we handle this scenario?

uftrace.h
namhyung: Right, maybe we need to design the communication protocol first.  I don't want to make things complicated but I guess we will add more messages later.  Also not sure if multiple clients want to talk to the agent at the same time.

uftrace.h
namhyung: I'm not sure if it's meaningful to separate the internal and public messages.  Hmm.. how about making a message can handle many options?  Something like below?

```c
enum uftrace_agent_msg {
    UAM_CLOSE, /* close the connection */
    UAM_QUERY, /* query supported features/options */
    UAM_GET_OPT, /* get current option value */
    UAM_SET_OPT, /* set new option value */
    ...
};
```

Each message should handle variable-length data then.

namhyung:
> uftrace doesn't expect to only have kernel data. Should we handle this scenario?

There's `--kernel-full` option.   But I don't think it's a good example to test the agent operation since the kernel tracing is handled separately.

uftrace.c
namhyung: Mark as deprecated, please.

uftrace.c
namhyung: Interesting, can you please file an issue separately?

namhyung:
> One more minor request is that we better print something in debug level 1 when agent receives a request whether it's --trace on or off.

Agreed.  It looks like a global state change and not repeated frequently.

uftrace.h
clementguidi: I'm thinking about discussing this in a separate issue, is it ok?

namhyung: Sure!

doc/ko/uftrace-live.md
namhyung, line 1: You don't need to update Korean man pages (unless you want). :)

namhyung (commented):
Please break the first commit because it does many things together.  I think you can leave mechanical changes together but adding new code or fixing a bug should be separated.

namhyung:
Can you please rebase the patchset?

cmds/live.c
```diff
@@ -56,7 +56,7 @@ static bool match_replay_triggers(const char *trigger)
 static void reset_live_opts(struct uftrace_opts *opts)
 {
 	/* this is needed to set display_depth at replay */
-	live_disabled = opts->disabled;
+	live_disabled = !(opts->trace == TRACE_STATE_ON);
```
namhyung, line 59: Wouldn't it be `live_disabled = (opts->trace == TRACE_STATE_OFF);` ?

namhyung (commented):
It'd be nice to have a test for this.

cmds/live.c
clementguidi: Yes, so the initialized `TRACE_STATE_NONE` defaults to ON for this function. I'm fixing it.

clementguidi:
I added a test. I had to refactor the `s-agent.c` test target so it has a loop.

If you accept this PR, it will be easier to merge it before #1644, #1645 and others, because conflicts will arise in tests names. Also I'm refactoring `s-agent.c` again when adding new features.

tests/t281_agent_trace_toggle.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
```
namhyung, line 8: It would be nice if you add some comment what this test expects.  IIUC the program calls the `func()` three times but the trace disabled for the second one so the result should be 2.

cmds/live.c
```diff
@@ -340,9 +340,13 @@ static int forward_options(struct uftrace_opts *opts)
 	if (capabilities < 0)
 		goto close;
 
-	/* FIXME Forward user options and set status */
-	if (0)
-		status = forward_option(sfd, capabilities, 0, NULL, 0);
+	if (opts->trace != TRACE_STATE_NONE) {
+		int trace = opts->trace == TRACE_STATE_ON;
```
namhyung, line 344: I think it's better to add a parentheses.

doc/ko/uftrace-live.md
```diff
@@ -65,7 +65,7 @@ uftrace [live] [*options*] COMMAND [*command-options*]
 :   타입(TYPE)으로 일치하는 패턴을 보여준다. 가능한 형태는 `regex`와 `glob`이다.
     기본 설정은 `regex`이다.
 
-\--disable
+\--disable <!-- TODO Deprecate and use --trace -->
```
namhyung, line 68: Not relevant here.  But the commit message can just start with `doc:`.

namhyung (approved):
LGTM.  But I have only a few nitpicks.

clementguidi:
Thank you for your time and for your feedback!

FYI I dropped the changes to changes to the Korean documentation.

PR #1644, opened 2023-03-03
mcount: Update time and depth filters at runtime

clementguidi:
Hi Namhyung and Honggyu,

Duplicate of #1588
I am sorry I had to open a new PR for these features. I don't have access to the original repository and GitHub won't let me change the remote.

I reused the code and added tests.

With this PR we can use the agent to update the maximum call depth and the time threshold.
```
$ uftrace -p <pid> -t <threshold>
$ uftrace -p <pid> -D <depth>
```

The new values are applied to every function call made after the update (not retroactive).

PR #1645, opened 2023-03-06
Control name filters at runtime

clementguidi:
These commits give the user full control over tracing filters at runtime. It
applies to the following filters:
 - opt-in filters set by `-F/--filter`
 - opt-out filters set by `-N/--notrace`
 - caller filters set by `-C/--caller-filter`

It allows the user to control the target process from the client as follows:
 - add or update an opt-in name filter
   - `$ uftrace -p PID -F func`
 - add or update an opt-out name filter
   - `$ uftrace -p PID -N func`
 - remove an opt-in or opt-out name filter
   - `$ uftrace -p PID -F func@clear`
 - add a caller filter
   - `$ uftrace -p PID -C func`
 - delete a caller filter
   - `$ uftrace -p PID -C func@clear`
 - set the pattern matching type
   - `$ uftrace -p PID --match <simple|regex|glob>`

 We also introduce an internal agent command `AGENT_CAPABILITIES`. It prompts
 the agent to indicate what features it supports.

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -360,6 +360,11 @@ static int forward_options(struct uftrace_opts *opts)
 			goto close;
 	}
 
+	status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN, &opts->patt_type,
+				sizeof(opts->patt_type));
+	if (status < 0)
+		goto close;
+
 close:
 	status_close = agent_message_send(sfd, UFTRACE_MSG_AGENT_CLOSE, NULL, 0);
 	if (status_close == 0) {
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -360,7 +360,7 @@ static int forward_options(struct uftrace_opts *opts)
 			goto close;
 	}
 
-	if (opts->filter) { /* provide a pattern type for options that need it */
+	if (opts->filter || opts->caller) { /* provide a pattern type for options that need it */
 		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN,
 					&opts->patt_type, sizeof(opts->patt_type));
 		if (status < 0)
@@ -370,6 +370,13 @@ static int forward_options(struct uftrace_opts *opts)
 	if (opts->filter) {
 		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_FILTER, opts->filter,
 					strlen(opts->filter) + 1);
+		if (status < 0)
+			goto close;
+	}
+
+	if (opts->caller) {
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_CALLER, opts->caller,
+					strlen(opts->caller) + 1);
 		if (status < 0)
 			goto close;
 	}
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -360,10 +360,19 @@ static int forward_options(struct uftrace_opts *opts)
 			goto close;
 	}
 
-	status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN, &opts->patt_type,
-				sizeof(opts->patt_type));
-	if (status < 0)
-		goto close;
+	if (opts->filter) { /* provide a pattern type for options that need it */
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN,
+					&opts->patt_type, sizeof(opts->patt_type));
+		if (status < 0)
+			goto close;
+	}
+
+	if (opts->filter) {
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_FILTER, opts->filter,
+					strlen(opts->filter) + 1);
+		if (status < 0)
+			goto close;
+	}
 
 close:
 	status_close = agent_message_send(sfd, UFTRACE_MSG_AGENT_CLOSE, NULL, 0);
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -86,20 +86,11 @@ static struct uftrace_filter_setting mcount_filter_setting = {
 /* boolean flag to turn on/off recording */
 static bool __maybe_unused mcount_enabled = true;
 
-/* count of registered opt-in filters (-F) */
-static int __maybe_unused mcount_filter_count;
-
-/* count of registered opt-in location filters (-L) */
-static int __maybe_unused mcount_loc_count;
-
-/* tree of trigger actions */
-static struct rb_root __maybe_unused *mcount_triggers;
+/* triggers definition and counters */
+static struct mcount_triggers_info __maybe_unused *mcount_triggers;
 
 /* bitmask of active watch points */
 static unsigned long __maybe_unused mcount_watchpoints;
-
-/* count of registered caller filters */
-static int __maybe_unused mcount_caller_count;
 
 /* address of function will be called when a function returns */
 unsigned long mcount_return_fn;
@@ -412,22 +403,21 @@ static void mcount_filter_init(struct uftrace_filter_setting *filter_setting, bo
 	}
 
 	mcount_triggers = xmalloc(sizeof(*mcount_triggers));
-	*mcount_triggers = RB_ROOT;
-	uftrace_setup_filter(filter_str, &mcount_sym_info, mcount_triggers, &mcount_filter_count,
-			     filter_setting);
-	uftrace_setup_trigger(trigger_str, &mcount_sym_info, mcount_triggers, &mcount_filter_count,
-			      filter_setting);
+	memset(mcount_triggers, 0, sizeof(*mcount_triggers));
+	mcount_triggers->root = RB_ROOT;
+	uftrace_setup_filter(filter_str, &mcount_sym_info, mcount_triggers, filter_setting);
+	uftrace_setup_trigger(trigger_str, &mcount_sym_info, mcount_triggers, filter_setting);
 	uftrace_setup_argument(argument_str, &mcount_sym_info, mcount_triggers, filter_setting);
 	uftrace_setup_retval(retval_str, &mcount_sym_info, mcount_triggers, filter_setting);
 
 	if (needs_debug_info) {
 		uftrace_setup_loc_filter(loc_str, &mcount_sym_info, mcount_triggers,
-					 &mcount_loc_count, filter_setting);
+					 filter_setting);
 	}
 
 	if (caller_str) {
 		uftrace_setup_caller_filter(caller_str, &mcount_sym_info, mcount_triggers,
-					    &mcount_caller_count, filter_setting);
+					    filter_setting);
 	}
 
 	if (autoargs_str) {
@@ -470,7 +460,7 @@ static void mcount_filter_release(struct mcount_thread_data *mtdp)
 
 static void mcount_filter_finish(void)
 {
-	uftrace_cleanup_filter(mcount_triggers);
+	uftrace_cleanup_filter(&mcount_triggers->root);
 	free(mcount_triggers);
 	finish_auto_args();
 
@@ -854,7 +844,7 @@ extern void *get_argbuf(struct mcount_thread_data *, struct mcount_ret_stack *);
  */
 static inline enum filter_mode mcount_get_filter_mode()
 {
-	return mcount_filter_count > 0 ? FILTER_MODE_IN : FILTER_MODE_OUT;
+	return mcount_triggers->filter_count > 0 ? FILTER_MODE_IN : FILTER_MODE_OUT;
 }
 
 /**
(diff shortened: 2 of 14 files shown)


cmds/live.c
```diff
@@ -340,9 +340,19 @@ static int forward_options(struct uftrace_opts *opts)
 	if (capabilities < 0)
 		goto close;
 
-	status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN, &opts->patt_type, sizeof(opts->patt_type));
-	if (status < 0)
-		goto close;
+	if (opts->filter) { /* provide a pattern type for options that need it */
```
namhyung: `s/filter/patt_type/`?

libmcount/mcount.c
```diff
@@ -1785,6 +1785,20 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
 }
 
+/**
+ * agent_setup_filter - update the registered filters from the agent
+ * @filter_str - filters to add or remove
+ */
+static void agent_setup_filter(char *filter_str)
+{
+	if (agent_guard_filter()) {
+		usleep(100); /* grace period */
```
namhyung, line 1808: I'm not sure if it's enough, especially on large machines.

libmcount/mcount.c
```diff
@@ -870,7 +890,10 @@ enum filter_result mcount_entry_filter_check(struct mcount_thread_data *mtdp, un
 	if (mtdp->filter.out_count > 0)
 		return FILTER_OUT;
 
-	uftrace_match_filter(child, &mcount_triggers, tr);
+	if (!agent_run || agent_guard_filter()) {
```
namhyung, line 906: I'm not sure if it's really safe.  Even with the full memory barrier, isn't it possible for concurrent readers and the agent can get the writer marker?

Also, I don't like the fact that it might miss the whole filters when agent makes the change.  Can we do it better?

namhyung (commented):
Thanks for your work!

cmds/live.c
clementguidi: No, I mean to send the pattern type whenever the user uses a filter (or caller filter, trigger, args in later commits). In other words don't send the pattern type if the agent doesn't need it.

libmcount/mcount.c
clementguidi: I agree this is not safe enough. There are a few ways we can tackle this.

### Disable tracing when agent is working

Our view on this is that if you are interacting with the agent to tweak tracing parameters, you do this for future events, because chances are the current calls are already over when the agent gets the instructions. So you may well skip instrumentation while the agent makes changes.

We can have the agent disable tracing by itself. This requires synchronizing.

We can have the user disabling tracing, and the agent would fail to apply filters if tracing is enabled. Without locks, this comes down to a grace period.

### Use read/write locks

We can introduce read/write locks, and skip instrumentation when data is read-protected. This requires a read lock in the fast path, which comes with an unwanted performance impact.

As discussed during our meeting in January, we could implement this solution in a separate libmcount build, e.g. `libmcount-sync.so`.

### Else

As for lockless algorithms, I don't think that would be simple to do.

cmds/live.c
namhyung: I see, thanks for the explanation.

libmcount/mcount.c
namhyung: I'm not happy with the both solutions.  Maybe we could have a simple RCU style algorithm to atomically update the whole filters and free the old copy a few seconds later.  Thoughts?

clementguidi: OK. I thought about RCU and we should give it a try. Having a RCU rbtree with per-node RCU is out of reach IMO (quiet complex, and the code I can find is old. See [1] and [2]). Instead we may copy the whole tree. I'll try that.

[1] https://lists.lttng.org/pipermail/lttng-dev/2017-December/027640.html
[2] https://github.com/philip-w-howard/RP-Red-Black-Tree

clementguidi: This is out of topic, but I'm curious: why use rbtrees in the first place? They better perform insertion/deletion than lookup, which is the opposite of what libmcount does. Is this from Ftrace?

namhyung: Yeah, I also meant replacing the whole tree.

The rbtree was needed because it needs to check the address range.  Since the location of `mcount` call varies among functions, it couldn't check a specific offset from the function start.  Maybe we can reconsider a hash table if it only supports dynamic tracing because it can control the location.

clementguidi:
I rebased on #1678.

I need your view on this: I replaced the flags for filtering mode and location filtering mode to counters (count the registered filters instead of just IN/OUT flags). Those are global variables, and match the rbtree structure. Since we use RCU, we need to have a set of counters (same if we keep flags) to match the RCU copy and one for the original. **I want to have a structure with both the rbtree and the counters, and handle it with RCU** instead of just the tree. What do you think? Something as follows:

```
struct mcount_triggers_info {
    struct rb_root triggers;
    int filter_count;
    int loc_count;
    int caller_count;
}
```

Note: Tests fail on optimization level 2 and 3 because two functions in the test program get merged into a single one, I'll fix it.

utils/filter.c
```diff
@@ -45,8 +45,10 @@ static void print_trigger(struct uftrace_trigger *tr)
 	if (tr->flags & TRIGGER_FL_FILTER) {
 		if (tr->fmode == FILTER_MODE_IN)
 			pr_dbg("\ttrigger: filter IN\n");
-		else
+		else if (tr->fmode == FILTER_MODE_OUT)
 			pr_dbg("\ttrigger: filter OUT\n");
+		else
+			pr_dbg("\ttrigger: filter NONE\n");
```
namhyung, line 51: I think it should `tr->fmode` set when it has `TRIGGER_FL_FILTER` flag.

utils/filter.c
```diff
@@ -249,6 +251,11 @@ static void add_arg_spec(struct list_head *arg_list, struct uftrace_arg_spec *ar
  */
 void update_trigger(struct uftrace_filter *filter, struct uftrace_trigger *tr, bool exact_match)
 {
+	if (tr->flags & TRIGGER_FL_CLEAR) {
+		filter->trigger.flags &= ~tr->flags;
```
namhyung, line 257: Hmm.. what's the intended use case?  I think we can simply do `filter->triggers.flags = 0`, right?

namhyung (commented):
I'm fine with having a struct to have all filter related settings.  By the way, why not renaming the test case to start from 281?

cmds/live.c
```diff
@@ -340,10 +340,19 @@ static int forward_options(struct uftrace_opts *opts)
...
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN,
+					&opts->patt_type, sizeof(opts->patt_type));
+		if (status < 0)
+			goto close;
+	}
+
+	if (opts->filter) {
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_FILTER, opts->filter,
+					strlen(opts->filter));
```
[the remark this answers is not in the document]
clementguidi: Yes that makes more sense.

utils/filter.h
```diff
@@ -40,6 +40,15 @@ enum trigger_flag {
...
 
+/**
+ * filter_mode - opt-in or opt-out mode
+ *
+ * When in opt-in mode, only trace functions that have an explicit filter. When
+ * in opt-out mode, trace all but explicitly excluded functions.
+
+ * @FILTER_MODE_NONE is neutral and is only used for initialization in the
+ * location filter.
+ */
```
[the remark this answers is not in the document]
clementguidi: No problem!

utils/filter.c
clementguidi: Sorry, what do you mean? That the `NONE` value isn't valid when `TRIGGER_FL_FILTER` is set?

utils/filter.c
clementguidi: I only want to remove a given filter type. E.g. if you have a filter and an trigger set for a symbol, you can remove either the filter or the trigger (triggers are not implemented yet in the agent). Setting the flags to `0` clears all filters/triggers/args.

The following example would preserve the depth trigger on `func`.
```
$ uftrace record -F func -T func@depth=1
$ uftrace -p <pid> -F func@clear
```

clementguidi:
> why not renaming the test case to start from 281?

I'm using tests 281 and 282 in #1644, and I expected said PR to go through first.

clementguidi:
For now I'm keeping the `mcount_triggers` name but for the new structure (rbtree + counters). And I suggest we discuss naming separately, because it requires many modifications, especially in the code doc. What do you think?

```
struct mcount_triggers_info {
    struct rb_root triggers; /* the former 'mcount_triggers' rbtree */
    int filter_count; /* use counters instead of boolean mode */
    int caller_count;
    int loc_count;

struct mcount_triggers_info *mcount_triggers; /* RCU triggers data */
```

Regarding the nomenclature, the name `mcount_triggers` is ambiguous, as it refers to a rbtree with `struct uftrace_filter` nodes. And `struct uftrace_filter` itself contains a `struct uftrace_trigger` field. Eventually, the naming is not consistent, as in other places (e.g. `fstack.c`) the rbtree is named `filters`.

utils/filter.c
namhyung: Right, it should not have `NONE` together with `TRIGGER_FL_FILTER`.

utils/filter.c
namhyung: Hmm.. ok.  Then `-T func@clear` would clear them all.

namhyung:
> I'm using tests 281 and 282 in https://github.com/namhyung/uftrace/pull/1644, and I expected said PR to go through first.

Ah, ok.  Let's do it first then :)

namhyung:
About the naming, I agree it's inconsistent.  It started with filters only, but later the triggers were added.  Now triggers are superset of filters with additional features.  To minimize the work, I think it's good to keep the `mcount_triggers`.  How about changing the name of tree as `root`?

clementguidi:
> How about changing the name of tree as `root`?

Yes sounds good.

namhyung (commented):
Can you please rebase this series?

I think we need to reconsider the filter mode/count change since it replaces the whole filter tree at once and no need to keep the number of filters.

clementguidi:
> I think we need to reconsider the filter mode/count change since it replaces the whole filter tree at once and no need to keep the number of filters.

I keep track of the number of filters to change the default filtering behavior on function entry. As long as there's a `-F` filter registered, functions are traced by default. I copy the counters so I don't have to parse the tree again. I don't understand how we can get rid of the mode/count?

libmcount/mcount.c
```diff
@@ -1931,6 +1932,14 @@ static int agent_apply_option(int opt, void *value, size_t size, struct rb_root
 			pr_dbg3("dynamic time threshold unchanged\n");
 		break;
 
+	case UFTRACE_AGENT_OPT_PATTERN:
+		opts.patt_type = *((int *)value);
+		if (opts.patt_type != mcount_filter_setting.ptype) {
+			mcount_filter_setting.ptype = opts.patt_type;
+			pr_dbg4("use pattern type %#x\n", opts.patt_type);
```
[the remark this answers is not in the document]
clementguidi: Yes, I didn't pay attention while rebasing.

clementguidi:
The counters are incremented when adding a filter, and decremented when removing one from the tree. As we copy the counters together with the tree, we don't need to parse the tree to get the count. Is that OK?

utils/filter.c
```diff
@@ -808,6 +822,10 @@ static const struct trigger_action_parser actions[] = {
 		"auto-args",
 		parse_auto_args_action,
 	},
+	{
+		"clear",
+		parse_clear_action,
+	},
```
namhyung, line 846: Can you please update the documentation?

namhyung (commented):
Looks good now.  I think it's ok to apply delta after changing the filters.

Btw, can you please add "agent" section in the uftrace live man page to explain how to use it with an example?

clementguidi:
> Btw, can you please add "agent" section in the uftrace live man page to explain how to use it with an example?

Yes, I'm sending another PR.

utils/filter.c
clementguidi: I added a few lines in `uftrace-live.md` and `uftrace-record.md` to explain that the `@clear` suffix exists. I plan to add a reference to the agent section when I create it, where I will show examples. Is it OK for now? Or do you want to have examples in the doc in this PR?

namhyung: You can do it in a separate PR.  Thanks!

tests/t284_agent_filter.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
```
namhyung, line 8: Again, please add a brief explanation for this test.

tests/t285_agent_caller_filter.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
```
namhyung, line 8: Ditto.

namhyung (commented):
(Probably) last nitpick.  Please add a comment in the tests to describe what to check.

clementguidi:
> Please add a comment in the tests to describe what to check.

I'm not sure how detailed it needs to be, as the client commands in the test script are quite intuitive.

namhyung (approved):
Simple description like you added should be fine.  LGTM, thanks!

PR #1646, opened 2023-03-06
Control triggers at runtime

clementguidi:
These commits give the user full control over triggers at runtime. It applies to
all actions.

It allows the user to control the target process from the client as follows:
 - add or update trigger actions
   - `$ uftrace -p PID -T func@act1[,act2]`
 - delete one or more trigger actions
   - `$ uftrace -p PID -T func@act1[,act2],clear`
 - delete all trigger actions for a symbol
   - `$ uftrace -p PID -T func@clear`

Trigger parameters can be omitted when deleting actions. For example:
- `$ uftrace -p PID -T func@depth,clear`
- `$ uftrace -p PID -T func@time,read,clear`

We also fix the `--trigger` option which was actually not implemented.

Based on #1645.

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -360,7 +360,8 @@ static int forward_options(struct uftrace_opts *opts)
 			goto close;
 	}
 
-	if (opts->filter || opts->caller) { /* provide a pattern type for options that need it */
+	/* provide a pattern type for options that need it */
+	if (opts->filter || opts->caller || opts->trigger) {
 		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_PATTERN,
 					&opts->patt_type, sizeof(opts->patt_type));
 		if (status < 0)
@@ -377,6 +378,13 @@ static int forward_options(struct uftrace_opts *opts)
 	if (opts->caller) {
 		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_CALLER, opts->caller,
 					strlen(opts->caller) + 1);
+		if (status < 0)
+			goto close;
+	}
+
+	if (opts->trigger) {
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_TRIGGER, opts->trigger,
+					strlen(opts->trigger) + 1);
 		if (status < 0)
 			goto close;
 	}
--- a/tests/t286_agent_trigger.py
+++ b/tests/t286_agent_trigger.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    """Add and remove triggers at runtime from the client.
+
+    The target runs a loop and waits for input from this script. The agent is
+    enabled. Calls to a(), b() and c() include a sleep period so we can test time
+    filters.
+
+    The client adds then removes a depth filter. Then adds a time filter. And
+    adds then removes an opt-out filter. """
+
+    def __init__(self):
+        TestBase.__init__(self, 'agent', """
+# DURATION     TID     FUNCTION
+            [  6999] | main() {
+            [  6999] |   func() {
+            [  6999] |     a() {
+            [  6999] |       b() {
+   1.056 ms [  6999] |         c();
+   2.113 ms [  6999] |       } /* b */
+   3.175 ms [  6999] |     } /* a */
+   3.175 ms [  6999] |   } /* func */
+            [  6999] |   func() {
+            [  6999] |     a() {
+   2.110 ms [  6999] |       b();
+   3.181 ms [  6999] |     } /* a */
+   3.182 ms [  6999] |   } /* func */
+            [  6999] |   func() {
+   3.178 ms [  6999] |     a();
+   3.179 ms [  6999] |   } /* func */
+   3.177 ms [  6999] |   func();
+            [  6999] |   func() {
+            [  6999] |     a() {
+            [  6999] |       b() {
+   1.057 ms [  6999] |         c();
+   2.119 ms [  6999] |       } /* b */
+   3.185 ms [  6999] |     } /* a */
+   3.185 ms [  6999] |   } /* func */
+  55.835 ms [  6999] | } /* main */
+""")
+
+
+    def client_send_command(self, pid, option):
+        self.subcmd = 'live'
+        self.option = '-p %d %s' % (pid, option)
+        self.exearg = ''
+        client_cmd = self.runcmd()
+        self.pr_debug('prerun command: ' + client_cmd)
+        client_p = sp.run(client_cmd.split())
+        return client_p.returncode
+
+    def prerun(self, timeout):
+        self.subcmd = 'record'
+        self.option  = '--keep-pid'
+        self.option += ' --no-libcall'
+        self.option += ' --agent'
+        self.exearg  = 't-' + self.name
+        self.exearg += ' --delay'
+        record_cmd  = self.runcmd()
+        self.pr_debug("prerun command: " + record_cmd)
+        # bufsize=0 to write characters one by one
+        record_p = sp.Popen(record_cmd.split(), stdin=sp.PIPE, stderr=sp.PIPE, bufsize=0)
+
+        sleep(.05)              # time for the agent to start
+
+        if self.client_send_command(record_p.pid, '-T func@depth=3') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        if self.client_send_command(record_p.pid, '-T func@clear=depth,time=3ms') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        if self.client_send_command(record_p.pid, '-T a@notrace') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        if self.client_send_command(record_p.pid, '-T a@clear -T func@clear') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        record_p.stdin.close()
+        record_p.wait()
+
+        return TestBase.TEST_SUCCESS
+
+    def setup(self):
+        self.subcmd = 'replay'
+        self.option = ''
+        self.exearg = ''
--- a/utils/filter.c
+++ b/utils/filter.c
@@ -747,7 +747,58 @@ static int parse_hide_action(char *action, struct uftrace_trigger *tr,
 static int parse_clear_action(char *action, struct uftrace_trigger *tr,
 			      struct uftrace_filter_setting *setting)
 {
+	struct strv acts = STRV_INIT;
+	char *pos = NULL;
+	int j;
+
 	tr->flags |= TRIGGER_FL_CLEAR;
+
+	if (strlen(action) == 5) {
+		tr->clear_flags = ~0;
+		return 0;
+	}
+
+	if (action[5] != '=') {
+		pr_use("skipping invalid action: %s\n", action);
+		return -1;
+	}
+
+	/* action = "clear=act1+act2+..." */
+	pos = action + 6;
+	strv_split(&acts, pos, "+");
+	strv_for_each(&acts, pos, j) {
+		if (!strcmp(pos, "arg") || !strcmp(pos, "fparg"))
+			tr->clear_flags |= TRIGGER_FL_ARGUMENT;
+		else if (!strcmp(pos, "retval"))
+			tr->clear_flags |= TRIGGER_FL_RETVAL;
+		else if (!strcmp(pos, "filter") || !strcmp(pos, "notrace"))
+			tr->clear_flags |= TRIGGER_FL_FILTER;
+		else if (!strcmp(pos, "depth"))
+			tr->clear_flags |= TRIGGER_FL_DEPTH;
+		else if (!strcmp(pos, "time"))
+			tr->clear_flags |= TRIGGER_FL_TIME_FILTER;
+		else if (!strcmp(pos, "size"))
+			tr->clear_flags |= TRIGGER_FL_SIZE_FILTER;
+		else if (!strcmp(pos, "hide"))
+			tr->clear_flags |= TRIGGER_FL_HIDE;
+		else if (!strcmp(pos, "trace"))
+			tr->clear_flags |= TRIGGER_FL_TRACE | TRIGGER_FL_TRACE_ON |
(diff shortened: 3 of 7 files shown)


tests/t286_agent_trigger.py
```diff
@@ -0,0 +1,97 @@
...
+        self.pr_debug('prerun command: ' + client_cmd)
+        client_p = sp.run(client_cmd.split())
+        return client_p.returncode
+
+    def prerun(self, timeout):
+        self.subcmd = 'record'
+        self.option  = '--keep-pid'
+        self.option += ' --no-libcall'
+        self.option += ' --agent'
+        self.exearg  = f't-{self.name}'
```
namhyung, line 60: Please avoid the f-string syntax.

tests/t286_agent_trigger.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python
+
+import subprocess as sp
+from time import sleep
+
+from runtest import TestBase
+
+class TestCase(TestBase):
```
namhyung, line 8: A brief description?

tests/t286_agent_trigger.py
```diff
@@ -0,0 +1,97 @@
...
+        # bufsize=0 to write characters one by one
+        record_p = sp.Popen(record_cmd.split(), stdin=sp.PIPE, stderr=sp.PIPE, bufsize=0)
+
+        sleep(.05)              # time for the agent to start
+
+        if self.client_send_command(record_p.pid, '-T a@depth=2') != 0:
+            return TestBase.TEST_NONZERO_RETURN
+        record_p.stdin.write(b'0')
+
+        if self.client_send_command(record_p.pid, '-T a@depth,clear') != 0:
```
namhyung, line 73: Just curious, `-T a@clear,depth` should also work, right?

namhyung (commented):
Thanks for the update.  It might be in v0.14 if it finishes in time.

clementguidi:
> It might be in v0.14 if it finishes in time.

When do you plan to release v0.14?

cmds/live.c
```diff
@@ -381,6 +382,13 @@ static int forward_options(struct uftrace_opts *opts)
 			goto close;
 	}
 
+	if (opts->trigger) {
+		status = forward_option(sfd, capabilities, UFTRACE_AGENT_OPT_TRIGGER, opts->trigger,
+					strlen(opts->trigger));
```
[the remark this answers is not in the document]
clementguidi: Yes, sorry, rebase mistake.

utils/filter.c
```diff
@@ -252,7 +252,13 @@ static void add_arg_spec(struct list_head *arg_list, struct uftrace_arg_spec *ar
 void update_trigger(struct uftrace_filter *filter, struct uftrace_trigger *tr, bool exact_match)
 {
 	if (tr->flags & TRIGGER_FL_CLEAR) {
-		filter->trigger.flags &= ~tr->flags;
+		/* No new flag other than CLEAR means delete all registered triggers */
+		if (tr->flags == TRIGGER_FL_CLEAR) {
+			tr->flags |= filter->trigger.flags; /* Remember what we remove */
```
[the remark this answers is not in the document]
namhyung: Oh, I meant if it needs to remember  the old flags.  I'm ok with the syntax that removes all actions at once.

But I'm a bit skeptical to change the existing trigger syntax.  As I said in other comment, `-T func@depth` can be confusing and/or resulting in misbehavior.  How about making `@clear` can take an argument.

* `-T func@clear` will remove all actions
* `-T func@clear=read+depth` will remove read and depth trigger actions only

namhyung:
> When do you plan to release v0.14?

No specific date, but I hope I can do it this weekend. :)

utils/filter.c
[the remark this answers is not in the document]
clementguidi: > I meant if it needs to remember the old flags.

OK my bad. This is useful when removing filters, to keep count of them.

> `-T func@depth` can be confusing and/or resulting in misbehavior. How about making `@clear` can take an argument.

I'm OK with this syntax, I'll use it. Otherwise I would have added a check at the end of `setup_trigger_action()` to invalidate e.g. the `depth` flag if no value is given, thus making `-T func@depth` illegal. Wdyt?

utils/filter.c
```diff
@@ -588,8 +594,10 @@ static int parse_float_argument_spec(char *str, struct uftrace_trigger *tr,
 static int parse_depth_action(char *action, struct uftrace_trigger *tr,
 			      struct uftrace_filter_setting *setting)
 {
+	/* action = "depth[=]..." */
 	tr->flags |= TRIGGER_FL_DEPTH;
-	tr->depth = strtoul(action + 6, NULL, 10);
+	if (action[5] == '=')
+		tr->depth = strtoul(action + 6, NULL, 10);
```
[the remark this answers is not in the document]
clementguidi: OK, will add. Yes, that is so `depth` doesn't require arguments.

tests/t286_agent_trigger.py
clementguidi: Yes, the parsers are independent. As long as `clear` is found, it reverses the flags for other actions.

utils/filter.c
```diff
@@ -252,7 +252,13 @@ static void add_arg_spec(struct list_head *arg_list, struct uftrace_arg_spec *ar
...
 {
 	if (tr->flags & TRIGGER_FL_CLEAR) {
-		filter->trigger.flags &= ~tr->flags;
+		/* No new flag other than CLEAR means delete all registered triggers */
+		if (tr->flags == TRIGGER_FL_CLEAR) {
+			tr->flags |= filter->trigger.flags; /* Remember what we remove */
+			filter->trigger.flags = 0;
+		}
+		else
+			filter->trigger.flags &= ~tr->flags;
```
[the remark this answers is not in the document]
clementguidi: I added a `clear_flags` field.

utils/filter.c
```diff
@@ -38,72 +38,131 @@ static void snprintf_trigger_read(char *buf, size_t len, enum trigger_read_type
...
-	if (tr->flags & TRIGGER_FL_CLEAR)
-		pr_dbg("\ttrigger: clear\n");
-	if (tr->flags & TRIGGER_FL_DEPTH)
-		pr_dbg("\ttrigger: depth %d\n", tr->depth);
+	if (tr->flags & TRIGGER_FL_DEPTH) {
+		if (tr->clear_flags & TRIGGER_FL_DEPTH)
+			pr_dbg("\tclear trigger: depth\n");
+		else
+			pr_dbg("\ttrigger: depth %d\n", tr->depth);
+	}
```
namhyung, line 46: I'm not sure if we need this whole changes.  Can we simply do like below and leave other code as is?

```
if (tr->flags & TRIGGER_FL_CLEAR)
    pr_dbg("\ttriggers: clear=%x\n", tr->clear_flags);
```

utils/filter.c
```diff
@@ -747,7 +805,10 @@ static int parse_hide_action(char *action, struct uftrace_trigger *tr,
 static int parse_clear_action(char *action, struct uftrace_trigger *tr,
 			      struct uftrace_filter_setting *setting)
 {
-	tr->flags |= TRIGGER_FL_CLEAR;
+	/* 'clear' suffix for any option other than -T/--trigger */
+	if (tr->flags)
+		tr->clear_flags = tr->flags;
```
namhyung, line 814: I'm afraid it can have other flags given together with the `@clear` action.  If you want to set the `orig_flags` only, then I think you need to do it in `setup_trigger_action`

utils/filter.h
```diff
@@ -38,7 +38,6 @@ enum trigger_flag {
 	TRIGGER_FL_HIDE = (1U << 17),
 	TRIGGER_FL_LOC = (1U << 18),
 	TRIGGER_FL_SIZE_FILTER = (1U << 19),
-	TRIGGER_FL_CLEAR = (1U << 20), /* Reverse other flags when set */
```
namhyung, line 41: And let's keep it.

utils/filter.c
```diff
@@ -940,6 +991,11 @@ int setup_trigger_action(char *str, struct uftrace_trigger *tr, char **module,
 
 			break;
 		}
+		if (!strncasecmp(pos, "clear", 5)) {
+			/* preserve orig flags overwritten by 'clear parser' */
+			tr->flags |= orig_flags;
+			break; /* 'clear' cannot be used along with other actions */
```
namhyung, line 997: No, this is not I wanted.

utils/filter.c
```diff
@@ -940,6 +991,11 @@ int setup_trigger_action(char *str, struct uftrace_trigger *tr, char **module,
 
 			break;
 		}
+		if (!strncasecmp(pos, "clear", 5)) {
+			/* preserve orig flags overwritten by 'clear parser' */
+			tr->flags |= orig_flags;
```
namhyung, line 996: I don't want to touch `tr->flags` here.  How about this and move it out of the loop?

```c
strv_for_each(&acts, pos, j) {
    ...
}
if (tr->flags & TRIGGER_FL_CLEAR) {
    if (orig_flags && tr->clear_flags == ~0)
        tr->clear_flags = orig_flags;
}
```

tests/t286_agent_trigger.py
```diff
@@ -0,0 +1,94 @@
...
+        self.exearg  = 't-' + self.name
+        self.exearg += ' --delay'
+        record_cmd  = self.runcmd()
+        self.pr_debug("prerun command: " + record_cmd)
+        # bufsize=0 to write characters one by one
+        record_p = sp.Popen(record_cmd.split(), stdin=sp.PIPE, stderr=sp.PIPE, bufsize=0)
+
+        sleep(.05)              # time for the agent to start
+
+        if self.client_send_command(record_p.pid, '-T func@depth=2') != 0:
```
namhyung, line 70: Just cosmetic but it'd be a bit nicer if we can use `depth=3` here. :)

namhyung (commented):
Thanks for the update.

utils/filter.c
namhyung: Yeah, I think that's reasonable.  There's no point to use clear for the same action, it seems like a user error.  But anyway, the clear action is handled first and then we add the new flags.  So it should be fine.

namhyung:
I'm ok with `compat_flags`, thanks!

utils/filter.c
```diff
@@ -251,14 +251,13 @@ static void add_arg_spec(struct list_head *arg_list, struct uftrace_arg_spec *ar
...
 {
-	if (tr->flags & TRIGGER_FL_CLEAR) {
-		filter->trigger.flags &= ~tr->flags;
-		tr->fmode = filter->trigger.fmode; /* Read from tree before deleting */
-		return;
-	}
-
 	filter->trigger.flags |= tr->flags;
 
+	if (tr->clear_flags) {
```
namhyung, line 256: I think we can check `TRIGGER_FL_CLEAR` for consistency.

utils/filter.c
```diff
@@ -369,7 +368,7 @@ static int update_filter(struct rb_root *root, struct uftrace_filter *filter,
 			if (auto_ret)
 				update_trigger(iter, &auto_ret->trigger, exact_match);
 			tr->flags = orig_flags;
-			if (tr->flags & TRIGGER_FL_CLEAR)
+			if (tr->clear_flags)
```
namhyung, line 371: Ditto.

utils/filter.c
```diff
@@ -746,7 +746,54 @@ static int parse_hide_action(char *action, struct uftrace_trigger *tr,
 static int parse_clear_action(char *action, struct uftrace_trigger *tr,
 			      struct uftrace_filter_setting *setting)
 {
+	struct strv acts = STRV_INIT;
+	char *pos = NULL;
+	int j;
+
 	tr->flags |= TRIGGER_FL_CLEAR;
+
+	if (strlen(action) > 5) {
```
namhyung, line 755: I like early returns and reduced nestings like below.

```c
if (strlen(action) == 5) {
    tr->clear_flags = ~0;
    return 0;
}

if (action[5] != '=') {
    pr_use(...);
    return -1;
}

...
```

utils/filter.c
```diff
@@ -746,7 +746,54 @@ static int parse_hide_action(char *action, struct uftrace_trigger *tr,
...
+			else if (!strcmp(pos, "finish"))
+				tr->clear_flags |= TRIGGER_FL_FINISH;
+			else if (!strcmp(pos, "read"))
+				tr->clear_flags |= TRIGGER_FL_READ;
+			else if (!strcmp(pos, "color"))
+				tr->clear_flags |= TRIGGER_FL_COLOR;
+			else if (!strcmp(pos, "backtrace"))
+				tr->clear_flags |= TRIGGER_FL_BACKTRACE;
+			else if (!strcmp(pos, "recover"))
+				tr->clear_flags |= TRIGGER_FL_RECOVER;
```
namhyung, line 790: I think the commit message should mention that it can support clearing selected actions only.

utils/filter.c
```diff
@@ -888,8 +935,15 @@ int setup_trigger_action(char *str, struct uftrace_trigger *tr, char **module,
 		}
 	}
 	if (tr->flags & TRIGGER_FL_CLEAR) {
-		if (orig_flags && (tr->clear_flags == 0))
+		if (orig_flags)
+			/* '@clear' suffix for options other then -T/--trigger */
 			tr->clear_flags = orig_flags;
+		else {
+			if (tr->clear_flags == 0) /* 'clear' action has no argument */
+				tr->clear_flags = ~0; /* clear all */
```
namhyung, line 943: This can move to the parse code.

utils/filter.c
```diff
@@ -888,8 +935,15 @@ int setup_trigger_action(char *str, struct uftrace_trigger *tr, char **module,
 		}
 	}
 	if (tr->flags & TRIGGER_FL_CLEAR) {
-		if (orig_flags && (tr->clear_flags == 0))
+		if (orig_flags)
+			/* '@clear' suffix for options other then -T/--trigger */
 			tr->clear_flags = orig_flags;
```
namhyung, line 940: This part I thought something like `-F func@clear=depth`.  But it's an advanced usage and maybe better not support it.

clementguidi: OK I see, to me clearing a trigger action should only happen in a trigger option.

namhyung (approved):
The second last commit message looks squashed, please update.  Otherwise looks good!

clementguidi:
> The second last commit message looks squashed

Yes I missed that. Thank you again Namhyung!

namhyung:
Can you please rebase to the current master?

clementguidi:
Done. I'm updating #1647 now.

namhyung (approved):
LGTM, thanks!

PR #1647, opened 2023-03-06
Control arguments and return values at runtime

clementguidi:
These changes give the user full control over the argument and return value
collection at runtime.

We can use the `-A/--argument` option from the client to collect new argument
values in the trace, or change existing specifications. We can use the
`-R/--retval` option from the client to collect new return values in the trace,
or change existing specifications. And we can use the `-a/--auto-args` option
from the client to collect all arguments and return values.

Example:
```
$ uftrace --agent record my-program
$ uftrace -p $(pidof my-program) -A func@arg1
$ uftrace -p $(pidof my-program) -R func@retval
$ uftrace -p $(pidof my-program) -a
```

Based on #1646.


clementguidi:
Test 286 with with optimizations enabled, because it interferes with the auto-args. Do you suggest a way to deal with this?

clementguidi:
I'll stop stop on this feature for now. Changing arg/retval spec at runtime requires a more elaborate data format.

I plan to work on session control at runtime, i.e. initiating and finishing recording sessions without terminating the target process (e.g. if the target is a daemon). Then each session can have different arg/retval specs, which solves the problem in this particular use case (multiple sessions).

PR #1653, opened 2023-03-22
README.md: Update the introduction to uftrace

bernhardkaindl:
Update the introduction to uftrace in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,50 @@
 uftrace
 =======
 
-The uftrace tool is to trace and analyze execution of a program
-written in C/C++.  It was heavily inspired by the ftrace framework
-of the Linux kernel (especially function graph tracer) and supports
-userspace programs.  It supports various kind of commands and filters
-to help analysis of the program execution and performance.
+uftrace is a function call graph tracer for C, C++, Rust and Python programs.
+
+It hooks into the entry and exit of each function, recording timestamps
+as well as the function's arguments and return values. uftrace is capable
+of tracing both user and kernel functions, as well as library functions and
+system events providing an integrated execution flow in a single timeline.
+
+Initially, uftrace only supported function tracing with compiler support.
+However, it now allows users to trace function calls without recompilation
+by analyzing instructions in each function prologue and dynamically and
+selectively patching those instructions.
+
+Users can also write and run scripts for each function entry and exit using
+python/luajit APIs to create custom tools for their specific purposes.
+
+uftrace offers various filters to reduce the amount of trace data and provides
+visualization using Chrome trace viewer and flame graph or call-graph diagrams
+for graphviz and mermaid, allowing for a big picture view of the execution flow.
+
+It was heavily inspired by the ftrace framework of the Linux kernel and
+the name uftrace stems from the combination of user and ftrace.
+
+It can record data from:
+- User space functions by dynamically patching functions at runtime, or with code
+  compiled with with `-pg`, `-fpatchable-function-entry`, `-mfentry`, `-mnop-mcount`,
+  `-finstrument-functions`, or using clang with `-fxray-instrument`.
+- Library functions (through PLT hooking)
+- Kernel functions (using the ftrace framework in Linux kernel)
+- Kernel trace events affecting the program, e.g. scheduling events (which affect the
+  execution timing of the program) and records nanosecond-exact timestamps
+  (using perf_event and systemtap SDT)
+
+With the recorded data, it:
+- Can show colored and nested function call graphs.
+- Can show arguments and return values symbolically using libc function prototypes
+  and DWARF debug information.
+- Can apply filters to minimize the amount of trace data in both record and replay time.
+- Can extract metadata from traces. (e.g. system information on which the trace was taken)
+- Can generate symbol tables and memory maps of the traced program and library functions.
+- When the execution of multiple programs was recorded,
+  it can generate a task relationship tree (parent and children) from the trace.
+
+It supports many commands and filters such as filtering by function call
+duration for analysis of program execution and performance.
 
 ![uftrace-live-demo](doc/uftrace-live-demo.gif)
 
@@ -16,35 +55,73 @@ to help analysis of the program execution and performance.
  * Tutorial: https://github.com/namhyung/uftrace/wiki/Tutorial
  * Chat: https://gitter.im/uftrace/uftrace
  * Mailing list: [uftrace@googlegroups.com](https://groups.google.com/forum/#!forum/uftrace)
+ * Lightning talk: https://youtu.be/LNav5qvyK7I
 
 
 Features
 ========
 
-It traces each function in the executable and shows time duration.  It
-can also trace external library calls - but usually entry and exit are
-supported.  Optionally it's possible to trace other (nested) external
-library calls and/or internal function calls in the library call.
+uftrace traces each function in the executable and shows time durations.
 
-It can show detailed execution flow at function level, and report which
-function has the highest overhead.  And it also shows various information
-related the execution environment.
+Usually, for this to be possible, the program needs to be compiled with
+`-pg` or `-fpatchable-function-entry=5` (`=2` is enough on aarch64).
+With full dynamic tracing (`-P.`|`--patch=.`), uftrace works on all executables (as
+long they are not stripped, or symbol information is available from a separate file).
+
+uftrace hooks into the PLT in the given executable file to trace library calls and with
+(`-l`|`--nest-libcall`), it also hooks into the procedure linkage tables (PLTs) of shared
+libraries. The depth can be limited using `-D<num>`, where 1 is flat call tracing.
+
+Using (`-a`|`--auto-args`), uftrace automatically records arguments and
+return values of known functions. Without extra debug information, this
+includes the API functions of standard (C language or system) libraries.
+This can be combined with `-P.` or `-l`:
+For example, `-la` traces nested library calls, even in stripped executables.
+
+In addition, `-a` implies `--srcline` so it records the source line location info and
+it can be shown by `uftrace replay --srcline` and in `uftrace tui`. Users can directly
+open the editor at the source location as shown in https://uftrace.github.io/slide/#120.
+
+If debug information for the program (`gcc -g`) is available, `--auto-args`
+works even on functions inside of the compiled the user programs.
+
+In case argument information is not available, argument specifiations like
+(`-A udev_new@arg1/s`) can be passed on the command line or an options file.
+
+Example:
+```py
+$ uftrace record -la -A udev_new@arg1/s lsusb >/dev/null
+$ uftrace replay -f+module
+or simply:
+$ uftrace -la -A udev_new@arg1/s -f+module lsusb  # -f+module adds the module name
+# DURATION     TID        MODULE NAME   FUNCTION
+ 306.339 us [ 23561]            lsusb | setlocale(LC_TYPE, "") = "en_US.UTF-8";
+   1.163 us [ 23561]            lsusb | getopt_long(1, 0x7fff7175f6a8, "D:vtP:p:s:d:Vh") = -1;
+            [ 23561]            lsusb | udev_new("POSIXLY_CORRECT") {
+   0.406 us [ 23561] libudev.so.1.7.2 |   malloc(16) = 0x55e07277a7b0;
+   2.620 us [ 23561]            lsusb | } /* udev_new */
+            [ 23561]            lsusb | udev_hwdb_new() {
+   0.427 us [ 23561] libudev.so.1.7.2 |   calloc(1, 200) = 0x55e07277a7d0;
+   5.829 us [ 23561] libudev.so.1.7.2 |   fopen64("/etc/systemd/hwdb/hwdb.bin", "re") = 0;
+```
+
+Furthermore, it can show detailed execution flow at function level, and report
+which functions had the longest execution time.  It also shows information about
+the execution environment.
 
 You can setup filters to exclude or include specific functions when tracing.
-In addition, it can save and show function arguments and return value.
+In addition, function arguments and return values can be saved and shown later.
 
 It supports multi-process and/or multi-threaded applications.  With root
-privilege, it can also trace kernel functions as well( with `-k` option)
-if the system enables the function graph tracer in the kernel
-(`CONFIG_FUNCTION_GRAPH_TRACER=y`).
-
+privileges and if the kernel was built with `CONFIG_FUNCTION_GRAPH_TRACER=y`,
+kernel functions can be traced as well.
 
 How to build and install uftrace
 ================================
 
-On Linux distros, [misc/install-deps.sh](misc/install-deps.sh) installs required
-software(s) on your system.  Those are for optional advanced features but highly
-recommend to install them together.
+On Linux distros, [misc/install-deps.sh](misc/install-deps.sh) can be used to
+install required software(s) for building uftrace.  Those are for optional
+and advanced features, but are highly recommended.
 
     $ sudo misc/install-deps.sh
 
@@ -55,7 +132,7 @@ installed like following:
     $ make
     $ sudo make install
 
-For more advanced setup, please refer [INSTALL.md](INSTALL.md) file.
+For details about installation and dependencies, please refer to [INSTALL.md](INSTALL.md)
 
 
 How to use uftrace


README.md
```diff
@@ -21,30 +24,29 @@ to help analysis of the program execution and performance.
 Features
 ========
 
-It traces each function in the executable and shows time duration.  It
-can also trace external library calls - but usually entry and exit are
-supported.  Optionally it's possible to trace other (nested) external
+The tool traces each function in the executable and shows time durations.  It
+can also trace external library calls - but usually only entry and exit are
```
namhyung, line 28: I think this part can be improved further.  I meant it only records PLT calls in the main executable so that we can see total library call durations.

README.md
```diff
@@ -4,11 +4,14 @@
...
-of the Linux kernel (especially function graph tracer) and supports
-userspace programs.  It supports various kind of commands and filters
-to help analysis of the program execution and performance.
+uftrace traces, records and analyzes the execution of programs written in
+C, C++ and Rust. It was heavily inspired by the ftrace framework of the
+Linux kernel (especially the function graph tracer) and supports userspace
+programs, and with elevated privileges, also kernel function call graphs.
+
+It supports many commands and filters for analysis of program execution
+and performance. Recorded data can be visualized as flame graphs in Chrome:
```
namhyung, line 13: I don't think the chrome trace viewer shows the flame graph by default.  It's a timeline view but I think it also supports flame graphs (in .svg format).

namhyung (commented):
Thanks a lot for updating the documentation.  It looks much better.

README.md
bernhardkaindl: Thanks, I've pushed a new update of this paragraph with this new content:

uftrace traces each function in the executable and shows time durations.
By hooking into the procedure linkage tables of shared libraries, it can
also trace shared library calls like the `ltrace` tool, but with a nicer,
more readable and colored output format which even includes resolving of
C FILE pointers for stdin, stdout and stderr.

README.md
bernhardkaindl: Thanks, in the the new update, I moved the last sentence about Chrome trace viewer into a new paragraph at the end of the feature list, and reworded it as:

Recorded data can be graphically visualized using Chrome trace viewer:
It allows to zoom and pan in the trace as well as display calls as a flame graph.

bernhardkaindl:
@namhyung: In this new push, I've tried to address your review comments.

Please review for any corrections or improvements to be made, thanks!

README.md
honggyukim: Hi @bernhardkaindl, thanks very much for your help!

I couldn't fully follow the changes, but this part doesn't include explanation of `--nest-libcall`, which allows uftrace to follow nested library function calls. `ltrace` doesn't support this as far as I know.

honggyukim: You can see the output difference of `ltrace` and `uftrace`.
```
$ ltrace -C clang 2>&1 | head
getenv("bar")                                    = nil
clang::noteBottomOfStack()(1, 0x7ffd3719af78, 0x7ffd3719af88, 0x420c78) = 0x7fd75a6d8778
llvm::InitLLVM::InitLLVM(int&, char const**&, bool)(0x7ffd3719ad00, 0x7ffd37199a64, 0x7ffd37199a58, 1) = 0
llvm::setBugReportMsg(char const*)(0x41df93, 0, 0, 1) = 0
memcpy(0x7ffd3719a500, "^\303\0317\375\177\0\0", 8) = 0x7ffd3719a500
llvm::sys::Process::FixupStandardFileDescriptors()(0x7ffd3719a500, 0x7ffd3719c35e, 8, 0x7ffd3719c35e) = 0
LLVMInitializeAArch64TargetInfo(2, 0x7fd7537d846f, 0x7fd753c26220, 0x7fd753713eee) = 0x7fd75a6a5908
LLVMInitializeAMDGPUTargetInfo(0x7fd75a6a5a08, 0x7fd757a834ee, 0x7fd757b82979, 0x7fd757b006eb) = 0x7fd75a6aba18
LLVMInitializeARMTargetInfo(0x7fd75a6abb18, 0x7fd757a87ab0, 0x7fd757b9ee85, 0x7fd757b86b9f) = 0x7fd75a6af5e0
LLVMInitializeAVRTargetInfo(0x7fd75a6af6e0, 0x7fd757a834ce, 0x7fd757bcf606, 0x7fd757b4cf5e) = 0x7fd75a6af6e0

$ uftrace --nest-libcall -a clang | head -12
clang: error: no input files
 292.220 us [ 82779] | getenv("bar") = "NULL";
            [ 82779] | clang::noteBottomOfStack() {
   1.160 us [ 82779] |   __tls_get_addr();
   7.834 us [ 82779] | } /* clang::noteBottomOfStack */
            [ 82779] | llvm::InitLLVM::InitLLVM() {
   0.570 us [ 82779] |   pthread_mutex_lock(0x7f04dfc50090) = 0;
            [ 82779] |   operator new(48) {
   0.297 us [ 82779] |     malloc(48) = 0x19f9720;
   1.596 us [ 82779] |   } = 0x19f9720; /* operator new */
   0.184 us [ 82779] |   pthread_mutex_unlock(0x7f04dfc50090) = 0;
   0.152 us [ 82779] |   pthread_mutex_lock(0x19f9720) = 0;
```
Especially, uftrace shows the library calls in more depth

bernhardkaindl: Thanks, I didn't even know about this nice combination of --nest-libcall and -a!
I guess a dedicated EXAMPLES.md which shows off each feature would be nice!

_**Do you accept this correction (below) of my omission? Pushed now.**_

Example with nested library calls:
```py
uftrace -la kdemv -h
   0.598 us [2064374] | std::ios_base::Init::Init();
   0.102 us [2064374] | __cxa_atexit();
            [2064374] | QApplication::QApplication() {
            [2064374] |   operator new(360) {
   0.263 us [2064374] |     malloc(360) = 0x55eddb5fad70;
   2.620 us [2064374] |   } = 0x55eddb5fad70; /* operator new */
            [2064374] |   QApplicationPrivate::QApplicationPrivate() {
            [2064374] |     QGuiApplicationPrivate::QGuiApplicationPrivate() {
            [2064374] |       QCoreApplicationPrivate::QCoreApplicationPrivate() {
   2.178 us [2064374] |         QObjectPrivate::QObjectPrivate();
   0.250 us [2064374] |         QReadWriteLock::QReadWriteLock();
   0.328 us [2064374] |         geteuid() = 1000;
   0.218 us [2064374] |         getuid() = 1000;
```

Direct link to the updated README.md:
https://github.com/bernhardkaindl/uftrace/blob/improve-README.md/README.md

bernhardkaindl:
This new push is an attempt to fix my omission of the mention of nested libcall tracing. Adds it with an example.

It follows the adage(used in at least in english an german): [A picture is worth more than 1000 words](https://en.wikipedia.org/wiki/A_picture_is_worth_a_thousand_words)

README.md
```diff
@@ -4,47 +4,70 @@
...
-It can show detailed execution flow at function level, and report which
-function has the highest overhead.  And it also shows various information
-related the execution environment.
+uftrace traces each function in the executable and shows time durations.
+By hooking into the procedure linkage tables of shared libraries, it can
+also trace shared library calls like the `ltrace` tool, but with a nicer,
+more readable and colored output format which even includes resolving of
+C FILE pointers for stdin, stdout and stderr. Example with nested library
+calls:
+```py
```
namhyung, line 33: py ?

README.md
```diff
@@ -4,47 +4,70 @@
...
-library calls and/or internal function calls in the library call.
-
-It can show detailed execution flow at function level, and report which
-function has the highest overhead.  And it also shows various information
-related the execution environment.
+uftrace traces each function in the executable and shows time durations.
+By hooking into the procedure linkage tables of shared libraries, it can
+also trace shared library calls like the `ltrace` tool, but with a nicer,
+more readable and colored output format which even includes resolving of
+C FILE pointers for stdin, stdout and stderr. Example with nested library
```
namhyung, line 31: I think you can also mention that it doesn't require recompilation for tracing.  The library calls were fine.  And with the full dynamic tracing (the `-P` option), you can also trace functions in the executable (as long as it has the symbol table).

namhyung (commented):
Thanks for the update!

README.md
honggyukim: Thanks for the update. It looks good but I guess it might be useful if you can add some examples with string argument printed together as it looks nicer.

README.md
honggyukim: I think it's to make the uftrace output prettier.

README.md
honggyukim: For example, this kind of output looks good as follows.
```py
$ uftrace record -la lsusb > /dev/null 
$ uftrace replay | head -30
 299.951 us [ 23303] | setlocale(LC_TYPE, "") = "en_US.UTF-8";
   1.434 us [ 23303] | getopt_long(1, 0x7ffcd0a4ea68, "D:vtP:p:s:d:Vh") = -1;
            [ 23303] | udev_new() {
   0.301 us [ 23303] |   malloc(16) = 0x55aef04df990;
   2.143 us [ 23303] | } /* udev_new */
            [ 23303] | udev_hwdb_new() {
   0.424 us [ 23303] |   calloc(1, 200) = 0x55aef04df9b0;
   5.724 us [ 23303] |   fopen64("/etc/systemd/hwdb/hwdb.bin", "re") = 0;
   0.069 us [ 23303] |   __errno_location() = 0x7fb68c6d56f8;
   0.231 us [ 23303] |   strlen("/etc/systemd/hwdb/hwdb.bin") = 26;
   2.068 us [ 23303] |   fopen64("/etc/udev/hwdb.bin", "re") = 0;
   0.056 us [ 23303] |   __errno_location() = 0x7fb68c6d56f8;
   0.111 us [ 23303] |   strlen("/etc/udev/hwdb.bin") = 18;
   1.989 us [ 23303] |   fopen64("/usr/lib/systemd/hwdb/hwdb.bin", "re") = 0;
   0.047 us [ 23303] |   __errno_location() = 0x7fb68c6d56f8;
   0.117 us [ 23303] |   strlen("/usr/lib/systemd/hwdb/hwdb.bin") = 30;
   1.940 us [ 23303] |   fopen64("/lib/systemd/hwdb/hwdb.bin", "re") = 0;
   0.049 us [ 23303] |   __errno_location() = 0x7fb68c6d56f8;
   0.116 us [ 23303] |   strlen("/lib/systemd/hwdb/hwdb.bin") = 26;
   2.596 us [ 23303] |   fopen64("/lib/udev/hwdb.bin", "re") = 0x55aef02ab740;
   0.185 us [ 23303] |   fileno(0x55aef02ab740) = 4;
   1.375 us [ 23303] |   fstat64();
   0.080 us [ 23303] |   fileno(0x55aef02ab740) = 4;
   3.973 us [ 23303] |   mmap64(0, 0xb0c1e7, PROT_READ, MAP_SHARED, 4, 0) = 0x7fb689800000;
   0.142 us [ 23303] |   malloc(24) = 0x55aef04dfa80;
   0.084 us [ 23303] |   malloc(24) = 0x55aef04dfaa0;
  31.215 us [ 23303] | } /* udev_hwdb_new */
            [ 23303] | libusb_init() {
   1.118 us [ 23303] |   pthread_mutex_lock(0x7fb68cb7a420) = 0;
```
It's off topic but we can improve the output  by handling `__errno_location`.

The prototype of `__errno_location` might be this.
```
int *__errno_location(void);
```
But uftrace needs to add support for dereferencing a type, then it can be shown as follows.
```
__errno_location() = 2;
```
Then we can give hint for the `errno` values from `errno -l`. Maybe I can create an issue for this.

README.md
bernhardkaindl: Added above now:
> With fully dynamic dynamic tracing (`-la`), it works on unmodified executables:

Let me just for completeness explain why (now that I know it, I was not even aware of -la before) I like to mention `-la` as example:

I give `-la` as an example because `-la` works universally, includes arguments and return values and is easier to use:

For unmodified, distributed and stripped executables, I usually only know one function: `main`, and apparently `-P<any argument>` does the same on those for me.

Linux distributions strip their executables by default so `-P FUNC`, so it seems, as long as it's stripped and there is no debug info installed (I guess), `-P<any argument>` does the same for but but I didn't find the documentation for (to stay that way).

Also, with `-l` / `--nest-libcalls`, I get the nesting of calls throughout libraries, so it gives much more information.

README.md
bernhardkaindl: Thanks, I've added it like this:
```py
$ uftrace record -la lsusb >/dev/null 
$ uftrace replay -f+module
or simply:
$ uftrace -la -f+module lsusb  # -f+module adds the module name column(man uftrace-live):
# DURATION     TID        MODULE NAME   FUNCTION
 306.339 us [ 23561]            lsusb | setlocale(LC_TYPE, "") = "en_US.UTF-8";
   1.163 us [ 23561]            lsusb | getopt_long(1, 0x7fff7175f6a8, "D:vtP:p:s:d:Vh") = -1;
            [ 23561]            lsusb | udev_new() {
   0.406 us [ 23561] libudev.so.1.7.2 |   malloc(16) = 0x55e07277a7b0;
   2.620 us [ 23561]            lsusb | } /* udev_new */
            [ 23561]            lsusb | udev_hwdb_new() {
   0.427 us [ 23561] libudev.so.1.7.2 |   calloc(1, 200) = 0x55e07277a7d0;
   5.829 us [ 23561] libudev.so.1.7.2 |   fopen64("/etc/systemd/hwdb/hwdb.bin", "re") = 0;
```

README.md
```diff
@@ -36,8 +36,10 @@ Features
 ========
 
 uftrace traces each function in the executable and shows time durations.
+
+With fully dynamic dynamic tracing (`-la`), it works on unmodified executables:
```
namhyung, line 40: Double dynamic? :)  Also the option should be `-P`.

bernhardkaindl: Thanks, I've fixed and updated the paragraph now by adding more explations:

> uftrace traces each function in the executable and shows time durations.
>
> Usually, for this to be possible, the program needs to be compiled with
`-pg` or `-fpatchable-function-entry=2`, but with fully dynamic tracing
(`-P.`|`--patch=.`), uftrace works on all executables (as long they are
not stripped, or symbol information is available from a separate file).
>
>uftrace is able to perform nested library call tracing even on stripped
executables with no symbol information using (`-l`|`--nest-libcall`)
by hooking into the procedure linkage tables of shared libraries.
>
>Using (`-a`|`--auto-args`), uftrace automatically records arguments and
return values of known functions. Without extra debug information, this
includes the API functions of standard (C language or system) libraries.
>
>If debug information for the program (`gcc -g`) is available, `--auto-args`
works even on functions inside of the compiled the user programs.
>
>In case argument information is not available, argument specifiations like
(`-A udev_new@arg1/s`) can be passed on the command line or an options file.

README.md
```diff
@@ -4,47 +4,97 @@
...
-supported.  Optionally it's possible to trace other (nested) external
-library calls and/or internal function calls in the library call.
-
-It can show detailed execution flow at function level, and report which
-function has the highest overhead.  And it also shows various information
-related the execution environment.
+uftrace traces each function in the executable and shows time durations.
+
+Usually, for this to be possible, the program needs to be compiled with
+`-pg` or `-fpatchable-function-entry=2`, but with fully dynamic tracing
```
honggyukim, line 43: The number of NOPs required depends on architecture.

`-fpatchable-function-entry=2` is only for AArch64. For x86_64, it must be `-fpatchable-function-entry=5`.

README.md
```diff
@@ -4,47 +4,97 @@
...
+(`-P.`|`--patch=.`), uftrace works on all executables (as long they are
+not stripped, or symbol information is available from a separate file).
+
+uftrace is able to perform nested library call tracing even on stripped
+executables with no symbol information using (`-l`|`--nest-libcall`)
+by hooking into the procedure linkage tables of shared libraries.
```
honggyukim, line 49: This description is only for the first level library function tracing. To explain about `--nest-libcall`, we better explain uftrace hooks PLTs inside libraries as well as the given executable file.

honggyukim (commented):
Hi @bernhardkaindl, thanks very much for your help! I've just made a few comments for improvement.

README.md
namhyung: Actually the dynamic tracing option `-P` takes a pattern (usually in regex, but you can use wildcard with additional `--match=glob` option) to match function names.  If the pattern doesn't contains any special character in regex, it'd try the simple full-string match.  If it contains any special character then it'd try the regex match.  That's why we use the single dot (`.`) mostly to match every function in the simplest way (with regex).

README.md
namhyung: It's the `--force` (or `-P.`) option which enables (flat) library call tracing on stripped binaries.

README.md
bernhardkaindl: I'll update it to use "full dynamic tracing" to be consistent but grammar-wise, it looks a bit strange:

"full" is mostly used to refer to a container or all of the something which has some boundary, like:
full moon, strength, full speed, full support, full report.

It is used to refer something which can be "partial", "half", or "minimal", like half moon, half speed, partial report, or minimal support.

See the Cambridge dictionary, for example:
https://dictionary.cambridge.org/dictionary/english/full
https://www.merriam-webster.com/dictionary/full

I'm not saying that "fully" is correct, because I know no definition of "full dynamic tracing" nor of "fully dynamic tracing". I don't think either one is good, I think "fully" just sounds less wrong. Omitting "full" and "fully" would be best I think, but that is a global change.

I know that "dynamic" in regarding to libraries and symbols refers to shared libraries which are dynamically linked into programs, but I do not actually know if "dynamic" tracing:

I have two interpretations:
- Tracing of shared (dynamically loaded) libraries.
- Tracing which can be dynamically configured (enabled, disabled, refined, changed, reconfigured)

I assume the latter. But what is the difference beween "dynamic tracing" (like normal? dynamic tracing" and "full dynamic tracing"?

README.md
```diff
@@ -4,47 +4,97 @@
 uftrace
 =======
 
-The uftrace tool is to trace and analyze execution of a program
-written in C/C++.  It was heavily inspired by the ftrace framework
-of the Linux kernel (especially function graph tracer) and supports
-userspace programs.  It supports various kind of commands and filters
-to help analysis of the program execution and performance.
+uftrace is a function call graph tracer for C, C++, Rust and Python programs.
+Using dynamic tracing, it can include nested library calls and
```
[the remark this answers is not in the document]
bernhardkaindl: Thank at lot, and according to `doc/uftrace-live.md`, it also supports `-mfentry`, `-mnop-mcount`, and using clang, `-fxray-instrument`.

With this information, I've updated the start of the README like this now:

> It can record data from:
> - User space functions by dynamically patching functions at runtime, or with code
  compiled with with `-pg`, `-fpatchable-function-entry`, `-mfentry`, `-mnop-mcount`,
  `-finstrument-functions`, or using clang with `-fxray-instrument`.
>- Library functions (through PLT hooking)
>- Kernel functions (using the ftrace framework in Linux kernel)
>- Kernel trace events affecting the program, e.g. scheduling events (which affect the
  execution timing of the program) and records nanosecond-exact timestamps
  (using perf_event and systemtap SDT)
>
>With the recorded data, it:
>- Shows colored nested function call graphs (instead of listings)!
>- Displays arguments symbolically using libc function prototypes and DWARF
  debug information values.
>- Can apply filters like minimal function call duration
>- Can display metadata about the system on which the trace was taken
>- Can generate a symbol table of the traced program and library functions
>- When the execution of multiple programs was recorded,
  it can generate a task relationship tree (parent and childs) from the trace.

README.md
bernhardkaindl: Thanks, I submitted https://github.com/namhyung/uftrace/pull/1664 to add this helpful information to doc/uftrace-live.md and doc/uftrace-record.md. It needed elsewhere as well, I propose to follow it ip in #1664.

README.md
bernhardkaindl: Yes, I was actually mention hooking into the PLT (jus using "procedure linkage tables"), added "PLT" now as well.

Thanks @namhyung for correcting the info Honggyu Kim, and I'd like to add that -D1 is more clear for flat/no-nesting. I've added this to the paragraph.

README.md
```diff
@@ -4,47 +4,109 @@
...
 
-The uftrace tool is to trace and analyze execution of a program
-written in C/C++.  It was heavily inspired by the ftrace framework
-of the Linux kernel (especially function graph tracer) and supports
-userspace programs.  It supports various kind of commands and filters
-to help analysis of the program execution and performance.
+uftrace is a function call graph tracer for C, C++, Rust and Python programs.
+
+uftrace traces, records and analyzes the execution of programs written in
+C, C++ and Rust. It was heavily inspired by the ftrace framework of the
```
namhyung, line 10: And Python, but I think this is redundant.

README.md
```diff
@@ -4,47 +4,109 @@
...
+
+With the recorded data, it:
+- Can show colored and nested function call graphs
+- Can show arguments and return values symbolically using libc function prototypes
+  and DWARF debug information.
+- Can apply filters like minimal function call duration
+- Can extract metadata from traces (e.g. system information on which the trace was taken)
+- Can generate a symbol table of the traced program and library functions
+- When the execution of multiple programs was recorded,
+  it can generate a task relationship tree (parent and children) from the trace.
```
namhyung, line 34: It also can generate chrome trace data, flame graph or call-graph diagrams for graphviz and mermaid.

namhyung (commented):
Looks much better now!

README.md
```diff
@@ -4,47 +4,109 @@
...
+  execution timing of the program) and records nanosecond-exact timestamps
+  (using perf_event and systemtap SDT)
+
+With the recorded data, it:
+- Can show colored and nested function call graphs
+- Can show arguments and return values symbolically using libc function prototypes
+  and DWARF debug information.
+- Can apply filters like minimal function call duration
+- Can extract metadata from traces (e.g. system information on which the trace was taken)
+- Can generate a symbol table of the traced program and library functions
```
honggyukim, line 32: This can be changed as follows.
> Can generate symbol tables and memory maps of the traced program and library functions

README.md
```diff
@@ -4,47 +4,109 @@
...
+- Kernel functions (using the ftrace framework in Linux kernel)
+- Kernel trace events affecting the program, e.g. scheduling events (which affect the
+  execution timing of the program) and records nanosecond-exact timestamps
+  (using perf_event and systemtap SDT)
+
+With the recorded data, it:
+- Can show colored and nested function call graphs
+- Can show arguments and return values symbolically using libc function prototypes
+  and DWARF debug information.
+- Can apply filters like minimal function call duration
```
honggyukim, line 30: This can be changed as follows.

> Can apply filters to minimize the amount of trace data in both record and replay time

README.md
```diff
@@ -4,47 +4,109 @@
...
-supported.  Optionally it's possible to trace other (nested) external
-library calls and/or internal function calls in the library call.
-
-It can show detailed execution flow at function level, and report which
-function has the highest overhead.  And it also shows various information
-related the execution environment.
+uftrace traces each function in the executable and shows time durations.
+
+Usually, for this to be possible, the program needs to be compiled with
+`-pg` or `-fpatchable-function-entry=5`, but with full dynamic tracing
```
honggyukim, line 54: `-fpatchable-function-entry=5` can be just `-fpatchable-function-entry=2` in AArch64, but it's not harmful even having more NOPs so let's leave it.

README.md
```diff
@@ -4,47 +4,109 @@
...
+(`-P.`|`--patch=.`), uftrace works on all executables (as long they are
+not stripped, or symbol information is available from a separate file).
+
+uftrace is able to perform nested library call tracing even on stripped
+executables with no symbol information using (`-l`|`--nest-libcall`)
+by hooking into the procedure linkage tables (PLTs) of shared libraries.
+The depth can be limited using `-D<num>`, where 1 is flat call tracing.
+
+Using (`-a`|`--auto-args`), uftrace automatically records arguments and
+return values of known functions. Without extra debug information, this
```
honggyukim, line 64: In addition, `-a` implies `--srcline` so it records the source line location info and it can be shown in `uftrace replay --srcline` and in `uftrace tui`, users can directly open their editor with the exact source location as shown in https://uftrace.github.io/slide/#120.

README.md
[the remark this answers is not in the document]
honggyukim: Looks great! Thanks.

honggyukim (commented):
Thanks very much for the update! It's getting much better.

There is another summary about uftrace at https://www.blackhat.com/asia-23/arsenal/schedule/#uftrace-dynamic-analysistracing-tool-for-ccrust-programs-30951.

It's for another presentation in Black Hat Asia conference this May. You can refer to the description.

README.md
bernhardkaindl: I've replaced this paragraph now with the text from https://www.blackhat.com/asia-23/arsenal/schedule/#uftrace-dynamic-analysistracing-tool-for-ccrust-programs-30951, which is even better.

README.md
bernhardkaindl: Thanks, merged into the new paragraph about visualization from https://www.blackhat.com/asia-23/arsenal/schedule/#uftrace-dynamic-analysistracing-tool-for-ccrust-programs-30951:

> uftrace offers various filters to reduce the amount of trace data and provides
visualization using Chrome trace viewer and flame graph or call-graph diagrams
for graphviz and mermaid, allowing for a big picture view of the execution flow.

README.md
bernhardkaindl: done.

README.md
bernhardkaindl: done.

README.md
bernhardkaindl: incorporated:

> Usually, for this to be possible, the program needs to be compiled with `-pg` or `-fpatchable-function-entry=5` (`=2` is enough on aarch64). With full dynamic tracing (`-P.`|`--patch=.`), uftrace works on all executables (as long they are not stripped, or symbol information is available from a separate file).

README.md
bernhardkaindl: added.

bernhardkaindl:
@namhyung I've incorporated all the suggested changes and feedback, the PR is ready for review again!

namhyung (approved):
LGTM, thanks a lot for your effort!

honggyukim (approved):
Thanks. I appreciate your valuable contribution!

PR #1654, opened 2023-03-22
Allow customizing the bash completions dir for system-RPM installs

bernhardkaindl:
@namhyung :

Fedora packaging dictates that `/etc` is dedicated for bash completions installed and managed by the system admin and bash completions installed by the packages shall live in `/usr/share` (referenced by `%{_datadir}`)

Support packaging for distros by using a common variable at all locations where we install the bash completions script, which distribution packages can override.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ libdir = $(prefix)/lib/uftrace
 etcdir = $(prefix)/etc
 mandir = $(prefix)/share/man
 docdir = $(srcdir)/doc
+completiondir = $(etcdir)/bash_completion.d
 
 ifeq ($(DOCLANG), ko)
   docdir = $(srcdir)/doc/ko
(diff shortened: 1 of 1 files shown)


bernhardkaindl:
@namhyung This would make packaging scripts for RedHat/Fedora/CentOS/Rocky/AlmaLinux/OracleLinux/AmazonLinux-family and Debain/Ubuntu/Mint-family distros easier:

With it, the build can specify to use the /usr/share/bash-completion/completions/ as installation directory:

These Distributions (probably >80% of all users, I guess) reserve `/etc/bash_completion.d/` for local system admins:

`/etc/` is also called the sysconfdir - it is for configuring the system by the local system admins, e.g. if they want to use special completion files locally and `/usr/` is reserved for the system-installed files.

IT would be good idea to switch to using /usr/share/bash-completion/completions/, but that's for a later PR, if at all!

This is just to give the possibility with the smallest update possible!

namhyung (approved):
I'm sorry for the late review.  I thought I already replied but I didn't.  LGTM

We may want to add it to `configure` script too.  But it can be done separately.

PR #1655, opened 2023-03-22
Catch IndexError from self.sort in test/runtest.py

bernhardkaindl:
@namhyung :

When packaging `uftrace` for Fedora using the Copr build service, I enabled running the test suite for checking the build.

I found that on some CentOS 8 build targets, a few tests caused `test/runtest.py` to crash because they referenced an `variable[1]` without first checking if `variable[1]` can be indexed.

Catch this exception in runtest.py and (hopefully) handle it correctly as a diff in the test case result.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -626,8 +626,12 @@ class TestBase:
             timed_out = True
         signal.alarm(0)
 
-        result_expect = self.sort(self.result)
-        result_tested = self.sort(result_origin)  # for python3
+        try:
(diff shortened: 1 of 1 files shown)


namhyung (approved):
LGTM

bernhardkaindl:
@namhyung Thanks for the approval.
with @honggyukim: I've no merge permission here, so maybe one of you have to merge for now.

honggyukim:
Hi @bernhardkaindl, thanks for the fix.

@namhyung normally merges PRs together at some points so please wait for a while.

PR #1656, opened 2023-03-25
misc: Add more prototypes enums for socket

honggyukim:
The following socket() shows some incorrect enum values for now.
```
  int socket(int domain, int type, int protocol);
```
The incorrect output looks like this.
```
  socket(AF_NETLINK, SOCK_PACKET|SOCK_DCCP|SOCK_SEQPACKET|SOCK_RDM|SOCK_RAW|SOCK_DGRAM|SOCK_STREAM+0x807e4, 15) = 7;
```
It's because the 'type' field can accept SOCK_CLOEXEC and SOCK_NONBLOCK, but current prototypes.h doesn't have them.

So this patch adds them then the output looks correct as follows.
```
  socket(AF_NETLINK, SOCK_CLOEXEC|SOCK_NONBLOCK|SOCK_RAW, 15) = 7;
```
This can be seen as follows.
```
  $ uftrace -la -C socket lsusb
      ...
              [ 18635] | libusb_init() {
              [ 18635] |   udev_monitor_new_from_netlink() {
     5.682 us [ 18635] |     socket(AF_NETLINK, SOCK_CLOEXEC|SOCK_NONBLOCK|SOCK_RAW, 15) = 7;
    13.064 us [ 18635] |   } /* udev_monitor_new_from_netlink */
     6.142 ms [ 18635] | } /* libusb_init */
```


misc/prototypes.h
```diff
@@ -371,9 +371,10 @@ enum uft_socket_domain {
 enum uft_socket_type {
 	SOCK_STREAM = 1, SOCK_DGRAM, SOCK_RAW, SOCK_RDM, SOCK_SEQPACKET, SOCK_DCCP,
 	SOCK_PACKET = 10,
+	SOCK_NONBLOCK = 04000, SOCK_CLOEXEC = 02000000,
 };
 enum uft_socket_flag {
-	 SOCK_NONBLOCK = 04000, SOCK_CLOEXEC = 02000000,
+	SOCK_NONBLOCK = 04000, SOCK_CLOEXEC = 02000000,
```
namhyung, line 377: I'm curious why they are in octal notation, can we just have them in hex?

honggyukim: There are written in octal notation in `/usr/include/x86_64-linux-gnu/bits/socket_type.h`. It's not a problem to convert to hex, but I think it looks fine to keep same as where it was written.

namhyung: Ok, that makes sense.

namhyung (approved):

PR #1659, opened 2023-03-27
aarch64/runtest.py: Fix tests where gcc/clang emits memcpy()

bernhardkaindl:
This fixes about a dozen test case failures using the s-arg.c and s-return.c files on aarch64:

On aarch64 (depending on a number of factors including size and optimisation level) gcc and clang use `memcpy()` to pass and `return` big `structs` to and from function calls.

For large 4k returns, `gcc` may emit `memcpy()` for all optimization level, while for other smaller structs, gcc and clang may emit memcpy() this only for lower optimisation levels.
For smaller structs, gcc and clang may emit memcpy() only    for lower optimisation levels.
In additon, it may change from one compiler version to the next.

It affects all tests tracing s-arg.c:pass() / return.c:return_large()

Two test cases did already filter memcpy(), but about a dozen other test failures were caused by unexpected memcpy() in the output.

Because emitting memcpy varies by different factors and can change from compiler to compiler, pass -N memcpy to exclude it from all traces.

memcpy() is not not expected in any test output yet, so the change to suppress memcpy is limited to tests/runtest.py.

### DIFF
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -300,6 +300,13 @@ class TestBase:
         if '-P' in self.option:
             self.p_flag = ''
 
+        # On aarch64, gcc and clang emit calls to memcpy() to pass and return structs.
+        # This depends on a number of factors (gcc, clang, size and optimisation levels).
+        # For now, it affects all tests tracing s-arg.c:pass() and return.c:return_large().
+        # Any test can be affected by it and it may change from compiler to compiler.
+        # To prevent inconsistencies, filter all calls to memcpy() for all tests:
+        self.option += ' -N memcpy'
+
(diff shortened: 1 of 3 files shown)


honggyukim:
I think we better add `-N memset` as well as `-N memcpy`. This case was handled in 65807751ee8992ee487cf1136de735123653d1f6.

Please see this.
```
$ git gp memset | grep replace
t088_graph_session.py:47:        return result.replace("readlink", """memset
t181_graph_full.py:44:        return result.replace("readlink", """memset
```
Those are handled in `fixup` routine.

honggyukim:
There is one more case at 403056ae682e3145eb298ca8f2bace84b71dbd82 and it tried to suppress unexpected `memset`.

namhyung:
Please send a new PR.

Issue #781, opened 2019-06-05
AArch64 first argument passing problem


honggyukim:
As I wrote in #777 #778, there is a problem passing the first argument in AArch64.

I just compiled and dumped for both original `s-fibonacci.c` and its `-pg` compiled version as follows:
```
$ aarch64-linux-gnu-gcc        -c -o s-fibonacci.o      tests/s-fibonacci.c
$ aarch64-linux-gnu-gcc -pg -g -c -o s-fibonacci.o.pg.g tests/s-fibonacci.c

$ aarch64-linux-objdump -d s-fibonacci.o      > s-fibonacci.o.asm
$ aarch64-linux-objdump -d s-fibonacci.o.pg.g > s-fibonacci.o.pg.g.asm
```
For better comparison, I just modified some offset labels in `s-fibonacci.o.asm` and its diff result is as follows:
```diff
--- s-fibonacci.o.adj.asm       2019-06-06 07:48:12.045889894 +0900
+++ s-fibonacci.o.pg.g.asm      2019-06-06 07:09:13.941035548 +0900
@@ -7,9 +7,12 @@
 0000000000000000 <fib>:
    0:  a9bd7bfd        stp     x29, x30, [sp,#-48]!
    4:  910003fd        mov     x29, sp
    8:  f9000bf3        str     x19, [sp,#16]
+   c:  aa1e03e1        mov     x1, x30
   10:  b9002fa0        str     w0, [x29,#44]
+  14:  aa0103e0        mov     x0, x1
+  18:  94000000        bl      0 <_mcount>
   1c:  b9402fa0        ldr     w0, [x29,#44]
   20:  7100081f        cmp     w0, #0x2
   24:  5400006c        b.gt    30 <fib+0x30>
   28:  52800020        mov     w0, #0x1                        // #1
@@ -28,10 +31,13 @@

 000000000000005c <main>:
   5c:  a9bd7bfd        stp     x29, x30, [sp,#-48]!
   60:  910003fd        mov     x29, sp
+  64:  aa1e03e2        mov     x2, x30
   68:  b9001fa0        str     w0, [x29,#28]
   6c:  f9000ba1        str     x1, [x29,#16]
+  70:  aa0203e0        mov     x0, x2
+  74:  94000000        bl      0 <_mcount>
   78:  52800100        mov     w0, #0x8                        // #8
   7c:  b9002fa0        str     w0, [x29,#44]
   80:  b9401fa0        ldr     w0, [x29,#28]
[...]

Issue #777, opened 2019-06-03
argument replay error in AArch64


honggyukim:
It seems that DWARF handling is broken in AArch64. The most of testcases are failed as follows:
```
honggyu@rpi3:~/uftrace$ ls tests/ | grep dwarf.*.py$
t201_arg_dwarf1.py
t202_arg_dwarf2.py
t203_arg_dwarf3.py
t204_arg_dwarf4.py
```
The test result is as follows:
```
Test case                 pg
------------------------: O0 O1 O2 O3 Os
201 arg_dwarf1          : NG NG NZ NZ NZ
202 arg_dwarf2          : NG NG NG NG NG
203 arg_dwarf3          : NG NG NG NG NG
204 arg_dwarf4          : NG NG NZ NZ NZ
```
The most simple case might be `204` as follows:
```
honggyu@rpi3:~/uftrace/tests$ ./runtest.py -vdp 204 -O0
Start 1 tests with 4 worker
Test case                 pg
------------------------: O0
build command: g++ -o t-exception4 -fno-inline -fno-builtin -fno-ipa-cp -fno-omit-frame-pointer -D_FORTIFY_SOURCE=0 -g -pg -O0  s-exception4.cpp   
test command: /home/honggyu/uftrace/uftrace --no-pager --no-event -L/home/honggyu/uftrace -A . -R . -F main -N personality_v. t-exception4
=========== original ===========
# DURATION     TID     FUNCTION
            [ 17758] | main() {
            [ 17758] |   foo(0xaaaae9a0dc08) {
   8.177 us [ 17758] |     __cxa_allocate_exception();
 242.239 us [ 17758] |   } /* foo */
 267.082 us [ 17758] | } = 0; /* main */

===========  result  ===========
 main() {
   foo(0xADDR) {
     __cxa_allocate_exception();
   } /* foo */
 } = 0; /* main */
=========== expected ===========
 main() {
   foo(1) {
     __cxa_allocate_exception();
   } /* foo */
 } = 0; /* main */
t204_arg_dwarf4: diff result of -pg -O0
--- expect      2019-06-03 22:25:19.940876188 +0900
+++ result      2019-06-03 22:25:19.940876188 +0900
@@ -1,3 +1,3 @@
  main() {
-   foo(1) {
+   foo(0xADDR) {
      __cxa_allocate_exception();

204 arg_dwarf4          : NG
```

honggyukim:
[...]

PR #1660, opened 2023-03-27
Workaround the AArch64 mcount-gcc first argument passing problem

bernhardkaindl:
Address the AArch64 `mcount` `gcc` first argument passing problem.

In https://github.com/namhyung/uftrace/issues/777#issuecomment-498304674, @namhyung wrote:
> AArch64 has a problem that it sets first argument of mcount function so the argument recording is broken. This may be fixed with dynamic tracing.

It has also been analysed in detail in https://github.com/namhyung/uftrace/issues/781

The summary is that on aarch64, as far as I can see, `gcc` passes the frame pointer in 0x (despite already passing it in x30),
thereby overwriting arg1 for mcount().

We need to use some form of dynamic code patching to address this issue.

Because the arguments work just fine using fully dynamic tracing, this initial version of handling it enables fully dynamic tracing and uses it instead of mcount (and even -finstrument-functions, if instrumented by it) to collect tracing data.

Thus, this inital workaround switches from `mcount()` to `-P.`, even when the traced code is in part instrumented using `-pg` or `-finstrument-functions`.

To avoid the duplicate data recording from these, we switch from collecting data from those to use fully dynamic tracing.


namhyung:
In my understanding, `_mcount()` on AArch64 requires an argument (caller address) according to the [glibc code](https://elixir.bootlin.com/glibc/glibc-2.37/source/sysdeps/aarch64/machine-gmon.h#L32).  The first argument should be passed by `x0` register that's why the compiler generates the code with `-pg`.

On the binaries with `-pg`, I don't think we can simply use `-P .` for full dynamic tracing because it would call the `_mcount` later as it's in the code.  I thought about replacing `mov x0, x30` to NOP before the call but I was not sure since it might have different code patterns too.

bernhardkaindl:
> In my understanding, _mcount() on AArch64 requires an argument (caller address) according to the [glibc code](https://elixir.bootlin.com/glibc/glibc-2.37/source/sysdeps/aarch64/machine-gmon.h#L32). The first argument should be passed by x0 register that's why the compiler generates the code with -pg.

Ah, good to know!

> On the binaries with -pg, I don't think we can simply use -P . for full dynamic tracing because it would call the _mcount later as it's in the code.

I know, this needs a program compiled with `-pg` and `-fpatchable-function-enty=5`, and a user who uses `-P`, but in this case, we have this already:

```py
gcc -g -pg -fpatchable-function-entry=5 float-test.c -o float-test-pg-patchable-function-entry
uftrace -P. -Amain@arg1/d -A.*_add@arg1/f,arg2/f ./float-test-pg-patchable-function-entry
float_add(-0.100000, 0.200000) = 0.100000
double_add(-0.110000, 0.220000) = 0.110000
   0.520 us [1284796] | __monstartup();
   0.080 us [1284796] | __cxa_atexit();
            [1284796] | main(1) {                     <- (I assume) from the patchable function entry
            [1284796] |   main(0xffffa5c5aeb4) {      <- (from mcount tracing)
            [1284796] |     float_add(0.000000, 0.000000) {
            [1284796] |       float_add(0.000000, 0.000000) {
  17.760 us [1284796] |         fprintf();
  18.440 us [1284796] |       } /* float_add */
  18.680 us [1284796] |     } /* float_add */
   1.760 us [1284796] |     fprintf();
            [1284796] |     double_add(0.000000, 0.000000) {
            [1284796] |       double_add(0.000000, 0.000000) {
   1.240 us [1284796] |         fprintf();
   1.520 us [1284796] |       } /* double_add */
   1.760 us [1284796] |     } /* double_add */
   0.960 us [1284796] |     fprintf();
  23.920 us [1284796] |   } /* main */
  24.640 us [1284796] | } /* main */
```
> I thought about replacing mov x0, x30 to NOP before the call but I was not sure since it might have different code patterns too.

Indeed: I've seen a disassembly where the `mov x0, x30` was about half a dozen instructions before the mcount in one of the three github issues describing it.

Yes I also thought, if mcount is available, patching the code before mcount would be the nicer change, but I've no idea how to find all the mcount calls (I think, if a shared library would happen to be compiled with `-pg` then they would need to be "fixed" as well, so I've given up on that idea for now).

Is there a big benefit on using `-pg` and not fully dynamic tracing with libcapstone? Thanks!

Meanwhile, I found that there appears to be a runtime flag to disable recording data from mcount calls! :-)
This this works, it could be used in this situation (optionally).

honggyukim:
> Is there a big benefit on using -pg and not fully dynamic tracing with libcapstone? Thanks!

I'm also thinking about it. I believe our full dynamic tracing works quite well in most cases so I think we can recommend it for general users in `README.md`. But the thing is that we don't have enough time to update all the documents for that.

namhyung:
The original concern of the full dynamic tracing is the coverage of instructions that it can patch (or modify).  So it might miss some functions in the outpu that failed to patch.  With compiler support like `-pg` or `-fpatchable-function-entry` we don't worry about the loss.

So I think we have options on AArch64 like below:
1. rebuild the binary with `-pg`: no missing functions, incorrect arg1
2. rebuild the binary with `-finstrument-functions`: no missing functions, cannot record any args,return values
3. rebuild the binary with `-fpatchable-function-entry=2`: no missing functions, no problems with arg1
4. no rebuild, full dynamic tracing with `-P .`: can miss some functions, no problems with arg1

PR #1664, opened 2023-03-29
doc/uftrace-live.md and doc/uftrace-record.md: Update -P FUNC: regex/glob

bernhardkaindl:
As mentioned in https://github.com/namhyung/uftrace/pull/1653#discussion_r1151424388 add that in -P *FUNC*, *FUNC* is actually a regular expression, and that it can be switched to using wildcards using --match=glob:

```ecmarkdown
 -P *FUNC*, \--patch=*FUNC*
 :   Patch FUNC dynamically.  This is only applicable binaries built by
    gcc with `-pg`, `-mfentry`, `-mnop-mcount` or by clang with `-fxray-instrument`.
    By default, *FUNC* is a regex. For example `-P.` matches every function to patch
    every function in the program, so it is a quick shorthand to patch all functions.
    Alternativley, `--match=glob` can be used to to switch to wildcard-based function matching.
    This option can be used more than once.  See *DYNAMIC TRACING* for details.
```

### DIFF
--- a/doc/uftrace-live.md
+++ b/doc/uftrace-live.md
@@ -103,7 +103,10 @@ RECORD OPTIONS
 
 -P *FUNC*, \--patch=*FUNC*
 :   Patch FUNC dynamically.  This is only applicable binaries built by
-    gcc with `-pg -mfentry -mnop-mcount` or clang with `-fxray-instrument`.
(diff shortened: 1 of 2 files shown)


namhyung:
I don't think the dynamic tracing (`-P`) would work with `-pg` nor `-mfentry`.  These options generate call instructions so no control for dynamic tracing.  Yeah, we can convert them to NOPs with `-P` and `-U`, but it doesn't happen yet.

Issue #1652, opened 2023-03-20
Define client/agent protocol


clementguidi:
Let's define the communication protocol between the client and the agent.

From previous discussions (see #1643), and upcoming pull requests (#1644, #1645, #1646, #1647 ATM), we can outline the following points.

- [Context](#context)
  * [Connection handling](#connection-handling)
- [Protocol phases](#protocol-phases)
  * [Handshake phase (once)](#handshake-phase--once-)
  * [Data exchange phase (repeated)](#data-exchange-phase--repeated-)
  * [Validation phase (repeated)](#validation-phase--repeated-)
- [Internal definition](#internal-definition)
  * [Message type](#message-type)
  * [Agent capabilities](#agent-capabilities)
  * [Naming](#naming)
- [Message types](#message-types)
  * [Protocol control](#protocol-control)
  * [State query](#state-query)
  * [Option forwarding](#option-forwarding)
  * [Error checking](#error-checking)
  * [Definition](#definition)
  * [Sequence diagram](#sequence-diagram)
- [Debug info](#debug-info)
  * [Agent output](#agent-output)
  * [Client output](#client-output)

## Context

When the agent is started (`--agent/-g`) in libmcount, it can receive messages from a client at runtime. We're introducing new features in the agent, so it can modify tracing parameters e.g. tracing state (on/off), tracing depth and triggers, but also patch and unpatch the running process.

### Connection handling

Currently, the agent can sequentially handle multiple client connections. The next one is processed when the current one is terminated.

Some operations like patching are heavy and require some time. Any incoming connection will hang until heavy operations are achieved.

The agent will support many independent features e.g. changing tracing depth and patching, which have no reason to be done in a particular order.

- [x] Do we need to handle multiple concurrent connections? &#8594; Not for now

## Protocol phases

The aim of the message exchange is to ultimately forward options from a client to the agent. To make this robust, we introduce control phases as follows.

### Handshake phase (once)

The handshake phase initializes the connection.

1. The client checks is the agent is ready and operational
2. The client checks the agent's capabilities (supported features)

If the handshake fails, the client aborts.

### Data exchange phase (repeated)

When the handshake succeeds, the client and the agent can exchange data.

The client is aware of the capabilities of the agent.

#### Client to agent

The client can instruct the agent to apply user options, as mentioned earlier.

The client skips actions that the agent is known not to support.

#### Agent to client

The client can fetch info from the agent. The agent can send requested info about its state (e.g. parameter values) to the agent.

[...]

PR #1665, opened 2023-03-31
Implement client/agent protocol

clementguidi:
Hi Namhyung,

I reworked the client/agent protocol, as discussed in #1652.

This PR doesn't implement any option in the client, it is only about the protocol.

Two things to notice:
- the `forward_option()` function is unused because we do not send options yet
- I added a `GET_OPT` message but its action is not implemented (only a placeholder for now)

Closes: #1652

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -238,6 +238,86 @@ static int query_agent_capabilities(int fd)
 	return status;
 }
 
+/**
+ * forward_option - forward a given option and its data to the agent
+ * @fd           - agent socket file descriptor
+ * @capabilities - features supported by the agent
+ * @opt          - option type to send
+ * @value        - option data load
+ * @value_size   - size of @value
+ * @return       - status: -1 on error, 0 on success
+ */
+static int forward_option(int fd, int capabilities, int opt, void *value, size_t value_size)
+{
+	void *data;
+	size_t data_size;
+	struct uftrace_msg ack;
+	int ret = -1;
+
+	if (!(opt & capabilities)) {
+		pr_warn("option not unsupported by the agent: %#x\n", opt);
+		return 0;
+	}
+
+	/* The agent needs to know which option to apply, and its associated value.
+	 * We pack both information into the data load of a single message, with
+	 * type UFTRACE_MSG_AGENT_SET_OPT.
+	 *
+	 * We use the following data structure:
+	 *
+	 * PACKED_DATA = OPTION_TYPE |   VALUE
+	 *               -----------   ----------
+	 *               sizeof(opt)   value_size
+	 *
+	 * Example:
+	 * PACKED_DATA = UFTRACE_AGENT_OPT_FILTER |  'func1,!func2'
+	 *               ------------------------   ---------------
+	 *                     sizeof(opt)          value_size = 12
+	 */
+	pr_dbg2("send option to agent: %#x\n", opt);
+	data_size = value_size + sizeof(opt);
+	data = malloc(data_size);
+	if (!data)
+		goto cleanup;
+	memcpy(data, &opt, sizeof(opt));
+	memcpy(data + sizeof(opt), value, value_size);
+
+	ret = agent_message_send(fd, UFTRACE_MSG_AGENT_SET_OPT, data, data_size);
+	if (ret < 0) {
+		pr_dbg("error sending option to agent\n");
+		goto cleanup;
+	}
+
+	ret = agent_message_read_response(fd, &ack);
+	if (ret < 0) {
+		pr_dbg3("error reading agent response\n");
+		goto cleanup;
+	}
+
+	switch (ack.type) {
+	case UFTRACE_MSG_AGENT_OK:
+		ret = 0;
+		pr_dbg4("option applied by agent: %#x\n", opt);
+		break;
+
+	case UFTRACE_MSG_AGENT_ERR:
+		if (ret == ENOTSUP)
+			pr_warn("option not unsupported by the agent: %#x\n", opt);
+		else
+			pr_warn("agent error: %s\n", strerror(ret));
+		ret = -1;
+		break;
+
+	default:
+		pr_dbg3("bad agent message type (expected ack)\n");
+		ret = -1;
+	}
+
+cleanup:
+	free(data);
+	return ret;
+}
+
 /* Forward all client options to the agent */
 static int forward_options(struct uftrace_opts *opts)
 {
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -210,22 +210,22 @@ static int forward_options(struct uftrace_opts *opts)
 	struct sockaddr_un addr;
 	int ret = 0;
 
-	sfd = socket_create(&addr, opts->pid);
+	sfd = agent_socket_create(&addr, opts->pid);
 	if (sfd == -1)
 		return -1;
 
-	if (socket_connect(sfd, &addr) == -1) {
+	if (agent_connect(sfd, &addr) == -1) {
 		ret = -1;
 		goto socket_error;
 	}
 
-	if (socket_send_option(sfd, UFTRACE_DOPT_CLOSE, NULL, 0) == -1) {
+	if (agent_message_send(sfd, UFTRACE_MSG_AGENT_CLOSE, NULL, 0) == -1) {
 		pr_warn("cannot terminate agent connection\n");
 		ret = -1;
 	}
 	else {
-		enum uftrace_dopt ack;
-		if (read(sfd, &ack, sizeof(enum uftrace_dopt)) < 0 || ack != UFTRACE_DOPT_CLOSE)
+		int ack;
+		if (read(sfd, &ack, sizeof(ack)) < 0 || ack != UFTRACE_MSG_AGENT_CLOSE)
 			ret = -1;
 	}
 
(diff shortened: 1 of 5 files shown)


utils/socket.c
```diff
@@ -78,3 +67,63 @@ int agent_accept(int fd)
...
+ *
+ * Fetch the message type and size so the data load can be read somewhere else.
+ *
+ * @fd     - socket file descriptor
+ * @msg    - received message head
+ * @return - status code, negative for error
+ */
+int agent_message_read_head(int fd, struct uftrace_msg *msg)
+{
+	if (read_all(fd, msg, sizeof(msg)) < 0) {
```
namhyung, line 118: Should be `sizeof(*msg)`.

utils/socket.c
```diff
@@ -78,3 +67,63 @@ int agent_accept(int fd)
...
+		},
+		{
+			.iov_base = data,
+			.iov_len = size,
+		},
+	};
+
+	pr_dbg4("send agent message (size=%d)\n", size);
+	size += sizeof(msg);
+	if (writev(fd, iov, 2) != (ssize_t)size) {
```
namhyung, line 99: There's `writev_all()`.

uftrace.h
```diff
@@ -419,6 +419,8 @@ enum uftrace_msg_type {
 	UFTRACE_MSG_SEND_END,
 
 	UFTRACE_MSG_AGENT_CLOSE = 200, /* close the connection */
+	UFTRACE_MSG_AGENT_OK, /* ack previous message */
+	UFTRACE_MSG_AGENT_ERR, /* signal error on previous message */
```
namhyung, line 426: We can just have `UFTRACE_MSG_AGENT_RESP` and pass 0 for OK, non-0 for ERR.  That'd simplify handling responses a bit.

cmds/live.c
```diff
@@ -238,6 +238,69 @@ static int query_agent_capabilities(int fd)
...
+		return 0;
+	}
+
+	pr_dbg2("send option to agent\n");
+	data_size = value_size + sizeof(opt);
+	data = malloc(data_size);
+	if (!data)
+		goto cleanup;
+	memcpy(data, &opt, sizeof(opt));
+	memcpy(data + sizeof(opt), value, value_size);
```
namhyung, line 268: I think we need to document this kind of details.

cmds/live.c
```diff
@@ -238,6 +238,69 @@ static int query_agent_capabilities(int fd)
...
+ * @return       - status: -1 on error, >= 0 on success
+ */
+static int forward_option(int fd, int capabilities, int opt, void *value, size_t value_size)
+{
+	void *data;
+	size_t data_size;
+	struct uftrace_msg ack;
+	int ret = -1;
+
+	if (opt && capabilities == 0) {
```
namhyung, line 257: I don't know what you want to check exactly.  How do you change it if we have non-zero capabilities?

cmds/live.c
```diff
@@ -238,6 +238,69 @@ static int query_agent_capabilities(int fd)
...
 }
 
+/**
+ * forward_option - forward a given option and its data to the agent
+ * @fd           - agent socket file descriptor
+ * @capabilities - features supported by the agent
+ * @opt          - option type to send
+ * @value        - option data load
+ * @value_size   - size of @value
+ * @return       - status: -1 on error, >= 0 on success
```
namhyung, line 248: I'm not sure the description of the return value is correct.

cmds/live.c
```diff
@@ -238,6 +238,69 @@ static int query_agent_capabilities(int fd)
...
+	case UFTRACE_MSG_AGENT_OK:
+		ret = 0;
+		pr_dbg4("option applied by agent\n");
+		break;
+
+	case UFTRACE_MSG_AGENT_ERR:
+		if (ret == ENOTSUP)
+			pr_dbg2("option not supported\n");
+		else
+			pr_dbg2("agent error: %s\n", strerror(ret));
```
namhyung, line 292: So this could be a positive number.

namhyung (commented):
Thanks for the update.

uftrace.h
clementguidi: I'm using the OK response to carry data. For example, on a query, the capabilities are returned as the data load of the OK message. Should I modify this?

> ```
>       client                              agent
> -------------------            ----------------------------
> QUERY []            --->
>                           <--- OK [features (DEPTH | TIME)]
>
> SET_OPT [DEPTH, 3]  --->
>                           <--- OK []
>
> SET_OPT [SIZE, 100] --->
>                           <--- ERR [NOT SUPPORTED]
> 
> CLOSE []            --->
>                           <--- OK []
> ```

cmds/live.c
clementguidi: Sorry, that is meant to be a bitwise AND.

I want to check if the option that the client is forwarding is supported by the agent.

`capabilities` stores the supported agent features as OR-ed flags, e.g. `capabilities = OPT_TRACE | OPT_FILTER | OPT_DEPTH`.

The above example means that the agent is able to change the tracing state (on/off), to apply filters, and to change the filter depth.

cmds/live.c
clementguidi: I changed it so it returns `0` or `-1`.

utils/socket.c
```diff
@@ -127,3 +127,28 @@ int agent_message_read_head(int fd, struct uftrace_msg *msg)
...
+ * @return - status: data or error code, negative on error
+ */
+int agent_message_read_response(int fd, struct uftrace_msg *response)
+{
+	int status = 0;
+
+	if (agent_message_read_head(fd, response) < 0)
+		return -1;
+
+	if (response->len > sizeof(status))
```
namhyung, line 144: This is just the current implementation issue.  I think we can have a response with bigger data payload.

utils/socket.c
```diff
@@ -127,3 +127,28 @@ int agent_message_read_head(int fd, struct uftrace_msg *msg)
...
+		return -1;
+
+	if (read_all(fd, &status, response->len) < 0) {
+		pr_dbg3("error reading agent socket\n");
+		return -1;
+	}
+	pr_dbg4("read agent response (size=%d)\n", response->len);
```
namhyung, line 151: Print the response type too.

cmds/live.c
```diff
@@ -220,6 +221,11 @@ static int forward_options(struct uftrace_opts *opts)
 	pr_dbg2("connected to agent %d\n", opts->pid);
 
 	status_close = agent_message_send(sfd, UFTRACE_MSG_AGENT_CLOSE, NULL, 0);
+	if (status_close == 0) {
+		status_close = agent_message_read_response(sfd, &ack);
```
namhyung, line 346: `s/status_close/status/` ?

utils/socket.c
```diff
@@ -78,3 +67,63 @@ int agent_accept(int fd)
...
+			.iov_base = &msg,
+			.iov_len = sizeof(msg),
+		},
+		{
+			.iov_base = data,
+			.iov_len = size,
+		},
+	};
+
+	pr_dbg4("send agent message (size=%d)\n", size);
```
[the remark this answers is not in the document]
clementguidi: Is the integer value enough? Or do you want a verbose output like 'UFTRACE_MSG_AGENT_SET_OPT'?

cmds/live.c
clementguidi: I'm using `status` to check the agent response to option forwarding (not implemented, I added a placeholder comment). Using it again to check is the connection closes properly would overwrite it. Forwarding option could fail, then the connection is closed properly, and vice-versa. In either case, I consider that a client error.

utils/socket.c
clementguidi: I wanted to keep things simple and only return an `int` in response messages. Should I allocate memory here for arbitrary data?

cmds/live.c
namhyung: Sounds reasonable.

utils/socket.c
namhyung: No, we can do that later if needed.

cmds/live.c
```diff
@@ -238,6 +238,86 @@ static int query_agent_capabilities(int fd)
...
+ */
+static int forward_option(int fd, int capabilities, int opt, void *value, size_t value_size)
+{
+	void *data;
+	size_t data_size;
+	struct uftrace_msg ack;
+	int ret = -1;
+
+	if (!(opt & capabilities)) {
+		pr_dbg2("skip unsupported option\n");
```
namhyung, line 258: Don't we need to warn the user about the unsupported options?

clementguidi: I'm note sure when to use `pr_warn()`, maybe you can a word about it in the wiki? Is it for major events, or unexpected states that the user should have a look at? I see it as an error prevents normal execution, but which is not fatal to the program. And that is not the case if we're skipping an option.

What about other feedback? The client command is not interactive so I don't expect it to communicate with the user outside of the debug messages, which the user has to ask for. For example if the agent fails to apply a filter, should we warn the user too? Or just log it in debug messages so the user can check?

namhyung: Basically it should call `pr_warn()` if it fails to do something user requested.  It might be a debug message if there's a way to work around it.  I think the client command just forwards the option to the agent, then failing on applying a filter would need `pr_warn()`.

namhyung (approved):
Looks good.  Can you please rebase this onto the current master?

clementguidi:
Thank you for your precious feedback, and for the time you put into this project.

PR #1666, opened 2023-04-05
tui: Change the loading message for report mode

qpakzk:
Even though tui command started with the report mode, the loading message was "Building graph for TUI...". This commit changes the loading message for the report mode.


namhyung (approved):
LGTM

PR #1667, opened 2023-04-06
Configuration to support Fedora Packit in GitHub CI using Fedora COPR as build service

bernhardkaindl:
Hi @namhyung!

This PR adds the configuration to support using Fedora Packit for GitHub CI using Fedora COPR.

----
As a demonstration, please have a look at the Checks executed for this PR here:

In https://github.com/linux-tracing/uftrace/pull/3, to the right of:

> All checks have passed
> 12 successful checks

-> click on the link "show all checks" to see all the checks, and scroll down the list of checks (of course it is configurable)

----
This PR provides this capability, when Fedora Packit (which is amed to help devs) is enabled for a given Repo,
it does nothing when Packit is not enabled.

What you get:
- Uses Fedora COPR, the free build service of Fedora builds RPMs for CI testing
- The testsuite is executed during each RPM build (it allows to put this into a separate step as well)
- The Output is checked and archived in the RPM package for analysis.
   - The output of the build is accessible directly from the GitHub Checks.
   - All output is preseved for 60 days for inspection, including testsuite logs.
- If the testsuite shows regressions, the build fails.
   - Nor now, the limits which cause the build to fails are defined in the spec file
   - Of course, this could best be moved into a script or test/run_testsuite.py
- For each PR, it builds with Fedora 36, 37, 38 and 39 (x86_64 and aarch64)
  - Because the builds happen on the latest Fedora code, you get early warning if something breaks.
- The duration all builds and full testsuite runs should be roughly within 20 minutes on average.

This page gives you a quick overview how it works: https://packit.dev (Quotes from it):
> #### [What is Packit?](https://packit.dev/#what-is-packit)
> An open source project aiming to ease the integration of your project with [Fedora Linux](https://getfedora.org/), [CentOS Stream](https://www.centos.org/centos-stream/) and other distributions.
> #### Who should use Packit?
> Upstream projects that want to build RPM packages out of pull requests, commits or releases, test their code changes and bring upstream releases to Fedora Linux.

This describes how to enable the Packit Service: https://packit.dev/docs/guide/, in short:
1. An account at https://accounts.fedoraproject.org/ (you can just sign up) needs set your user name as his Github user.

2. The administrator of the repo (likely you), would need to "order" the free Packit-as-a-service on Github.
3. Then, using a special comment on a Github PR, you confirm that the user on accounts.fedoraproject.org should be used to run the builds,
4. Create a pull request for master and have fun ;-)

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -20,4 +20,10 @@ jobs:
   branch: main
   targets:
     - fedora-36-x86_64
+    - fedora-37-x86_64
     - fedora-38-x86_64
+    - fedora-36-aarch64
+    - fedora-37-aarch64
+    - fedora-38-aarch64
+    - fedora-rawhide-x86_64
+    - fedora-rawhide-aarch64
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -17,7 +17,7 @@ downstream_package_name: uftrace
 jobs:
 - job: copr_build
   trigger: pull_request
-  branch: main
+  branch: master
   targets:
     - fedora-36-x86_64
     - fedora-37-x86_64
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,23 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+upstream_project_url: https://github.com/namhyung/uftrace
+specfile_path: uftrace.spec
+
+# add or remove files that should be synced
+files_to_sync:
+    - uftrace.spec
+    - .packit.yaml
+
+# name in upstream package repository
+upstream_package_name: uftrace
+# downstream (Fedora) RPM package name
+downstream_package_name: uftrace
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  branch: main
+  targets:
+    - fedora-36-x86_64
+    - fedora-38-x86_64
--- a/uftrace.spec
+++ b/uftrace.spec
@@ -63,7 +63,7 @@ rm t147_event_sdt.py t150_recv_event.py
 rm *pmu_*.py t140_dynamic_xray.py # Need to run on a machine with PMU access.
 rm t014_ucontext.py *taskname*.py *sched.py # Kernel API access dependency
 %ifarch x86_64
-rm t271_script_event.py  # Linux perf event diff possble: to be filtered out
+rm t271_script_event.py  # Linux perf event diff possible: to be filtered out
 rm t212_noplt_libcall.py *_report_* t121_malloc_fork.py # Diff, To be fixed
 rm t220_trace_script.py  # Very long output diff on native FC37 install
 rm t225_dynamic_size.py  # all gcc tests fail due to diff on c9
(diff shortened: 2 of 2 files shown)


.packit.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# name in upstream package repository
+upstream_package_name: uftrace
+# downstream (Fedora) RPM package name
+downstream_package_name: uftrace
+
+jobs:
+- job: copr_build
+  trigger: pull_request
+  branch: main
```
namhyung: Is it an exact branch name or an alias?  I think it should be `master`.

.packit.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - .packit.yaml
+
+# name in upstream package repository
+upstream_package_name: uftrace
+# downstream (Fedora) RPM package name
+downstream_package_name: uftrace
+
+jobs:
+- job: copr_build
+  trigger: pull_request
```
namhyung, line 19: It'd be nice if I can trigger it when I push some codes to a branch.  For example, I often push some code to the branches under `check/` or `review/`.  I usually don't create a PR by myself.

Also I want to run different set of tests depending on the trigger type.  For PR check, maybe it's enough run the tests on the latest Fedora only.  For nightly test, it can run on most versions.

uftrace.spec
```diff
@@ -0,0 +1,201 @@
+%bcond_without   check
+%bcond_without   python
+Name:            uftrace
+Version:         0.13
+Release:         16%{?dist}
+
+Summary:         Function graph tracer for C and C++ with many features
+# https://github.com/namhyung/uftrace/issues/1343
+%global          _lto_cflags %nil
```
namhyung, line 9: I wish I can fix the LTO build but I don't know what's the issue exactly.

uftrace.spec
```diff
@@ -72,7 +72,7 @@ rm t225_dynamic_size.py  # all gcc tests fail due to diff on c9
 rm t216_no_libcall_report.py t217_no_libcall_dump.py t218_no_libcall_graph.py
 %ifarch aarch64
 rm t232_dynamic_unpatch.py # x86_64 skips most, but not all, aarch skips all
-rm *_retval.py  *float*.py *nested_func*.py *signal* # Different ouput, to be fixed
+rm *_retval.py  *float*.py *nested_func*.py *signal* # Different output, to be fixed
```
namhyung, line 75: Please move typo changes to the original commit.

namhyung (commented):
Nice work!  I'm happy to see CI working on Fedora.

.packit.yaml
bernhardkaindl: Yes, it's possible to trigger builds by commits to specific branches:
https://packit.dev/docs/configuration/#packit-service-jobs
Example snippet for `.packit.yml` from this page:
```yaml
- job: copr_build
  trigger: commit
  branch: main
  targets:
  - centos-stream-8-x86_64
  - centos-stream-9-x86_64
  - fedora-all
```
So yes, depending on the branch, you can configure specific targets to be built.

I don't know or understand yet, where the builds will be linked so you can access them, but I assume that it just has to be tried, and then we'll find out.

namhyung: Is it also possible to trigger the build manually like from nightly tests?

bernhardkaindl: There seems to be a possiblitily run jobs on a daily basis. At least, there is a project for daily LLVM snapshot builds:
https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-snapshots/

A likewise also interesting topic is the Fedora testing farm which can be also used thruogh Packit:
https://fedoramagazine.org/test-github-projects-with-github-actions-and-testing-farm/

For public users, the tests are run only by AWS instances, but it has a nice test results viewer: 
https://docs.testing-farm.io/general/0.1/test-results.html

namhyung: Can these files live in the `misc/` directory?

PR #1668, opened 2023-04-07
README.md: Fix duplicated "with"

bernhardkaindl:
Fix a duplicated "with" and even with it fixed, the sentence was not clear and easy to read.

Therefore, split the description of tracing possibilities into two:
- One for dynamic tracing, mentioning or `-P.`
- one for tracing by using instrumentation.

While at it, I also recoginzed that the part regarding `-fxray-instrument` can be made clearer and that the readability of the options by adding a space after `-f` (even tough we'd normally omit the space of course), when so many of the are given in one long sentence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ It was heavily inspired by the ftrace framework of the Linux kernel and
 the name uftrace stems from the combination of user and ftrace.
 
 It can record data from:
-- User space C/C++/Rust functions (by dynamically patching functions at runtime,
-  or with code compiled with with `-pg`, `-finstrument-functions` or using
-  `-fpatchable-function-entry=N` for selective NOP patching)
+- User space C/C++/Rust functions, by either dynamically patching functions
+  using `-P.`, or else selective NOP patching using code compiled with `-pg`,
+  `-finstrument-functions` or `-fpatchable-function-entry=N`.
 - C/C++/Rust Library functions (through PLT hooking)
 - Python functions (using Python's trace/profile infrastructure)
 - Kernel functions (using the ftrace framework in Linux kernel)


README.md
```diff
@@ -27,9 +27,9 @@ It was heavily inspired by the ftrace framework of the Linux kernel and
 the name uftrace stems from the combination of user and ftrace.
 
 It can record data from:
-- User space functions by dynamically patching functions at runtime, or with code
-  compiled with with `-pg`, `-fpatchable-function-entry`, `-mfentry`, `-mnop-mcount`,
-  `-finstrument-functions`, or using clang with `-fxray-instrument`.
+- Functions in the executable by dynamically patching functions at runtime using `-P.`.
+- Code compiled with `-pg`, `-f patchable-function-entry`, `-m fentry`, `-m nop-mcount`,
+  `-f instrument-functions` and by the clang-specific `-f xray-instrument`.
```
namhyung, line 32: Maybe we can remove the details about the compiler options and just say that function can come from either instrumented binary or regular binary.

README.md
```diff
@@ -27,9 +27,9 @@ It was heavily inspired by the ftrace framework of the Linux kernel and
 the name uftrace stems from the combination of user and ftrace.
 
 It can record data from:
-- User space functions by dynamically patching functions at runtime, or with code
-  compiled with with `-pg`, `-fpatchable-function-entry`, `-mfentry`, `-mnop-mcount`,
-  `-finstrument-functions`, or using clang with `-fxray-instrument`.
+- Functions in the executable by dynamically patching functions at runtime using `-P.`.
+- Code compiled with `-pg`, `-f patchable-function-entry`, `-m fentry`, `-m nop-mcount`,
```
honggyukim, line 31: This change makes the compiler option incorrect.
- `-f patchable-function-entry` -> `-fpatchable-function-entry`
- `-m fentry` -> `-mfentry`
- `-m nop-mcount` -> `-mnop-mcount`

honggyukim (commented):
Thanks for your work again. I also have a few comments.

README.md
bernhardkaindl: Thank you very much, done and pushed!

README.md
bernhardkaindl: Thanks, I've removed the spaces again!

README.md
```diff
@@ -27,9 +27,8 @@ It was heavily inspired by the ftrace framework of the Linux kernel and
 the name uftrace stems from the combination of user and ftrace.
 
 It can record data from:
-- User space functions by dynamically patching functions at runtime, or with code
-  compiled with with `-pg`, `-fpatchable-function-entry`, `-mfentry`, `-mnop-mcount`,
-  `-finstrument-functions`, or using clang with `-fxray-instrument`.
+- Functions in the executable by dynamically patching functions at runtime using `-P.`.
```
honggyukim, line 30: uftrace can trace functions inside share libraries as well so I think the "executable" is not enough. I think we better go back to "User space functions" instead rather than "Functions in the executable and shared libraries".

namhyung (commented):
Please add your Signed-off-by at the end of the commit message.  Now I've realized that you missed it for previous patches too.  I won't change the commit history just because of that.  Can you confirm that I can assume your sign-off on your earlier changes?

bernhardkaindl:
@namhyung Added the missing Sign-off in the commit message added declaration that I Sign-off on my previous commits as well. You can merge it now.

namhyung:
The same comment for the another PR.  Please update the subject prefix to "doc:" and use the real email address.

bernhardkaindl:
> Please update the subject prefix to "doc:" and use the real email address.
@namhyung Done.

namhyung (approved):
LGTM

Issue #1406, opened 2022-02-04
Support recording with pre-saved symbols


namhyung:
We have `--with-syms` option for `uftrace replay` to symbolize data using external symbol files.  Extend it to `uftrace record` so we can use stripped binary for recording.  Symbols and debug info should be recorded separately before the strip.  And filenames should not be changed.

Example:
```
$ gcc -o myprog -g -pg myprog.c
$ uftrace record -d myprog-sym.data  myprog

$ strip myprog
$ uftrace myprog
# ... does not have symbols

$ uftrace record --with-syms myprog-sym.data  myprog
$ uftrace replay
# ... it has symbols!!
```

It also supports filtering and debug info.  During the record, libmcount will read symbols and debug info from the files in the symbol directory.  After that it copies the files into the current data directory so that uftrace can read the data as if it has symbols from the beginning.

I've pushed the code to [review/with-syms-v3](../compare/review/with-syms-v3).

honggyukim:
I didn't carefully look into the changes in details, but I tested and found it works as expected. Thanks for the work!

PR #1671, opened 2023-04-10
Obsolete Draft: RFC for split debuginfo using libdw's frontend API

bernhardkaindl:
Update: This was only a first attempt to serve as a request for comments! - Thanks!

I plan to submit a new PR based on a new example (which I have to test first).

See my answer to the review of this PR for more information:  https://github.com/namhyung/uftrace/pull/1671#issuecomment-1505403698 

----
First step towards supporting split or separate DWARF debuginfo!

Convert dwarf/debug.c to use libdwfl (dwarf frontend lib) functions of libdw. uftrace now loads DWARF debuginfo from executables created with:
```ini
cp $exe $exe.full
objcopy --add-gnu-debuglink=$exe.full $exe
strip -g $exe
```
Using "strip $exe" instead of "strip -g $exe" is not yet supported, but could be after further investigation. For now this means that the DWARF debuginfo for programs shipped by most Linux distro is not yet used, but should be possible later.

TODO: Need to compare the test suite before/after this change.

In order to support showing stripped symbols, further updates appear to be needed.
As far as i can see, uftrace currently uses the ELF symbol table exclusively to generate the `uftrace.data/filename.sym` files and this excludes stripped debug info of functions which aren't in the ELF symbol table:
https://github.com/namhyung/uftrace/blob/master/utils/dwarf.c#L1486

This would likely have to be changed to generate the content of the symtab array from the debug symbols, I guess.

### DIFF
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -19,6 +19,15 @@
 #include "utils/symbol.h"
 #include "utils/utils.h"
 
+// Used by libdw in build_debug_info_cb after setup_debug_info returns,
+// so this may not be allocated temporary on setup_debug_info's stack (which is freed):
+char *debuginfo_path = NULL;
+Dwfl_Callbacks callbacks = {
+	.find_elf = dwfl_linux_proc_find_elf,
+	.find_debuginfo = dwfl_standard_find_debuginfo,
+	.debuginfo_path = &debuginfo_path,
+};
+
 bool debug_info_has_argspec(struct uftrace_dbg_info *dinfo)
 {
 	if (dinfo == NULL)
@@ -208,8 +217,9 @@ struct cu_files {
 static int elf_file_type(struct uftrace_dbg_info *dinfo)
 {
 	GElf_Ehdr ehdr;
+	GElf_Addr bias;
 
-	if (dinfo->dw && gelf_getehdr(dwarf_getelf(dinfo->dw), &ehdr))
+	if (gelf_getehdr(dwfl_module_getelf(dinfo->dwfl_module, &bias), &ehdr))
 		return ehdr.e_type;
 
 	return ET_NONE;
@@ -262,18 +272,18 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
 	if (force || check_trace_functions(filename) != TRACE_CYGPROF)
 		dinfo->needs_args = true;
 
-	pr_dbg2("setup dwarf debug info for %s\n", filename);
-
 	fd = open(filename, O_RDONLY);
 	if (fd < 0) {
 		pr_dbg2("cannot open debug info for %s: %m\n", filename);
 		return -1;
 	}
 
-	dinfo->dw = dwarf_begin(fd, DWARF_C_READ);
+	dinfo->dwfl = dwfl_begin(&callbacks);
+	dinfo->dwfl_module = dwfl_report_elf(dinfo->dwfl, filename, filename, fd, 0, true);
+	dwfl_report_end(dinfo->dwfl, NULL, NULL);
 	close(fd);
 
-	if (dinfo->dw == NULL) {
+	if (dinfo->dwfl == NULL || dinfo->dwfl_module == NULL) {
 		pr_dbg2("failed to setup debug info: %s\n", dwarf_errmsg(dwarf_errno()));
 		return -1;
 	}
@@ -295,11 +305,11 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
 
 static void release_dwarf_info(struct uftrace_dbg_info *dinfo)
 {
-	if (dinfo->dw == NULL)
+	if (dinfo->dwfl == NULL)
 		return;
 
-	dwarf_end(dinfo->dw);
-	dinfo->dw = NULL;
+	dwfl_end(dinfo->dwfl);
+	dinfo->dwfl = NULL;
 }
 
 #define ARGSPEC_MAX_SIZE 256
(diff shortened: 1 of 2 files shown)


utils/dwarf.c
```diff
@@ -1797,6 +1825,10 @@ void prepare_debug_info(struct uftrace_sym_info *sinfo, enum uftrace_pattern_typ
 	struct uftrace_mmap *map;
 	struct strv dwarf_args = STRV_INIT;
 	struct strv dwarf_rets = STRV_INIT;
+	const char *debuginfod_urls = getenv("DEBUGINFOD_URLS");
+
+	if (debuginfod_urls)
+		unsetenv("DEBUGINFOD_URLS");
```
namhyung, line 1831: I think this part can be separated.

utils/dwarf.c
```diff
@@ -19,6 +19,15 @@
 #include "utils/symbol.h"
 #include "utils/utils.h"
 
+// Used by libdw in build_debug_info_cb after setup_debug_info returns,
+// so this may not be allocated temporary on setup_debug_info's stack (which is freed):
```
namhyung, line 23: C style comments please, here and other places.

utils/dwarf.c
```diff
@@ -19,6 +19,15 @@
...
+char *debuginfo_path = NULL;
+Dwfl_Callbacks callbacks = {
+	.find_elf = dwfl_linux_proc_find_elf,
+	.find_debuginfo = dwfl_standard_find_debuginfo,
+	.debuginfo_path = &debuginfo_path,
+};
```
namhyung, line 29: It seems they both can be declared as static.

utils/dwarf.c
```diff
@@ -262,18 +272,18 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
...
 	fd = open(filename, O_RDONLY);
 	if (fd < 0) {
 		pr_dbg2("cannot open debug info for %s: %m\n", filename);
 		return -1;
 	}
 
-	dinfo->dw = dwarf_begin(fd, DWARF_C_READ);
+	dinfo->dwfl = dwfl_begin(&callbacks);
+	dinfo->dwfl_module = dwfl_report_elf(dinfo->dwfl, filename, filename, fd, 0, true);
+	dwfl_report_end(dinfo->dwfl, NULL, NULL);
```
namhyung, line 283: I was worrying about the potential version compatibility issues, but it seems the dwfl APIs were supported long ago.

utils/dwarf.c
```diff
@@ -1624,9 +1634,16 @@ static char *get_base_comp_dir(struct rb_root *dirs)
 	return max->name;
 }
 
-static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symtab *symtab,
-			     enum uftrace_pattern_type ptype, struct strv *args, struct strv *rets)
+static int build_dwarf_info_cb(Dwfl_Module *mod __attribute__((unused)),
+			       void **userdata __attribute__((unused)), const char *name,
```
namhyung, line 1638: uftrace uses `-Wno-unused-parameter`, I think the `unused` attributes can go away.

utils/dwarf.c
```diff
@@ -1707,14 +1721,28 @@ static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symt
...
+}
+
+// wrap build_dwarf_info_cb() using dwfl_getdwarf() for slpit debuginfo support:
+static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symtab *symtab,
+			     enum uftrace_pattern_type ptype, struct strv *args, struct strv *rets)
+{
+	ptrdiff_t p = 0;
+	struct dwarf_info_args callargs = {
+		.dinfo = dinfo, symtab = symtab, ptype = ptype, args = args, rets = rets
+	};
```
namhyung, line 1734: Please add a blank line between declarations and the statements.

utils/dwarf.c
```diff
@@ -1707,14 +1721,28 @@ static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symt
 	for (i = 0; i < rets->nr; i++)
 		free_filter_pattern(&ret_patt[i]);
 	free(ret_patt);
+	return DWARF_CB_OK;
+}
+
+// wrap build_dwarf_info_cb() using dwfl_getdwarf() for slpit debuginfo support:
```
namhyung, line 1727: C style please, and there's a typo `split`.

utils/dwarf.c
```diff
@@ -1707,14 +1721,28 @@ static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symt
...
+static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symtab *symtab,
+			     enum uftrace_pattern_type ptype, struct strv *args, struct strv *rets)
+{
+	ptrdiff_t p = 0;
+	struct dwarf_info_args callargs = {
+		.dinfo = dinfo, symtab = symtab, ptype = ptype, args = args, rets = rets
+	};
+	do {
+		p = dwfl_getdwarf(dinfo->dwfl, &build_dwarf_info_cb, &callargs, p);
+	} while (p > 0);
```
namhyung, line 1737: Why is the loop needed?  Mind adding some comments?

utils/dwarf.h
```diff
@@ -34,8 +34,9 @@ struct uftrace_dbg_loc {
 };
 
 struct uftrace_dbg_info {
-	/* opaque DWARF info pointer */
-	Dwarf *dw;
+	/* opaque DWARF frontend library pointers */
+	Dwfl *dwfl;
+	Dwfl_Module *dwfl_module;
```
namhyung, line 39: Note that `struct uftrace_dbg_info` is used even when `HAVE_LIBDW` is NOT defined.  You'd better defined the data types as `void` like above.

utils/dwarf.h
```diff
@@ -62,6 +63,15 @@ struct uftrace_dbg_info {
...
 };
 
+// arg struct for passing infos to and from build_dwarf_info_cb():
+struct dwarf_info_args {
+	struct uftrace_dbg_info *dinfo;
+	struct uftrace_symtab *symtab;
+	enum uftrace_pattern_type ptype;
+	struct strv *args;
+	struct strv *rets;
+};
```
namhyung, line 73: I think this is needed only if it has libdw dependency.  How about moving it to the .c file?

namhyung (commented):
Thanks a lot for working on this.  We had #1406 as a work around for the stripped binaries.  But it'd be much better if it can support split debug files.

honggyukim:
@bernhardkaindl Thanks very much for this work! I couldn't look into the details now but I have tested it and it works great!

bernhardkaindl:
Thanks for the encouraging comments!

This was based on my initial limited understanding of libdw.

The more I look, the more I am able to understand.

Like the elfutils developes admit as well, there is no documentation at all for how to
use the libdw frontend API. I was only able to go by examples and `elfutils/libdwfl.h`.

However, I've not found a better example which I could be able to apply:
https://bugzilla.redhat.com/show_bug.cgi?id=450218#c9

If this example approach works for uftrace as well, using libdwfl becomes simpler.

Thanks to your questions, I've seen a lot of fixes and improvements to make.

I plan to let this stand as a first RFC for now and refactor it based on this link.

Due to the amount of improvements, I'll submit a new PR, maybe by next week or so.

namhyung:
Hi @bernhardkaindl I'd like to try something based on your change.  Can I assume your Signed-off-by here?

namhyung:
Closed by 18d225ad.

PR #1672, opened 2023-04-12
tests: Apply aarch64 fix only on aarch64 machines

clementguidi:
Hi,

@bernhardkaindl you add `-N memcpy` for all test commands in  #1659. If you agree, we'd better do it on aarch64 only so the commands don't get too verbose on other platforms.

Thank you.


honggyukim:
Thanks for the patch, but I think it might be better to keep the filter same across all the architectures. We can't test examples that have `memcpy` but we don't really have to test it.

PR #1678, opened 2023-04-24
mcount: Add RCU-like capabilities to the rbtree

clementguidi:
Hello Namhyung and Honggyu,

I'm submitting this pull request to make it easier to implement #1645.

The idea is to interact with the `mcount_triggers` rbtree in an RCU-like fashion. The approach is simple: I deep copy the whole tree when the agent want to modify it. Then I atomically swap the pointers when the agent is done (on connection end).

I use an arbitrary grace period of `1s`, which blocks the agent. What do you think?

Note: If we are to use hash maps instead of an rbtree in the future, we could take advantage of [liburcu](https://liburcu.org/), which defines RCU-capable hash maps (but not rbtrees). I don't plan to be able to work on this though.

### DIFF
--- a/gdb/uftrace/rbtree.py
+++ b/gdb/uftrace/rbtree.py
@@ -17,6 +17,160 @@ from uftrace import utils
 rb_root = utils.CachedType("struct rb_root")
 rb_node = utils.CachedType("struct rb_node")
 
+def rb_color(node):
+    """Return the color of a node.
+    red -> 0 | black -> 1"""
+    if node.address == 0:
+        return 1
+
+    if node['rb_parent_color'] % 2 == 0:
+        return 0
+    else:
+        return 1
+
+def rb_check(node, val_min=-1, val_max=-1, gdbtype=None, val_field="start"):
+    if node.address == 0:
+        return 1
+
+    # check order
+    if gdbtype is not None:
+        node_container = utils.container_of(node.address, gdbtype.pointer(), "node").dereference()
+        val = int(node_container[val_field])
+        if val < val_min:
+            gdb.write(f"node {node.address} is not ordered (val={val} < min={val_min})\n")
+            return -1
+        if val > val_max and val_max != -1: # use -1 as infinity value for val_max
+            gdb.write(f"node {node.address} is not ordered (val={val} > max={val_max})\n")
+            return -1
+    else:
+        val = -1
+
+    left = node['rb_left'].dereference()
+    right = node['rb_right'].dereference()
+
+    # check that a red node has black children
+    if rb_color(node) == 0:
+        if rb_color(left) == 0:
+            gdb.write(f"red node {node.address} has red left child {left}\n")
+            return -1
+        if rb_color(right) == 0:
+            gdb.write(f"red node {node.address} has red right child {right}\n")
+            return -1
+
+    # recursively check that paths to NULL leafs have as many black nodes
+    left_black_count = rb_check(left, val_min, val, gdbtype)
+    if left_black_count == -1:
+        return -1
+
+    right_black_count = rb_check(right, val, val_max, gdbtype)
+    if right_black_count == -1:
+        return -1
+
+    if left_black_count != right_black_count:
+        gdb.write(f"node @ {node.address}: {left_black_count} on left != {right_black_count} on right\n")
+        return -1
+    else:
+        black_count = left_black_count
+        if rb_color(node) == 1:
+            black_count += 1
+        return black_count
+
+
+class UftRbtreeCheck(gdb.Command):
+    """Check if a rbtree has a valid structure.
+
+    A red-black tree is a binary search tree with the following constraints:
+        1. Every node is either red or black
+        2. All NULL leafs are defined as black
+        3. A red node does not have a red child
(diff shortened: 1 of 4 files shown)


clementguidi:
I'll be checking the ASAN error soon.

libmcount/mcount.c
```diff
@@ -2404,6 +2404,9 @@ static void setup_mcount_test(void)
 	mcount_exename = read_exename();
 	pthread_key_create(&mtd_key, mtd_dtor);
 	mcount_global_flags = 0;
+
+	mcount_triggers = xmalloc(sizeof(*mcount_triggers));
+	*mcount_triggers = RB_ROOT;
```
namhyung, line 2409: Can you please move it to the first commit?

libmcount/mcount.c
```diff
@@ -1837,6 +1837,21 @@ static struct rb_root *deep_copy_triggers(struct rb_root *src)
...
+ * swap_triggers - atomically swap the pointer to a filter rbtree and free the
+ * old one
+ * @old - pointer to the tree to deprecate
+ * @new - new version of the tree to use
+ */
+static void swap_triggers(struct rb_root **old, struct rb_root *new)
+{
+	struct rb_root *tmp;
+	tmp = __sync_val_compare_and_swap(old, *old, new);
+	sleep(1); /* RCU-like grace period */
```
namhyung, line 1850: I'm not sure 1 second sleep is good enough.  But I don't have a good alternative now.  Maybe we can consider more sophisticated grace period handling later, but let's start with the simple approach.

libmcount/mcount.c
```diff
@@ -1762,6 +1762,81 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+ * 'struct uftrace_filter' nodes
+ * @dest - pointer to the address of the copy (modified in place)
+ * @src  - original rbtree
+ */
+static void __deep_copy_triggers(struct rb_node **dest, struct rb_node *src)
+{
+	struct uftrace_filter *old, *new;
+
+	if (!src)
+		return;
```
namhyung, line 1807: Doesn't it need to reset the `*dest` here?

gdb/uftrace/rbtree.py
```diff
@@ -17,6 +17,115 @@
...
+        if rb_color(node) == 1:
+            black_count += 1
+        return black_count
+
+
+class RbTreeCheck(gdb.Command):
+    """Check is a rbtree is valid."""
+
+    def __init__(self):
+        super(RbTreeCheck, self).__init__("rb-tree-check", gdb.COMMAND_DATA)
```
namhyung, line 70: I think the convention is to start the command name with "uft-" prefix.  How about "uft-rbtree-check"?

gdb/uftrace/rbtree.py
```diff
@@ -17,6 +17,115 @@
...
+        return -1
+    else:
+        black_count = left_black_count
+        if rb_color(node) == 1:
+            black_count += 1
+        return black_count
+
+
+class RbTreeCheck(gdb.Command):
+    """Check is a rbtree is valid."""
```
namhyung, line 67: Can you please describe what to check briefly?  I almost forgot about the detail of the RB tree constraints :-p

gdb/uftrace/rbtree.py
```diff
@@ -17,6 +17,115 @@
...
+
+def rb_print(node):
+    __rb_print(node, 0)
+
+
+class RbTreePrint(gdb.Command):
+    """Display a textual representation of an rbtree."""
+
+    def __init__(self):
+        super(RbTreePrint, self).__init__("rb-tree-print", gdb.COMMAND_DATA)
```
namhyung, line 113: Ditto.  "uft-rbtree-print" ?

namhyung (commented):
Thanks for working on this!

libmcount/mcount.c
namhyung: I was wondering if some kind of atomic read is required for the read side, but it seems `mcount_entry_filter_check()` only passes the root of the tree once.  Looks ok for now.

libmcount/mcount.c
```diff
@@ -1762,6 +1762,83 @@ static void mcount_script_init(enum uftrace_pattern_type patt_type)
...
+ * @return - deep copy
+ */
+static struct rb_root *deep_copy_triggers(struct rb_root *src)
+{
+	struct rb_root *new;
+	new = xmalloc(sizeof(*new));
+	*new = RB_ROOT;
+	__deep_copy_triggers(&new->rb_node, src->rb_node);
+	return new;
+}
```
namhyung, line 1840: Can you please move this logic to utils/filter.c?  I think it's better to keep the related code together for the future changes don't miss to update any fields.  Also please add a unit test for the deep copy to verify there's no leak (we run the test with asan).

clementguidi: I added a unit test, with a simple value check on the deep copy. I don't check is the tree is valid though.

namhyung: I think that's fine, thanks.

namhyung (approved):
LGTM

PR #1687, opened 2023-04-30
record: Honour --report option to record command

honggyukim:
The current --report option is only used in live command, but it's more useful when users want to see only report result without running uftrace record and uftrace report twice.

The simple usage is as follows.
```
  $ uftrace record --report t-abc
      1.480 us    0.170 us           1  main
      1.310 us    0.141 us           1  a
      1.169 us    0.157 us           1  b
      1.012 us    0.309 us           1  c
      0.703 us    0.703 us           1  getpid
      0.666 us    0.666 us           1  __monstartup
      0.132 us    0.132 us           1  __cxa_atexit
```

### DIFF
--- a/cmds/live.c
+++ b/cmds/live.c
@@ -426,6 +426,7 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 		return forward_options(opts);
 
 	ret = command_record(argc, argv, opts);
+	pr_dbg("live-record finished.. \n");
 	if (!can_skip_replay(opts, ret)) {
 		int ret2;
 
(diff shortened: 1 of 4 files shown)


namhyung (commented):
Please do not do this.  It should belong to the live command and there's no point to have it in the record.

honggyukim:
Hmm.. then we better show report output only without replay. I just wanted to have a way to show report only.

namhyung:
Why not running `uftrace record` and `uftrace report`? :)

honggyukim:
Can do but I think the current --report is quite unhelpful in live command. It’s a matter of preference. :)

namhyung:
I believe the record command should do its job only.  When we add `--report` to it, there's no reason not to add others like replay or graph.  I admit that adding `--report` to live is quite an ad hoc solution.  Maybe we can remove the replay output there.

Issue #1690, opened 2023-05-01
unable to identify shebang when it has initial spaces


honggyukim:
uftrace detects shebang in the first line as follows.
```
#!/usr/bin/env python3
```
But if it has some whitespaces at the beginning, it can't detect the shebang as follows.
```
#! /usr/bin/env python3
```
For example, when there is a script as follows.
```py
$ cat p-abc.py
#! /usr/bin/env python3

import os

def c():
    return os.getpid()

def b():
    return c()

def a():
    return b()

a()
```
The tracing result looks as follows.
```
$ uftrace p-abc.py
uftrace: /home/honggyu/work/uftrace/cmds/record.c:1580:find_in_path
 ERROR: Cannot trace '': No such executable file.
```
The shebang with whitespaces is legal so we should fix our shebang detection logic.

PR #1696, opened 2023-05-10
fix: #1690 unable to identify shebang when it has initial spaces

yihong0618:
fix: #1690 
Signed-off-by:  yihong0618 <zouzou0208@gmail.com>

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1608,7 +1608,8 @@ again:
 		pr_err("Cannot read '%s'", opts->exename);
 
 	if (memcmp(elf_ident, ELFMAG, SELFMAG)) {
-		char *script = check_script_file(opts->exename);
+		// for issue 1690 unable to identify shebang when it has initial spaces
+		char *script = str_trim(check_script_file(opts->exename));
 		char *p;
 
 		if (script == NULL)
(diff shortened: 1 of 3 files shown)


cmds/record.c
```diff
@@ -19,6 +19,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#include "libtraceevent/event-utils.h"
```
namhyung: I don't want to include libtraceevent headers just because of this.  Maybe we can implement `strim()` instead.

namhyung (commented):
Can you please squash the commits into one and rewrite the commit message like this?

```
record: skip whitespaces after shebang for scripts

Python tracing won't work when the shebang line has a space like below:

    #! /usr/bin/python3

Fixed: #1690 

Signed-off-by: ...
```

yihong0618:
> Can you please squash the commits into one and rewrite the commit message like this?
> 
> ```
> record: skip whitespaces after shebang for scripts
> 
> Python tracing won't work when the shebang line has a space like below:
> 
>     #! /usr/bin/python3
> 
> Fixed: #1690 
> 
> Signed-off-by: ...
> ```

of course

cmds/record.c
yihong0618: will do this tonight

namhyung:
Next time, you can reuse this PR instead of creating a new one.  You don't need to create a new PR whenever you update the code.  Just force-push is fine.

yihong0618:
copy that, thanks

Issue #1369, opened 2021-10-05
Runtime dynamic patching on x86_64


yottapulsky:
# Introduction

We would like to bring _safe and concurrent runtime patching_ for the x86_64 architecture, so binaries that are not compiled with special options can be instrumented at runtime.

This issue is a place to discuss informally about pull request #1274.

We propose a progressive strategy, in which we want to gradually improve the efficiency of our methods.

# Previous work

PR #1274 relies on other work. The most recent commits are based on other pull requests, which should be accepted first. See below. Once they are merged, PR #1274 can be stripped from the commits that belong to previous PR, so it is actually smaller.

## Client server architecture

The client command discussed in #1269 serves as an entry point for dynamic patching. It provides a way to forward the name of the functions to (un)patch at runtime to a running uftrace instance.

- [ ] Need to solve issue #1330 before

## Meson build system

The Meson build system was introduced in #1214. It is actually __independent__ from this work, but is currently used to compile the new code. Makefile can be supported too.

# Patching strategy

## Progressive approach

We believe that we should start by implementing simpler methods, which performance may not be production-ready. Users can thus get familiar with the new features and use them in basic cases.

Once these changes get accepted, we can work on more sophisticated and fine grained methods, that enhance the performance (coverage, overhead) but need more effort.

See for example NOProbe[1] and Instruction punning[2].

## Current implementation

The proposed PR #1274 works as follows. Various hashmaps are used to store the correspondance between original instructions and the location of trampolines.

### Step 1 - Insert temporary int3 trap 

Fist we insert a 1 byte int3 trap, so incoming threads will be interrupted before reaching the critical section located after it, that is to be modified (the patching region).

The signal handler redirects the threads to an out of line execution (OLX) buffer.

### Step 2 - Move threads out of the critical section

Then, we make sure that no thread is currently executing code in the patching region. We send a SIGRTMIN+n signal to each thread. The signal handler checks that the instruction pointer isn't in any patching region. If so, it redirects it to an OLX.

### Step 3 - Patch the functions

The third step is to patch the now-safe regions with the address for the jump instruction. The int3 trap remains untouched, and will be replaced at the end of the process.

Before doing so, the processors are synchronized (membarrier system call with MEMBARRIER\_CMD\_PRIVATE\_EXPEDITED\_SYNC\_CORE).

### Step 4 - Remove the int3 trap

Eventually, the trap is replaced _atomically_ by the first byte of the jump instruction, effectively redirecting to the trampoline.

[...]

PR #1698, opened 2023-05-10
dynamic: Runtime dynamic instrumentation on x86_64

clementguidi:
Hello Namhyung and Honggyu,

This is a attempt to implement runtime dynamic instrumentation on x86_64 architecture. That is, performing full dynamic instrumentation on a live target.

This PR is long. If you want me to break it down, please advise me.

This feature, and technical details were already discussed in #1274 and #1369, so I'll be relying on this past material. Please ask me if you need me to elaborate on this PR.

It consists in 8 blocks of commits, as follows.

### 1. Refactor code for multiple runtime executions

The idea is that the agent will trigger the patching process at arbitrary moments, and an arbitrary amount of times. We refactor initialization and cleanup functions so they can be executed anytime at runtime, instead of when the target starts. We also make sure that they support multiple calls.

### 2. Add utils and wrappers

We introduce configuration options to use specific memory barriers when available. And also system call wrappers for older glibc versions. And last, we implement signal-related helpers.

### 3. Implement naive runtime dynamic patching

Core feature: we add safety mechanisms that allow to perform dynamic patching while code is executing. We make use of memory barriers or serializing instructions, and signals to ensure safe execution of cross-modifying code. This first implementation is based on the current patching process. We make heavy use of signals, and this has a strong negative impact on performance. We optimize the process later.

The patching process is seen as atomic regarding cores' execution. It follows these steps:
1. Critical region locking: insert a trap a the start of the patching region so no new thread can enter it
2. Serialization: synchronize execution across cores so each one is up to date on cross-modified instructions
3. Thread eviction: clear the critical region of thread that might have entered it before the previous step
4. Patching: move original code out of line at replace it with a relative call instruction (except the first byte, preserving the trap)
5. Critical region unlocking: replace the trap with the first byte of the call instruction

### 4. Optimize signal use

We batch operations in a way to minimize the amount of signals that are emitted. After the patching regions are locked, we issue a single signal to clear them all from threads that could still execute them. Another signal is used for serialization when the required membarrier is unavailable. Then we patch the locked regions.

This approach uses only one (or two depending on the kernel version) signal, but increases the amount of time during which traps are installed and can be executed.

### 5. Implement optimized unpatching

We implement unpatching in a similar way that we did for unpatching, but instrumentation is replaced by original instructions.

### 6. Trigger (un)patching from the agent

We hook the relevant to the agent so it can perform runtime dynamic instrumentation when running.

### 7. Preserve original behavior

We make sure that our new process in only used when the target is execution, falling back to the original behavior of uftrace when the target is instrumented before being executed.

### 8. Display stats

We eventually reset and display dynamic statistics every time patching or unpatching occurs.

Based on: #1274
Related: #1369 
See also: #870

### DIFF
--- a/arch/aarch64/mcount-insn.c
+++ b/arch/aarch64/mcount-insn.c
@@ -8,10 +8,19 @@
 #include <capstone/capstone.h>
 #include <capstone/platform.h>
 
+/**
+ * mcount_disasm_init - initialize the capstone engine once
+ * @disasm - mcount disassembly engine
+ */
 void mcount_disasm_init(struct mcount_disasm_engine *disasm)
 {
+	if (disasm->engine)
+		return;
+
+	pr_dbg2("initialize disassembly engine\n");
+
 	if (cs_open(CS_ARCH_ARM64, CS_MODE_ARM, &disasm->engine) != CS_ERR_OK) {
-		pr_dbg("failed to init Capstone disasm engine\n");
+		pr_dbg("failed to init capstone disasm engine\n");
 		return;
 	}
 
--- a/arch/x86_64/mcount-dynamic.c
+++ b/arch/x86_64/mcount-dynamic.c
@@ -1,4 +1,5 @@
 #include <errno.h>
+#include <semaphore.h>
 #include <string.h>
 #include <sys/mman.h>
 #include <unistd.h>
@@ -7,7 +8,6 @@
 #include <linux/membarrier.h>
 #else
 #include <cpuid.h>
-#include <semaphore.h>
 #endif
 
 /* This should be defined before #include "utils.h" */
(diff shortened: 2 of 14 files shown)


clementguidi:
Sure. I believe it's a good thing to have an overview of the full feature set, because the first patches don't make much sense otherwise. Should I leave this PR open (maybe mark it as draft)? So we can refer to it when discussing design choices in the first patches that are useful for later patches.

honggyukim:
Sure. If you think that’s more sensible then just leave it.

namhyung:
I think I can take a look after finishing the agent changes and make a release.

clementguidi:
Cool! I also want to let you know that I'm working on x86 full dynamic and runtime dynamic instrumentation.

Issue #1685, opened 2023-04-30
python tracing can't have symbols with `os._exit(0)`


honggyukim:
When the target python has `os._exit(0)` inside, uftrace can't generate symbol tables so the tracing result looks bad.

If `p-abc.py` has `os._exit(0)` at the end as follows.
```py
$ cat p-abc.py 
#!/usr/bin/env python3

import os

def c():
    return os.getpid() % 100000

def b():
    return c() + 1

def a():
    return b() - 1

a()
os._exit(0)
```
The tracing result looks as follows.
```
$ uftrace p-abc.py 
# DURATION     TID     FUNCTION
            [ 38580] | <1>() {
            [ 38580] |   <2>() {
            [ 38580] |     <3>() {
            [ 38580] |       <4>() {
   1.350 us [ 38580] |         <5>();
   8.483 us [ 38580] |       } /* <4> */
  11.374 us [ 38580] |     } /* <3> */
  14.137 us [ 38580] |   } /* <2> */
            [ 38580] |   /* linux:task-exit */

uftrace stopped tracing with remaining functions
================================================
task: 38580
[0] <1>
```

honggyukim:
It looks `os._exit`doesn't call destructor so can't write the symbol table at the end.

> Exit the process with status n, without calling cleanup handlers, flushing stdio buffers, etc. 

https://docs.python.org/3/library/os.html#os._exit

It would be helpful if we can override `os._exit` but not sure if it's possible.

honggyukim:
Maybe with the same reason, if a user press `Ctrl+C` then uftrace doesn't have a chance to write symbols.

honggyukim:
We have to do something like in `uftrace.py`.
```py
os_exit = os._exit

def on_exit(n):
    uftrace_python.dtor()    # calls uftrace_trace_python_finish() in trace-python.c
    os_exit(n)

os._exit = on_exit
```
This might also have to be done in C level inside `trace-python.c` directly.

namhyung:
Hmm.. that means we need to intercept every possible path to the exit.  I'm not sure we can do that.

honggyukim:
After install `mypy`, it shows an error in normal execution as follows.
```
$ python3 -m pip install mypy
Defaulting to user installation because normal site-packages is not writeable
Collecting mypy
  Downloading mypy-1.2.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (12.2 MB)
[...]

PR #1700, opened 2023-05-11
python tracing can't have symbols with os._exit(0)

yihong0618:
This patch use a tricky way by replace os._exit to python can trace symbol exit that fixed the problem.

Signed-off-by: Yi Hong [zouzou0208@gmail.com](mailto:zouzou0208@gmail.com)

Fixed: #1685

### DIFF
--- a/python/uftrace.py
+++ b/python/uftrace.py
@@ -31,5 +31,7 @@ sys.path.insert(0, os.path.dirname(pathname))
 
 code = open(sys.argv[0]).read()
 sys.setprofile(uftrace_python.trace)
+# Python tracing can't have symbols with os._exit(0), fix issue #1685
+code = code.replace("os._exit", "exit")
 exec(code, new_globals)
 sys.setprofile(None)


honggyukim:
Thanks for the PR. My test result looks good.

I know that this can just be a workaround because it replaces `os._exit` string to `exit` blindly so it replaces all the matches even if the string `os._exit` is in comments.

Moreover, if user's code has a module name `eos` and if it also has `_exit` inside by any chance, this patch will replace `eos._exit` to `eexit` and this is clearly an incorrect replacement.

Although having those problems, I think this will serve well for our problem in most cases so it would be better to get this in and find a proper solution having more time. We can create a follow up issue for this.

@namhyung What do you think?

yihong0618:
> Thanks for the PR. My test result looks good.
> 
> 
> 
> I know that this can just be a workaround because it replaces `os._exit` string to `exit` blindly so it replaces all the matches even if the string `os._exit` is in comments.
> 
> 
> 
> Moreover, if user's code has a module name `eos` and if it also has `_exit` inside by any chance, this patch will replace `eos._exit` to `eexit` and this is clearly an incorrect replacement.
> 
> 
> 
> Although having those problems, I think this will serve well for our problem in most cases so it would be better to get this in and find a proper solution having more time. We can create a follow up issue for this.
> 
> 
> 
> @namhyung What do you think?

maybe we can use re to replace, to ignore the exactly os._exit to avoid eos._exit or something like that.

namhyung:
https://docs.python.org/3/library/os.html#os._exit says it can be used by a child process after fork().  I'm afraid it'd change the behavior for the legitimate case.

namhyung:
Also I'm not sure if it fixes(?) the mypy case.

yihong0618:
@namhyung seems you are right, will take a look

yihong0618:
closing since have better one

Issue #1692, opened 2023-05-05
full dynamic tracing patching problem


honggyukim:
The original instructions before dynamic patching looks as follows.
```asm
0000000000ef0ac0 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE>:
  ef0ac0:       f3 0f 1e fa             endbr64
  ef0ac4:       85 d2                   test   %edx,%edx
  ef0ac6:       75 08                   jne    ef0ad0 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x10>
  ef0ac8:       c3                      ret
  ef0ac9:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
  ef0ad0:       48 81 ef b8 d2 00 00    sub    $0xd2b8,%rdi
  ef0ad7:       83 fa 08                cmp    $0x8,%edx
  ef0ada:       74 64                   je     ef0b40 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x80>
  ef0adc:       83 fa 10                cmp    $0x10,%edx
  ef0adf:       74 3f                   je     ef0b20 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x60>
  ef0ae1:       48 8b 87 f0 01 00 00    mov    0x1f0(%rdi),%rax
  ef0ae8:       48 89 06                mov    %rax,(%rsi)
  ef0aeb:       48 89 d0                mov    %rdx,%rax
  ef0aee:       48 c1 e0 20             shl    $0x20,%rax
  ef0af2:       48 89 46 08             mov    %rax,0x8(%rsi)
  ef0af6:       85 c9                   test   %ecx,%ecx
  ef0af8:       75 ce                   jne    ef0ac8 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x8>
```
In the above example, the `ret` instruction was patched and it makes the target program crashed.
```asm
ef0ac8:       c3                      ret
```
The `ret` could be moved to somewhere else but the instruction at `ef0af8` jumps to the address `ef0ac8` to make return so it might cause the crash because the `ret` instruction was patched. The following show the instructions after patching.
```asm
[...]

PR #1701, opened 2023-05-11
dynamic: x86_64: Do not allow ret insn for patch target

honggyukim:
Since 'ret' instruction can be a target of another jump instruction, it is not safe to patch and move the instruction somewhere else.

For example, the following instruction pattern must not be patched because the branch target of 'jne' instruction at ef0af8 is the address of the 'ret' instruction at ef0ac8.
```asm
  0000000000ef0ac0 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE>:
    ef0ac0:       f3 0f 1e fa             endbr64
    ef0ac4:       85 d2                   test   %edx,%edx
    ef0ac6:       75 08                   jne    ef0ad0 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x10>
    ef0ac8:       c3                      ret
        ...
    ef0af8:       75 ce                   jne    ef0ac8 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+0x8>
```
If the 'ret' is patched, then it gets crashed with SIGILL and the prologue of the function becomes as follows.
```asm
  $ cgdb --args uftrace record -P v8::internal::Heap::CreateFillerObjectAt --no-libcall ./node -e ''
        ...
  Thread 2.1 "node" received signal SIGILL, Illegal instruction.
  0x00005555562f0aca in v8::internal::Heap::CreateFillerObjectAt(unsigned long, int, v8::internal::ClearFreedMemoryMode) ()

   1├>Dump of assembler code for function _ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE:
   2│    0x00005555562f0ac0 <+0>:     endbr64
   3│    0x00005555562f0ac4 <+4>:     call   0x5555578eeff0
         ...
  16│    0x00005555562f0af8 <+56>:    jne    0x5555562f0ac8 <_ZN2v88internal4Heap20CreateFillerObjectAtEmiNS0_20ClearFreedMemoryModeE+8>
```
The 'call' instruction at 5555562f0ac4 is the patched one, but the target of 'jne' is pointing the address in the middle of the 'call' instruction, which causes the SIGILL.

To avoid this problem, we rejects the 'ret' instruction and more candidates to make it safer.

Fixed: #1692

### DIFF
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -40,6 +40,7 @@ enum fail_reason {
 	INSTRUMENT_FAIL_RELJMP = (1U << 1),
 	INSTRUMENT_FAIL_RELCALL = (1U << 2),
 	INSTRUMENT_FAIL_PIC = (1U << 3),
+	INSTRUMENT_FAIL_RET = (1U << 4),
 };
 
 enum branch_group {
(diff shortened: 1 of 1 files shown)


namhyung (approved):
LGTM

PR #1702, opened 2023-05-11
dynamic: Refactor logic to perform multiple updates

clementguidi:
This is the first PR of a series of patches intended to bring runtime dynamic tracing on x86_64 to uftrace.

First, we refactor the life cycle of dynamic structures so a dynamic update can be triggered at anytime during execution, any amount of times. This will allow the agent to call `mcount_dynamic_update()` multiple times at runtime.

Related: #1698

PR #1703, opened 2023-05-11
dynamic: Refactor patch pattern logic

clementguidi:
This is the second PR in a series of patches intended to bring runtime dynamic tracing on x86_64 to uftrace.
1. #1702 
2. #1703 🠈
3. #1704
4. #1705
5. #1745
6. #1746
7. #1747
8. #1748
9. #1749
10. #1750
11. #1751

We redefine the pattern handling logic to remove default behaviors that conflict with runtime dynamic patching. We adapt it to runtime dynamic unpatching.

Related: #1698

### DIFF
--- a/libmcount/dynamic.c
+++ b/libmcount/dynamic.c
@@ -384,10 +384,17 @@ static bool match_pattern_module(char *pathname)
 	return ret;
 }
 
-static bool match_pattern_list(struct uftrace_mmap *map, char *soname, char *sym_name)
+/**
+ * match_pattern_list - match a symbol name against a pattern list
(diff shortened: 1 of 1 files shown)


namhyung (approved):
I thought it should work similar to `-F` and `-N` so patch all functions when `-U` is given.  But I think it's different and we don't usually use `-P` and `-U` at the same time.  For example, `-pg` and `-mfentry` only needs `-U` option since all functions are already enabled (traceable).  While regular binary and `-fpatchable-function-entry` only needs `-P`.

So I think it's ok to change.

PR #1704, opened 2023-05-11
configure: Check for membarrier support

clementguidi:
This is the third PR in a series of patches intended to bring runtime dynamic tracing on x86_64 to uftrace.
1. #1702 
2. #1703 
3. #1704 🠈

Runtime dynamic patching requires cache synchronization. This is best achieved by using the `MEMBARRIER_CMD_PRIVATE_EXPEDITED_SYNC_CORE` memory barrier. However it was introduced in Linux 4.16. We set a flag
indicating whether this membarrier can be used or if libmcount has to rely on other mechanisms (e.g. signals).

We introduce a configuration option `--without-membarrier` to disable the use of `MEMBARRIER_CMD_PRIVATE_EXPEDITED_SYNC_CORE` even if it is available.


check-deps/__have_membarrier.c
```diff
@@ -0,0 +1,10 @@
+#include <linux/membarrier.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+int main()
+{
+	syscall(__NR_membarrier, MEMBARRIER_CMD_REGISTER_PRIVATE_EXPEDITED_SYNC_CORE, 0, 0);
+	syscall(__NR_membarrier, MEMBARRIER_CMD_PRIVATE_EXPEDITED_SYNC_CORE, 0, 0);
```
namhyung, line 8: By using `_PRIVATE_EXPEDITED_SYNC_CORE`, can we delete the old copy right after the syscall?

namhyung: Sorry for the late reply.  I think you want to use membarrier to update global state like in a RCU fashion.  Something like below:

```c
new_data = copy_data(shared_data);
add_some_data(new_data);

old_data = shared_data;
shared_data = new_data;  // WRITE_ONCE ?

membarrier();

free(old_data);
```

azharivs: That is somewhat true. We want to use membarriers to stop all running threads before entering patching/unpatching code as will be done in this other commit:
https://github.com/namhyung/uftrace/pull/1747/commits/f07ab57e33e6c2b33e8a45af83318a3712bbf4e0

Note that this particular PR merely checks for availability of membarrier system call. The actual usage is left for PR #1747

namhyung: I know, I was just curious how it's gonna be used.

PR #1705, opened 2023-05-11
utils: Add helpers for runtime dynamic patching

clementguidi:
This is the third PR in a series of patches intended to bring runtime dynamic tracing on x86_64 to uftrace.
1. #1702 
2. #1703 
3. #1704
4. #1705 🠈

We add syscall wrappers and signal-related functions that will be useful for runtime dynamic instrumentation.

Related: #1698


utils/utils.c
```diff
@@ -1032,6 +1033,31 @@ int copy_file(const char *path_in, const char *path_out)
 	return 0;
 }
 
+#ifndef gettid
+/**
+ * gettid - gettid syscall wrapper (glibc < 2.30)
+  * @return - thread id
+ */
+pid_t gettid()
```
namhyung, line 1041: Please add `void`.

utils/utils.c
```diff
@@ -1058,6 +1058,96 @@ int tgkill(pid_t tgid, pid_t tid, int signal)
 }
 #endif
 
+/**
+ * find_unused_sigrt - find a real-time signal with no associated action
+ * @return - unused RT signal, or -1 if none found
+ */
+int find_unused_sigrt()
```
namhyung, line 1065: `void` please.

utils/utils.c
```diff
@@ -1058,6 +1058,96 @@ int tgkill(pid_t tgid, pid_t tid, int signal)
...
+	int sig;
+	struct sigaction oldact;
+
+	for (sig = SIGRTMIN; sig <= SIGRTMAX; sig++) {
+		if (sigaction(sig, NULL, &oldact) < 0) {
+			pr_dbg3("failed to check RT signal handler\n");
+			continue;
+		}
+
+		if (!oldact.sa_handler)
```
namhyung, line 1076: Also this can be racy if other thread installs a new sighandler at the same time.  But there's not much we can do..

azharivs: True, and furthermore, there is no guarantee that we're going to keep this signal to ourselves even if we don't get into a race condition here. The application can change our signal's disposition at any other point without notice. However, this function is only triggered upon dynamic runtime patching and unpatching via (do_dynamic_update) which we expect to be 1) infrequent 2) not happening at any particular time, e.g., application startup. Hence, the likelihood of this condition is low. Let me know what you think.

utils/utils.c
```diff
@@ -1058,6 +1058,96 @@ int tgkill(pid_t tgid, pid_t tid, int signal)
...
+		/*
+		 * By reading /proc/<pid>/task directory, there is the possibility of
+		 * a race condition where a thread exits before we send the signal.
+		 */
+		pr_dbg4("send SIGRT%d to %u\n", sig, task);
+		if (tgkill(pid, task, sig) < 0) {
+			if (errno != ESRCH) {
+				pr_dbg2("cannot signal thread %u: %s\n", task, strerror(errno));
+				errno = 0;
+			}
```
namhyung, line 1133: It seems we should not take `ESRCH` as an error.  Something like this?

```c
if (tgkill(pid, task, sig) < 0) {
    if (errno != ESRCH)
        pr_dbg2(...);
    errno = 0;
}
```

utils/utils.c
namhyung: Right, then we need to take care of releasing the rt signal after use.

PR #1708, opened 2023-05-12
Fix #1707 x86 build 

### LINKED ISSUE #1707: [bug] build failed on x86-64
uftrace/arch/x86_64/mcount-insn.c:479:3: error: label at end of compound statement
   default:
   ^~~~~~~

yihong0618:
This PR fix build on x86_64 fix #1707

### DIFF
--- a/arch/x86_64/mcount-insn.c
+++ b/arch/x86_64/mcount-insn.c
@@ -478,6 +478,7 @@ static int check_instrumentable(struct mcount_disasm_engine *disasm, cs_insn *in
 			return INSTRUMENT_FAIL_NO_DETAIL;
 		default:
 			/* do nothing for legit instructions. */
+			break;
 		}
 	}
 


arch/x86_64/mcount-insn.c
```diff
@@ -476,7 +476,7 @@ static int check_instrumentable(struct mcount_disasm_engine *disasm, cs_insn *in
 		case CS_GRP_PRIVILEGE:
 		case CS_GRP_BRANCH_RELATIVE:
 			return INSTRUMENT_FAIL_NO_DETAIL;
-		default:
+		default:;
```
namhyung, line 479: Please don't mix two changes in a commit.  This change should be separate from the Github workflow.  Also you'd better have `break` statement in the following line.

.github/workflows/pr-test.yml
```diff
@@ -1,9 +1,20 @@
...
 jobs:
   unit-test-with-asan:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x86_64
+          - aarch64
+          - arm64
```
namhyung, line 17: Aren't aarch64 and arm64 the same?

namhyung (commented):
Please rebase your changes on to the current master (after running `git pull`).  And the commit message should have a proper prefix.  'fix' is not a proper one.  It can be 'build' or 'arch/x86' or 'dynamic'.  In addition, please consider adding the actual compiler error message to see what's going on here since it might depend on the compiler and its version.  It's a convention to indent by two spaces those messages.

yihong0618:
copy that will fix it tonight thanks

.github/workflows/pr-test.yml
```diff
@@ -1,9 +1,19 @@
...
+
 jobs:
   unit-test-with-asan:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x86_64
+          - aarch64
```
namhyung, line 16: Actually I'm not sure if it's really needed to run the unittest on multiple archs.

arch/x86_64/mcount-insn.c
```diff
@@ -477,6 +477,7 @@ static int check_instrumentable(struct mcount_disasm_engine *disasm, cs_insn *in
 		case CS_GRP_BRANCH_RELATIVE:
 			return INSTRUMENT_FAIL_NO_DETAIL;
 		default:
+			break;
```
namhyung, line 480: Please move it after the comment.

.github/workflows/pr-test.yml
namhyung: Could please check the actual result?  It seems it didn't run the test on an ARM (64-bit) machine.

yihong0618: will check it today,  sorry for the inconvenience

yihong0618:
@namhyung you are right,  the `arch` did not work...
I think it needs a lot of work to test on different arch, so I want to change back and do not change the actions things on different arch, do you agree?


maybe need this:
https://github.com/marketplace/actions/run-on-architecture

namhyung:
I'm ok to drop the github actions and just to fix the build problem.

yihong0618:
Done

namhyung:
For this kind of change, it'd be nice if you could add the actual compiler warning message to the commit like in dae3c721d2d.   Note that you may want to indent them by 2 spaces to prevent accidental commenting out if it starts with the hash (#).   Anyway I prefer having them indented (even without the hash) to indicate they are copied from elsewhere.

yihong0618:
of course, I will learn your commit and revise mine, thanks for the advice again~

namhyung (approved):
LGTM, thanks for your endurance.

honggyukim:
Hi @yihong0618, it was my mistake. Thanks for the fix!

Issue #1719, opened 2023-06-02
fstack_read unit test runtime error : member access within null pointer of type 'struct uftrace_session' 


paranlee:
I review the code with `make -j4 DEBUG=1 ASAN=1 unittest`
`/home/kim/uftrace/utils/fstack.c:706:7: runtime error: member access within null pointer of type 'struct uftrace_session'`

Their is null check routine on top of function `fstack_entry()`
but the middle of function access `fstack_get_loc_mode(&sess->filters)` this function is void argument type.
```
static inline int fstack_get_loc_mode()
{
	int loc_count = fstack_triggers.loc_count;
	return loc_count > 0 ? FILTER_MODE_IN : FILTER_MODE_OUT;
}

static inline int fstack_get_filter_mode()
{
	int filter_count = fstack_triggers.filter_count;
	return filter_count > 0 ? FILTER_MODE_IN : FILTER_MODE_OUT;
}
```

I don't know in my opinion  is correct, but in Linux kernel modules, [If i don't mark an argument on non-args void type function, I will get a compile error](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstrict-prototypes), so wouldn't it be nice to add a warning on Makefile?

Something like this example.
```
/home/kim/linux-modules/workqueue/driver.c:87:13: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
   87 | static void func()
      |             ^~~~
```

paranlee:
I unit tested and PR #1720
```
kim@raspberrypi:~/uftrace $ make -j4 DEBUG=1 ASAN=1 unittest
  CC       utils/fstack.o
  LINK     uftrace
  CC       utils/fstack.ot
  LINK     unittest
  TEST     test_unit
Running 95 test cases
======================
[001] unittest_framework            : PASS
[002] argspec_auto_args             : PASS
[003] argspec_extract               : PASS
[004] argspec_parse_enum            : PASS
[005] argspec_parse_struct          : PASS
[006] data_basic                    : PASS
[007] demangle_rust1                : PASS
[008] demangle_simple1              : PASS
[009] demangle_simple2              : PASS
[010] demangle_simple3              : PASS
[011] demangle_simple4              : PASS
[012] demangle_simple5              : PASS
[013] demangle_simple6              : PASS
[014] demangle_simple7              : PASS
[015] demangle_simple8              : PASS
[016] dwarf_same_file_name1         : PASS
[...]

PR #1720, opened 2023-06-02
uftrace: Add 'void' to functions without arguments

paranlee:
This patch remove argument on void non-args function when using
`fstack_get_loc_mode()`, fstack_get_filter_mode()`.

Fixed: #1719

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ LIB_LDFLAGS        = $(COMMON_LDFLAGS) $(LDFLAGS_$@) $(LDFLAGS_lib) -Wl,--no-und
 TEST_LDFLAGS       = $(COMMON_LDFLAGS) -L$(objdir)/libtraceevent -ltraceevent
 
 ifeq ($(DEBUG), 1)
-  COMMON_CFLAGS += -O0 -g3 -DDEBUG_MODE=1
+  COMMON_CFLAGS += -O0 -g3 -DDEBUG_MODE=1 -Werror -Wstrict-prototypes
 else
   COMMON_CFLAGS += -O2 -g -DDEBUG_MODE=0
 endif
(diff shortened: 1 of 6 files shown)


namhyung (commented):
OK, we have a lot of such functions.  I think you can make the subject in the second commit more compact like "uftrace: Add 'void' to functions without arguments".  You can use 'uftrace:' prefix when it doesn't belong to a specific area.

Also can we check this void argument before merging the code?  I think we can add a warning in the Makefile or update the clang-format check.  Thoughts?

paranlee:
@namhyung  I add compile options at Makefile. I tested with [without void](https://github.com/namhyung/uftrace/issues/1719#issuecomment-1575558281) at `DEBUG=1`

Makefile
```diff
@@ -104,7 +104,7 @@ LIB_LDFLAGS        = $(COMMON_LDFLAGS) $(LDFLAGS_$@) $(LDFLAGS_lib) -Wl,--no-und
 TEST_LDFLAGS       = $(COMMON_LDFLAGS) -L$(objdir)/libtraceevent -ltraceevent
 
 ifeq ($(DEBUG), 1)
-  COMMON_CFLAGS += -O0 -g3 -DDEBUG_MODE=1
+  COMMON_CFLAGS += -O0 -g3 -DDEBUG_MODE=1 -Werror=strict-prototypes
```
namhyung, line 107: Please do it in a separate commit.  Also I think we can add `-Werror` here and add `-Wstrict-prototypes` to all conditions.

paranlee: OK, Thanks for code review!

namhyung (commented):
You don't need to add 'Fixed:' tag to every commit.  Anyway, looks ok.

PR #1725, opened 2023-06-03
python: Handle os._exit() properly

honggyukim:
When os._exit() is invoked in the target python program, uftrace can't write the symbol and dbginfo files because it doesn't invoke the destructor.

This patch fixes the problem by hooking os._exit and invoke the destructor manually then call _exit().

Fixed: #1685

### DIFF
--- a/python/trace-python.c
+++ b/python/trace-python.c
@@ -151,9 +151,14 @@ static void (*cygprof_exit)(unsigned long child, unsigned long parent);
 /* main trace function to be called from python interpreter */
 static PyObject *uftrace_trace_python(PyObject *self, PyObject *args);
 
+/* hooking function of os._exit() for proper cleanup */
+static PyObject *uftrace_trace_python_exit(PyObject *self, PyObject *obj);
+
 static __attribute__((used)) PyMethodDef uftrace_py_methods[] = {
 	{ "trace", uftrace_trace_python, METH_VARARGS,
 	  PyDoc_STR("trace python function with uftrace.") },
+	{ "exit", uftrace_trace_python_exit, METH_O,
+	  PyDoc_STR("exit the target program with cleanup.") },
 	{ NULL, NULL, 0, NULL },
 };
 
@@ -1017,6 +1022,14 @@ static void __attribute__((destructor)) uftrace_trace_python_finish(void)
 	free(main_dir);
 }
 
+static PyObject *uftrace_trace_python_exit(PyObject *self, PyObject *obj)
+{
+	int n = PyLong_AsLong(obj);
+	uftrace_trace_python_finish();
+	_exit(n);
+	return NULL;
+}
+
 #else /* UNIT_TEST */
 
 static PyObject *uftrace_trace_python(PyObject *self, PyObject *args)
@@ -1025,6 +1038,11 @@ static PyObject *uftrace_trace_python(PyObject *self, PyObject *args)
 	skip_first_frame = false;
 	code_tree = code_tree;
 
+	return NULL;
+}
+
+static PyObject *uftrace_trace_python_exit(PyObject *self, PyObject *obj)
+{
 	return NULL;
 }
 
--- a/python/uftrace.py
+++ b/python/uftrace.py
@@ -25,6 +25,14 @@ else:
 # UFTRACE_PYMAIN must be set before importing uftrace_python
 import uftrace_python
 
+# Symbol and debug files are finally written at uftrace_trace_python_finish()
+# when program exits, but os._exit() terminates the program immediately so there
+# is no chance to write symbol and debug files at the destructor.
+# The os._exit() is hooked here to prevent this problem.
+def os_exit(n):
+    uftrace_python.exit(n)
+os._exit = os_exit
+
 new_globals = globals()
 new_globals["__file__"] = pathname
 sys.path.insert(0, os.path.dirname(pathname))
--- a/tests/p012_os_exit.py
+++ b/tests/p012_os_exit.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+
+from runtest import PyTestBase
+
+class TestCase(PyTestBase):
+    def __init__(self):
+        PyTestBase.__init__(self, 'abc-exit', """
+# DURATION     TID     FUNCTION
+            [  6461] | __main__.<module>() {
+            [  6461] |   a() {
+            [  6461] |     b() {
+            [  6461] |       c() {
+   1.114 us [  6461] |         posix.getpid();
+   4.746 us [  6461] |       } /* c */
+   7.101 us [  6461] |     } /* b */
+   9.897 us [  6461] |   } /* a */
+
+uftrace stopped tracing with remaining functions
+================================================
+task: 6461
+[0] __main__.<module>
+""", sort='task')
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -318,7 +318,7 @@ class TestBase:
         order = 1
         before_main = True
         for ln in output.split('\n'):
-            if ln.find(' | main()') > 0:
+            if ln.find(' | main()') > 0 or ln.find(' | __main__.<module>') > 0:
                 before_main = False
             if before_main:
                 continue
--- a/tests/p011_libcall_nested.py
+++ b/tests/p011_libcall_nested.py
@@ -27,3 +27,6 @@ class TestCase(PyTestBase):
 
     def setup(self):
         self.option = '--nest-libcall -N ^importlib -D 6'
+
+    def fixup(self, cflags, result):
+        return result.replace(".JSONEncoder", "")
--- a/tests/s-abc-exit.py
+++ b/tests/s-abc-exit.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+import os
+
+def a():
+    b()
+
+def b():
+    c()
+
+def c():
+    return os.getpid()
+
+a()
+os._exit(0)


honggyukim:
just updated to fix a unittest failure.

honggyukim:
@namhyung Could you please review before v0.14 release?

python/uftrace.py
```diff
@@ -25,6 +25,11 @@
 # UFTRACE_PYMAIN must be set before importing uftrace_python
 import uftrace_python
 
+def os_exit(n):
+    uftrace_python.exit(n);
+
+os._exit = os_exit
```
namhyung, line 31: Please add a comment why this is needed.

python/trace-python.c
```diff
@@ -1017,6 +1022,13 @@ static void __attribute__((destructor)) uftrace_trace_python_finish(void)
 	free(main_dir);
 }
 
+static PyObject *uftrace_trace_python_exit(PyObject *self, PyObject *obj)
+{
+	int n = PyLong_AsLong(obj);
+	uftrace_trace_python_finish();
+	_exit(n);
```
namhyung, line 1029: Maybe not necessary, but to prevent possible compiler complaint, return NULL here.

namhyung (commented):
Can you add a test case?

honggyukim:
I've added a test for this and the full test result looks as follows.
```
$ ./runtest.py -P
Start 12 tests with 6 worker

Test case                 Result
--------------------------------
001 basic               : OK
002 filter_F            : OK
003 filter_N            : OK
004 filter_FN           : SG
005 srcline             : OK
006 filter_time         : OK
007 filter_depth        : OK
008 file_var            : OK
009 libcall_single      : OK
010 libcall_none        : OK
011 libcall_nested      : OK
012 os_exit             : OK
```
The `004 filter_FN` shows `SG`, but this test result looks really weird because the test fails with comments. If I remove those comments then the test is passed.
```diff
diff --git a/python/uftrace.py b/python/uftrace.py
index 1c0db525..12a5133f 100644
--- a/python/uftrace.py
+++ b/python/uftrace.py
@@ -25,10 +25,6 @@ else:
 # UFTRACE_PYMAIN must be set before importing uftrace_python
 import uftrace_python
 
-# Symbol and debug files are finally written at uftrace_trace_python_finish()
-# when program exits, but os._exit() terminates the program immediately so there
-# is no chance to write symbol and debug files at the destructor.
-# The os._exit() is hooked here to prevent this problem.
 def os_exit(n):
     uftrace_python.exit(n)
 os._exit = os_exit
```
```
$ ./runtest.py -P 004

Test case                 Result
--------------------------------
004 filter_FN           : OK
```
This problem can be reproduced as follows.
```
$ uftrace record -F a -N c s-abc.py 
```
The backtrace looks as follows.
```
Thread 2.1 "python3" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7cf5840 (LWP 30052)]
0x0000555555720a18 in PyDict_Clear ()
(gdb) bt
#0  0x0000555555720a18 in PyDict_Clear ()
#1  0x00005555557b1b2a in ?? ()
#2  0x0000555555680185 in ?? ()
#3  0x00005555557bda00 in ?? ()
#4  0x00005555557b94f6 in Py_FinalizeEx ()
#5  0x00005555557a9193 in Py_RunMain ()
#6  0x000055555577f32d in Py_BytesMain ()
#7  0x00007ffff7a29d90 in __libc_start_call_main (main=main@entry=0x55555577f2f0, argc=argc@entry=4, argv=argv@entry=0x7fffffffdaa8)
    at ../sysdeps/nptl/libc_start_call_main.h:58
#8  0x00007ffff7a29e40 in __libc_start_main_impl (main=0x55555577f2f0, argc=4, argv=0x7fffffffdaa8, init=<optimized out>, fini=<optimized out>,
    rtld_fini=<optimized out>, stack_end=0x7fffffffda98) at ../csu/libc-start.c:392
#9  0x000055555577f225 in _start ()
```
@namhyung Can you reproduce the same problem?

tests/runtest.py
```diff
@@ -705,7 +705,7 @@ def __del__(self):
             sp.call(['rm', '-rf', self.test_dir])
 
 class PyTestBase(TestBase):
-    def __init__(self, name, result, lang='Python', cflags='', ldflags='', sort='simple', serial=False):
+    def __init__(self, name, result, lang='Python', cflags='', ldflags='', sort='task', serial=False):
```
namhyung, line 708: I think calling os._exit() is an exceptional case.  Why not just change the upcoming exit test and leave others as is?

tests/p012_os_exit.py
```diff
@@ -0,0 +1,22 @@
...
+   1.114 us [  6461] |         posix.getpid();
+   4.746 us [  6461] |       } /* c */
+   7.101 us [  6461] |     } /* b */
+   9.897 us [  6461] |   } /* a */
+
+uftrace stopped tracing with remaining functions
+================================================
+task: 6461
+[0] __main__.<module>
+""")
```
namhyung, line 22: Add 'sort=task' here.

namhyung:
I'm seeing a different failure.  I think the test code runs installed version, not the developing version.  Maybe we can install it first and then run the test again.  We need something like this:

```diff
diff --git a/tests/runtest.py b/tests/runtest.py
index ff89f3f8..48e385c6 100755
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -707,6 +707,13 @@ import time
 class PyTestBase(TestBase):
     def __init__(self, name, result, lang='Python', cflags='', ldflags='', sort='task', serial=False):
         TestBase.__init__(self, name, result, lang, cflags, ldflags, sort, serial)
+        # setup PYTHONPATH to load the new code before the installed one
+        orig_path = os.environ["PYTHONPATH"]
+        os.environ["PYTHONPATH"] = TestBase.objdir + '/python'
+        if TestBase.objdir != TestBase.srcdir:
+            os.environ["PYTHONPATH"] += ':' + TestBase.objdir + '/python'
+        if orig_path != "":
+            os.environ["PYTHONPATH"] += ':' + orig_path
 
 RED     = '\033[1;31m'
 GREEN   = '\033[1;32m'
```

honggyukim:
> I think the test code runs installed version, not the developing version.

Yes, right. That's why I ran the test after installation but problem shows 

> Maybe we can install it first and then run the test again. We need something like this:

It doesn't fix my problem and it fails to be executed as follows.
```
$ ./runtest.py -P 004

Test case                 Result
--------------------------------
  File "/home/honggyu/work/uftrace/tests/./runtest.py", line 1108, in <module>
    results = run_single_case(name, flags, opts.split(), arg, compilers)
  File "/home/honggyu/work/uftrace/tests/./runtest.py", line 795, in run_single_case
    ret, dif = run_python_case(T, case, timeout)
  File "/home/honggyu/work/uftrace/tests/./runtest.py", line 769, in run_python_case
    tc = T.TestCase()
  File "/home/honggyu/work/uftrace/tests/p004_filter_FN.py", line 7, in __init__
    PyTestBase.__init__(self, 'abc', """
  File "/home/honggyu/work/uftrace/tests/runtest.py", line 711, in __init__
    orig_path = os.environ["PYTHONPATH"]
  File "/usr/lib/python3.10/os.py", line 679, in __getitem__
    raise KeyError(key) from None
KeyError: 'PYTHONPATH'
```

honggyukim:
Do you see the segfault disappears if the comments are removed?

namhyung:
I don't see any segfaults before and after the change.  But removing comments should not affect that.

namhyung:
Without the above patch I'm seeing this.

```
$ make pytest TESTARG='-v exit'
  TEST     test_python

Test case                 Result
--------------------------------
test command: /home/namhyung/project/uftrace/uftrace live --no-pager --no-event --libmcount-path=/home/namhyung/project/uftrace   -N memcpy /home/namhyung/project/uftrace/tests/s-abc-exit.py
=========== original ===========
            [2512461] | <1>() {
            [2512461] |   <2>() {
            [2512461] |     <3>() {
            [2512461] |       <4>() {
   2.074 us [2512461] |         <5>();
   6.097 us [2512461] |       } /* <4> */
   8.080 us [2512461] |     } /* <3> */
  11.559 us [2512461] |   } /* <2> */

task: 2512461
[0] <1>


__main__.<module>() {
   a() {
     b() {
       c() {
         posix.getpid();
       } /* c */
     } /* b */
   } /* a */
012 os_exit             : NG
```

After applying the change, it succeeds.

namhyung:
What about this then?

```diff
diff --git a/tests/runtest.py b/tests/runtest.py
index ff89f3f8..1349a752 100755
--- a/tests/runtest.py
+++ b/tests/runtest.py
@@ -707,6 +707,13 @@ import time
 class PyTestBase(TestBase):
     def __init__(self, name, result, lang='Python', cflags='', ldflags='', sort='task', serial=False):
         TestBase.__init__(self, name, result, lang, cflags, ldflags, sort, serial)
+        # setup PYTHONPATH to load the new code before the inst
+        orig_path = os.environ["PYTHONPATH"] if "PYTHONPATH" in os.environ else ""
+        os.environ["PYTHONPATH"] = TestBase.objdir + '/python'
+        if TestBase.objdir != TestBase.srcdir:
+            os.environ["PYTHONPATH"] += ':' + TestBase.objdir + '/python'
+        if orig_path != "":
+            os.environ["PYTHONPATH"] += ':' + orig_path
 
 RED     = '\033[1;31m'
 GREEN   = '\033[1;32m'
```

honggyukim:
I have tested in another x86_64 and aarch64 machines and they don't show the segfault. It might be some weird issue in my previous machine.
```
$ uftrace -F a -N c s-abc.py 
            [139968] | a() {
   4.795 us [139968] |   b();
   7.565 us [139968] | } /* a */
```
All the tests are passed now. Please have a look again.

python/trace-python.c
```diff
@@ -1017,6 +1022,14 @@ static void __attribute__((destructor)) uftrace_trace_python_finish(void)
 	free(main_dir);
 }
 
+static PyObject *uftrace_trace_python_exit(PyObject *self, PyObject *obj)
+{
+	int n = PyLong_AsLong(obj);
```
namhyung, line 1027: I think it's better to emit an entry for "os._exit" so that users can know it's exited by that.

honggyukim: Hmm.. I dont’t get what you mean. Could you explain what you mean by “emit an entry”?

namhyung: Something like this

```
__main__.<module>() {
   a() {
     b() {
       c() {
         posix.getpid();
       } /* c */
     } /* b */
   } /* a */
  os._exit() {          <--- here
```

Of course it'd also show remaining functions, but I think that's clearer.

```
task: 1234
[1] os._exit
[0] __main__.<module>
```

namhyung: Nevermind, I can add that later.

namhyung (approved):
LGTM

python/trace-python.c
honggyukim: It actually captures the entry of `os_exit`, but didn't fully write the record. I can see that with the following change.
```diff
diff --git a/python/trace-python.c b/python/trace-python.c
index 84ffb873..9ddd003f 100644
--- a/python/trace-python.c
+++ b/python/trace-python.c
@@ -981,14 +981,18 @@ static PyObject *uftrace_trace_python(PyObject *self, PyObject *args)
                Py_RETURN_NONE;
 
        if (!strcmp(event, "call") || !strcmp(event, "c_call")) {
-               if (can_trace(true, sym))
+               if (can_trace(true, sym)) {
+                       fprintf(stderr, "[ call ] sym->name = %s\n", sym->name);
                        cygprof_enter(sym->addr, 0);
+               }
                else
                        Py_RETURN_NONE;
        }
        else if (!strcmp(event, "return") || !strcmp(event, "c_return")) {
-               if (can_trace(false, sym))
+               if (can_trace(false, sym)) {
+                       fprintf(stderr, "[return] sym->name = %s\n", sym->name);
                        cygprof_exit(0, 0);
+               }
                else
                        Py_RETURN_NONE;
        }
```
It shows the `os_exit` and `uftrace_python.exit` as well.
```
$ uftrace tests/s-abc-exit.py 
[ call ] sym->name = __main__.<module>
[ call ] sym->name = a
[ call ] sym->name = b
[ call ] sym->name = c
[ call ] sym->name = posix.getpid
[return] sym->name = posix.getpid
[return] sym->name = c
[return] sym->name = b
[return] sym->name = a
[ call ] sym->name = os_exit
[ call ] sym->name = uftrace_python.exit
            [ 12145] | __main__.<module>() {
            [ 12145] |   a() {
            [ 12145] |     b() {
            [ 12145] |       c() {
   2.254 us [ 12145] |         posix.getpid();
   8.856 us [ 12145] |       } /* c */
  14.264 us [ 12145] |     } /* b */
  20.208 us [ 12145] |   } /* a */
            [ 12145] |   /* linux:task-exit */

task: 12145
[0] __main__.<module>
```
I'm not sure why it's not fully written in uftrace record.

namhyung: Maybe it didn't flush the rstack until it sees a exit record.

PR #1714, opened 2023-05-30
New release: v0.14

namhyung:
It's time for a new release.  Please test and report any issues!

In this development cycle, we've seen many interesting changes including:
 * Python language support
 * Runtime option changes with agent
 * Android build support

And of course, a lot more fixes and improvements are there as well.  Thanks for all the contributors.

PR #1731, opened 2023-06-16
filter: Cleanup debug code

clementguidi:
Hi, I spotted a printf debug line that I forgot to remove in a previous PR. You probably want to get rid of it before v0.14.

Related: #1714


honggyukim (approved):
LGTM. Thanks for the fix.

PR #1756, opened 2023-07-10
misc: Add gen-compile-commands.py for code analysis

honggyukim:
This patch adds gen-compile-commands.py script that generates compile_commands.json, which is a JSON Compilation Database Format Specification[1].

This compile_commands.json can be used by static analysis tools such as clang-tidy[2] for advanced diagnosis or lintting.

The usage is as follows.
```
  $ ./configure

  $ make clean    # To print build commands for all source files.
  $ ./misc/gen-compile-commands.py
  (... compile_commands.json generated ...)

  $ run-clang-tidy
```
[1] https://clang.llvm.org/docs/JSONCompilationDatabase.html
[2] https://clang.llvm.org/extra/clang-tidy


honggyukim:
You can see that the second patch is based on the automatic fix by clang-tidy. It's just show an example. I will drop the second patch before merging this PR.

honggyukim:
There are many useful checkers in clang-tidy.
https://clang.llvm.org/docs/analyzer/checkers.html

misc/gen-compile-commands.py
```diff
@@ -0,0 +1,44 @@
...
+        return lines
+
+with open('compile_commands.json', 'w') as f:
+    f.write('[')
+    needs_comma = False
+    directory = os.getcwd()
+    lines = do_dry_make()
+    for line in lines:
+        command = line
+        if command[:3] != 'gcc':
```
namhyung, line 24: Why do you need this?  I guess you wanted clang as well, no?

namhyung (commented):
The changes in the second commit are too many so I don't know which one is meaningful.  It'd be nice if we can control clang-tidy what to report.

misc/gen-compile-commands.py
honggyukim: I've added a comment as follows.

> We're only interested in compilation commands so skip otherwise.

honggyukim:
> The changes in the second commit are too many so I don't know which one is meaningful. It'd be nice if we can control clang-tidy what to report.

The second commit was to give you an idea what kind of changes can be made by `clang-tidy`. I've just dropped it and left only the `compile_commands.json` generation script.

You can still see what kind of warnings can be shown by running `run-clang-tidy`. We can set up only the checkers we're interested in at `.clang-tidy`.

You can take a look at the following as references.
- https://github.com/honggyukim/heaptrace/blob/main/.clang-tidy
- https://github.com/honggyukim/heaptrace/pull/31

honggyukim:
I've added one more commit that applies [readability-non-const-parameter](https://clang.llvm.org/extra/clang-tidy/checks/readability/non-const-parameter.html) checker for reference.

I'm fine to drop the second commit. It's just to give an idea what clang-tidy can do.

honggyukim:
I've just added a few more commits that applies each clang-tidy checkers.

.clang-tidy
```diff
@@ -6,3 +6,4 @@ Checks: >-
     readability-inconsistent-declaration-parameter-name,
     readability-redundant-declaration,
     bugprone-macro-parentheses,
+    bugprone-suspicious-string-compare,
```
namhyung, line 9: I don't think it's better. :(

namhyung (commented):
Is it possible to disable clang-tidy in some specific code?

.clang-tidy
```diff
@@ -1,3 +1,4 @@
 Checks: >-
     -*,
     readability-non-const-parameter,
+    readability-isolate-declaration,
```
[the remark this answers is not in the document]
honggyukim: Ack. I will remove it.

.clang-tidy
honggyukim: Ack. will remove it as well.

honggyukim (commented):
> Is it possible to disable clang-tidy in some specific code?

Yes, we can ignore clang-tidy warning as follows.
> If a specific suppression mechanism is not available for a certain warning, or its use is not desired for some reason, clang-tidy has a generic mechanism to suppress diagnostics using NOLINT, NOLINTNEXTLINE, and NOLINTBEGIN … NOLINTEND comments.

https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

PR #1759, opened 2023-07-13
Refactor the kernel depth option and documentation

ChoKyuWon:
To process the `-k` option, the `kernel_depth` in opt field is not set and is set to 1 when used (`setup_writers`).
This PR initializes the `opts->kernel_depth` to 1 at the case  `-k` and `-K` if the input value is invalid.

This refactoring clarifies the code semantic when the user only uses the `-k`, and invalid `-K`.


namhyung (commented):
Looks good, but we could improve the message a little bit.

namhyung (approved):
LGTM

Issue #1753, opened 2023-07-09
Unable to find tracefs in different kernel so kernel tracing fails


honggyukim:
The current way of finding tracefs for kernel tracing was implemented in #1476.

It assumes that `/proc/<pid>/mountinfo` looks as follows.
```
36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 .... ....
```
So it detects the file system type from the `/proc/self/mountinfo`. For example, the following input can be properly parsed and successfully finds `mount_point` and `fs_type` as written in [`find_tracing_dir`](https://github.com/namhyung/uftrace/blob/v0.14/utils/tracefs.c#L28-L47).
```
$ cat /proc/self/mountinfo | grep tracefs
40 23 0:12 / /sys/kernel/tracing rw,nosuid,nodev,noexec,relatime shared:18 - tracefs tracefs rw
```
However, it fails detecting `fs_type` from the following my qemu rootfs and it makes kernel tracing disabled. The following line has one less word as shown as `shared:18` in the above example.
```
root@qemux86-64:~# cat /proc/self/mountinfo | grep tracefs
26 22 0:10 / /sys/kernel/debug/tracing rw,relatime - tracefs tracefs rw
```

As I mentioned at https://github.com/namhyung/uftrace/issues/1471#issuecomment-1179689118, I think we should use [`getmntent`](https://man7.org/linux/man-pages/man3/getmntent.3.html) to find the mount point.

honggyukim:
As written above, the following example shows the output as follows.
```c
$ cat getmntent.c
#include <stdio.h>
#include <stdlib.h>
#include <mntent.h>

int main(void)
{
  struct mntent *ent;
  FILE *aFile;

  aFile = setmntent("/proc/mounts", "r");
  if (aFile == NULL) {
    perror("setmntent");
    exit(1);
  }
  while (NULL != (ent = getmntent(aFile))) {
    printf("%s %s\n", ent->mnt_fsname, ent->mnt_dir);
  }
  endmntent(aFile);
}
```
It can simply be compiled as follows.
```
$ gcc getmntent.c
```
Then the output shows it find `tracefs` correctly.
```
$ ./a.out | grep tracefs
tracefs /sys/kernel/tracing
```
It also successfully finds the `tracefs` in my qemu rootfs.
```
root@qemux86-64:~# ./a.out | grep tracefs
tracefs /sys/kernel/debug/tracing
```

gichoel:
[...]

PR #1762, opened 2023-07-16
utils: Fix an issue not finding tracefs on some kernels

gichoel:
When parsing /proc/self/mountinfo on some kernels, the number of fields is different and tracefs cannot be obtained.

changed method to check if paths of "/sys/kernel/tracing (tracefs)" and "/sys/kernel/debug/tracing (debugfs)" exist

Fixed: namhyung/uftrace#1753

### DIFF
--- a/utils/tracefs.c
+++ b/utils/tracefs.c
@@ -1,67 +1,41 @@
 #include <fcntl.h>
+#include <mntent.h>
 #include <stdio.h>
 #include <string.h>
 
 #include "utils/tracefs.h"
 #include "utils/utils.h"
 
-#define PROC_MOUNTINFO "/proc/self/mountinfo"
+#define PROC_MOUNTS_DIR_PATH "/proc/mounts"
 
 static char *TRACING_DIR = NULL;
 
 static bool find_tracing_dir(void)
 {
 	FILE *fp;
-	char *line = NULL, fs_type[NAME_MAX], mount_point[PATH_MAX];
-	static char debugfs_suffix[] = "tracing";
-	bool debugfs_found = false;
-	size_t len;
+	struct mntent *ent;
 
 	if (TRACING_DIR)
-		return false;
+		return true;
 
-	fp = fopen(PROC_MOUNTINFO, "r");
+	fp = setmntent(PROC_MOUNTS_DIR_PATH, "r");
 	if (fp == NULL)
 		return false;
 
-	while (getline(&line, &len, fp) > 0) {
-		/*
-		 * /proc/<pid>/mountinfo format:
-		 * 36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 .... ....
-		 * (1)(2)(3)   (4)   (5)      (6)      (7)   (8) (9) (10) (11)
-		 *                mount_point                  fs_type
-		 *
-		 * (9) is the file system type, (5) is the mount point relative
-		 * to self's root directory.
-		 */
-		sscanf(line, "%*i %*i %*u:%*u %*s %s %*s %*s - %s %*s %*s\n", mount_point, fs_type);
-
-		if (!strcmp(fs_type, "tracefs")) {
-			/* discard previously kept debugfs tracing dir */
-			if (TRACING_DIR)
-				free(TRACING_DIR);
-			xasprintf(&TRACING_DIR, "%s", mount_point);
-			pr_dbg2("Found tracefs at %s\n", mount_point);
-			pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
-			return true;
-		}
-
-		if (!strcmp(fs_type, "debugfs")) {
-			xasprintf(&TRACING_DIR, "%s/%s", mount_point, debugfs_suffix);
-			pr_dbg2("Found debugfs at %s\n", mount_point);
-			pr_dbg2("Keep searching for tracefs...\n");
-			debugfs_found = true;
+	while ((ent = getmntent(fp)) != NULL) {
+		if (!strcmp(ent->mnt_fsname, "tracefs")) {
+			xasprintf(&TRACING_DIR, "%s", ent->mnt_dir);
+			break;
 		}
 	}
+	endmntent(fp);
 
-	/* we couldn't find a tracefs, but found a debugfs... */
-	if (debugfs_found) {
-		pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
-		return true;
+	if (!TRACING_DIR) {
+		pr_dbg2("No tracefs or debugfs found..!\n");
+		return false;
 	}
 
-	pr_dbg2("No tracefs or debugfs found..!\n");
-	return false;
+	return true;
 }
 
 char *get_tracing_file(const char *name)
--- a/utils/tracefs.c
+++ b/utils/tracefs.c
@@ -1,12 +1,16 @@
 #include <fcntl.h>
+#include <linux/magic.h>
 #include <mntent.h>
 #include <stdio.h>
 #include <string.h>
+#include <sys/vfs.h>
 
 #include "utils/tracefs.h"
 #include "utils/utils.h"
 
 #define PROC_MOUNTS_DIR_PATH "/proc/mounts"
+#define TRACEFS_DIR_PATH "/sys/kernel/tracing"
+#define OLD_TRACEFS_DIR_PATH "/sys/kernel/debug/tracing"
 
 static char *TRACING_DIR = NULL;
 
@@ -14,9 +18,19 @@ static bool find_tracing_dir(void)
 {
 	FILE *fp;
 	struct mntent *ent;
+	struct statfs fs;
 
 	if (TRACING_DIR)
 		return true;
+
+	if (!statfs(TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
+		return true;
+	}
+	else if (!statfs(OLD_TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", OLD_TRACEFS_DIR_PATH);
+		return true;
+	}
 
 	fp = setmntent(PROC_MOUNTS_DIR_PATH, "r");
 	if (fp == NULL)


utils/tracefs.c
```diff
@@ -1,66 +1,44 @@
...
+	struct statfs fs;
+
+	/* check tracefs path is mounted */
+	statfs(TRACEFS_DIR_PATH, &fs);
+	if (fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
+		pr_dbg2("TRACING_DIR : %s\n", TRACING_DIR);
+
+		return true;
 	}
```
honggyukim, line 30: You should check whether `statfs` is successfully returned and you can rewrite it as follows.
```c
if (!statfs(TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
         xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
         return true;
}
```

utils/tracefs.c
```diff
@@ -1,66 +1,44 @@
...
 
-	/* we couldn't find a tracefs, but found a debugfs... */
-	if (debugfs_found) {
-		pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
+	/* check debugfs path is mounted */
+	statfs(DEBUGFS_DIR_PATH, &fs);
+	if (fs.f_type == DEBUGFS_MAGIC) {
```
honggyukim, line 34: This check actually fails because `/sys/kernel/debug` is debugfs, but  `/sys/kernel/debug/tracing` is mounted as tracefs on top of the debugfs.

There might be a chance that old kernel uses `/sys/kernel/debug/tracing` as debugfs so you might be better to check both `DEBUGFS_MAGIC` and `TRACEFS_MAGIC`.

honggyukim:
And please reword the commit message to make it begin with a verb instead of past tense. i.e. `utils: Fixed ...` -> `utils: Fix ...`.

Please read other commit message to get more references by running `git log` and you will see all the other commit messages begin with a verb.

gichoel:
Hi @honggyukim

of course, statfs was marked as deprecated as I mentioned, so I wasn't sure if it was right to use it, but I used it for the following reasons.

1. `statvfs` is an alternative to `statfs`, but it did not provide magic numbers for checking `tracefs` and `debugfs`.
2. Rather than just checking the path, I wanted to do an additional check to make sure this physical path is indeed the path we are looking for.
3. In Kernel version 5.19.0 running on Ubuntu 22.04, `statfs` was still running.

Thank you for always reviewing my code and comments and pointing out anything I've missed. I'm learning a lot thanks!

I will re-apply and commit push the feedback you suggested. :smiley:

utils/tracefs.c
```diff
@@ -1,67 +1,37 @@
...
-	/* we couldn't find a tracefs, but found a debugfs... */
-	if (debugfs_found) {
-		pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
-		return true;
+	/* check tracefs or old tracefs is mounted */
+	if (!statfs("/sys/kernel/tracing", &fs) && fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
+		goto found;
+	}
+	else if (!statfs("/sys/kernel/debug", &fs) && fs.f_type == DEBUGFS_MAGIC) {
```
honggyukim, line 24: Why not testing `/sys/kernel/debug/tracing` here? i.e. `OLD_TRACEFS_DIR_PATH`.

utils/tracefs.c
```diff
@@ -1,67 +1,37 @@
...
-		}
-	}
+	struct statfs fs;
 
-	/* we couldn't find a tracefs, but found a debugfs... */
-	if (debugfs_found) {
-		pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
-		return true;
+	/* check tracefs or old tracefs is mounted */
+	if (!statfs("/sys/kernel/tracing", &fs) && fs.f_type == TRACEFS_MAGIC) {
```
honggyukim, line 20: You can use `TRACEFS_DIR_PATH` instead of the string.

namhyung:
Please don't remove the original logic and use it as a fallback when the quick check for the known path fails.

I'm not sure why `statfs` is deprecated but I guess it won't go away anytime soon.  Especially if there's no better way to verify the file system type.  At least linux perf tool is using `statfs` to find some filesystem mount locations.

Also it's not true that having debugfs at `/sys/kernel/debug` guarantees tracefs at `/sys/kernel/debug/tracing`.  Actually you don't need to check `/sys/kernel/debug`, just check if `/sys/kernel/debug/tracing` is tracefs type.

honggyukim:
> Please don't remove the original logic and use it as a fallback when the quick check for the known path fails.

Do you want to keep the parsing logic from `/proc/self/mountinfo`? I think it can be error-prone and we've already found broken base because `/proc/self/mountinfo` sometimes has different number of fields.

> I'm not sure why `statfs` is deprecated but I guess it won't go away anytime soon. Especially if there's no better way to verify the file system type. At least linux perf tool is using `statfs` to find some filesystem mount locations.

Then I'm fine using `statfs` here.

> Also it's not true that having debugfs at `/sys/kernel/debug` guarantees tracefs at `/sys/kernel/debug/tracing`.

The new tracefs location seems to be fixed at `/sys/kernel/tracing`. I see that from the commit at https://github.com/torvalds/linux/commit/cc31004a4aa784d89054ec07b87eae05cecf7121. Steven mentioned the location as "a consistent place".

As you  mentioned I'm not sure about the tracefs location at `/sys/kernel/debug/tracing`, but I think we can blindly try it, then go to a fallback routine if it fails.

> Actually you don't need to check `/sys/kernel/debug`, just check if `/sys/kernel/debug/tracing` is tracefs type.

I also think so.

gichoel:
Hi @namhyung and @honggyukim.

I agree with honggyukim, it's better to use `/proc/mounts` because `/proc/self/mountinfo` can cause errors due to the different number of fields. 

So I thought it would be better to remove the existing logic using `/proc/self/mountinfo` and use `/proc/mounts` and `statfs()` to trace a path of tracefs.

I still don't understand why you don't remove the existing logic, can you explain why?

namhyung:
OK, I agree to use `/proc/mounts` and I was confused the existing logic already used it.  I thought you just wanted to check the well-known locations like `/sys/kernel/tracing` without checking `/proc/mounts`.

utils/tracefs.c
```diff
@@ -1,67 +1,59 @@
...
+
+			while (found != true && NULL != (ent = getmntent(fp))) {
+				if (!strcmp(ent->mnt_fsname, "tracefs")) {
+					xasprintf(&TRACING_DIR, "%s", ent->mnt_dir);
+					found = true;
+				}
+				else if (!strcmp(ent->mnt_fsname, "debugfs")) {
+					xasprintf(&TRACING_DIR, "%s/tracing", ent->mnt_dir);
+					found = true;
+				}
```
namhyung, line 42: I think this 'else if' part should be removed because there's no guarantee for tracefs even if it found debugfs.

utils/tracefs.c
```diff
@@ -1,67 +1,59 @@
...
 static bool find_tracing_dir(void)
 {
-	FILE *fp;
-	char *line = NULL, fs_type[NAME_MAX], mount_point[PATH_MAX];
-	static char debugfs_suffix[] = "tracing";
-	bool debugfs_found = false;
-	size_t len;
-
-	if (TRACING_DIR)
-		return false;
```
namhyung, line 21: You removed this caching logic..

namhyung (commented):
The commit message is too long, please break long lines not to exceed 72 characters.

utils/tracefs.c
```diff
@@ -1,67 +1,59 @@
...
+					found = true;
+				}
+				else if (!strcmp(ent->mnt_fsname, "debugfs")) {
+					xasprintf(&TRACING_DIR, "%s/tracing", ent->mnt_dir);
+					found = true;
+				}
+			}
+
+			endmntent(fp);
+		} while (0);
```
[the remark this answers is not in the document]
gichoel: First, I wanted to go down without executing the `while() { ... }` and `endmntent()` below when the result of the `setmntent()` function is NULL.

Second, I didn't want to keep the `if ~ else` statements too long. If the `if ~ else` statement is too long or deep it becomes hard to read. so I wanted to use `do ~ while(0)` which is used by the Linux kernel to avoid this.

I thought this was best, but if there is a better alternative, I'm open to it.

utils/tracefs.c
gichoel: I'll revert the caching logic.

utils/tracefs.c
gichoel: I'll accept the feedback and upload the changed code.

utils/tracefs.c
namhyung: Hmm.. I'm not sure why the original code returns false.

utils/tracefs.c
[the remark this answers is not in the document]
namhyung: I think you can split the check for TRACEFS_DIR_PATH into a separate commit and focus on the mntent change here.  Then it can be like this.

```c
	fp = setmntent("/proc/mounts", "r");
	if (fp == NULL)
		return false;

	while ((ent = getmntent(fp)) != NULL) {
		if (strcmp(ent->mnt_fsname, "tracefs"))
			continue;

		xasprintf(&TRACING_DIR, "%s", ent->mnt_dir);
		found = true;
		break;
	}
	endmntent(fp);
	return found;
```

utils/tracefs.c
gichoel: I couldn't figure out why the original code was returning false either. 

The original code works fine, but I think it's semantically correct for the caching logic to return true.

namhyung (commented):
I'm ok with the change roughly but want to have the commits in the reverse order.  The first commit removed the original logic and added a new code to quickly check the well-known path.  That's +1 and -1 in terms of the functionality.  And the second commit added /proc parsing back, changing -1 to 0.

I'd like to have a commit that fixes /proc parsing first so that we don't lose any functionality in the meantime (like +1 to +1, no change).  Then add the the quick check in the second commit (+1 to +2).  Even though the net result is the same (+1), this way we can gaurantee it only adds and never loses. :)

utils/tracefs.c
```diff
@@ -1,65 +1,37 @@
...
-	if (debugfs_found) {
-		pr_dbg2("Use %s as TRACING_DIR\n", TRACING_DIR);
+		xasprintf(&TRACING_DIR, "%s", ent->mnt_dir);
 		return true;
 	}
+	endmntent(fp);
 
+	/* we couldn't find a tracefs, but found a debugfs... */
 	pr_dbg2("No tracefs or debugfs found..!\n");
 	return false;
```
honggyukim, line 45: It looks you missed closing the stream, which was opened with `setmntent` when `tracefs` is found.

This part can be rewritten as follows.
```c
	while ((ent = getmntent(fp)) != NULL) {
		if (!strcmp(ent->mnt_fsname, "tracefs")) {
			xasprintf(&TRACING_DIR, "%s", ent->mnt_dir);
			break;
		}
	}
	endmntent(fp);

	if (!TRACING_DIR) {
		pr_dbg2("No tracefs or debugfs found..!\n");
		return false;
	}
	return true;
```

honggyukim:
I've merged a few commits into master so please rebase and update this PR onto the current master.

gichoel:
@honggyukim, I replaced the broken commit link and re-uploaded the commit with the feedback.

utils/tracefs.c
```diff
@@ -1,67 +1,50 @@
...
+	struct statfs fs;
 
 	if (TRACING_DIR)
-		return false;
+		return true;
+
+	if (!statfs(TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
+		return true;
+	}
```
honggyukim, line 28: Could you add one more quick check logic for `OLD_TRACEFS_DIR_PATH "/sys/kernel/debug/tracing"`?

honggyukim (commented):
Thanks for the update. I have two last comments. One is the inlined comment below.

The other is that I would like to ask you to use URL from the official kernel repo instead of github one.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2455f0e

In addition, the link can be clicked in github, but not from the commit message itself because it's shown as just `torvalds/linux@2455f0e`, which is not a complete URL.

Instead of it, please add the line below and just above `Signed-off-by` line as follows.
```
Link: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2455f0e
Signed-off-by: Gichoel Choi <gichoel3101@gmail.com>
```

namhyung:
> The other is that I would like to ask you to use URL from the official kernel repo instead of github one. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2455f0e

Actually 7 digit is not enough for kernel commits due to possible collision, kernel community recommends 12.  And there's a short form of the URL like: https://git.kernel.org/torvalds/c/2455f0e124d3

utils/tracefs.c
gichoel: > Also it's not true that having debugfs at /sys/kernel/debug guarantees tracefs at /sys/kernel/debug/tracing. Actually you don't need to check /sys/kernel/debug, just check if /sys/kernel/debug/tracing is tracefs type.

Incorporated feedback based on that comment.
Added code to check if the path `/sys/kernel/debug/tracing` is of type `tracefs`.

Additionally, I've also added a link to the official kernel repository with a 12-character representation of the length of the commit hash.

utils/tracefs.c
```diff
@@ -1,67 +1,52 @@
...
 
 	if (TRACING_DIR)
-		return false;
+		return true;
+
+	if ((!statfs(TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) ||
+	    (!statfs(OLD_TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC)) {
+		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
+		return true;
+	}
```
honggyukim, line 30: Hmm.. Does it really handle when only `/sys/kernel/debug/tracing` is available? This logic sets `TRACING_DIR` to `/sys/kernel/tracing` in both cases.

gichoel: I've incorporated feedback on what I missed.

utils/tracefs.c
```diff
@@ -1,67 +1,59 @@
...
+		found = true;
+	}
+	else if (!statfs(OLD_TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
+		xasprintf(&TRACING_DIR, "%s", OLD_TRACEFS_DIR_PATH);
+		found = true;
+	}
 
-	fp = fopen(PROC_MOUNTINFO, "r");
+	if (found)
+		return true;
```
honggyukim, line 37: I would rather write this part as follows.
```c
	if (!statfs(TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
		xasprintf(&TRACING_DIR, "%s", TRACEFS_DIR_PATH);
                return true;
	}
	if (!statfs(OLD_TRACEFS_DIR_PATH, &fs) && fs.f_type == TRACEFS_MAGIC) {
		xasprintf(&TRACING_DIR, "%s", OLD_TRACEFS_DIR_PATH);
                return true;
	}
```

utils/tracefs.c
honggyukim: It's not that important but the expression "incorporated feedback" sounds awkward. You can just say that "I've updated it based on your feedback".

honggyukim (approved):
LGTM! Thanks very much for your patience.

gichoel:
Thank you @namhyung and @honggyukim for the great feedback each time.

I learned a lot from this PR and issue :)

namhyung (approved):
LGTM

PR #1744, opened 2023-07-04
docs: uftrace slides update

paranlee:
Update explanation of configure.

Fixed: #1743
Signed-off-by: Paran Lee <p4ranlee@gmail.com>

PR #1763, opened 2023-07-17
doc: Update explanation of configure

SEOKMIN83:
Add the part about the libunwind package that being missed from the slide and an install command if you want to use the that package.

Rather than subtracting the libunwind package portion of I think it would be less confusing for users if we described it correctly.

### DIFF
--- a/utils/filter.h
+++ b/utils/filter.h
@@ -172,7 +172,6 @@ const char *get_filter_pattern(enum uftrace_pattern_type ptype);
 
 char *uftrace_clear_kernel(char *filter_str);
 
-void add_trigger(struct uftrace_filter *filter, struct uftrace_trigger *tr, bool exact_match);
 int setup_trigger_action(char *str, struct uftrace_trigger *tr, char **module,
 			 unsigned long orig_flags, struct uftrace_filter_setting *setting);
 


doc/uftrace.html
```diff
@@ -247,6 +247,7 @@
 ...     perf_event: [ .green[on]  ] - perf (PMU) event support
 ...       schedule: [ .green[on]  ] - scheduler event support
 ...       capstone: [ .red[OFF] ] - full dynamic tracing support
+...      libunwind: [ .red[OFF] ] - stacktrace support support (optional for debugging)
```
honggyukim, line 250: It's not required here. Please read the previous discussion at  https://github.com/namhyung/uftrace/pull/1744#discussion_r1252985345.

doc/uftrace.html
```diff
@@ -274,6 +275,10 @@
 ...     perf_event: [ .green[on]  ] - perf (PMU) event support
 ...       schedule: [ .green[on]  ] - scheduler event support
 ...       capstone: [ .green[on]  ] - full dynamic tracing support
+...      libunwind: [ .red[OFF] ] - stacktrace support support (optional for debugging)
+
+# If you want to turn on libunwind option
+$ sudo apt-get install -y libunwind-dev
```
honggyukim, line 281: This tutorial slide doesn't have to cover all the details so no need to have this change.

honggyukim (commented):
Thanks for the PR, but this change was already discussed and decided not to have it.

namhyung:
As @honggyukim said, we intentionally dropped it.  Please add a comment instead to prevent confusion later.

paranlee:
If `libunwind-N` package have dashed number, then `configure` can't find libunwind.
So, configure didn't find `libunwind-N`.
```sh
libunwind-15/unstable,now 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64 [installed,automatic]
  production-quality unwinder

libunwind-15-dbgsym/unstable 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64
  debug symbols for libunwind-15

libunwind-15-dev/unstable,now 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64 [installed]
  production-quality unwinder
```

SEOKMIN83:
> As @honggyukim said, we intentionally dropped it. Please add a comment instead to prevent confusion later.

Then can I put a comment on the slide like below?
"libunwind is optional. so we skipped this part. in tutorial."

SEOKMIN83:
> If `libunwind-N` package have dashed number, then `configure` can't find libunwind. So, configure didn't find `libunwind-N`.
> 
> ```shell
> libunwind-15/unstable,now 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64 [installed,automatic]
>   production-quality unwinder
> 
> libunwind-15-dbgsym/unstable 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64
>   debug symbols for libunwind-15
> 
> libunwind-15-dev/unstable,now 1:15~++20220509105605+460fc79a080b-1~exp1 riscv64 [installed]
>   production-quality unwinder
> ```

It would be nice if you could explain in a little more detail that it is a problem that occurs in RISC-V.
And it would be better to subtract this part from the slide, right?

honggyukim:
> > As @honggyukim said, we intentionally dropped it. Please add a comment instead to prevent confusion later.
> 
> Then can I put a comment on the slide like below? "libunwind is optional. so we skipped this part. in tutorial."

I think it's too verbose adding a comment inside markdown slide because it hurts the readability. So I'm sorry but it's better to be dropped.

honggyukim:
> If libunwind-N package have dashed number, then configure can't find libunwind.
So, configure didn't find libunwind-N.

@paranlee Thanks for the report. But I think it's not the one we have to handle.

honggyukim:
Besides that, I actually think about adding libunwind-dev by default in both release and debug build. But it could be a future work item so we can recall this PR when it's really needed.

PR #1780, opened 2023-07-26
utils: fix demangle_full to only demangle mangled C++ symbols

### LINKED ISSUE #1778: --demangle=full (__cxa_demangle) confusedly translates unmangled C symbols, e.g. "f" is translated to "float"
A simple C code `t.c` with symbol "f" and "g":
```c
void f() {}
void g() {}
int main() {
  f();
  g();
  return 0;
}
```

--demangle=full translates "f" and "g" to "float" and "__float128" respectively, which is confusing:
```shell
$ gcc -pg t.c -o t
$ uftrace --demangle=full ./t
   0.611 us [ 17718] | __monstartup();
   0.210 us [ 17718] | __cxa_atexit();
            [ 17718] | main() {
   0.100 us [ 17718] |   float();
   0.060 us [ 17718] |   __float128();
   0.762 us [ 17718] | } /* main */
```

To prevent this, directly `strdup` the `str` argument not starting with `_Z` in [`demangle_full(char *str)`](https://github.com/namhyung/uftrace/blob/master/utils/demangle.c#L1747) (refer to https://reviews.llvm.org/D22939, since uftrace only resolves function symbols, the problems they discuss do not arise).

jyf111:
Currently, uftrace demangles all symbols, regardless of whether they are mangled C++ symbols or unmangled C symbols.

But when setting --demangle=full, uftrace calls __cxa_demangle(), which translates unmangled symbols confusingly; e.g., "f" is translated to "float".

To address this, now we only call __cxa_demangle() if a symbol is mangled (starts with "_Z").

Fixed #1778

### DIFF
--- a/utils/demangle.c
+++ b/utils/demangle.c
@@ -1750,6 +1750,10 @@ static char *demangle_full(char *str)
 	size_t len = 64; /* minimum length */
 	int status;
 
+	/* str is not mangled C++ symbol */
+	if (str[0] != '_' || str[1] != 'Z')
+		return xstrdup(str);
+
 	__cxa_demangle(str, NULL, &len, &status);
 	if (status < 0)
 		return xstrdup(str);


jyf111:
The commit message should be "skip symbols **not** starting with _Z"

honggyukim:
Thanks for the fix! But please rewrite the commit message so that other readers can understand why this change is made with complete sentences.

Please read other commit messages with `git log`.

namhyung (approved):
LGTM

PR #1781, opened 2023-07-27
docker: Add make install in Dockerfile when test=yes

3godhalmang:
Added code because docker file is missing 'make install' in case test=yes

Alpine and Arch could not be added due to the following error. 
I'm going to give a try to fix these problems.

Alpine(only if test =yes)
[Makefile:49: test_unit] Segmentation fault (core dumped)

Arch(for both condition)
no matching manifest for linux/arm64/v8 in the manifest list entries


honggyukim:
This PR includes a weird merge commit. Please rebase it correctly.

Issue #1766, opened 2023-07-20
Unable to build docker image of ubuntu 


haileynam:
Hello, currently I have a problem while building docker image with Dockerfile which is based on Ubuntu image.
```
root@IPG-PC-56:/home/yoojung/uftrace/misc/docker/ubuntu/18.04# docker image build -t uftrace_ubuntu:18.04 .
[+] Building 130.7s (8/9)
 => [internal] load build definition from Dockerfile                                                   0.0s
 => => transferring dockerfile: 597B                                                                   0.0s
 => [internal] load .dockerignore                                                                      0.0s
 => => transferring context: 2B                                                                        0.0s
 => [internal] load metadata for docker.io/library/ubuntu:18.04                                        2.1s
 => [1/6] FROM docker.io/library/ubuntu:18.04@sha256:152dc042452c496007f07ca9127571cb9c29697f42acbfad  0.0s
 => CACHED [2/6] RUN apt-get update     && apt-get install -y --no-install-recommends git gcc make ca  0.0s
 => CACHED [3/6] RUN mkdir -p /usr/src                                                                 0.0s
 => CACHED [4/6] RUN git clone https://github.com/namhyung/uftrace /usr/src/uftrace                    0.0s
 => ERROR [5/6] RUN if [ "$test" = "yes" ] ; then         cd /usr/src/uftrace         && ./misc/ins  128.6s
------
 > [5/6] RUN if [ "$test" = "yes" ] ; then         cd /usr/src/uftrace         && ./misc/install-deps.sh -y         && ./configure && make ASAN=1 && make ASAN=1 unittest;     else         cd /usr/src/uftrace && ./misc/install-deps.sh -y && ./configure && make && make install;     fi:
#0 0.352 Reading package lists...
#0 1.269 Building dependency tree...
#0 1.377 Reading state information...
#0 1.461 The following additional packages will be installed:
#0 1.461   build-essential dh-python dirmngr dpkg-dev fakeroot file g++ g++-7 gnupg
#0 1.461   gnupg-l10n gnupg-utils gpg gpg-agent gpg-wks-client gpg-wks-server gpgconf
#0 1.461   gpgsm libalgorithm-diff-perl libalgorithm-diff-xs-perl
#0 1.461   libalgorithm-merge-perl libassuan0 libc-dev-bin libc6-dev libdpkg-perl
#0 1.461   libdw1 libelf-dev libelf1 libexpat1-dev libfakeroot libfile-fcntllock-perl
[...]

PR #1783, opened 2023-07-28
misc: modify manual to build dockerfile

haileynam:
Currently make install on some Ubuntu versions failed like below:
```
    GEN      uftrace.1
  pandoc: blocks is null
  CallStack (from HasCallStack):
    error, called at src/Text/Pandoc/Writers/Man.hs:287:47 in pandoc-1.19.2.4-HbfKWUyODESBIy0vGktOwX:Text.Pandoc.Writers.Man
  Makefile:18: recipe for target 'uftrace.1' failed
```
It's because the description of --task option has incompatible format in
the pandoc syntax.  We don't need that part so let's get rid of it.

Fixed: #1766

### DIFF
--- a/doc/uftrace.md
+++ b/doc/uftrace.md
@@ -356,7 +356,7 @@ record and `replay` internally. Thus, it describes most regular option in detail
 :   Hide small functions run less than the *TIME*
 
 \--task
-:   [info]: Print task relationship in a tree form instead of the tracing info.
+:   Print task relationship in a tree form instead of the tracing info.
 
 \--task-newline
 :   Interleave a newline when task is changed


honggyukim:
Please rewrite the commit message with complete sentences so that everyone who reads the message can understand why this change is needed.

doc/uftrace.md
```diff
@@ -356,7 +356,7 @@ record and `replay` internally. Thus, it describes most regular option in detail
 :   Hide small functions run less than the *TIME*
 
 \--task
-:   [info]: Print task relationship in a tree form instead of the tracing info.
+:   [info] Print task relationship in a tree form instead of the tracing info.
```
honggyukim, line 359: I found that the same problem was there and the `[info]:` part was removed by 5c307433aa34d7788382e67423443565118cb0d6.

This is better to be removed.
```suggestion
:   Print task relationship in a tree form instead of the tracing info.
```

honggyukim:
And it's weird that two different commits have the same commit title. Please rewrite the commit message.

haileynam:
@honggyukim I have updated my commit and included your code review. Thank you for your comment.

namhyung (commented):
Looks good.  But the commit prefix should be "doc:" instead of "misc:".  Also "Modify" has no value in the commit message and we want to know what or why.  So the commit message line can be like below:

```
doc: Fix pandoc error in uftrace.md

Currently make install on some Ubuntu versions failed like below:

    GEN      uftrace.1
  pandoc: blocks is null
  CallStack (from HasCallStack):
    error, called at src/Text/Pandoc/Writers/Man.hs:287:47 in pandoc-1.19.2.4-HbfKWUyODESBIy0vGktOwX:Text.Pandoc.Writers.Man
  Makefile:18: recipe for target 'uftrace.1' failed

It's because the description of --task option has incompatible format in
the pandoc syntax.  We don't need that part so let's get rid of it.

Fixed: #1766
```

honggyukim:
Hi @haileynam, the change looks good but I would like to ask you to change the commit title from
```
doc: Modify manual to build dockerfile
```
to the one @namhyung suggested.
```
doc: Fix pandoc error in uftrace.md
```
I'm asking because any kind of commit does modify so `modify` doesn't provide useful information to the commit message readers. And please let us know when you change the commit.

haileynam:
@honggyukim  Sorry for my mistake. I just changed my commit including you and @namhyung suggested. Please check for it.

honggyukim (approved):
Thanks for the update. LGTM!

PR #1785, opened 2023-07-29
doc: Fix --task name in uftrace graph man page

honggyukim:
The man pages generated by .md files require to have an escape character at the beginning for two dash options.

This patch fixes the problem with the output as follows.

Before:
```
       –task  Print task graph instead of normal function graph.  ...
```
After:
```
       --task  Print task graph instead of normal function graph.  ...
```


namhyung (approved):
LGTM

PR #1787, opened 2023-07-30
docker: Add make install in Dockerfile when test=yes

3godhalmang:
Added code because docker file is missing 'make install' in case test=yes

Alpine and Arch could not be added due to the following error
I'm going to give a try to fix these problems.

Alpine(only if test =yes)
[Makefile:49: test_unit] Segmentation fault (core dumped)

Arch(for both condition)
no matching manifest for linux/arm64/v8 in the manifest list entries


namhyung:
It seems the failure in Alpine is not related to `make install`.  Also I'm not sure if we really want to run `make install` for tests.

3godhalmang:
It would be better to include the `make install` process for tests
because there was a `make install fail` problem with PANDOC in 2.0 and earlier.

related: https://github.com/namhyung/uftrace/issues/1766, https://github.com/namhyung/uftrace/pull/1783

Issue #1793, opened 2023-08-02
Mermaid doesn't work caused of deprecated `mermaid.render` method


mirusu400:
With the release of Mermaid v10 on February 21, 2023, a change has been made to the `mermaid.render` function where it no longer accepts callbacks and only supports async operations. [changelog](https://github.com/mermaid-js/mermaid/blob/develop/CHANGELOG.md)

Due to this change, an issue has arisen where JavaScript in the HTML document created via the `uftrace dump --mermaid` command fails to function as expected. The error message provided is shown below:

```
mermaid.min.js:6 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'createElementNS')
    at Function.<anonymous> (mermaid.min.js:6:16723)
    ...
    at mermaid.min.js:87:669
    at new Promise (<anonymous>)
    at Object.UOt [as render] (mermaid.min.js:87:479)
    at renderGraph (a.html:270:10)
    at drawBaseGraph (a.html:246:2)
    at HTMLButtonElement.onclick (a.html:40:61)
```
![Honeycam 2023-08-03 02-54-13](https://github.com/namhyung/uftrace/assets/25397908/4cf44788-77ce-4656-b18b-e3941d0d5a26)

We can solve the problem by simply modifying the source code to use v9, but it seems necessary to rewrite the entire source code to ESM for maintenance.

PR #1794, opened 2023-08-02
dump: Fix Mermaid to work properly by specifying the version of Mermaid

mirusu400:
After Mermaid v10, the function `mermaid.render` does not support callback, but only supports async operation.

This commit fixed the Mermaid to work by temporarily specifying the version of Mermaid as v9.

Ps. The PR has temporarily specified a version of the mermaid. Please do not close issue as we are planning to rewrite `mermaid.js` for ESM in the future.

Ref: #1793

### DIFF
--- a/cmds/dump.c
+++ b/cmds/dump.c
@@ -1354,7 +1354,7 @@ static void dump_mermaid_footer(struct uftrace_dump_ops *ops, struct uftrace_dat
 	pr_out("flowchart TB\n");
 	print_graph_node_mermaid(&mermaid_graph, &mermaid_graph.root);
 	pr_out("</div>\n");
-	pr_out("<script src=\"https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js\"></script>\n");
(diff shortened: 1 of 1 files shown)


honggyukim:
Hi @mirusu400, thanks for the patch. Do you mean this patch fixes the `Sub-Graph Selectcion` button?

Could you please explain more which mermaid feature is broken in the current master?

honggyukim:
I have tested the current master doesn't allow me to click a node for `Sub-Graph Selection`, but this patch fixes the problem.

mirusu400:
@honggyukim Every option - `Full-Graph`, `Sub-Graph Selection`, `Depth` are not work because all of them use `mermaid.render` internally.

namhyung (approved):
I'm not familiar with Mermaid, but the change looks ok.

honggyukim:
@mirusu400 Thanks for the update. I'm fine with the change but could you enhance the commit message with better description? I think the description in the issue #1793 is better than the current commit message.

In addition, we don't use `Ref: #1793`. Please change it to `Fixed: #1793`.

mirusu400:
@honggyukim I fixed commit message, could you please review again?

honggyukim (approved):
LGTM. Thanks for the fix!

PR #1798, opened 2023-08-05
mcount-dynamic: Use MAP_FIXED_NOREPLACE for trampoline

oldiob:
By trying to allocate the extra page for a trampoline with MAP_FIXED, Uftrace could end up overlapping already mapped pages, which are discarded.  Typically, if data was there with rw- protection, now the data is undefined and has rwx protection.

Since Linux 4.17, the flag MAP_FIXED_NOREPLACE can be used to ensure that the fixed location is mapped atomically, preventing any clobbering.

For earlier kernel, the call to mmap(2) will fall back to a random address, by not honoring the MAP_FIXED request at all.  So callers should ensure that the returned address matches the requested one.

### DIFF
--- a/arch/aarch64/mcount-dynamic.c
+++ b/arch/aarch64/mcount-dynamic.c
@@ -63,13 +63,21 @@ int mcount_setup_trampoline(struct mcount_dynamic_info *mdi)
 	mdi->trampoline -= sizeof(trampoline);
 
 	if (unlikely(mdi->trampoline < mdi->text_addr + mdi->text_size)) {
+		void *trampoline_check;
+
 		mdi->trampoline += sizeof(trampoline);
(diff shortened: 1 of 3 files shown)


honggyukim:
Hi @oldiob, thanks for the PR. We need to think a bit more about the new `MAP_FIXED_NOREPLACE` so we will get back to you later. Especially, we need to think if this is backward compatible on the old kernels before v4.17.

namhyung:
It seems the header defines the flag a preprocessor macro.  Then we can simply check it like:

```c
#ifndef MAP_FIXED_NOREPLACE
#define MAP_FIXED_NOREPLACE  0
#endif
```

namhyung (approved):
I'll merge this for now and add a follow up separately.

Issue #1503, opened 2022-07-31
Add RISC-V architecture support


paranlee:
I think add uftrace riscv support like other architectures already implemented is futuristic fantasy.

I am currently researching arm64 and riscv on the topic of architecture-dependent `mcount` implementation. and I found ftrace mcount implementation of the riscv architecture Linux kernel.

Even if there is no riscv target board, full virtualization is possible in the QEMU environment, so the development environment does not seem to be a problem.

[`arch/riscv/kernel/mcount.S`](https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/mcount.S)

[`arch/riscv/kernel/mcount-dyn.S`](https://github.com/torvalds/linux/blob/master/arch/riscv/kernel/mcount-dyn.S)
```asm
#ifdef CONFIG_DYNAMIC_FTRACE_WITH_REGS
	.macro SAVE_ALL
	addi	sp, sp, -SZREG
	addi	sp, sp, -PT_SIZE_ON_STACK

	REG_S x1,  PT_EPC(sp)
	addi	sp, sp, PT_SIZE_ON_STACK
	REG_L x1,  (sp)
	addi	sp, sp, -PT_SIZE_ON_STACK
	REG_S x1,  PT_RA(sp)
	REG_L x1,  PT_EPC(sp)

	REG_S x2,  PT_SP(sp)
	REG_S x3,  PT_GP(sp)
	REG_S x4,  PT_TP(sp)
	REG_S x5,  PT_T0(sp)
	REG_S x6,  PT_T1(sp)
	REG_S x7,  PT_T2(sp)
	REG_S x8,  PT_S0(sp)
	REG_S x9,  PT_S1(sp)
	REG_S x10, PT_A0(sp)
	REG_S x11, PT_A1(sp)
	REG_S x12, PT_A2(sp)
	REG_S x13, PT_A3(sp)
	REG_S x14, PT_A4(sp)
	REG_S x15, PT_A5(sp)
	REG_S x16, PT_A6(sp)
	REG_S x17, PT_A7(sp)
	REG_S x18, PT_S2(sp)
	REG_S x19, PT_S3(sp)
	REG_S x20, PT_S4(sp)
	REG_S x21, PT_S5(sp)
	REG_S x22, PT_S6(sp)
	REG_S x23, PT_S7(sp)
	REG_S x24, PT_S8(sp)
	REG_S x25, PT_S9(sp)
	REG_S x26, PT_S10(sp)
[...]

PR #1815, opened 2023-08-29
arch/riscv64: Add basic mcount tracing support for RISC-V 64bit

gichoel:
This pull request is a porting effort to enable successful, error-free builds on the RISC-V 64-bit architecture environment and to support tracing using the `mcount` function.

This feature has also been tested on the `VisionFive2` RISC-V board. 
 
The following features are currently supported in the PR.
- Basic mcount Tracing
- Argument Handling

The following features are not yet supported
- PLT hooking (this feature is in progress)
- Dynamic tracing
<br>

This PR was born out of this issue.
- https://github.com/namhyung/uftrace/issues/1503
------

This work has been discussed previously with @honggyukim in the PR link below.
- https://github.com/honggyukim/uftrace/pull/18

The Argument Handling feature was split into the PRs below and then merged back together.
- https://github.com/namhyung/uftrace/pull/1824

### DIFF
--- a/arch/riscv64/mcount-support.c
+++ b/arch/riscv64/mcount-support.c
@@ -4,24 +4,218 @@
 #include "utils/filter.h"
 #include "utils/utils.h"
 
-/* TODO: not implemented yet (Start) */
+static int mcount_get_register_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
+{
+	struct mcount_regs *regs = ctx->regs;
+	int reg_idx;
+
+	switch (spec->type) {
+	case ARG_TYPE_REG:
+		reg_idx = spec->reg_idx;
+		break;
+	case ARG_TYPE_INDEX:
+		reg_idx = spec->idx; /* for integer arguments */
+		break;
+	case ARG_TYPE_FLOAT:
+		reg_idx = spec->idx + UFT_RISCV64_REG_FLOAT_BASE;
+		break;
+	case ARG_TYPE_STACK:
+	default:
+		return -1;
+	}
+
+	ctx->val.i = 0;
+
+	switch (reg_idx) {
+	case UFT_RISCV64_REG_A0:
+		ctx->val.i = ARG1(regs);
+		break;
+	case UFT_RISCV64_REG_A1:
+		ctx->val.i = ARG2(regs);
+		break;
+	case UFT_RISCV64_REG_A2:
+		ctx->val.i = ARG3(regs);
+		break;
+	case UFT_RISCV64_REG_A3:
+		ctx->val.i = ARG4(regs);
+		break;
+	case UFT_RISCV64_REG_A4:
+		ctx->val.i = ARG5(regs);
+		break;
+	case UFT_RISCV64_REG_A5:
+		ctx->val.i = ARG6(regs);
+		break;
+	case UFT_RISCV64_REG_A6:
+		ctx->val.i = ARG7(regs);
+		break;
+	case UFT_RISCV64_REG_A7:
+		ctx->val.i = ARG8(regs);
+		break;
+	case UFT_RISCV64_REG_FA0:
+		asm volatile("fsd fa0, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA1:
+		asm volatile("fsd fa1, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA2:
+		asm volatile("fsd fa2, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA3:
+		asm volatile("fsd fa3, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA4:
+		asm volatile("fsd fa4, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA5:
+		asm volatile("fsd fa5, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA6:
+		asm volatile("fsd fa6, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA7:
+		asm volatile("fsd fa7, %0\n" : "=m"(ctx->val.v));
+		break;
+	default:
+		return -1;
+	}
+
+	return 0;
+}
+
+static void mcount_get_stack_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
+{
+	int offset;
+	unsigned long *addr = ctx->stack_base;
+
+	switch (spec->type) {
+	case ARG_TYPE_STACK:
+		offset = spec->stack_ofs;
+		break;
+	case ARG_TYPE_INDEX:
+		offset = spec->idx - ARCH_MAX_REG_ARGS;
+		break;
+	case ARG_TYPE_FLOAT:
+		offset = (spec->idx - ARCH_MAX_FLOAT_REGS) * 2 - 1;
+		break;
+	case ARG_TYPE_REG:
+	default:
+		/* should not reach here */
+		pr_err_ns("invalid stack access for arguments\n");
+		break;
+	}
+
+	if (offset < 1 || offset > 100) {
+		pr_dbg("invalid stack offset: %d\n", offset);
+		mcount_memset4(ctx->val.v, 0, sizeof(ctx->val));
+		return;
+	}
+
+	addr += offset;
+
+	if (check_mem_region(ctx, (unsigned long)addr)) {
+		/* save long double arguments properly */
+		mcount_memcpy4(ctx->val.v, addr, ALIGN(spec->size, 4));
+	}
+	else {
+		pr_dbg("stack address is not allowed: %p\n", addr);
+		mcount_memset4(ctx->val.v, 0, sizeof(ctx->val));
+	}
+}
+
+static void mcount_get_struct_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
+{
+	struct uftrace_arg_spec reg_spec = {
+		.type = ARG_TYPE_REG,
+	};
+	void *ptr = ctx->val.p;
+	int i;
+
+	for (i = 0; i < spec->struct_reg_cnt; i++) {
+		reg_spec.reg_idx = spec->struct_regs[i];
+
+		mcount_get_register_arg(ctx, &reg_spec);
+		mcount_memcpy4(ptr, ctx->val.v, sizeof(long));
+		ptr += sizeof(long);
+	}
+
+	if (spec->stack_ofs > 0) {
+		unsigned long *addr = ctx->stack_base + spec->stack_ofs;
+
+		/*
+		 * it cannot call mcount_get_stack_arg() since the struct
+		 * might be bigger than the ctx->val.  It directly updates
+		 * the argument buffer (in the ptr).
+		 */
+		if (check_mem_region(ctx, (unsigned long)addr))
+			mcount_memcpy4(ptr, addr, spec->size);
+		else {
+			pr_dbg("stack address is not allowed: %p\n", addr);
+			mcount_memset4(ptr, 0, spec->size);
+		}
+	}
+	else if (spec->struct_reg_cnt == 0) {
+		mcount_get_register_arg(ctx, spec);
+		mcount_memcpy4(ptr, ctx->val.v, sizeof(long));
+	}
+}
+
 void mcount_arch_get_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
 {
-	return;
+	if (spec->fmt == ARG_FMT_STRUCT) {
+		mcount_get_struct_arg(ctx, spec);
+		return;
+	}
+
+	if (mcount_get_register_arg(ctx, spec) < 0)
+		mcount_get_stack_arg(ctx, spec);
 }
 
 void mcount_arch_get_retval(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
 {
-	return;
+	if (spec->fmt == ARG_FMT_STRUCT)
+		mcount_memcpy4(ctx->val.v, ctx->retval, sizeof(long));
+	/* type of return value cannot be FLOAT, so check format instead */
+	else if (spec->fmt == ARG_FMT_FLOAT) {
+		long *float_retval = ctx->retval - 2;
+
+		if (spec->size <= 4) {
+			asm volatile("flw fa0, %1\n"
+				     "fsw fa0, %0\n"
+				     : "=m"(ctx->val.v)
+				     : "m"(*float_retval));
+		}
+		else {
+			asm volatile("fld fa0, %1\n"
+				     "fsd fa0, %0\n"
+				     : "=m"(ctx->val.v)
+				     : "m"(*float_retval));
+		}
+	}
+	else
+		mcount_memcpy4(ctx->val.v, ctx->retval, spec->size);
 }
 
 void mcount_save_arch_context(struct mcount_arch_context *ctx)
 {
-	return;
+	asm volatile("fsd fa0, %0\n" : "=m"(ctx->f[0]));
+	asm volatile("fsd fa1, %0\n" : "=m"(ctx->f[1]));
+	asm volatile("fsd fa2, %0\n" : "=m"(ctx->f[2]));
+	asm volatile("fsd fa3, %0\n" : "=m"(ctx->f[3]));
+	asm volatile("fsd fa4, %0\n" : "=m"(ctx->f[4]));
+	asm volatile("fsd fa5, %0\n" : "=m"(ctx->f[5]));
+	asm volatile("fsd fa6, %0\n" : "=m"(ctx->f[6]));
+	asm volatile("fsd fa7, %0\n" : "=m"(ctx->f[7]));
 }
 
 void mcount_restore_arch_context(struct mcount_arch_context *ctx)
 {
-	return;
+	asm volatile("fld fa0, %0\n" ::"m"(ctx->f[0]));
+	asm volatile("fld fa1, %0\n" ::"m"(ctx->f[1]));
+	asm volatile("fld fa2, %0\n" ::"m"(ctx->f[2]));
+	asm volatile("fld fa3, %0\n" ::"m"(ctx->f[3]));
+	asm volatile("fld fa4, %0\n" ::"m"(ctx->f[4]));
+	asm volatile("fld fa5, %0\n" ::"m"(ctx->f[5]));
+	asm volatile("fld fa6, %0\n" ::"m"(ctx->f[6]));
+	asm volatile("fld fa7, %0\n" ::"m"(ctx->f[7]));
 }
 /* TODO: not implemented yet (End) */
--- a/arch/riscv64/mcount-support.c
+++ b/arch/riscv64/mcount-support.c
@@ -1,0 +1,27 @@
+#include <stdlib.h>
+
+#include "libmcount/internal.h"
+#include "utils/filter.h"
+#include "utils/utils.h"
+
+/* TODO: not implemented yet (Start) */
+void mcount_arch_get_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
+{
+	return;
+}
+
+void mcount_arch_get_retval(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
+{
+	return;
+}
+
+void mcount_save_arch_context(struct mcount_arch_context *ctx)
+{
+	return;
+}
+
+void mcount_restore_arch_context(struct mcount_arch_context *ctx)
+{
+	return;
+}
+/* TODO: not implemented yet (End) */
--- a/arch/riscv64/mcount.S
+++ b/arch/riscv64/mcount.S
@@ -2,10 +2,85 @@
 
 .text
 
-/* TODO: not implemented yet (Start) */
 GLOBAL(_mcount)
+	/* setup frame pointer & return address */
+	addi sp, sp, -80
+	sd ra, 72(sp)
+	sd fp, 64(sp)
+	addi fp, sp, 80
+
+	/* save arguments */
+	sd a7, 56(sp)
+	sd a6, 48(sp)
+	sd a5, 40(sp)
+	sd a4, 32(sp)
+	sd a3, 24(sp)
+	sd a2, 16(sp)
+	sd a1, 8(sp)
+	sd a0, 0(sp)
+
+	/* parent location */
+	ld t1, 64(sp)
+	addi t1, t1, -8
+	mv a0, t1
+
+	/* child addr */
+	mv a1, ra
+
+	/* mcount_args */
+	mv a2, sp
+
+	/* call mcount_entry func */
+	call mcount_entry
+
+	/* restore argunents */
+	ld a0, 0(sp)
+	ld a1, 8(sp)
+	ld a2, 16(sp)
+	ld a3, 24(sp)
+	ld a4, 32(sp)
+	ld a5, 40(sp)
+	ld a6, 48(sp)
+	ld a7, 56(sp)
+
+	/* restore frame pointer */
+	ld fp, 64(sp)
+	ld ra, 72(sp)
+
+	addi sp, sp, 80
+
+	ret
 END(_mcount)
 
 ENTRY(mcount_return)
+	/* setup frame pointer & return address */
+	addi sp, sp, -32
+	sd ra, 24(sp)
+	sd fp, 16(sp)
+	addi fp, sp, 32
+
+	/* save return values */
+	sd a1, 8(sp)
+	sd a0, 0(sp)
+
+	/* set the first argument of mcount_exit as pointer to return values */
+	addi a0, sp, 0
+
+	/* call mcount_exit func */
+	call mcount_exit
+
+	mv t1, a0
+
+	/* restore return values */
+	ld a0, 0(sp)
+	ld a1, 8(sp)
+
+	/* restore frame pointer */
+	ld fp, 16(sp)
+	ld ra, 24(sp)
+
+	addi sp, sp, 32
+
+	/* call return address */
+	jr t1
 END(mcount_return)
-/* TODO: not implemented yet (End) */
--- a/arch/riscv64/mcount.S
+++ b/arch/riscv64/mcount.S
@@ -1,0 +1,11 @@
+#include "utils/asm.h"
+
+.text
+
+/* TODO: not implemented yet (Start) */
+GLOBAL(_mcount)
+END(_mcount)
+
+ENTRY(mcount_return)
+END(mcount_return)
+/* TODO: not implemented yet (End) */
--- a/utils/arch.h
+++ b/utils/arch.h
@@ -12,6 +12,7 @@ enum uftrace_cpu_arch {
 	UFT_CPU_ARM,
 	UFT_CPU_AARCH64,
 	UFT_CPU_I386,
+	UFT_CPU_RISCV64,
 };
 
 static inline enum uftrace_cpu_arch host_cpu_arch(void)
@@ -24,6 +25,8 @@ static inline enum uftrace_cpu_arch host_cpu_arch(void)
 	return UFT_CPU_AARCH64;
 #elif defined(__i386__)
 	return UFT_CPU_I386;
+#elif defined(__riscv) && __riscv_xlen == 64
+	return UFT_CPU_RISCV64;
 #else
 	return UFT_CPU_NONE;
 #endif
@@ -34,6 +37,7 @@ static inline bool arch_is_lp64(enum uftrace_cpu_arch arch)
 	switch (arch) {
 	case UFT_CPU_X86_64:
 	case UFT_CPU_AARCH64:
+	case UFT_CPU_RISCV64:
 		return true;
 	default:
 		return false;
--- a/utils/arch.h
+++ b/utils/arch.h
@@ -157,6 +157,30 @@ enum uftrace_i386_reg_index {
 	UFT_I386_REG_XMM7,
 };
 
+enum uftrace_riscv64_reg_index {
+	UFT_RISCV64_REG_INT_BASE = 0,
+	/* integer argument registers */
+	UFT_RISCV64_REG_A0,
+	UFT_RISCV64_REG_A1,
+	UFT_RISCV64_REG_A2,
+	UFT_RISCV64_REG_A3,
+	UFT_RISCV64_REG_A4,
+	UFT_RISCV64_REG_A5,
+	UFT_RISCV64_REG_A6,
+	UFT_RISCV64_REG_A7,
+
+	UFT_RISCV64_REG_FLOAT_BASE = 100,
+	/* floating-point argument registers */
+	UFT_RISCV64_REG_FA0,
+	UFT_RISCV64_REG_FA1,
+	UFT_RISCV64_REG_FA2,
+	UFT_RISCV64_REG_FA3,
+	UFT_RISCV64_REG_FA4,
+	UFT_RISCV64_REG_FA5,
+	UFT_RISCV64_REG_FA6,
+	UFT_RISCV64_REG_FA7,
+};
+
 int arch_register_number(enum uftrace_cpu_arch arch, char *reg_name);
 int arch_register_at(enum uftrace_cpu_arch arch, bool integer, int idx);
 int arch_register_index(enum uftrace_cpu_arch arch, int idx);
--- a/utils/compiler.h
+++ b/utils/compiler.h
@@ -37,6 +37,29 @@
 #endif
 #endif
 
+/* TODO: not implemented yet (Start)
+ *
+ *	From RISC-V's "The RISC-V Instruction Set Manual, Volume I: User-Level
+ *	ISA, Document Version 20191213", the G Extension consists of "I, M, A,
+ *	F, D, Zicsr, Zifencei".
+ *
+ *	In RISC-V, two instructions exist for memory barriers: the fence
+ *	instruction, defined in the I Extension, and the fence.i instruction,
+ *	defined in the Zifencei Extension.
+ *
+ *	So the memory barrier commands we can use are fence, fence.i, and
+ *	we'll have to figure out which one we need later when we implement
+ *	the functions that call those macro functions.
+ *
+ */
+#if defined(__riscv)
+#define cpu_relax() asm volatile("nop" ::: "memory")
+#define full_memory_barrier() asm volatile("nop" ::: "memory")
+#define read_memory_barrier() asm volatile("nop" ::: "memory")
+#define write_memory_barrier() asm volatile("nop" ::: "memory")
+#endif
+/* TODO: not implemented yet (End) */
+
 /* ignore 'restrict' keyword if not supported (before C99) */
 #if !defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L
 #define restrict
--- a/arch/riscv64/Makefile
+++ b/arch/riscv64/Makefile
@@ -1,0 +1,34 @@
+LINKFLAGS := -r -z noexecstack
+
+sdir := $(srcdir)/arch/riscv64
+odir := $(objdir)/arch/riscv64
+
+include $(srcdir)/Makefile.include
+
+ARCH_ENTRY_SRC = $(wildcard $(sdir)/*.S)
+ARCH_MCOUNT_SRC = $(wildcard $(sdir)/mcount-*.c)
+ARCH_UFTRACE_SRC = $(sdir)/cpuinfo.c
+
+ARCH_MCOUNT_OBJS  = $(patsubst $(sdir)/%.S,$(odir)/%.op,$(ARCH_ENTRY_SRC))
+ARCH_MCOUNT_OBJS += $(patsubst $(sdir)/%.c,$(odir)/%.op,$(ARCH_MCOUNT_SRC))
+ARCH_UFTRACE_OBJS = $(patsubst $(sdir)/%.c,$(odir)/%.o,$(ARCH_UFTRACE_SRC))
+
+all: $(odir)/entry.op
+
+$(odir)/mcount-entry.op: $(ARCH_MCOUNT_OBJS)
+	$(QUIET_LINK)$(LD) $(LINKFLAGS) -o $@ $^
+
+$(odir)/uftrace.o: $(ARCH_UFTRACE_OBJS)
+	$(QUIET_LINK)$(LD) $(LINKFLAGS) -o $@ $^
+
+$(odir)/%.op: $(sdir)/%.S
+	$(QUIET_ASM)$(CC) $(LIB_CFLAGS) -c -o $@ $<
+
+$(odir)/%.op: $(sdir)/%.c
+	$(QUIET_CC_FPIC)$(CC) $(LIB_CFLAGS) -c -o $@ $<
+
+$(odir)/%.o: $(sdir)/%.c
+	$(QUIET_CC)$(CC) $(UFTRACE_CFLAGS) -c -o $@ $<
+
+clean:
+	$(RM) $(odir)/*.op $(odir)/*.o
--- a/arch/riscv64/cpuinfo.c
+++ b/arch/riscv64/cpuinfo.c
@@ -1,0 +1,24 @@
+#include <stdio.h>
+#include <string.h>
+
+int arch_fill_cpuinfo_model(int fd)
+{
+	char buf[1024];
+	FILE *fp;
+	int ret = -1;
+
+	fp = fopen("/proc/cpuinfo", "r");
+	if (fp == NULL)
+		return -1;
+
+	while (fgets(buf, sizeof(buf), fp) != NULL) {
+		if (!strncmp(buf, "isa\t\t: rv64", 11)) {
+			dprintf(fd, "cpuinfo:desc=RISCV64_%s", &buf[12]);
+			ret = 0;
+			break;
+		}
+	}
+
+	fclose(fp);
+	return ret;
+}
--- a/arch/riscv64/mcount-arch.h
+++ b/arch/riscv64/mcount-arch.h
@@ -1,0 +1,52 @@
+#ifndef MCOUNT_ARCH_H
+#define MCOUNT_ARCH_H
+
+#define mcount_regs mcount_regs
+
+struct mcount_regs {
+	unsigned long a0;
+	unsigned long a1;
+	unsigned long a2;
+	unsigned long a3;
+	unsigned long a4;
+	unsigned long a5;
+	unsigned long a6;
+	unsigned long a7;
+};
+
+#define ARG1(x) ((x)->a0)
+#define ARG2(x) ((x)->a1)
+#define ARG3(x) ((x)->a2)
+#define ARG4(x) ((x)->a3)
+#define ARG5(x) ((x)->a4)
+#define ARG6(x) ((x)->a5)
+#define ARG7(x) ((x)->a6)
+#define ARG8(x) ((x)->a7)
+
+#define ARCH_MAX_REG_ARGS 8
+#define ARCH_MAX_FLOAT_REGS 8
+
+#define HAVE_MCOUNT_ARCH_CONTEXT
+struct mcount_arch_context {
+	double f[ARCH_MAX_FLOAT_REGS];
+};
+
+#if defined(__riscv_compressed)
+#define NOP_INSN_SIZE 2
+#else
+#define NOP_INSN_SIZE 4
+#endif
+
+/* TODO: not implemented yet (Start) */
+#define ARCH_PLT0_SIZE 0
+#define ARCH_PLTHOOK_ADDR_OFFSET 0
+
+struct mcount_disasm_engine;
+struct mcount_dynamic_info;
+struct mcount_disasm_info;
+
+int disasm_check_insns(struct mcount_disasm_engine *disasm, struct mcount_dynamic_info *mdi,
+		       struct mcount_disasm_info *info);
+/* TODO: not implemented yet (End) */
+
+#endif /* MCOUNT_ARCH_H */
--- a/arch/riscv64/plthook.S
+++ b/arch/riscv64/plthook.S
@@ -1,0 +1,11 @@
+#include "utils/asm.h"
+
+.text
+
+/* TODO: not implemented yet (Start) */
+ENTRY(plt_hooker)
+END(plt_hooker)
+
+ENTRY(plthook_return)
+END(plthook_return)
+/* TODO: not implemented yet (End) */
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -281,6 +281,9 @@ static int read_cpuinfo(void *arg)
 			else if (!strncmp(info->cpudesc, "ARM64", 5)) {
 				handle->arch = UFT_CPU_AARCH64;
 			}
+			else if (!strncmp(info->cpudesc, "RISCV64", 7)) {
+				handle->arch = UFT_CPU_RISCV64;
+			}
 			else if (data_is_lp64(handle)) {
 				handle->arch = UFT_CPU_X86_64;
 			}
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -1589,7 +1589,7 @@ static void check_binary(struct uftrace_opts *opts)
 	static char altname[PATH_MAX]; // for opts->exename to be persistent
 	uint16_t e_type;
 	uint16_t e_machine;
-	uint16_t supported_machines[] = { EM_X86_64, EM_ARM, EM_AARCH64, EM_386 };
+	uint16_t supported_machines[] = { EM_X86_64, EM_ARM, EM_AARCH64, EM_386, EM_RISCV };
 
 again:
 	/* if it cannot be found in PATH, then fails inside */
--- a/uftrace.c
+++ b/uftrace.c
@@ -1389,7 +1389,10 @@ int main(int argc, char *argv[])
 	struct uftrace_opts opts = {
 		.mode = UFTRACE_MODE_INVALID,
 		.dirname = UFTRACE_DIR_NAME,
+#if !defined(__riscv)
+		/* FIXME: disable libcall until PLT hooking is implemented in riscv64. */
 		.libcall = true,
+#endif
 		.bufsize = SHMEM_BUFFER_SIZE,
 		.max_stack = OPT_RSTACK_DEFAULT,
 		.port = UFTRACE_RECV_PORT,
--- a/utils/dwarf.c
+++ b/utils/dwarf.c
@@ -399,6 +399,13 @@ static void setup_arg_data(struct arg_data *ad, const char *name, struct uftrace
 		ad->struct_arg_needs_ptr = true;
 		/* struct return will use 'x8' register */
 		break;
+	case UFT_CPU_RISCV64:
+		ad->reg_max = 8;
+		ad->fpreg_max = 8;
+		ad->struct_arg_needs_ptr = true;
+		ad->struct_return_needs_ptr = true;
+		ad->struct_uses_fpreg = true;
+		break;
 	default:
 		/* TODO */
 		ad->broken = true;
@@ -585,6 +592,10 @@ static void setup_param_data(struct param_data *data)
 		break;
 	case UFT_CPU_AARCH64:
 		data->max_struct_size = 16 * 8;
+		break;
+	case UFT_CPU_RISCV64:
+		data->max_struct_size = 16 * 8;
+		data->use_fpregs = true;
 		break;
 	default:
 		/* TODO */
--- a/utils/regs.c
+++ b/utils/regs.c
@@ -131,8 +131,42 @@ static const struct uftrace_reg_table uft_i386_reg_table[] = {
 #undef X86_REG
 };
 
+static const struct uftrace_reg_table uft_riscv64_reg_table[] = {
+#define RISCV64_REG(_r)                                                                            \
+	{                                                                                          \
+#_r, UFT_RISCV64_REG_##_r                                                          \
+	}
+
+	/* integer registers */
+	RISCV64_REG(A0),
+	RISCV64_REG(A1),
+	RISCV64_REG(A2),
+	RISCV64_REG(A3),
+	RISCV64_REG(A4),
+	RISCV64_REG(A5),
+	RISCV64_REG(A6),
+	RISCV64_REG(A7),
+
+	/* floating-point registers */
+	RISCV64_REG(FA0),
+	RISCV64_REG(FA1),
+	RISCV64_REG(FA2),
+	RISCV64_REG(FA3),
+	RISCV64_REG(FA4),
+	RISCV64_REG(FA5),
+	RISCV64_REG(FA6),
+	RISCV64_REG(FA7),
+
+#undef RISCV64_REG
+};
+
 static const struct uftrace_reg_table *arch_reg_tables[] = {
-	NULL, uft_x86_64_reg_table, uft_arm_reg_table, uft_aarch64_reg_table, uft_i386_reg_table,
+	NULL,
+	uft_x86_64_reg_table,
+	uft_arm_reg_table,
+	uft_aarch64_reg_table,
+	uft_i386_reg_table,
+	uft_riscv64_reg_table,
 };
 
 static const size_t arch_reg_sizes[] = {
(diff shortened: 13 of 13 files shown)


gichoel:
If you are interested in doing basic tracing with the mcount function in a RISC-V environment, please see the branch at the link below.
- https://github.com/gichoel/uftrace/tree/arch/riscv-support-mcount

SEOKMIN83:
> If you are interested in doing basic tracing with the mcount function in a RISC-V environment, please see the branch at the link below.
> 
> * https://github.com/gichoel/uftrace/tree/arch/riscv-support-mcount

I  tried basic tracing on RISC-V with your implementation, 
and it worked fine.
Thank you so much for making uftrace usable 
in the RISC-V environment.

SEOKMIN83:
In my opinion,
it would be more useful to have a detailed manual 
on how to do basic tracing tests in a RISC-V environment.

gichoel:
Thanks for the good point @seokmin83. I had forgotten the part about needing a way to run this. 

Currently, this task can be executed as follows after building from the root directory of the uftrace source code.

1. compile the target program
- Using gcc
  ```
  $ gcc -o t-abc -pg tests/s-abc.c
  ```

- Using clang
  ```
  $ clang -o t-abc -pg tests/s-abc.c`
  ```

2. Trace the target program with uftrace
- Run uftrace with Target Program
  ```
  $ ./uftrace --libmcount-path=. --no-libcall t-abc
  ```

SEOKMIN83:
> Thanks for the good point @SEOKMIN83. I had forgotten the part about needing a way to run this.
> 
> Currently, this task can be executed as follows after building from the root directory of the uftrace source code.
> 
> 1. compile the target program
> 
> * Using gcc
>   ```
>   $ gcc -o t-abc -pg tests/s-abc.c
>   ```
> * Using clang
>   ```
>   $ clang -o t-abc -pg tests/s-abc.c`
>   ```
> 
> 2. Trace the target program with uftrace
> 
> * Run uftrace with Target Program
>   ```
>   $ ./uftrace --libmcount-path=. --no-libcall t-abc
>   ```

Thank you☺️☺️🌸🌸 
It'll be great help to those who want to test it.

gichoel:
I believe the code can be reduced to the following after testing.

```diff
diff --git a/arch/riscv64/mcount.S b/arch/riscv64/mcount.S
index 91d677bb..498f38e0 100644
--- a/arch/riscv64/mcount.S
+++ b/arch/riscv64/mcount.S
@@ -59,20 +59,12 @@ ENTRY(mcount_return)
        sd fp, 8(sp)
        addi fp, sp, 24

-       /* save return values */
-       sd a0, 0(sp)
-
        /* set the first argument of mcount_exit as pointer to return values */
        addi a0, sp, 0

        /* call mcount_exit func */
        call mcount_exit

-       mv t1, a0
-
-       /* restore return values */
-       ld a0, 0(sp)
-
        /* restore frame pointer */
        ld s0, 8(sp)
        ld ra, 16(sp)
@@ -80,5 +72,5 @@ ENTRY(mcount_return)
        addi sp, sp, 24

        /* call return address */
-       jr t1
+      jr a0
 END(mcount_return)
```

If you try this code and see any issues, please let me know in the comments.

honggyukim (commented):
In d8ad79e4009bd44b35ca931d221510759e6fc02a `arch/riscv64: Update 'mcount.S' to support tracking based on mcount func`,

> Info : When compiling the target program with gcc, you need to add
> '-O0' or add '-fno-omit-frame-pointer' as an option because it will
> disable frame pointers when optimization options are enabled.

Users do not have to specify `-O0` because it is the default setting. You better explain it as follows.
```
In order to figure out the address of parent_loc, we need the frame pointer,
but compiler optimization such as `-O2` in gcc removes the `fp` so we won't
be able know where to change to hijack the return address to `mcount_return`.
This problem only happens in gcc, but not in clang.

To avoid the problem, `-fno-omit-frame-pointer` must be used when gcc
optimization option is used in riscv64.
```

In addition, we better have the following change in this PR.
```diff
diff --git a/uftrace.c b/uftrace.c
index 69e6cefc..69486e94 100644
--- a/uftrace.c
+++ b/uftrace.c
@@ -1389,7 +1389,10 @@ int main(int argc, char *argv[])
        struct uftrace_opts opts = {
                .mode = UFTRACE_MODE_INVALID,
                .dirname = UFTRACE_DIR_NAME,
+#if !defined(__riscv)
+               /* FIXME: disable libcall until PLT hooking is implemented in riscv64. */
                .libcall = true,
+#endif
                .bufsize = SHMEM_BUFFER_SIZE,
                .max_stack = OPT_RSTACK_DEFAULT,
                .port = UFTRACE_RECV_PORT,
```
This is needed to avoid segfault when `--no-libcall` is not used.

honggyukim:
It looks the argument handling is incorrect even for integer types for now.
```
$ gcc -pg -g tests/s-abc.c -o t-abc
$ uftrace -a --no-libcall t-abc 
            [265212] | main(0x3f80d79808, 0x3fc59ab208) {
            [265212] |   a() {
            [265212] |     b() {
   2.500 us [265212] |       c() = 65212;
   7.250 us [265212] |     } = 0xffffffffc7cc78c5; /* b */
   8.500 us [265212] |   } = 0xffffffffc7cc789f; /* a */
  13.500 us [265212] | } = 0; /* main */
```
The return address of `b` and `a` are incorrect.

gichoel:
I will update the commit to incorporate your feedback.

Also, the argument handling issue has been confirmed in my environment and seems to be related to the mcount_return part, which I recently removed as unnecessary.

As such, I will be reverting that code back.

gichoel:
And I was able to see the following results after the patch.

Thanks to this, we no longer need to use the `--no-libcall` option!
```
$ gcc -pg -g tests/s-abc.c -o t-abc
$ ./uftrace --libmcount-path=. -a t-abc
            [274658] | main(0x3f8da21808, 0x3fe7908478) {
            [274658] |   a() {
            [274658] |     b() {
   2.000 us [274658] |       c() = 74658;
   6.250 us [274658] |     } = 74659; /* b */
   8.250 us [274658] |   } = 74658; /* a */
  13.000 us [274658] | } = 0; /* main */
```

SEOKMIN83:
> And I was able to see the following results after the patch.
> 
> Thanks to this, we no longer need to use the `--no-libcall` option!
> 
> ```
> $ gcc -pg -g tests/s-abc.c -o t-abc
> $ ./uftrace --libmcount-path=. -a t-abc
> # DURATION     TID     FUNCTION
>             [274658] | main(0x3f8da21808, 0x3fe7908478) {
>             [274658] |   a() {
>             [274658] |     b() {
>    2.000 us [274658] |       c() = 74658;
>    6.250 us [274658] |     } = 74659; /* b */
>    8.250 us [274658] |   } = 74658; /* a */
>   13.000 us [274658] | } = 0; /* main */
> ```

It works well. !!!!!
The below is the result of no use the `--no-libcall` option.
``` shell
sm_ple38@starfive:~/uftrace$ gcc -pg -g tests/s-abc.c -o t-abc
sm_ple38@starfive:~/uftrace$ ./uftrace --libmcount-path=. -a t-abc
            [285395] | main(0x3f8ba48808, 0x3fe94d3338) {
            [285395] |   a() {
            [285395] |     b() {
   1.750 us [285395] |       c() = 85395;
   6.000 us [285395] |     } = 85396; /* b */
   7.250 us [285395] |   } = 85395; /* a */
  12.750 us [285395] | } = 0; /* main */
```

coyotek:
If it is a string, it will not work. 
The second argument of the bar() function is different.
Below are the test results. 
The test source file is `./tests/s-arg.c`.
* for x86_64
```bash
$ ./uftrace -a t-arg
   0.691 us [224359] | __monstartup();
   0.140 us [224359] | __cxa_atexit();
            [224359] | main(1, 0x7fff378d7338) {
            [224359] |   foo(3) {
            [224359] |     bar(2, "c") {
   0.561 us [224359] |       strcmp("c", "c") = 0;
 184.333 us [224359] |     } = 0; /* bar */
            [224359] |     bar(1, "b") {
   0.220 us [224359] |       strcmp("b", "b") = 0;
   0.872 us [224359] |     } = 0; /* bar */
            [224359] |     bar(0, "a") {
   0.190 us [224359] |       strcmp("a", "a") = 0;
   0.682 us [224359] |     } = 0; /* bar */
 187.098 us [224359] |   } = 0; /* foo */
   0.270 us [224359] |   many(12) = 0;
            [224359] |   pass(3) {
   0.531 us [224359] |     check(big{...}, 0x7fff378d7190) = 0;
   1.082 us [224359] |   } = 0; /* pass */
 190.775 us [224359] | } = 0; /* main */
```

* for RISC-V
```bash
$ ./uftrace -a t-arg
            [286418] | main(1, 0x3ff0657448) {
            [286418] |   foo(3) {
   2.501 us [286418] |     bar(2, 68944) = 0;
   0.500 us [286418] |     bar(1, 68942) = 0;
   0.500 us [286418] |     bar(0, 68940) = 0;
  10.001 us [286418] |   } = 0; /* foo */
   1.250 us [286418] |   many(12) = 0;
            [286418] |   pass(3) {
   1.250 us [286418] |     check(big{...}, 0x3ff0657200) = 0;
   4.250 us [286418] |   } = 0; /* pass */
  21.501 us [286418] | } = 0; /* main */
```
Also, if you change the order of the arguments in the bar() function, the result is the same.
```bash
$ ./uftrace -a t-arg2                   
            [287700] | main(1, 0x3fd905b438) {                  
            [287700] |   foo(3) {                               
   3.000 us [287700] |     bar(68944, 2) = 0;                   
   0.750 us [287700] |     bar(68942, 1) = 0;                   
   0.750 us [287700] |     bar(68940, 0) = 0;                   
  10.750 us [287700] |   } = 0; /* foo */                       
   1.000 us [287700] |   many(12) = 0;                          
            [287700] |   pass(3) {                              
   1.500 us [287700] |     check(big{...}, 0x3fd905b1f0) = 0;   
   4.750 us [287700] |   } = 0; /* pass */                      
  24.000 us [287700] | } = 0; /* main */                        
  ```

**Comment by gichoel** [comment]
Hmm... I found the same problem on RISC-V and tested characters and strings.

- Testing character output
  ```c
  /* test2.c */
  #include <stdio.h>
  
  void foo(int i, char c){
          printf("%d %c\n",i, c);
  }
  
  int main(void){
          foo(15, 'c');
          return 0;
  }
  ```
  
  ```sh
  $ clang -pg -g test2.c -o t-arg3
  $ ./uftrace --libmcount-path=. -a t-arg3

              [288561] | main() {
    12.501 us [288561] |   foo(15, 'c');
    34.001 us [288561] | } = 0; /* main */
  ```

- Testing string output
  ```sh
  $ clang -pg -g tests/s-arg.c -o t-arg
  $ ./uftrace --libmcount-path=. -a t-arg

              [288583] | main(1, 0x3ff7bc7478) {
              [288583] |   foo(3) {
     3.250 us [288583] |     bar(2, 68944) = 0;
     1.250 us [288583] |     bar(1, 68942) = 0;
     1.000 us [288583] |     bar(0, 68940) = 0;
    91.001 us [288583] |   } = 0; /* foo */
     1.250 us [288583] |   many(12) = 0;
              [288583] |   pass(3) {
     1.500 us [288583] |     check(big{...}, 0x3ff7bc7230) = 0;
    32.001 us [288583] |   } = 0; /* pass */
   223.004 us [288583] | } = 0; /* main */
  ```

- Debugging with gdb
  ```gdb
  $ gdb --args ./uftrace --libmcount-path=. -a t-arg

  ......

  (gdb) b _mcount:1
  (gdb) r

  ......

  /* Skip main() func */
  Thread 2.1 "t-arg" hit Breakpoint 1, _mcount () at ...
  7               addi sp, sp, -80
  (gdb) c

  ......

  /* Skip foo() func */
  Thread 2.1 "t-arg" hit Breakpoint 1, _mcount () at  ...
  7               addi sp, sp, -80
  (gdb) c

  ......

  /* in bar() func */
  Thread 2.1 "t-arg" hit Breakpoint 1, _mcount () at ...
  7               addi sp, sp, -80
  (gdb) n

  ......

  (gdb) n
  20              sd a0, 0(sp)

  (gdb) info register
  ra             0x107e0  0x107e0 <bar+20>
  sp             0x3ffffff1b0     0x3ffffff1b0
  gp             0x12800  0x12800
  tp             0x3ff7afe7b0     0x3ff7afe7b0
  t0             0x3fffffec08     274877901832
  t1             0x106dc  67292
  t2             0x10     16
  fp             0x3ffffff200     0x3ffffff200
  s1             0x0      0
  a0             0x2      2
  a1             0x10d50  68944
  a2             0x0      0
  a3             0x0      0
  a4             0x3ffffff2f8     274877903608
  a5             0x10b72  68466
  a6             0x0      0
  a7             0x3ff7e95600     274742203904
  s2             0x3ffffff548     274877904200
  s3             0x0      0
  s4             0x3      3
  s5             0x0      0
  s6             0x0      0
  s7             0x2aaaaf9548     183252260168
  s8             0x3ff7cd2450     274740356176
  s9             0x3ffffff1bc     274877903292
  s10            0x3      3
  s11            0x3ffffff7e0     274877904864
  t3             0x3ff7fcbaa2     274743474850
  t4             0x17     23
  t5             0x1483357adb5cdd 5773765251914973
  t6             0x2cb1572ace4    3071261453540
  pc             0x3ff7fcbab8     0x3ff7fcbab8 <_mcount+22>

  (gdb) x/b 0x10d50
  0x10d50:        99 
  ```

In the ASCII code table, the decimal 99 corresponds to 'c', and debugging result that the value of the string argument to the _mcount function is normal.

So I think there is a problem with the code in the mcount-support.c file for the risc-v 64-bit architecture that checks if an argument is a string when it is a pointer, but I'm not familiar with the code so it may take me a while to find the cause.

**Comment by coyotek** [comment]
Like you said
It should print the value `0x10d50`, but it seems to be printing an address.

File: arch/riscv64/mcount-support.c [hunk 1]
```diff
@@ -0,0 +1,219 @@
...
+	case UFT_RISCV64_REG_FA6:
+		asm volatile("fsd fa6, %0\n" : "=m"(ctx->val.v));
+		break;
+	case UFT_RISCV64_REG_FA7:
+		asm volatile("fsd fa7, %0\n" : "=m"(ctx->val.v));
+		break;
+	default:
+		return -1;
+	}
+}
```
honggyukim, line 82: My compiler complains this part as follows.
```
/home/user/uftrace/arch/riscv64/mcount-support.c: In function 'mcount_get_register_arg':
/home/user/uftrace/arch/riscv64/mcount-support.c:82:1: error: control reaches end of non-void function [-Werror=return-type]
   82 | }
      | ^
cc1: all warnings being treated as errors
```
You better remove `default` case, then move `return -1` to the end of this function.

honggyukim (commented):
I get the following error when running it in debug build.
```
$ uftrace record -a a.out 
/home/user/uftrace/utils/regs.c:437: arch_register_dwarf_name: ASSERT `arch < ARRAY_SIZE(arch_dwarf_tables)' failed.
Stack trace:
  #1  0x003fb30d3ca2 stacktrace + 0x34

WARN: child terminated by signal: 5: Trace/breakpoint trap
```
Attaching gdb shows backtrace info as follows.
```
(gdb) r
    ...
/home/user/uftrace/utils/regs.c:437: arch_register_dwarf_name: ASSERT `arch < ARRAY_SIZE(arch_dwarf_tables)' failed.
Stack trace:
  #1  0x003ff7fbaca2 stacktrace + 0x34

Thread 2.1 "a.out" received signal SIGTRAP, Trace/breakpoint trap.
[Switching to Thread 0x3ff7ac9010 (LWP 290922)]
__GI_raise (sig=<optimized out>) at ../sysdeps/unix/sysv/linux/raise.c:49
49      in ../sysdeps/unix/sysv/linux/raise.c
(gdb) bt
#0  __GI_raise (sig=<optimized out>) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x0000003ff7fb0092 in arch_register_dwarf_name (arch=UFT_CPU_RISCV64, dwarf_reg=90) at /home/user/uftrace/utils/regs.c:437
#2  0x0000003ff7fc1456 in add_location (spec=0x3fffffea10 "arg1", len=256, die=0x3fffffeb18, ad=0x3fffffeb80)
    at /home/user/uftrace/utils/dwarf.c:1184
#3  0x0000003ff7fc185e in get_argspec (die=0x3fffffec98, data=0x3fffffeb80) at /home/user/uftrace/utils/dwarf.c:1290
#4  0x0000003ff7fc1fec in get_dwarfspecs_cb (die=0x3fffffec98, data=0x3fffffee10) at /home/user/uftrace/utils/dwarf.c:1512
#5  0x0000003ff7c348b8 in ?? () from /usr/lib/riscv64-linux-gnu/libdw.so.1
```
It looks you need to implement `utf_riscv64_dwarf_table` at https://github.com/namhyung/uftrace/blob/v0.14/utils/regs.c#L415-L421.
```c
static const struct uftrace_reg_table *arch_dwarf_tables[] = {
        NULL,   
        uft_x86_64_dwarf_table,
        uft_arm_dwarf_table,
        uft_aarch64_dwarf_table,
        uft_i386_dwarf_table,
};
```

arch/riscv64/mcount-support.c
gichoel: I also got an error message after the `make DEBUG=1` debug build, which looked like it needed fixing.

Your suggestion is good, but I was using `return 0;` in mcount-support.c on a different architecture.

This was my mistake during the porting process and I think adding `return 0;` is a good way to be consistent, so I will apply the same method as the other architectures.

gichoel:
I was referring to the psABI documentation for RISC-V to implement the DWARF-specific code that was the feedback regarding the Argument Handling feature. 

However, I've realized that the list of changes to be made is longer than I thought and that I need to reorder the previous commits to make this work.

-------------

Therefore, I reordered the commits and made the modifications in the `arch/riscv-support-mcount-v2` branch, but there is a problem when building using the `make DEBUG=1` command. 

I followed the problem backwards using `git bisect` and encountered the following error in the first commit of the RISC-V porting work, so I will fix it and get it reviewed again.

```make
/home/test/workdir/uftrace/arch/riscv64/mcount-support.c:18:13: error: 'mcount_get_struct_arg' defined but not used [-Werror=unused-function].
   18 | static void mcount_get_struct_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
      | ^~~~~~~~~~~~~~~~~~~~~
/home/test/workdir/uftrace/arch/riscv64/mcount-support.c:13:13: error: 'mcount_get_stack_arg' defined but not used [-Werror=unused-function].
   13 | static void mcount_get_stack_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
      | ^~~~~~~~~~~~~~~~~~~~
/home/test/workdir/uftrace/arch/riscv64/mcount-support.c:8:12: error: 'mcount_get_register_arg' defined but not used [-Werror=unused-function].
    8 | static int mcount_get_register_arg(struct mcount_arg_context *ctx, struct uftrace_arg_spec *spec)
      | ^~~~~~~~~~~~~~~~~~~~~~~

```

namhyung:
I think we can merge the basic RISC-V support code first without argument handling.  The new architecture support is not a simple task, we can approach it step by step.  You may want to file a separate issue for the arguments and return value tracing.

gichoel:
Based on feedback, this PR has been reduced to a commit for basic RISC-V support.

If you would like to see commits related to argument handling, please see the list of related PRs in this PR message.

namhyung:
> I will leave only the basic RISC-V support code in the current PR and create a new PR for the argument handling.

Sounds good to me.  So does the current code still have an error on debug builds?

honggyukim:
The current updated version fails as follows.
```
$ uftrace record t-abc 
$ uftrace replay
uftrace: /home/user/work/uftrace/utils/data-file.c:430:open_info_file
```
I didn't fully compare with the previous working version, but `arch_fill_cpuinfo_model` is empty in this version.

gichoel:
I was able to find the cause by referring to @honggyukim's comment. Thanks.

The cause was two changes I made during my recent work on porting the argument handling functionality.
- Changing `arch/riscv64/cpuinfo.c` to implement the code that goes into `cmds/info.c`.
- Changing the commit order of `cpuinfo.c` to implement argument handling.
<br>

Old code
```
/* in arch/riscv64/cpuinfo.c */
......
if (!strncmp(buf, "isa\t\t:", 6)) {
	dprintf(fd, "cpuinfo:desc=%s", &buf[7]);
......
```

Changed code (with error print)
```
/* in arch/riscv64/cpuinfo.c */
......
if (!strncmp(buf, "isa\t\t: rv64", 11)) {
	dprintf(fd, "cpuinfo:desc=RISCV64 (%s)", &buf[12]);
......
```
<br>

--------------

**The reason why I changed to this code** is that in `cmds/info.c`, the architectures except `x86_64` and `i386` are shown below, and I wanted the `RISC-V 64bit` architecture to follow the format of `cpuinfo.c` and `cmds/info.c` for the `aarch64` architecture.
```
/* in cmds/info.c */
if (!strncmp(info->cpudesc, "ARMv6", 5) ||
    !strncmp(info->cpudesc, "ARMv7", 5)) {
	handle->arch = UFT_CPU_ARM;
}
else if (!strncmp(info->cpudesc, "ARM64", 5)) {
	handle->arch = UFT_CPU_AARCH64;
}
```

And I didn't find the reason after several attempts, but I solved it by replacing it like below.
```
/* in arch/riscv64/cpuinfo.c */
......
if (!strncmp(buf, "isa\t\t: rv64", 11)) {
	dprintf(fd, "cpuinfo:desc=RISCV64_%s", &buf[12]);
......
```
<br>

And since the first commit was only for a successful build with no functional behavior, it doesn't work, and both commits succeeded in a debug build.

In conclusion, I don't know why `dprintf` throws an error when `(%s)` is included, but I have solved the problem.

gichoel:
@namhyung

Since both commits were fine in the debug build and the above error was also resolved, we were able to test again to confirm normal behavior.

arch/riscv64/mcount.S
```diff
@@ -2,10 +2,83 @@
...
 
 ENTRY(mcount_return)
+	/* setup frame pointer & return address */
+	addi sp, sp, -24
+	sd ra, 16(sp)
+	sd fp, 8(sp)
+	addi fp, sp, 24
+
+	/* save return values */
+	sd a0, 0(sp)
```
namhyung, line 63: It seems `a1` register also can be used for return value according to the RISC-V calling convention (RVG).

utils/compiler.h
```diff
@@ -37,6 +37,19 @@
 #endif
 #endif
 
+#if defined(__riscv)
+#define cpu_relax() asm volatile("nop" ::: "memory")
+#if __riscv_xlen == 64
```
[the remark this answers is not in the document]
gichoel: Well, when I wrote this code, I didn't know what to do, so I used the `nop` instruction as a temporary fix.

However, it appears that in RISC-V, the `fence` instruction will be used for this part, and this is the same for both `RV32I` and `RV64I`.
<br>

In conclusion, I would combine the 32 and 64 into one and implement the code to use the `fence` instruction.

arch/riscv64/mcount.S
gichoel: You're right, I'll add some code to allow the `a1` register to save and restore the return values as well.

utils/compiler.h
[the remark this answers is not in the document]
gichoel: I was going to implement it with the `fence` instruction, but then I realized that RISC-V has an additional `fence.i` instruction for memory barriers.

So when I implement a function that calls this macro function, I want to implement that macro function as well.

Do you think this would be okay?

utils/arch.h
```diff
@@ -24,6 +25,10 @@ static inline enum uftrace_cpu_arch host_cpu_arch(void)
 	return UFT_CPU_AARCH64;
 #elif defined(__i386__)
 	return UFT_CPU_I386;
+#elif defined(__riscv)
+#if __riscv_xlen == 64
+	return UFT_CPU_RISCV64;
+#endif
```
namhyung, line 31: It'd make it undefined if `__riscv` is defined but `__riscv_xlen` is not 64.  You need to make it in a single condition like

```c
#elif defined(__riscv) && __riscv_xlen == 64
        return UFT_CPU_RISCV64;
#else
```

utils/arch.h
```diff
@@ -153,6 +159,30 @@ enum uftrace_i386_reg_index {
...
+	UFT_RISCV64_REG_FA0,
+	UFT_RISCV64_REG_FA1,
+	UFT_RISCV64_REG_FA2,
+	UFT_RISCV64_REG_FA3,
+	UFT_RISCV64_REG_FA4,
+	UFT_RISCV64_REG_FA5,
+	UFT_RISCV64_REG_FA6,
+	UFT_RISCV64_REG_FA7,
+};
+
```
namhyung, line 185: I think this part can come with proper argument supports if it's not essential for build.

namhyung (commented):
Can anyone confirm if it works after the change?

gichoel:
The last commit in the current PR, `17213c2`, did not develop the argument handling feature, but I ran it further and here's what it looks like.

Basic tracing
```sh
$ ./uftrace --libmcount-path=. t-arg
            [552810] | main() {
            [552810] |   foo() {
   2.500 us [552810] |     bar();
   0.500 us [552810] |     bar();
   0.750 us [552810] |     bar();
  10.001 us [552810] |   } /* foo */
   0.750 us [552810] |   many();
            [552810] |   pass() {
   1.000 us [552810] |     check();
   3.500 us [552810] |   } /* pass */
  18.251 us [552810] | } /* main */
```

Basic tracing with argument handling
```sh
$ ./uftrace --libmcount-path=. -a t-arg
/home/gichoel/workdir/uftrace/utils/regs.c:437: arch_register_dwarf_name: ASSERT `arch < ARRAY_SIZE(arch_dwarf_tables)' failed.
Stack trace:
  #1  0x003f82ad5ca2 stacktrace + 0x34

WARN: child terminated by signal: 5: Trace/breakpoint trap
WARN: cannot open record data: /tmp/uftrace-live-DrmwKD: No data available
```

------

<br>

**Previously, the inclusion of the `-a` option did not cause any issues, but we can see that it does now.**

This is likely due to the changes in `cmds/info.c` and `utils/arch.h` being imported into the current PR based on feedback from the https://github.com/namhyung/uftrace/pull/1824#discussion_r1318070143 PR.

If you remove the changes to `cmds/info.c` and `utils/arch.h` from the code in the current PR and run it, you can see that it runs normally as shown below.
```
$ ./uftrace --libmcount-path=. -a t-arg
            [553654] | main(0, 0) {
            [553654] |   foo(0) {
   4.000 us [553654] |     bar(0, 0) = 0;
   1.000 us [553654] |     bar(0, 0) = 0;
   1.250 us [553654] |     bar(0, 0) = 0;
  18.001 us [553654] |   } = 0; /* foo */
   1.500 us [553654] |   many(0) = 0;
            [553654] |   pass(0) {
   1.750 us [553654] |     check(big{...}, 0x3fb36f5814) = 0;
   6.750 us [553654] |   } = 0; /* pass */
  37.501 us [553654] | } = 0; /* main */
```

So I would suggest moving the changes in `cmds/info.c` and `utils/arch.h` back to the "argument handling PR", what do you think @namhyung?

utils/arch.h
gichoel: The `__riscv` macro seems to point to all RISC-V architectures (riscv32, riscv64, riscv128).

So I thought it would be a good idea to separate them, considering that `riscv32` should be added in the future.

Would it still be better to configure the if statement with a single condition, as you suggested?

gichoel: According to the link below, older GCCs are using `__riscv__` instead of `__riscv`.

Therefore, we will add a condition that also checks for `__riscv__`.
<br>

Ref. https://github.com/riscv-non-isa/riscv-toolchain-conventions#cc-preprocessor-definitions

namhyung: > Would it still be better to configure the if statement with a single condition, as you suggested?

It should return a sane value for any configuration.  But it seems it doesn't return anything for `riscv32`.

> older GCCs are using __riscv__ instead of __riscv.

How old are those GCCs?  I'm not sure if we really care about them.

namhyung:
For the assertion failure, I think you need to add an empty definition for the registers.  But I don't think it's important as we know it's already broken.

Also, I'm seeing you are working on the arguments at the same time.  If you can add it together no really need for the separate PR then (I assumed it'd take more time).  We can simply merge #1824 at once.

utils/arch.h
gichoel: > But it seems it doesn't return anything for riscv32.

Hmm...  I didn't understand this part yet.

But I'm guessing that in the future, if we say we support both riscv64 and riscv32, the code with the above feedback would look like this, what do you think?
```
#elif defined(__riscv) && __riscv_xlen == 64
        return UFT_CPU_RISCV64;
#elif defined(__riscv) && __riscv_xlen == 32
        return UFT_CPU_RISCV32;
#else
```
<br>

------

Ref. the code in the official GCC Mirror Repository is implemented as follows.

in gcc/config/riscv/riscv.h
```
#define TARGET_64BIT (__riscv_xlen == 64)

......

#define UNITS_PER_WORD (TARGET_64BIT ? 8 : 4)
```

in gcc/config/riscv/riscv-c.cc
```
builtin_define_with_int_value ("__riscv_xlen", UNITS_PER_WORD * 8);
```

gichoel: > How old are those GCCs? I'm not sure if we really care about them.

To conclude, I think we shouldn't bother with `__riscv__` as @namhyung said.

This is because the official GCC Mirror Repository first supported RISC-V in `gcc-7.1.0`, and that version only supports the `__riscv` macro.
  - https://github.com/gcc-mirror/gcc/blob/releases/gcc-7.1.0/gcc/config/riscv/riscv-c.c
<br>

------
However, the reference link I referenced above says that older GCCs use `__riscv__`.

However, there is very little information about `__riscv__` and it was hard to find.

Therefore, my conclusion is that older GCC is riscv-gcc, which seems to have been maintained on its own before RISC-V was integrated into official GCC, and the relevant commit is below. 

riscv-collab/riscv-gcc@b6fc2268f59

gichoel:
I thought it would be a long time before I could separate the argument handling PR based on the feedback, but luckily it was resolved quickly.

(This is a re-listing of the above feedback) So I'm going to do the following things to that PR as per the feedback.

  1. Change the if statement in `utils/arch.h` that checks for RISC-V architecture to a single conditional like `defined(__riscv) && __riscv_xlen == 64`.
  2. (Optional) Add empty definitions for registers in case of assertion failure
  3. exit the PR for argument handling and bring the argument handling commit into the current PR
  4. move changes to `enum uftrace_riscv64_reg_index` in `utils/arch.h` to argument handling commit

utils/arch.h
namhyung: >>   But it seems it doesn't return anything for riscv32.

> Hmm... I didn't understand this part yet.

My understanding is that current code is doing this:

```c

static inline enum uftrace_cpu_arch host_cpu_arch(void)
{
#if defined(__x86_64__)
	return UFT_CPU_X86_64;
#elif defined(__arm__)
	return UFT_CPU_ARM;
#elif defined(__aarch64__)
	return UFT_CPU_AARCH64;
#elif defined(__i386__)
	return UFT_CPU_I386;
#elif defined(__riscv)
# if __riscv_xlen == 64
	return UFT_CPU_RISCV64;
# endif
#else
	return UFT_CPU_NONE;
#endif
}
```

If `__riscv` is defined but `__riscv_xlen` is 32 (or other than 64) there's no return statement for it.  So yeah, you need to check the xlen size properly and handle the else case.

namhyung:
Sounds good to me.

utils/arch.h
gichoel: Ah! I got it now.


If the system is `riscv32`, it satisfies `#elif defined(__riscv)` to get into that `#elif` statement.
But it doesn't return UFT_CPU_NONE because it doesn't satisfy `#if __riscv_xlen == 64`.

I will apply that part as well, thanks.

gichoel:
> For the assertion failure, I think you need to add an empty definition for the registers. But I don't think it's important as we know it's already broken.

It is my understanding that we can ignore assertion failure errors in this part of the code for the following reasons
 1. you understand that argument handling is not yet implemented in that mcount function tracing support commit.
 2. the argument handling commit was applied immediately after that mcount function tracing support commit, and they are merged together because they are inside the same PR. 

------

Other than that, I have implemented all the feedback and have verified that all the commits work with the `make DEBUG=1` command and that each feature works.

honggyukim:
I didn't look into the details this time, but tested the current version and I see it works fine most cases including argument handling.

However, it breaks unittest as follows.
```
$ make unittest
  TEST     test_unit
Running 95 test cases
======================
[001] ELF                       : SIG
[002] ELF                       : SIG
[003] ELF                       : SIG
[004] ELF                       : SIG
[005] ELF                       : SIG
    ...
[093] ELF                       : SIG
[094] ELF                       : SIG
[095] ELF                       : SIG

unit test stats
====================
  0 ran successfully
  0 failed
  0 skipped
 95 signal caught
  0 unknown result

make[1]: *** [Makefile:49: test_unit] Error 1
make: *** [Makefile:415: unittest] Error 2
```
For some reasons, all the unittest names are shown as `ELF`. @gichoel You will see the difference by running the same `make unittest` in x86_64.

gichoel:
I have verified that it works fine on x86_64 except for one skip, and I have verified that it outputs the same error on RISC-V environment.

I don't know how it starts after the execution of the `make unittest` command, so it needs to be analyzed.

Additionally, it outputs **100 test cases on x86_64**, but only **95 test cases on RISC-V**.

Is this something that needs to be checked as well?

namhyung:
The unittest is a separate binary including all the code but with a different main function.  Some test codes are in the `arch/` directory and only included for the architecture (x86_64).

honggyukim:
I have investigated the unittest and found that it doesn't work only in gcc.

Switching the compiler to clang makes it work fine. You can manually modify `CC := gcc` to `CC := clang` in `.config`.
```
    ...
unit test stats
====================
 94 ran successfully
  0 failed
  1 skipped
  0 signal caught
  0 unknown result
```

The reason looks the `uftrace.unit_test` section is not properly filled out.

**1. gcc built `unittest`**
```
$ objdump -D unittest
      ...
Disassembly of section uftrace.unit_test:

00000000000e0cb8 <test_unittest_framework>:
        ...

00000000000e0cc8 <test_option_parsing1>:
        ...

00000000000e0cd8 <test_option_parsing2>:
        ...

00000000000e0ce8 <test_option_parsing3>:
        ...

00000000000e0cf8 <test_option_parsing4>:
        ...
```

**2. clang built `unittest`**
```
$ objdump -D unittest
      ...
Disassembly of section uftrace.unit_test:

00000000000a4358 <test_unittest_framework>:
   a4358:       c88c                    sw      a1,16(s1)
   a435a:       00000007                .4byte  0x7
   a435e:       0000                    unimp
   a4360:       6eb8                    ld      a4,88(a3)
   a4362:       0001                    nop
   a4364:       0000                    unimp
        ...

00000000000a4368 <test_option_parsing1>:
   a4368:       0111                    addi    sp,sp,4
   a436a:       0008                    .2byte  0x8
   a436c:       0000                    unimp
   a436e:       0000                    unimp
   a4370:       8334                    .2byte  0x8334
   a4372:       0001                    nop
   a4374:       0000                    unimp
        ...

00000000000a4378 <test_option_parsing2>:
   a4378:       07e0                    addi    s0,sp,972
   a437a:       0008                    .2byte  0x8
   a437c:       0000                    unimp
   a437e:       0000                    unimp
   a4380:       9406                    add     s0,s0,ra
   a4382:       0001                    nop
   a4384:       0000                    unimp
        ...
```
@gichoel It looks a bug in gcc on riscv so I think you can ignore this issue. I think clang better supports RISC-V for our purpose because of multiple reasons.

clang doesn't have the following problems that gcc have.
1. broken unittest
2. `-fno-omit-frame-pointer` required in optimization builds
3. the first argument capture is broken

Let's focus on clang support for RISC-V as of now.

gichoel:
As you confirmed, switching to clang has resulted in all test cases working fine except for one skip. (The skip occurs in the same test case on x86_64).

All future RISC-V porting work will be done with clang as you suggested.

Thanks.

namhyung:
I'm ok with proceeding for clang, but could you please share more details of the GCC problems?  Specially I'd like to see the actual `objdump -D` output and ELF section headers (like `readelf -SW`).

gichoel:
The results of your requested `objdump -D` and `readelf -SW` are shown below, and gcc's `objdump -D` result was outputting `...` literally, rather than omitting the contents.

Additionally, these results were run on a RISC-V board.

**1. `gcc built` unittest**
```
$ objdump -D unittest

......

Disassembly of section uftrace.unit_test:

000000000009e5a0 <test_unittest_framework>:
	...

000000000009e5b0 <test_option_parsing5>:
	...

000000000009e5c0 <test_option_parsing4>:
	...

000000000009e5d0 <test_option_parsing3>:
	...

000000000009e5e0 <test_option_parsing2>:
	...

000000000009e5f0 <test_option_parsing1>:
	...
```
```
$ readelf -SW unittest

There are 40 section headers, starting at offset 0x3c8210:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .interp           PROGBITS        00000000000002a8 0002a8 000021 00   A  0   0  1
  [ 2] .note.gnu.build-id NOTE            00000000000002cc 0002cc 000024 00   A  0   0  4
  [ 3] .note.ABI-tag     NOTE            00000000000002f0 0002f0 000020 00   A  0   0  4
  [ 4] .gnu.hash         GNU_HASH        0000000000000310 000310 0001d0 00   A  5   0  8
  [ 5] .dynsym           DYNSYM          00000000000004e0 0004e0 0020d0 18   A  6   2  8
  [ 6] .dynstr           STRTAB          00000000000025b0 0025b0 000f13 00   A  0   0  1
  [ 7] .gnu.version      VERSYM          00000000000034c4 0034c4 0002bc 02   A  5   0  2
  [ 8] .gnu.version_r    VERNEED         0000000000003780 003780 000190 00   A  6   9  8
  [ 9] .rela.dyn         RELA            0000000000003910 003910 00b538 18   A  5   0  8
  [10] .rela.plt         RELA            000000000000ee48 00ee48 001950 18  AI  5  24  8
  [11] .plt              PROGBITS        00000000000107a0 0107a0 001100 10  AX  0   0 16
  [12] .text             PROGBITS        00000000000118a0 0118a0 05e69a 00  AX  0   0  4
  [13] .rodata           PROGBITS        000000000006ff40 06ff40 0270fa 00   A  0   0  8
  [14] .eh_frame_hdr     PROGBITS        000000000009703c 09703c 000014 00   A  0   0  4
  [15] .eh_frame         PROGBITS        0000000000097050 097050 00002c 00   A  0   0  8
  [16] .tbss             NOBITS          0000000000098360 097360 0010f0 00 WAT  0   0  8
  [17] .preinit_array    PREINIT_ARRAY   0000000000098360 097360 000008 08  WA  0   0  1
  [18] .init_array       INIT_ARRAY      0000000000098368 097368 000008 08  WA  0   0  8
  [19] .fini_array       FINI_ARRAY      0000000000098370 097370 000008 08  WA  0   0  8
  [20] .data.rel.ro      PROGBITS        0000000000098378 097378 002a08 00  WA  0   0  8
  [21] .dynamic          DYNAMIC         000000000009ad80 099d80 000280 10  WA  6   0  8
  [22] .data             PROGBITS        000000000009b000 09a000 00359c 00  WA  0   0  8
  [23] uftrace.unit_test PROGBITS        000000000009e5a0 09d5a0 0005f0 00  WA  0   0  8
  [24] .got              PROGBITS        000000000009eb90 09db90 000a40 08  WA  0   0  8
  [25] .bss              NOBITS          000000000009f5d0 09e5d0 003368 00  WA  0   0  8
  [26] .comment          PROGBITS        0000000000000000 09e5d0 00001e 01  MS  0   0  1
  [27] .riscv.attributes RISCV_ATTRIBUTES 0000000000000000 09e5ee 000037 00      0   0  1
  [28] .debug_aranges    PROGBITS        0000000000000000 09e625 000aa0 00      0   0  1
  [29] .debug_info       PROGBITS        0000000000000000 09f0c5 11fb40 00      0   0  1
  [30] .debug_abbrev     PROGBITS        0000000000000000 1bec05 013545 00      0   0  1
  [31] .debug_line       PROGBITS        0000000000000000 1d214a 0bab1a 00      0   0  1
  [32] .debug_frame      PROGBITS        0000000000000000 28cc68 011128 00      0   0  8
  [33] .debug_str        PROGBITS        0000000000000000 29dd90 015fb8 01  MS  0   0  1
  [34] .debug_line_str   PROGBITS        0000000000000000 2b3d48 00116a 01  MS  0   0  1
  [35] .debug_loclists   PROGBITS        0000000000000000 2b4eb2 0d0ce1 00      0   0  1
  [36] .debug_rnglists   PROGBITS        0000000000000000 385b93 029e51 00      0   0  1
  [37] .symtab           SYMTAB          0000000000000000 3af9e8 00f600 18     38 1534  8
  [38] .strtab           STRTAB          0000000000000000 3befe8 009072 00      0   0  1
  [39] .shstrtab         STRTAB          0000000000000000 3c805a 0001b1 00      0   0  1
```
<br>

**2. `clang built` unittest**

```
$ objdump -D unittest

......

Disassembly of section uftrace.unit_test:

00000000000a3760 <test_unittest_framework>:
   a3760:	c49c                	sw	a5,8(s1)
   a3762:	00000007          	.4byte	0x7
   a3766:	0000                	unimp
   a3768:	6eb8                	ld	a4,88(a3)
   a376a:	0001                	nop
   a376c:	0000                	unimp
	...

00000000000a3770 <test_option_parsing1>:
   a3770:	fcd5                	bnez	s1,a372c <type_size.table+0xe5c>
   a3772:	00000007          	.4byte	0x7
   a3776:	0000                	unimp
   a3778:	82f8                	.2byte	0x82f8
   a377a:	0001                	nop
   a377c:	0000                	unimp
	...

00000000000a3780 <test_option_parsing2>:
   a3780:	03a4                	addi	s1,sp,456
   a3782:	0008                	.2byte	0x8
   a3784:	0000                	unimp
   a3786:	0000                	unimp
   a3788:	93ca                	add	t2,t2,s2
   a378a:	0001                	nop
   a378c:	0000                	unimp
	...

00000000000a3790 <test_option_parsing3>:
   a3790:	051f 0008 0000      	.byte	0x1f, 0x05, 0x08, 0x00, 0x00, 0x00
   a3796:	0000                	unimp
   a3798:	97fc                	.2byte	0x97fc
   a379a:	0001                	nop
   a379c:	0000                	unimp
	...

00000000000a37a0 <test_option_parsing4>:
   a37a0:	0698                	addi	a4,sp,832
   a37a2:	0008                	.2byte	0x8
   a37a4:	0000                	unimp
   a37a6:	0000                	unimp
   a37a8:	9d22                	add	s10,s10,s0
   a37aa:	0001                	nop
   a37ac:	0000                	unimp
	...

00000000000a37b0 <test_option_parsing5>:
   a37b0:	0008078f          	.4byte	0x8078f
   a37b4:	0000                	unimp
   a37b6:	0000                	unimp
   a37b8:	a1fc                	fsd	fa5,192(a1)
   a37ba:	0001                	nop
   a37bc:	0000                	unimp
	...
```
```
$ readelf -SW unittest

There are 40 section headers, starting at offset 0x5cbd18:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .interp           PROGBITS        00000000000102a8 0002a8 000021 00   A  0   0  1
  [ 2] .note.gnu.build-id NOTE            00000000000102cc 0002cc 000024 00   A  0   0  4
  [ 3] .note.ABI-tag     NOTE            00000000000102f0 0002f0 000020 00   A  0   0  4
  [ 4] .hash             HASH            0000000000010310 000310 0008e0 04   A  6   0  8
  [ 5] .gnu.hash         GNU_HASH        0000000000010bf0 000bf0 0009dc 00   A  6   0  8
  [ 6] .dynsym           DYNSYM          00000000000115d0 0015d0 001c68 18   A  7   1  8
  [ 7] .dynstr           STRTAB          0000000000013238 003238 000c8c 00   A  0   0  1
  [ 8] .gnu.version      VERSYM          0000000000013ec4 003ec4 00025e 02   A  6   0  2
  [ 9] .gnu.version_r    VERNEED         0000000000014128 004128 000190 00   A  7   9  8
  [10] .rela.dyn         RELA            00000000000142b8 0042b8 0000d8 18   A  6   0  8
  [11] .rela.plt         RELA            0000000000014390 004390 001968 18  AI  6  24  8
  [12] .plt              PROGBITS        0000000000015d00 005d00 001110 10  AX  0   0 16
  [13] .text             PROGBITS        0000000000016e10 006e10 065688 00  AX  0   0  4
  [14] .rodata           PROGBITS        000000000007c498 06c498 0272c5 00   A  0   0  8
  [15] uftrace.unit_test PROGBITS        00000000000a3760 093760 0005f0 00   A  0   0  8
  [16] .eh_frame_hdr     PROGBITS        00000000000a3d50 093d50 000014 00   A  0   0  4
  [17] .eh_frame         PROGBITS        00000000000a3d68 093d68 00002c 00   A  0   0  8
  [18] .tbss             NOBITS          00000000000a4d78 094d78 0010f0 00 WAT  0   0  8
  [19] .preinit_array    PREINIT_ARRAY   00000000000a4d78 094d78 000008 08  WA  0   0  1
  [20] .init_array       INIT_ARRAY      00000000000a4d80 094d80 000008 08  WA  0   0  8
  [21] .fini_array       FINI_ARRAY      00000000000a4d88 094d88 000008 08  WA  0   0  8
  [22] .dynamic          DYNAMIC         00000000000a4d90 094d90 000270 10  WA  7   0  8
  [23] .data             PROGBITS        00000000000a5000 095000 0029c8 00  WA  0   0  8
  [24] .got              PROGBITS        00000000000a79c8 0979c8 000898 08  WA  0   0  8
  [25] .sdata            PROGBITS        00000000000a8260 098260 000158 00  WA  0   0  8
  [26] .sbss             NOBITS          00000000000a83b8 0983b8 000459 00  WA  0   0  8
  [27] .bss              NOBITS          00000000000a8818 0983b8 002f48 00  WA  0   0  8
  [28] .comment          PROGBITS        0000000000000000 0983b8 00003c 01  MS  0   0  1
  [29] .riscv.attributes RISCV_ATTRIBUTES 0000000000000000 0983f4 000039 00      0   0  1
  [30] .debug_info       PROGBITS        0000000000000000 09842d 0b0987 00      0   0  1
  [31] .debug_abbrev     PROGBITS        0000000000000000 148db4 00bc54 00      0   0  1
  [32] .debug_line       PROGBITS        0000000000000000 154a08 0754fb 00      0   0  1
  [33] .debug_frame      PROGBITS        0000000000000000 1c9f08 00b7b8 00      0   0  8
  [34] .debug_str        PROGBITS        0000000000000000 1d56c0 014dc3 01  MS  0   0  1
  [35] .debug_loc        PROGBITS        0000000000000000 1ea483 0a74be 00      0   0  1
  [36] .debug_ranges     PROGBITS        0000000000000000 291941 01cbe0 00      0   0  1
  [37] .symtab           SYMTAB          0000000000000000 2ae528 3147d8 18     38 133482  8
  [38] .strtab           STRTAB          0000000000000000 5c2d00 008e88 00      0   0  1
  [39] .shstrtab         STRTAB          0000000000000000 5cbb88 00018b 00      0   0  1
```

gichoel:
umm... In the last discussion, we discussed the issue with running the `make unittest` command, and since @honggyukim's analysis concluded that it was just a `gcc` issue in the RISC-V environment, we decided to move on to `clang`.

I don't see anything further mentioned in that PR, so I'm curious.
My recollection is that there were no unresolved issues prior to the `make unittest` issue.
Are there any further issues being investigated, or should we wait for others to analyze this further? 

PS : I've been working on researching and analyzing material for a porting of the `PLT hooking` feature for a few weeks now, and I'm still working on it.

So if more testing is needed or there are additional issues, there's no need to rush this PR.

namhyung:
Sorry I didn't follow the development lately and thought there are more remaining issues.  The test failures on GCC are unfortunate but as long as it works on clang, there's should be no fundamental problems.  I'm more concerning about the PLT problems and looking forward to seeing a solution soon.

If @honggyukim is ok, I can merge this.

honggyukim:
Sorry @gichoel, we missed further discussion after going into the unittest topic. I'm okay for almost most of the implementation.

But I just found that the floating point argument handling is broken. You can reproduce the problem as follows.
```diff
$ ./runtest.py -vdp -O0 -c clang 083
      ...
t083_arg_float: diff result of clang -pg -O0
--- expect      2023-10-05 13:19:42.213047344 +0000
+++ result      2023-10-05 13:19:42.213047344 +0000
@@ -1,6 +1,6 @@
 main() {
-   float_add(-0.100000, 0.200000) = 0.100000;
-   float_sub(0.100000, 0.200000) = -0.100000;
-   float_mul(300.000000, 400.000000) = 120000.000000;
-   float_div(40000000000.000000, -0.020000) = -2000000000000.000000;
+   float_add(-0.100000, 0.200000) = nan;
+   float_sub(0.100000, 0.200000) = nan;
+   float_mul(300.000000, 400.000000) = 0.000000;
+   float_div(40000000000.000000, -0.020000) = 0.000000;
  } /* main */

083 arg_float           : NG
```
Except for this issue, I'm fine with this work for merging into master.

gichoel:
Hi, @namhyung and @honggyukim.

I really appreciate you both taking the time to review and give feedback each time, I'm learning a lot from you :)

`make unittest` with `clang` there was no problem, so I think the new issue is an exceptional case that only happens in the Python test code.

Hmm... I wonder if it might be related to the `x8, x18` registers issue identified in the `aarch64` architecture, but I'm not sure, so I'll have to keep my mind open and analyze various possibilities.

gichoel:
I am investigating a previously presented issue that occurs when running the command `$ ./runtest.py -vdp -O0 -c clang 083`.

let's summarize conclusions, which are twofold
  1. `$ ./runtest.py -vdp -O0 -c clang 083` was also unable to get a return value in the `RISC-V QEMU` environment.
  2. `make unittest` failed with the same `ELF, SIG` error with `clang` on `RISC-V QEMU`.
    - On the `RISC-V QEMU` environment, even with `clang`, the results of running `objdump` and `readelf` are the same as running them with `gcc` on the `visionfive2`.
 <br/>


the reason I did the second task was to see if other RISC-V environments had the same issue with the `make unittest` command.

Previously, it didn't happen when using `clang` in `make unittest`, so we just let it go.

However, it turns out that this is not the case in the `RISC-V QEMU` environment.
Should we look into `make unittest` issue first and analyze the cause?

------
Each environment used for testing is shown below.

**[visionfive2 Board - Debian 12 ]**
```
$ clang --version
Debian clang version 13.0.1-6
Target: riscv64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

uname -a
Linux starfive 5.15.0-starfive #1 SMP Mon Dec 19 07:56:37 EST 2022 riscv64 GNU/Linux

$ cat /etc/issue
Debian GNU/Linux bookworm/sid
```

**[RISC-V QEMU - Ubuntu 23.04 ]**
```
$ clang --version
Ubuntu clang version 15.0.7
Target: riscv64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

uname -a
Linux ubuntu 6.2.0-19-generic #19.1-Ubuntu SMP Fri Mar 31 12:41:53 UTC 2023 riscv64 riscv64 riscv64 GNU/Linux

$ cat /etc/issue
Ubuntu 23.04
```

**[RISC-V QEMU - Ubuntu 22.04 ]**
```
$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: riscv64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

uname -a
Linux ubuntu 5.19.0-1021-generic #23~22.04.1-Ubuntu SMP Thu Jun 22 12:49:35 UTC 2023 riscv64 riscv64 riscv64 GNU/Linux

$ cat /etc/issue
Ubuntu 22.04.3 LTS
```

namhyung:
Can you please build tests/s-exp-float.c manually and check the output?

```
$ gcc -pg -g -o t-exp-float  tests/s-exp-float.c
$ uftrace -a ./t-exp-float
```

gichoel:
I manually compiled `tests/s-exp-float.c` as advised and ran it twice as shown below.

```
$ gcc -pg -g -o t-exp-float tests/s-exp-float.c

$ ./uftrace --libmcount-path=. -a ./t-exp-float
# duration tid function
            [ 881] | main(0x3fa04631d4, 0x3fc2ac3298) {
   5.500 us [ 881] | float_add(-0.100000, 0.200000) = -86.095337;
   2.500 us [ 881] | float_sub(0.100000, 0.200000) = -86.095337;
   1.700 us [ 881] | float_mul(300.000000, 400.000000) = 0.000000;
   1.500 us [ 881] | float_div(40000000000.000000, -0.020000) = 0.000000;
  34.401 us [ 881] | } = 0; /* main */

$ ./uftrace --libmcount-path=. -a ./t-exp-float
# duration tid function
            [ 934] | main(0x3f88fca1d4, 0x3fda4cb298) {
   5.700 us [ 934] | float_add(-0.100000, 0.200000) = -14403825662164992.000000;
   2.500 us [ 934] | float_sub(0.100000, 0.200000) = -14403825662164992.000000;
   1.400 us [ 934] | float_mul(300.000000, 400.000000) = 0.000000;
   1.400 us [ 934] | float_div(40000000000.000000, -0.020000) = 0.000000;
  33.500 us [ 934] | } = 0; /* main */
```

namhyung:
Yeah, the return value seems not correct.  But I don't think it should be a blocker at this point.

honggyukim (approved):
I just wanted to fix it but I also think that it doesn't have block it from being merged into master. I will give LGTM now. Thanks very much for your effort and great contribution!

namhyung:
Good, can you please rebase your changes onto the current master?

gichoel:
It's great to see the early work on RISC-V support finally coalescing into master, but I'm a little confused.

That's because there are still issues with `make unittest` and `return value`.

The `make unittest` issue seems to be a compiler issue, which I understand.
But I wonder why they are merged when the `return value issue` exists.

 @honggyukim Can we talk about this issue soon?

 @namhyung Yes, I will deal with it right away.

honggyukim:
> But I wonder why they are merged when the return value issue exists.

It's because we don't want to hold this PR too long and so we can merge this early then handle floating point return value issue as a bug fix later.

We're also aware that this work doesn't include plthook support so we know that this is incomplete somehow, but supporting all the features for a new architecture at once is difficult. We rather think that it'd be better to support RISC-V port incrementally.

So I think it's good to consider this PR has completed the initial milestone for the RISC-V support.

honggyukim (approved):
The rebase is done so I can give LGTM again.

gichoel:
I now understand why it was merged, thank you.

namhyung:
I've pushed [check/riscv-fpret](../compare/check/riscv-fpret).  Can you please take a look?

gichoel:
I ran the `check/riscv-fpret` branch code on my RISC-V board, and got the following results.

```
$ clang -pg -g -o t-exp-float tests/s-exp-float.c
$ ./uftrace --libmcount-path=. -a ./t-exp-float
            [  3784] | main(1, 0x3fe497a218) {
   1.250 us [  3784] |   float_add(-0.100000, 0.200000) = 0.100000;
   0.750 us [  3784] |   float_sub(0.100000, 0.200000) = -0.100000;
   0.750 us [  3784] |   float_mul(300.000000, 400.000000) = 120000.000000;
   0.750 us [  3784] |   float_div(40000000000.000000, -0.020000) = -2000000000000.000000;
  14.750 us [  3784] | } = 0; /* main */
```

Wow... this surprisingly solved the `floating point return value issue`.

------
I was curious as to how it was done, so I analyzed the commit and came up with the following conclusions.
  - Since this is a `return value` issue, I had to modify the `mcount_return` function.
  - (in mcount.S) Save `fa0`, the first return register of the floating point, to the stack before calling the `mcount_exit` function.
  - (in mcount-support.c) Save the value of the `fa0` register directly to `ctx->val.v` in `mcount_arch_get_retval`, the function that gets the return value of the function.

However, I couldn't figure out why the problem occurred with the existing method of using the `float_retval` variable.

namhyung:
You used `ctx->retval - 2`  to record the return value but forgot to save it in `mcount_return()`.  In the above code, it'd be +2 instead -2 though.  Also we can just use `fa0` directly (should be same as `ctx->retval + 2`).

FYI, the return value is passed to `mcount_exit()` as the first argument in `a0`.  The `sp` should point to a stack location that contains the following data.

```
        ...
        |--------------------------|
     40 | ra (return address)      |
     32 | fp (frame pointer)       |
        |     (rest of fa0       ) |
     16 | fa0 (return value in FP) |
      8 | a1 (return value2)       |
sp    0 | a0 (return value1)       |   <---  ctx->retval
        |--------------------------|
```

gichoel:
Thanks for the explanation. 

Your explanation above helped me understand why it is possible to avoid using the `float_retval` variable and what I was missing.

Issue #1777, opened 2023-07-26
`-fno-plt` disturbs tracing the libc call and default Rust build flag use it.


ChoKyuWon:
I want to track the allocation(`malloc`) in Rust, so I use the simple code as follows:
```rust
fn main() {
    let buf : [u8;64] = [0;64];
    let x = Box::<[u8;64]>::new(buf);
    println!("{:?}", *x);
}
```
And trace it with the following command:
```
$ RUSTFLAGS="-Zinstrument-mcount -Cpasses=ee-instrument<post-inline>" cargo build -Zbuild-std --target x86_64-unknown-linux-gnu
$ uftrace -la ./target/x86_64-unknown-linux-gnu/debug/test-rs
...
          [102876] |                                 __rdl_alloc() {
            [102876] |                                   core::alloc::layout::Layout::from_size_align_unchecked() {
   0.110 us [102876] |                                     core::ptr::alignment::Alignment::new_unchecked();
   0.461 us [102876] |                                   } /* core::alloc::layout::Layout::from_size_align_unchecked */
            [102876] |                                   std::sys::unix::alloc::_<impl core::alloc::global::GlobalAlloc for std::alloc::System>::alloc() {
            [102876] |                                     core::alloc::layout::Layout::align() {
   0.080 us [102876] |                                       core::ptr::alignment::Alignment::as_usize();
   0.361 us [102876] |                                     } /* core::alloc::layout::Layout::align */
            [102876] |                                     core::alloc::layout::Layout::align() {
   0.081 us [102876] |                                       core::ptr::alignment::Alignment::as_usize();
   0.331 us [102876] |                                     } /* core::alloc::layout::Layout::align */
   0.080 us [102876] |                                     core::alloc::layout::Layout::size();
   0.080 us [102876] |                                     core::alloc::layout::Layout::size();
   1.603 us [102876] |                                   } /* std::sys::unix::alloc::_<impl core::alloc::global::GlobalAlloc for std::alloc::System>::alloc */
   2.524 us [102876] |                                 } /* __rdl_alloc */
...
```
I use the `-Zbuild-std` option because the allocation layout(e.g., `exchange_malloc`, `__rdl_alloc`) is also not traced with this option.
As far as I know, `__rdl_alloc` calls the `malloc` eventually, but the trace misses it in my environment.

Additionally, I use the following code to test the direct libc call:
```rust
use libc::getpid;

fn main() {
    let pid = unsafe{getpid()};
[...]

PR #1816, opened 2023-08-29
mcount:Temporary fix find_got for no-plt binary

ChoKyuWon:
Issue #1777 reported that no-plt binary is handled as plt-enable binary because some functions are still left in plt section. The patch fixes this bug by comparing the name of self with a list of known functions.

However, current `mcount_arch_hook_no_plt` make segfault on x86 now, so I comment it out and print the debug message if the binary was built with `-fno-plt` option.

### DIFF
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -189,9 +189,14 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
 		    const char *modname, unsigned long offset)
 {
 	bool plt_found = false;
+	bool skip_found = false;
 	unsigned long pltgot_addr = 0;
 	unsigned long plt_addr = 0;
 	struct plthook_data *pd;
+	static const char *const skip_got_syms[] = {
+		"__monstartup",
+	};
+	size_t k;
 
 	elf_for_each_shdr(elf, iter) {
 		if (iter->shdr.sh_type == SHT_DYNAMIC)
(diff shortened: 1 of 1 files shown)


namhyung:
Is it only for debugging?  I remember some binary has both no-PLT libraries calls and PLT for a few of calls.  I'm not sure if checking `__monstartup` is enough.  But I agree that we should ignore some special functions.

ChoKyuWon:
Yes, it was only for debugging. I didn't figure out how to make it work now.

However, after I skip the functions inside the `mcount_arch_hook_no_plt` as follows:
```diff
--- a/arch/x86_64/mcount-noplt.c
+++ b/arch/x86_64/mcount-noplt.c
@@ -95,6 +95,12 @@ struct plthook_data *mcount_arch_hook_no_plt(struct uftrace_elf_data *elf, const
 				break;
 			}
 		}
+		for (k = 0; k < skip_got_syms_nr; k++){
+			if (!strcmp(sym->name, skip_got_syms[k])) {
+				skip = true;
+				break;
+			}
+		}
 		if (skip)
 			continue;

--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -104,6 +104,13 @@ const struct plthook_skip_symbol plt_skip_syms[] = {
 };
 size_t plt_skip_nr = ARRAY_SIZE(plt_skip_syms);
 
+const char *const skip_got_syms[] = {
+	"__monstartup",
+	"__cxa_atexit",
+	"__cxa_finalize",
+};
+size_t skip_got_syms_nr = ARRAY_SIZE(skip_got_syms);

@@ -209,10 +213,9 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
 			pltgot_addr = (unsigned long)iter->dyn.d_un.d_val + offset;
 			break;
 		case DT_JMPREL:
-			for (k = 0; k < ARRAY_SIZE(skip_got_syms); k++) {
-				if (!strncmp(skip_got_syms[k],
-					     elf_get_name(elf, iter, iter->sym.st_name),
-					     sizeof(skip_got_syms[k]))) {
+			for (k = 0; k < skip_got_syms_nr; k++) {
+				if (!strcmp(skip_got_syms[k],
+					     elf_get_name(elf, iter, iter->sym.st_name))) {
 					skip_found = true;
 					break;
 				}
```
Now for `fno-plt` binary, it doesn't make segfault now, but it is still not able to trace the library functions.
Furthermore, now the patched version of uftrace makes segfault for normal binary(a.k.a `-fplt`).
I revert this PR and dig thorough the root cause.

PR #1817, opened 2023-08-30
utils: Remove unused function: get_event_name

SEOKMIN83:
The function `get_event_name` in utils/utils.h
is not used within uftrace.
This commit deletes the unused `get_event_name` function.


namhyung (approved):
Right, `get_event_name()` was renamed to `event_get_name()` and moved to utils/event.[ch].

PR #1820, opened 2023-09-01
libmcount: support no-plt binary

ChoKyuWon:
Issue #1777 reported that no-plt binary is handled as plt-enable binary because some functions are still left in plt section. The patch fixes this bug by comparing the name of self with a list of known functions.

### DIFF
--- a/arch/x86_64/symbol.c
+++ b/arch/x86_64/symbol.c
@@ -29,12 +29,11 @@ int arch_load_dynsymtab_noplt(struct uftrace_symtab *dsymtab, struct uftrace_elf
 
 	memset(dsymtab, 0, sizeof(*dsymtab));
 
-	/* assumes there's only one RELA section (rela.dyn) for no-plt binary */
 	elf_for_each_shdr(elf, &sec_iter) {
-		if (sec_iter.shdr.sh_type == SHT_RELA) {
+		if (strcmp(elf_get_name(elf, &sec_iter, sec_iter.shdr.sh_name), ".rela.dyn") == 0) {
 			memcpy(&rel_iter, &sec_iter, sizeof(sec_iter));
-			pr_dbg2("found RELA section: %s\n",
-				elf_get_name(elf, &sec_iter, sec_iter.shdr.sh_name));
+			pr_dbg2("found rela.dyn section with %ld entry.\n",
+				sec_iter.shdr.sh_entsize);
 
 			reloc_start = rel_iter.shdr.sh_addr + offset;
 			reloc_entsize = rel_iter.shdr.sh_entsize;
@@ -93,6 +92,7 @@ int arch_load_dynsymtab_noplt(struct uftrace_symtab *dsymtab, struct uftrace_elf
 		pr_dbg3("[%zd] %c %lx + %-5u %s\n", dsymtab->nr_sym, sym->type, sym->addr,
 			sym->size, sym->name);
 	}
+	sort_dynsymtab(dsymtab);
 
 	return dsymtab->nr_sym;
 }
--- a/libmcount/plthook.c
+++ b/libmcount/plthook.c
@@ -104,6 +104,17 @@ const struct plthook_skip_symbol plt_skip_syms[] = {
 };
 size_t plt_skip_nr = ARRAY_SIZE(plt_skip_syms);
 
+const char *const noplt_skip_syms[] = {
+	/* For GCC/CLANG */
+	"__stack_chk_fail",
+	"__monstartup",
+	"__cxa_atexit",
+	/* For Rust */
+	"__tls_get_addr",
+	"_Unwind_Resume",
+};
+size_t noplt_skip_nr = ARRAY_SIZE(noplt_skip_syms);
+
 #undef SKIP_SYM
 #undef ALIAS_DECL
 
@@ -191,7 +202,12 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
 	bool plt_found = false;
 	unsigned long pltgot_addr = 0;
 	unsigned long plt_addr = 0;
+	unsigned long jmprel_addr = 0;
+	struct uftrace_elf_iter sec_iter;
+	size_t jmprel_nr = 0;
 	struct plthook_data *pd;
+	bool is_rela = true;
+	const char *fname;
 
 	elf_for_each_shdr(elf, iter) {
 		if (iter->shdr.sh_type == SHT_DYNAMIC)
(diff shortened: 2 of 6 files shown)


ChoKyuWon:
Hmm, the CI test failed on `[67] mcount_setup` due to the memory leak, but in my local environment, it succeeded without any error. 
I'll fix it and push an additional patch for it.

utils/symbol.c
```diff
@@ -476,6 +476,8 @@ void sort_dynsymtab(struct uftrace_symtab *dsymtab)
 {
 	unsigned i, k;
 
+	if (dsymtab->nr_sym == 0)
```
ChoKyuWon, line 479: Due to the ASAN error, malloc(0) actually allocates 1 byte memory and is detected by LeakSan.
I'm not sure this code is really needed.

namhyung:
I need to refresh my memory on this.  Please expect some delays..

arch/x86_64/symbol.c
```diff
@@ -31,9 +31,9 @@ int arch_load_dynsymtab_noplt(struct uftrace_symtab *dsymtab, struct uftrace_elf
 
 	/* assumes there's only one RELA section (rela.dyn) for no-plt binary */
 	elf_for_each_shdr(elf, &sec_iter) {
-		if (sec_iter.shdr.sh_type == SHT_RELA) {
+		if (strcmp(elf_get_name(elf, &sec_iter, sec_iter.shdr.sh_name), ".rela.dyn") == 0) {
 			memcpy(&rel_iter, &sec_iter, sizeof(sec_iter));
-			pr_dbg2("found RELA section: %s\n",
+			pr_dbg2("found rela.plt section: %s\n",
```
ChoKyuWon, line 36: I think this is typo. Will be fixed in next commit.

libmcount/plthook.c
```diff
@@ -104,6 +104,17 @@ const struct plthook_skip_symbol plt_skip_syms[] = {
 };
 size_t plt_skip_nr = ARRAY_SIZE(plt_skip_syms);
 
+const char *const additional_plt_skip_syms[] = {
```
namhyung, line 107: I prefre short names, how about 'noplt_skip_syms'?

arch/x86_64/symbol.c
```diff
@@ -31,9 +31,9 @@ int arch_load_dynsymtab_noplt(struct uftrace_symtab *dsymtab, struct uftrace_elf
 
 	/* assumes there's only one RELA section (rela.dyn) for no-plt binary */
```
namhyung, line 32: This comment should be updated too.

libmcount/plthook.c
```diff
@@ -246,7 +296,10 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
 
 	memset(&pd->dsymtab, 0, sizeof(pd->dsymtab));
 	/* do not demangle symbol names since it might call dlsym() */
-	load_elf_dynsymtab(&pd->dsymtab, elf, pd->base_addr, 0);
+	if (!plt_found)
```
namhyung, line 299: This condition cannot be true here, right?

libmcount/plthook.c
```diff
@@ -204,13 +220,47 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
 			pltgot_addr = (unsigned long)iter->dyn.d_un.d_val + offset;
 			break;
 		case DT_JMPREL:
-			plt_found = true;
```
namhyung, line 207: Please add a comment here why we cannot simply set `plt_found = true` here.

arch/x86_64/symbol.c
```diff
@@ -93,6 +93,7 @@ int arch_load_dynsymtab_noplt(struct uftrace_symtab *dsymtab, struct uftrace_elf
 		pr_dbg3("[%zd] %c %lx + %-5u %s\n", dsymtab->nr_sym, sym->type, sym->addr,
 			sym->size, sym->name);
 	}
+	sort_dynsymtab(dsymtab);
```
namhyung, line 96: I think this change can be applied separately.

namhyung (commented):
From a quick glance, it looks good.  Let me test it more..

libmcount/plthook.c
ChoKyuWon: I don't think so. Even modern compilers and x86_64 always retain some PLT entry, we can't exclude the case that the compiler does not retain the PLT entry and totally relies on the GOT table directly. So this `if` expression should be there IMO.

ChoKyuWon: > This condition cannot be true here, right?

I don't think so. Even modern compilers and x86_64 always retain some PLT entry, we can't exclude the case that the compiler does not retain the PLT entry and totally relies on the GOT table directly. So this if expression should be there IMO.

namhyung: Hmm.. but I think you checked it on line 265 and returned.

ChoKyuWon: Oh now I understand what you meant and it makes sound. I'll revert this diff.

arch/x86_64/symbol.c
namhyung: Oh, actually I meant the `sort_dynsymtab()` only.  I think you can split the changes in the following way.

1. add `sort_dynsymtab()` at the end of `arch_load_dynsymtab_noplt()`.  It's an independent issue and can go separately.
2. add `arch_load_dynsymtab_noplt()` to `load_dynsymtab()`.  This should have no effect to normal PLT binaries.
3. check PLT with additional skip symbols and read from `.rela.dyn`.

libmcount/plthook.c
```diff
@@ -204,13 +220,48 @@ static int find_got(struct uftrace_elf_data *elf, struct uftrace_elf_iter *iter,
...
+		if (sec_iter.shdr.sh_type == SHT_DYNSYM) {
+			elf_get_strtab(elf, &sec_iter, sec_iter.shdr.sh_link);
+			elf_get_secdata(elf, &sec_iter);
+			break;
+		}
+	}
+
+	for (size_t i = 0; i < jmprel_nr; i += (is_rela ? sizeof(Elf64_Rela) : sizeof(Elf64_Rel))) {
+		bool found = false;
+		Elf64_Rela *rel = (void *)jmprel_addr + i;
```
namhyung, line 248: I prefer having a blank line after the declaration.

namhyung (commented):
Looks good, can you update the TBU part in the first commit message?

namhyung (commented):
Would you please rebase your changes onto the current master?  It's already encouraged to do so (at least for uftrace) when you update your branch, thanks!

namhyung:
Sorry about the long delay.

Issue #1822, opened 2023-09-04
Memory leak with wrong command arguments


ChoKyuWon:
When I built the uftrace with ASAN and ran it with the wrong command arguments, the LEAKSAN reported there was a memory leak.
```sh
$ ./uftrace -F abcdefg           
 uftrace -- function (graph) tracer for userspace

 usage: uftrace [COMMAND] [OPTION...] [<program>]
...
=================================================================
==570670==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 7 byte(s) in 1 object(s) allocated from:
    #0 0x7f77db9bd887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x55bee96331d8 in strjoin /home/kyuwoncho18/sandbox/uftrace/utils/utils.c:588

SUMMARY: AddressSanitizer: 7 byte(s) leaked in 1 allocation(s).
```

It seems when the `parse_option()` returns a negative value, the switch loop returns directly whether there is heap allocation or not, so the leak happens.
This kind of fix may resolve the problem:
```diff
--- a/uftrace.c
+++ b/uftrace.c
@@ -1424,6 +1424,7 @@ int main(int argc, char *argv[])
        case -2:
                pr_out(uftrace_usage);
                pr_out(uftrace_footer);
+               free(...);
                return 0;
        case -3:
```

But should this leak be fixed? Because the leak happens on the main function, the OS automatically collects the garbage; normally, nothing bad happens in this case.

namhyung:
I believe it's better to reduce the noise in the sanitizer output in general.  Please send a PR for the fix.

PR #1823, opened 2023-09-05
uftrace: fix a memory leak when option parsing failed

ChoKyuWon:
When option parsing failed or binary name is not given, main function returns directly without freeing memory.
This commit fix it with goto cleanup when the main function exit.

Fixes: #1822


namhyung (approved):
LGTM

paranlee:
LGTM

Issue #1827, opened 2023-09-25
replay: some symbols are printed with incorrect parentheses, due to logic errors in setting NEEDS_PAREN


jyf111:
## To Reproduce
```cpp
$ cat func.cpp
class F {
public:
  int operator()() {
    return 1;
  }
  int operator+(int) {
    return 2;
  }
};

int main() {
  F()();
  F() + 1;
  auto f = [](int) {
    return 3;
  }(1);
  return 0;
}
$ g++ func.cpp -pg -o func
$ uftrace func
# DURATION     TID     FUNCTION
   0.771 us [  4742] | __monstartup();
   0.290 us [  4742] | __cxa_atexit();
            [  4742] | main() {
   0.120 us [  4742] |   F::operator();
   0.091 us [  4742] |   F::operator+();
   0.080 us [  4742] |   main::$_0::operator();
   1.032 us [  4742] | } /* main */
$ uftrace --demangle=full func
# DURATION     TID     FUNCTION
   0.702 us [  4788] | __monstartup();
   0.191 us [  4788] | __cxa_atexit();
            [  4788] | main() {
   0.080 us [  4788] |   F::operator()();
   0.051 us [  4788] |   F::operator+(int);
   0.050 us [  4788] |   main::{lambda(int)#1}::operator()(int) const();
   0.781 us [  4788] | } /* main */
```

When using simple demangle, `F::operator()` and `main::$_0::operator()` should be `F::operator()()` and `main::$_0::operator()()` respectively. 

When using full demangle, `main::{lambda(int)#1}::operator()(int) const()` should be `main::{lambda(int)#1}::operator()(int) const`.

## To Fix
https://github.com/namhyung/uftrace/blob/58e73072be45f64c0973cb1c9b9a083f7a0977b4/cmds/replay.c#L741-L744

Currently, `NEEDS_PAREN` is set when the last character of `symname` is not ')', since full demangle may add parentheses at the end.

But this does not work with `operator()` (and lambda) under simple demangle, and when using full demangle, the last few characters may be function cv-qualifer (this is a bit annoying).

honggyukim:
Hi @jyf111, it's a bit late response but thanks for the analysis. The logic was introduced at the following commits.
- 48321555 demangle: Use __cxa_demangle for full demangling
- bca429ff replay: Cleanup argument printing

I can also reproduce the problem as follows.
```
$ clang++ -pg -g -o func.clang func.cpp
$ uftrace func.clang 
[...]

PR #1832, opened 2023-10-13
replay: Revise the logic of setting NEEDS_PAREN to correctly add parentheses

jyf111:
The full demangler __cxa_demangle() may or may not add parentheses at the end of symbol names, so NEEDS_PAREN was introduced.

Previously, NEEDS_PAREN was set when the last character of the symbol name was not ')'. Now, it is also set when the last few characters are "operator()", to handle functors and lambdas in C++.

Related to #1827


namhyung (approved):
LGTM

PR #1841, opened 2023-10-30
mcount: Restore/reset rstack only when the return address is hijacked

jyf111:
When specifying estimate_return, the return address is not hijacked. This patch ensures that mcount_rstack_reset_exception() is not called in this case, to avoid recording incorrect end times.

Before:
I got an inverted time bug in estimate_return mode (it only occurs when using dynamic tracing). 
I found that `mcount_rstack_reset_exception()` was called unexpectedly, which sets the end time using `mcount_gettime()`, but the end time should be estimated by `mcount_rstack_inject_return()`. 
This PR should fix this small problem.

```shell
$ clang++ tests/s-exception3.cpp -o test/exp
$ uftrace -P. -e test/exp
            [117574] | main() {
            [117574] |   A::A() {
            [117574] |     foo1() {
            [117574] |       foo2() {
            [117574] |         foo4() {
   5.050 us [117574] |           C::C();
            [117574] |           foo5() {
 129.894 us [117574] |             __cxa_allocate_exception();
  81.546 us [117574] |           } /* foo5 */
  36.698 us [117574] |           C::~C();
 515.376 us [117574] |         } /* foo4 */
 547.074 us [117574] |       } /* foo2 */
                     |          /* inverted time: broken data? */
 289.136 us [117574] |     } /* foo1 */
  13.199 us [117574] |     B::B();
  17.199 us [117574] |     B::~B();
                     |        /* inverted time: broken data? */
 335.734 us [117574] |   } /* A::A */
   5.700 us [117574] |   A::~A();
            [117574] |   catch_exc() {
            [117574] |     bar() {
   3.450 us [117574] |       B::B();
            [117574] |       bar1() {
            [117574] |         bar2() {
   3.549 us [117574] |           C::C();
   7.199 us [117574] |           __cxa_allocate_exception();
  10.399 us [117574] |           C::~C();
  64.897 us [117574] |         } /* bar2 */
                     |            /* inverted time: broken data? */
  42.348 us [117574] |       } /* bar1 */
   3.749 us [117574] |       B::~B();
            [117574] |       catch_exc() {
  73.298 us [117574] |         baz();
  86.498 us [117574] |       } /* catch_exc */
 235.792 us [117574] |     } /* bar */
 261.492 us [117574] |   } /* catch_exc */
 988.659 us [117574] | } /* main */
```

### DIFF
--- a/libmcount/wrap.c
+++ b/libmcount/wrap.c
@@ -105,6 +105,9 @@ void mcount_rstack_reset_exception(struct mcount_thread_data *mtdp, unsigned lon
 	int idx;
 	struct mcount_ret_stack *rstack;
 
+	if (unlikely(mcount_estimate_return))
+		return;
+
 	/* it needs to find how much stack frame unwinds */
 	for (idx = mtdp->idx - 1; idx >= 0; idx--) {
 		rstack = &mtdp->rstack[idx];
@@ -393,7 +396,7 @@ __visible_default void *__cxa_begin_catch(void *exception)
 	obj = real_cxa_begin_catch(exception);
 
 	mtdp = get_thread_data();
-	if (!mcount_estimate_return && !check_thread_data(mtdp) && unlikely(mtdp->in_exception)) {
+	if (!check_thread_data(mtdp) && unlikely(mtdp->in_exception)) {
 		unsigned long *frame_ptr;
 		unsigned long frame_addr;
 


libmcount/mcount.c
```diff
@@ -1356,7 +1356,7 @@ static int __mcount_entry(unsigned long *parent_loc, unsigned long child, struct
 		return -1;
 	}
 
-	if (unlikely(mtdp->in_exception)) {
+	if (!mcount_estimate_return && unlikely(mtdp->in_exception)) {
```
namhyung, line 1359: Thanks for the fix.  I think it's better to move this check to `mcount_rstack_reset_exception()` to minimize the change and to be in sync with `mcount_rstack_reset()`.

jyf111: Ok, I also found that if I don't do this, `mtdp->in_exception` will not be reset to false, which may cause other issues.
Similarly, I removed the check for `mcount_estimate_return` in `__cxa_begin_catch()`.

namhyung: Actually I don't remember why I put it in `__cxa_begin_catch()` but it seems ok.

jyf111: This line is introduced by [27b0f44](https://github.com/namhyung/uftrace/commit/27b0f44ef2fae114a3b14bedcfd71e4861269abf)

honggyukim (approved):
Thanks for the fix. The change looks correct and I've confirmed that the `inverted time: broken data?` message got disappeared by this change.

PR #1844, opened 2023-11-02
misc: Enhance and rename wget-pr script

kangtegong:
I improved the recently added wget-pr.sh script.
It's really useful to get the patch file by pull request number. However, some systems (e.g. CentOS 7 or 8, fedora 33 or 34) do not provide wget unless install manually.

This commit improves the script to get the patch with curl if wget is not available. This also renames the file to a more appropriate name.

### DIFF
--- a/misc/wget-pr.sh
+++ b/misc/wget-pr.sh
@@ -6,4 +6,14 @@ if [ $# -ne "1" ]; then
 fi
 
 pr=$1
-wget https://github.com/namhyung/uftrace/pull/$pr.patch
+
+if [ -x "$(command -v wget)" ]; then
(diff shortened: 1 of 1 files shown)


kangtegong:
This script is tested on 

- ubuntu 22.04
- ubuntu 20.04
- macOS
- alpine 3.14
- alpine 3.13
- fedora 33
- fedora 34
- centos 7
- centos 8

namhyung (commented):
Can we keep the file name as is (wget-pr.sh) even if it uses curl internally?  I think `wget` gives more information than just `get`.

kangtegong:
Okay, I just renamed the file name as is :)

namhyung (approved):
LGTM, thanks!

PR #1849, opened 2023-11-24
Fix broken kernel tracing in presence of offline CPUs

shoffmeister:
Offline CPUs may be encountered in a number of scenarios, for instance in the presence of virtualization through VMware Workstation, where the product advertises 128 CPUs as present, but then only makes a (configurable) subset of these available and online for use.


shoffmeister:
With the PR merged,
```
sudo ./uftrace record -P --logfile=uftrace.log --auto-args --nest-libcall --kernel --libname --libmcount-path=/usr/lib/uftrace ls
```
will produce kernel trace output (note the hard-coded path to libmcount, I am running this from the build environment)

utils/kernel.c
```diff
@@ -617,8 +617,13 @@ int record_kernel_trace_pipe(struct uftrace_kernel_writer *kernel, int cpu, int
 			goto retry;
 		if (errno == EAGAIN)
 			return 0;
-		else
-			return -errno;
+
+		// offline CPU
+		if (errno == ENODEV) {
```
namhyung, line 622: No need for the parentheses and you'd better put it with the above lines.

namhyung (approved):
LGTM

utils/kernel.c
```diff
@@ -615,10 +615,10 @@ int record_kernel_trace_pipe(struct uftrace_kernel_writer *kernel, int cpu, int
 	if (n < 0) {
 		if (errno == EINTR)
 			goto retry;
-		if (errno == EAGAIN)
+		if ((errno == EAGAIN) || (errno == ENODEV))
```
bernhardkaindl, line 618: There is no need (and it's not usual) to add the extra braces around `==` because it precedence and order of evaluation comes before the logical or `||`: 
```suggestion
		if (errno == EAGAIN || errno == ENODEV)
```
A detailed reference: https://learn.microsoft.com/en-us/cpp/c-language/precedence-and-order-of-evaluation?view=msvc-170

Otherwise, LGTM.

bernhardkaindl (approved):

utils/kernel.c
namhyung: Oh, cool.  I didn't notice this feature.  Yes I prefer no parenthesis.

shoffmeister: Removed the redundant brackets and force pushed for a simple commit history.

honggyukim: It looks the brackets are still there. Please check again.

shoffmeister: Indeed, sorry + fixed!

PR #1854, opened 2023-12-08
utils: Add DTRAP macro and use it in pr_err

honggyukim:
We don't have much information when uftrace hits pr_err, but it just shows the given error message.

Instead, it'd be useful if we can attach gdb inside pr_err so that we can see its backtrace and so on.

This patch adds DTRAP, which enables TRAP only when debug mode is on.

We can add DTRAP just before pr_err goes to exit().

### DIFF
--- a/utils/debug.c
+++ b/utils/debug.c
@@ -214,6 +214,7 @@ void __pr_err(const char *fmt, ...)
 
 	color(TERM_COLOR_RESET, logfp);
 
+	DTRAP();
 	exit(1);
 }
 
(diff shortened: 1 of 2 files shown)


paranlee:
Thank you @honggyukim, It looks good to me.

paranlee:
I think making a demo `gif` animation on the debugging tab and adding it to the `wiki` for other contributors will be helpful.

honggyukim:
Or we can add just `stacktrace` when `DEBUG_MODE` is on.

namhyung (approved):
LGTM

Issue #1859, opened 2023-12-14
warning: implicit declaration of function 'close'


honggyukim:
There is a compilation warning in RISC-V.
```
/home/honggyu/work/uftrace/git/uftrace/utils/kernel-parser.c:151:9: warning: implicit declaration of function 'close'; did you mean 'pclose'? [-Wimplicit-function-declaration]
  151 |         close(kp->fds[cpu]);
      |         ^~~~~
      |         pclose
```
This can be fixed by including `unistd.h`.

paranlee:
I included fixing this issue with `utils/kernel-parser.c` adding header `unistd.h`.

namhyung:
Please separate the fix from the dead-store changes.

PR #1860, opened 2023-12-16
utils: Fix warnings implicit declaration of function 'close'

paranlee:
This patch fix warning on compile time with including unistd.h header file.
```
  utils/kernel-parser.c:151:9:
  warning: implicit declaration of function 'close';
  did you mean 'pclose'?
  [-Wimplicit-function-declaration]
    151 |         close(kp->fds[cpu]);
        |         ^~~~~
        |         pclose
```
Fixed: #1859

### DIFF
--- a/utils/kernel-parser.c
+++ b/utils/kernel-parser.c
@@ -4,6 +4,7 @@
 #include <string.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


namhyung (approved):
LGTM.

p.s. Warning messages like you added in the commit message can be longer than the normal messages as I think it's ok to keep the original shape.  But I'm also fine with your change, thanks!

PR #1871, opened 2023-12-30
libmcount: Remove UFTRACE_PIPE related code

honggyukim:
Since the following commit removed setting UFTRACE_PIPE, there is no need to read the env from libmcount side.

  fa064544 record: Get rid of UFTRACE_PIPE env variable

This patch also removes the local variable "pipefd_str" related code.


namhyung (approved):
LGTM

Issue #1877, opened 2024-01-11
Clang randomly crashing due to -mno-sse2 applied on x86_64


truboxl:
Hi, it seems that https://github.com/namhyung/uftrace/commit/d17c3e4b538941b3728abb0dca8bce6f1e8ba3a2

Is causing problem for https://github.com/llvm/llvm-project/issues/42971, https://github.com/llvm/llvm-project/issues/29774, https://github.com/android/ndk/issues/1987

Can you help comment on them?

namhyung:
Sure, thanks for the report.

MaskRay:
`-mgeneral-regs-only` (https://reviews.llvm.org/D103943) is probably suitable for your use case. It will place floating-point arguments in general purpose registers.

`-mno-sse2` (without `-mno-sse`) may cause crashes when using floating points.

namhyung:
@MaskRay thanks for your reply.  It'd be good as long as it suppresses the FP/SIMD registers.

One question, is it ok (IOW not to crash) if I add `-mno-sse` too?

namhyung:
Also I realized that script support (both python and luajit) in libmcount uses FP types. :(

It seems the minimal fix is to add `-mno-sse`.  We might consider to drop script support at record.  @honggyukim wdyt?

MaskRay:
My memory has been fuzzy... You probably need to try different code patterns (local variable/parameter float/double, floating point expressions, etc) to figure out whether the combinations of `-mno-sse -mno-sse2`  work as intended...

namhyung:
At the point, I think it's ok not to work as intended.  Higher priority would be not to crash compilers! :)

In my local testing, adding `-mno-sse` makes it stop crashing.  So I'll add it for now.  And then I'll check if it supports `-mgeneral-regs-only` and use it.

namhyung:
Pushed 95cb1112.  Maybe we need to release v0.15.1 with this.

honggyukim:
> We might consider to drop script support at record. @honggyukim wdyt?

Sorry for the late response. I think disabling the entire record time script is not a good idea. How about just dropping floating point argument handling instead?

truboxl:
Btw I think 32bit build need this: https://github.com/termux/termux-packages/pull/18963
Credit to @2096779623

namhyung:
Sure, can you please send a PR?

namhyung:
> How about just dropping floating point argument handling instead?

[...]

PR #1878, opened 2024-01-16
build: Add -mno-sse to fix i386 build

2096779623:
Part of #1877 .


namhyung:
Can you please add a little more detail to the commit message if we don't have this change?

2096779623:
> Can you please add a little more detail to the commit message if we don't have this change?

Done.

namhyung (approved):
LGTM

PR #1887, opened 2024-02-12
slide: Add tensorflow and keras tracing example

honggyukim:
It'd be better to have more practical python tracing example in slide document so this patch adds tensorflow and keras tracing example.

The example python program can be found at [1].

[1] https://github.com/keras-team/keras-io/blob/master/examples/vision/mnist_convnet.py

### DIFF
--- a/doc/uftrace.html
+++ b/doc/uftrace.html
@@ -161,6 +161,13 @@
       }
       x-tui-blue-line  { background-color: blue; }
       x-tui-white-line { background-color: white; color: black; }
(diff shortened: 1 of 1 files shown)


honggyukim:
The slide can be found from https://uftrace.github.io/slide/#199 and tui output is at https://uftrace.github.io/slide/#203 as follows.

image

namhyung (approved):
LGTM

paranlee:
Awesome!

Issue #1886, opened 2024-02-08
python tracing failure with RuntimeError: Too many open files


honggyukim:
This python code is based on a GPT generated script, but removed some unuseful parts for this issue.
```py
$ cat cnn-image.py
#!/usr/bin/env python3

import torch
import torchvision
import torchvision.transforms as transforms

# Step 1: Load and Normalize CIFAR-10 Data
transform = transforms.Compose(
    [transforms.ToTensor(),
     transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])

trainset = torchvision.datasets.CIFAR10(root='./data', train=True,
                                        download=True, transform=transform)
trainloader = torch.utils.data.DataLoader(trainset, batch_size=4,
                                          shuffle=True, num_workers=2)

# Step 4: Train the network
for epoch in range(2):  # loop over the dataset multiple times
    running_loss = 0.0
    for i, data in enumerate(trainloader, 0):
        pass
```
This can be normally executed.
```
$ chmod +x cnn-image.py
$ ./cnn-image.py
Files already downloaded and verified
```
But it shows `RuntimeError: Too many open files` when running with uftrace as follows.
```
$ uftrace record ./cnn-image.py
Files already downloaded and verified
WARN: unpaired cygprof exit: dropping...
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/honggyu/usr/lib/uftrace/uftrace.py", line 42, in <module>
    exec(code, new_globals)
  File "<string>", line 20, in <module>
  File "/home/honggyu/.local/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 630, in __next__
    data = self._next_data()
  File "/home/honggyu/.local/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 1328, in _next_data
    idx, data = self._get_data()
  File "/home/honggyu/.local/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 1294, in _get_data
    success, data = self._try_get_data()
  File "/home/honggyu/.local/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 1159, in _try_get_data
    raise RuntimeError(
[...]

PR #1888, opened 2024-02-14
python: Fix GC isn't working correctly.

yihong0618:
The uftrace_python_symbol keeps code PyObject in rb tree so GC can not run correctly in the trace process. Fixed the problem by dropping it.

Fixed: #1886

### DIFF
--- a/python/trace-python.c
+++ b/python/trace-python.c
@@ -102,7 +102,6 @@ static union uftrace_python_symtab *dbg_info;
 /* symbol table entry to maintain mappings from code to addr */
 struct uftrace_python_symbol {
 	struct rb_node node;
-	PyObject *code;
 	char *name;
 	uint32_t addr;
 	uint32_t flag;
@@ -653,42 +652,14 @@ static void init_uftrace(void)
 /* this is called during GC traversal */
 static int uftrace_py_traverse(PyObject *m, visitproc visit, void *arg)
 {
-	struct uftrace_py_state *state;
-	struct rb_node *node;
-	struct uftrace_python_symbol *sym;
-
-	state = PyModule_GetState(m);
-
-	Py_VISIT(state->trace_func);
-
-	node = rb_first(&code_tree);
-	while (node) {
-		sym = rb_entry(node, struct uftrace_python_symbol, node);
-		Py_VISIT(sym->code);
-
-		node = rb_next(node);
-	}
+	/* do nothing for now */
 	return 0;
 }
 
 /* this is called before the module is deallocated */
 static int uftrace_py_clear(PyObject *m)
 {
-	struct uftrace_py_state *state;
-	struct rb_node *node;
-	struct uftrace_python_symbol *sym;
-
-	state = PyModule_GetState(m);
-
-	Py_CLEAR(state->trace_func);
-
-	node = rb_first(&code_tree);
-	while (node) {
-		sym = rb_entry(node, struct uftrace_python_symbol, node);
-		Py_CLEAR(sym->code);
-
-		node = rb_next(node);
-	}
+	/* do nothing for now */
 	return 0;
 }
 
@@ -878,6 +849,7 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
 	const char *file_name = NULL;
 	char *func_name;
 	bool is_main = false;
+	int cmp;
 
 	if (is_pyfunc) {
 		code = PyObject_GetAttrString(frame, "f_code");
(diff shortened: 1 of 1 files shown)


python/trace-python.c
```diff
@@ -886,33 +858,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
 	}
 	else {
 		code = args;
-		Py_INCREF(code);
```
namhyung, line 889: This is needed to maintain the same reference count with  the`if` case.

python/trace-python.c
```diff
@@ -886,33 +858,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
...
 	}
 
+	if (is_pyfunc)
+		func_name = get_python_funcname(frame, code, &is_main);
+	else
+		func_name = get_c_funcname(frame, code);
```
namhyung, line 866: After this, it should decrease the reference count of `code` as it's not used anymore.

python/trace-python.c
```diff
@@ -886,33 +858,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
...
+	new_addr = get_new_sym_addr(func_name, !is_main);
+
 	while (*p) {
 		parent = *p;
 		iter = rb_entry(parent, struct uftrace_python_symbol, node);
 
-		/* just compare pointers of the code object */
-		if (iter->code == code) {
-			Py_DECREF(code);
+		if (iter->addr == new_addr) {
```
namhyung, line 877: It should free `func_name` here.

namhyung (commented):
Thanks for working on this!  But I'm afraid it won't work well because `get_new_sym_addr()` always returns a new address even for the same name.

I think you need to compare `func_name` instead.  It'd have little more overhead but probably better to make GC working properly.

python/trace-python.c
yihong0618: added it

python/trace-python.c
yihong0618: nice catch.

yihong0618:
Fixed thanks for the review, And use func_name is indeed a better way. and uftrace_py_traverse and uftrace_py_clean is no need either.

python/trace-python.c
```diff
@@ -889,30 +860,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
...
+	if (is_pyfunc)
+		func_name = get_python_funcname(frame, code, &is_main);
+	else
+		func_name = get_c_funcname(frame, code);
+
+	if (func_name == NULL)
+		return NULL;
+
+	/* code is not used anymore */
+	Py_DECREF(code);
```
namhyung, line 872: Please move this above the `if` statement so that we can release the `code` object even if `func_name` is NULL.

python/trace-python.c
```diff
@@ -889,30 +860,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
...
+
 	while (*p) {
 		parent = *p;
 		iter = rb_entry(parent, struct uftrace_python_symbol, node);
 
-		/* just compare pointers of the code object */
-		if (iter->code == code) {
-			Py_DECREF(code);
+		/* compare func_name */
+		if (iter->name == func_name) {
```
namhyung, line 879: I don't think we can do this.  It should call `strcmp()`.

python/trace-python.c
```diff
@@ -889,30 +860,32 @@ static struct uftrace_python_symbol *convert_function_addr(PyObject *frame, PyOb
...
 			return iter;
```
namhyung, line 880: Please free `func_name` before return;

yihong0618: fixed

python/trace-python.c
yihong0618: fixed

python/trace-python.c
yihong0618: fixed

namhyung (approved):
LGTM

honggyukim:
Hi @yihong0618, thanks very much for the fix. I've also found that it fixes the problem I reported.

But you don't have to put my `Signed-off-by` when I didn't involve the patch itself. I think `Reported-by` would be better in this case.

yihong0618:
> Hi @yihong0618, thanks very much for the fix. I've also found that it fixes the problem I reported.
> 
> But you don't have to put my `Signed-off-by` when I didn't involve the patch itself. I think `Reported-by` would be better in this case.

learned. will remember that next thanks a lot^_^

Issue #1890, opened 2024-02-18
feat: improve python trace user experience.


yihong0618:
for now, for python tracing work, we need the following steps.

1. chmod +x xxx.py
2. make sure we have `shebang` in front of python file

but in 2020s most python scripts do not executable and some of them do not have shebang,
if we want to trace it, we must make sure the two steps done.

if we can just run 

```cli
# like uftrace a.out would be better 
uftrace xxx.py # this support all files ends with .py  
```

that would be better.
I would like to give a simple method and open one PR for that.

namhyung:
The basic idea is to trace exactly what users do.  Non-executable python script means it needs a python interpreter so I didn't support tracing the script directly.

PR #1891, opened 2024-02-18
python: improve python trace user experience

yihong0618:
Add a simple way to support trace python file without shebang and be an executable file.

Fixed: #1890

with this path we can trace python file just like

g.py like
```python3
import sys
import urllib.request

def open_file(name):
    try:
        return open(name)
    except:
        return urllib.request.urlopen(name)

def count_word(file_obj, word):
    n = 0
    content = str(file_obj.read())
    for ln in content.split('\n'):
        n += ln.count(word)
    return n

if __name__ == '__main__':
    url = 'https://github.com/namhyung/uftrace'
    word = 'uftrace'
    if len(sys.argv) > 1:
        url = sys.argv[1]
    if len(sys.argv) > 2:
        word = sys.argv[2]

    f = open_file(url)
    n = count_word(f, word)
    print(n)
```

then

```console
uftrace g.py
```

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -24,6 +24,7 @@
 #include "utils/kernel.h"
 #include "utils/list.h"
 #include "utils/perf.h"
+#include "utils/script.h"
 #include "utils/shmem.h"
 #include "utils/symbol.h"
 #include "utils/utils.h"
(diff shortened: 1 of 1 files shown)


yskelg:
Hi, @yihong0618,

I agree that it's convenient, but we need to consider this feature carefully.
We’ve had similar discussions before. For instance, let's assume there's an access control feature like `SE Linux`or `App Armor` enabled.

If we allow the use of `/bin/uftrace` but not `/bin/python3` in specific user, there's a corner case where uftrace could execute a Python script without a shebang.

Yes, you could still run the script with `python g.py`, but that would be specific user to allowed to using python executable.

Issue #1892, opened 2024-02-20
Support a new argument format for octal values


honggyukim:
I've seen a weird enum type print as follows.
```
$ uftrace record --nest-libcall -a /usr/bin/clang /home/honggyu/hello.c
$ uftrace replay -t 3us
        ...
            [1591009] |   bfd_close_all_done() {
            [1591009] |     bfd_hash_table_free() {
   9.868 us [1591009] |       objalloc_free();
  10.134 us [1591009] |     } /* bfd_hash_table_free */
            [1591009] |     bfd_cache_close() {
   3.879 us [1591009] |       fclose(0x560e22a2b790) = 0;
   4.496 us [1591009] |     } /* bfd_cache_close */
   4.750 us [1591009] |     chmod("a.out", mod_755|mod_011+0x7) = 0;
  35.290 us [1591009] |   } /* bfd_close_all_done */
```
The `mod_755|mod_011+0x7` looks very much incorrect.

honggyukim:
`uftrace dump` shows it as `mod_755|mod_011+0x7 (509)`.
```
$ uftrace dump -t 3us -F chmod --no-event -f +module
        ...
3627691.156404887  1591009: [entry] chmod(7fa3537337c0) depth: 2
3627691.156404887  1591009: [args ] length = 16
  args[0] str: a.out
  args[1] enum uft_mode: mod_755|mod_011+0x7 (509)
3627691.156409637  1591009: [exit ] chmod(7fa3537337c0) depth: 2
3627691.156409637  1591009: [retval] length = 4
  retval d32: 0x00000000
```
`509` means `0775` in oct format.
```
$ python -c 'print(oct(509))'
0o775
```

honggyukim:
There is a simple example as follows.
```c
$ cat test-chmod.c
#include <sys/stat.h>

int main()
{
    chmod(__FILE__, 0775);
    return 0;
}
```
This issue can be simply reproduced as follows.
```
$ gcc -pg test-chmod.c

$ uftrace -a a.out
# DURATION     TID     FUNCTION
   0.972 us [1597819] | __monstartup();
   0.247 us [1597819] | __cxa_atexit();
            [1597819] | main() {
 256.815 us [1597819] |   chmod("test-chmod.c", mod_755|mod_011+0x7) = 0;
 258.027 us [1597819] | } /* main */
```

honggyukim:
Thinking it again, how about having a new argument format for octal values?
```diff
[...]

PR #1895, opened 2024-03-02
replay: Support a new argument format for octal values

nadongguri:
Enumeration members absent in uft_mode may cause unexpected output.
Adding a new argument format for octal values can handle it.

Fixed: #1892


nadongguri:
- Update document.
- I'm reading https://github.com/namhyung/uftrace/wiki/Testing, and I'm wondering if it's possible to modify a test command like adding -a option.

honggyukim:
Thanks for the update.

>  I'm wondering if it's possible to modify a test command like adding -a option.

You can get some ideas with f73febfc2d17efd5b5586ca4c68e32b6a102c013.

namhyung (commented):
Looks good.  You can add a new test case that calls `creat ()`and `unlink()` and see if the file mode is printed in octal as expected.

nadongguri:
Added test case. :)

```bash
./runtest.py -vdp -O0 -c gcc 288

Compiler                  gc
Runtime test case         pg
------------------------: O0
build command: gcc -o t-arg_oct -fno-inline -fno-builtin -fno-ipa-cp -fno-omit-frame-pointer -D_FORTIFY_SOURCE=0 -g -pg -O0 -fno-ipa-sra  s-arg_oct.c
test command: /home/dongjin/uftrace/uftrace live --no-pager --no-event --libmcount-path=/home/dongjin/uftrace  -F main -a -N memcpy t-arg_oct
=========== original ===========
            [1587656] | main() {
  78.651 us [1587656] |   creat("bar.foo", 0755) = 4;
   2.280 us [1587656] |   chmod("bar.foo", 0777) = 0;
   3.910 us [1587656] |   unlink("bar.foo") = 0;
  87.391 us [1587656] | } = 0; /* main */

main() {
   creat("bar.foo", 0755) = 4;
   chmod("bar.foo", 0777) = 0;
   unlink("bar.foo") = 0;
 } = 0; /* main */
main() {
   creat("bar.foo", 0755) = 4;
   chmod("bar.foo", 0777) = 0;
   unlink("bar.foo") = 0;
 } = 0; /* main */
288 arg_oct             : OK
```

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+import re
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'arg_oct', result="""
+# DURATION     TID     FUNCTION
+   0.560 us [1587104] | __monstartup();
+   0.170 us [1587104] | __cxa_atexit();
```
honggyukim, line 12: This part can be removed because `-F main` is used at `setup`.

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+            [1587104] | main() {
+ 120.272 us [1587104] |   creat("bar.foo", 0755) = 4;
+   3.210 us [1587104] |   chmod("bar.foo", 0777) = 0;
+   6.350 us [1587104] |   unlink("bar.foo") = 0;
+ 131.752 us [1587104] | } = 0; /* main */
+""", cflags='-g')
```
honggyukim, line 18: The `cflags='-g'` is not needed because this tests only for known library calls, which are described at `misc/prototypes.h`.
```suggestion
""")
```

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self):
+        TestBase.__init__(self, 'arg_oct', result="""
+# DURATION     TID     FUNCTION
+   0.560 us [1587104] | __monstartup();
+   0.170 us [1587104] | __cxa_atexit();
+            [1587104] | main() {
+ 120.272 us [1587104] |   creat("bar.foo", 0755) = 4;
+   3.210 us [1587104] |   chmod("bar.foo", 0777) = 0;
+   6.350 us [1587104] |   unlink("bar.foo") = 0;
+ 131.752 us [1587104] | } = 0; /* main */
```
honggyukim, line 17:
```suggestion
 131.752 us [1587104] | } /* main */
```

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+""", cflags='-g')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not "dwarf" in self.feature:
+            return TestBase.TEST_SKIP
```
honggyukim, line 22: It's not dwarf test so this should be removed.

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+        # cygprof doesn't support arguments now
+        if cflags.find('-finstrument-functions') >= 0:
+            return TestBase.TEST_SKIP
```
honggyukim, line 25: Function arguments for library calls can be captured even for `-finstrument-functions` compiled binaries. So this is not needed as well.

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,30 @@
...
+
+        return TestBase.build(self, name, cflags, ldflags)
```
honggyukim, line 27: This is same as default `TestBase.build` so the entire `def build` can be removed from line number 20 to here at 27.

honggyukim (commented):
Thanks for your work! It looks good but please break the changes into the following 3 patches.
1. implementation
2. document update
3. test

In addition, please make sure to fetch and rebase your changes onto the current upstream master. Some patches were pushed to master today by @namhyung.

nadongguri:
@honggyukim ,
- The existing commit was separated into 3 commits.
- Revised a test script. :)

honggyukim (approved):
LGTM. Thanks very much for your work!

tests/t288_arg_oct.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'arg_oct', result="""
```
namhyung, line 7: Nit, I prefer using `-` instead of `_` in the test file name (`arg-oct`).  The test case is different though, but please follow the convention.

namhyung (commented):
Looks good, but can you please further split the first commit to have the core logic change and prototype/autoarg change separately?

tests/s-arg_oct.c
```diff
@@ -0,0 +1,12 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+int main()
```
namhyung, line 5: I think valid C requires `void` in the argument.

nadongguri: Added it. :)

tests/t288_arg_oct.py
nadongguri: Done. :)

namhyung:
I understand your feeling but please try to focus what's the main change in each commit.  I really want to give a feedback on the commit message directly, but it seems there's no way to do that in Github.

For the first commit: it's not just for the replay command.  'record' should understand the new format.  So I'd like to have general "uftrace:" prefix.  Also I prefer shorter subject lines.  Otherwise the message in the first commit is fine.  And I think we need to mention the actual format specifier ("o").  And I think the actual fix for the issue in the second commit so let's have the "Fixed" there.  So it'd be

```
uftrace: Support octal argument format

Enumeration values not present in uft_mode may cause unexpected output.
Adding a new argument format for octal values ("argN/o") can handle it.

Signed-off-by: ...
```

The second commit message copied the first one.  It should mention you are applying it to some filesystem library functions.  Something like this

```
misc: Use octal format for filesystem functions

Some library functions deal with a file mode which is usually represented
in octal.  Let's use the octal format instead of enum to avoid broken output.

Fixed: ...
Signed-off-by: ...
```

The third commit is fine.

The final commit can include the test output as if you run it manually like below.  Note that you'd better indent when you insert a code block or shell output.

```
test: Add a test case for octal argument format

    $ uftrace -F main -a  t-arg-oct
                [1587104] | main() {
     120.272 us [1587104] |   creat("bar.foo", 0755) = 4;
       3.210 us [1587104] |   chmod("bar.foo", 0777) = 0;
       6.350 us [1587104] |   unlink("bar.foo") = 0;
     131.752 us [1587104] | } /* main */

Signed-off-by: ...
```

Hope this helps.

tests/s-arg-oct.c
```diff
@@ -0,0 +1,12 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+int main(void)
+{
+	const char *pathname = "bar.foo";
+	creat(pathname, 0755);
```
honggyukim, line 8: We usually give an empty line after declarations.

honggyukim (commented):
I actually wanted to ask enhancing the commit messages but didn't add feedback. 
@namhyung Thanks for the guide in details.

tests/s-arg-oct.c
nadongguri: Add empty between lines 7 and 8 and before `return 0;` as well.

nadongguri:
I appreciate your kind and informative feedback!

- I reorganized the commit titles and messages based on @namhyung's comment.
One difference is in the 4th commit message,  the binary name was changed from `t-oct-arg` to `s-oct-arg `because the file name is `s-oct-arg.c`. If `t-oct-arg` is an appropriate binary name, I will re-push it again.
- Added new lines in `s-arg-oct.c`

honggyukim:
> the binary name was changed from t-oct-arg to s-oct-arg because the file name is s-oct-arg.c. If t-oct-arg is an appropriate binary name, I will re-push it again.

Please use `t-oct-arg` for binary name. We use this prefix for test binaries and use `s-` prefix for source code.

honggyukim:
And the commit message for the forth commit is kind of empty except for the execution output. You can think that the commit message as an email to the patch readers. So I would encourage you to write complete sentences in the commit message.

nadongguri:
Regarding fourth commit message,
- Used s- prefix.
- Added a description.

honggyukim:
The commit message is misaligned. Could you please fix it?
```diff
-    $ uftrace -F main -a  t-arg-oct
+    $ uftrace -F main -a t-arg-oct
     # DURATION     TID     FUNCTION
-            [1630269] | main() {
+                [1630269] | main() {
       87.371 us [1630269] |   creat("bar.foo", 0755) = 4;
        3.990 us [1630269] |   chmod("bar.foo", 0777) = 0;
        4.540 us [1630269] |   unlink("bar.foo") = 0;
       97.791 us [1630269] | } /* main */
```

namhyung:
The first line has two spaces intentionally to split uftrace options from the target program.  But it actually doesn't matter much, so feel free to change it. :)

nadongguri:
Thanks for the check.
I modified the fourth commit message.

honggyukim (approved):
LGTM again from my side. Thanks for the update for all our comments.

PR #1909, opened 2024-03-26
include libgen.h for basename

kraj:
basename prototype has been removed from string.h from latest musl [1] compilers e.g. clang-18 flags the absense of prototype as error. therefore include libgen.h for providing it.

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

### DIFF
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -11,6 +11,9 @@
 
 #include <ctype.h>
 #include <endian.h>
+#ifndef __GLIBC__
+#include <libgen.h>
+#endif
 #include <limits.h>
 #include <signal.h>
 #include <stdbool.h>
@@ -21,6 +24,7 @@
 #include <time.h>
 
 #include "compiler.h"
+
 
 #ifndef container_of
 #define container_of(ptr, type, member)                                                            \


utils/utils.h
```diff
@@ -11,6 +11,9 @@
 
 #include <ctype.h>
 #include <endian.h>
+#ifndef __GLIBC__
+#include <libgen.h>
+#endif
```
namhyung, line 16: I think we can just remove the `__GLIBC__` check as glibc also provides `libgen.h`.

kraj: We can but then it will start using posix basename API for glibc as well, I just wanted to keep the glibc to keep using GNU extention of basename as it is doing so without it. Another option was to include it after `string.h` unconditionally, but then CI build breaks since it wants includes to be alphabetical :(

namhyung: I see.  Probably there's a way (maybe in a comment?) to suppress the pre-commit check failure.  But I'm not familiar with that, so let's have this for now.

honggyukim: > I just wanted to keep the glibc to keep using GNU extention of basename as it is doing so without it.

Is there a strong reason of using GNU extention of basename? And what are the differences?

> I think we can just remove the __GLIBC__ check as glibc also provides libgen.h.

I also think we can remove `#ifndef __GLIBC__` guard. Besides that, `utils/utils.h` doesn't use anything about `basename` and `dirname` so we can add `#include <libgen.h>` other places where it's used other than `utils.h`.

> Probably there's a way (maybe in a comment?) to suppress the pre-commit check failure.

Adding an empty line ignores `#include` reordering.

namhyung: According to the [man page](https://man7.org/linux/man-pages/man3/basename.3.html), GNU version never modifies the input string while POSIX version (from `libgen.h`) might.

It seems we use `basename(3)` in many places so adding the header to all there may not be a good idea.  We can add it to `"utils.h"` after a newline with a comment to explain it's because we prefer GNU version and musl needs separate `libgen.h` after `string.h` due to glibc feature check ordering.

kraj: hmm adding `libgen.h` to include list regardless where, it seems to pick the posix implementation in glibc too and it seems project relies on GNU extension behavior. So Perhaps another option is to provide an implementation of `basename` which is passthrough for glibc but is defined for other libc impls e.g.

```c
static const char *uftrace_basename(const char *path)
{
  const char *base = strrchr(path, '/');
  return base ? base + 1 : path;
}
```

namhyung: Looks ok.

namhyung: @kraj will you update the PR?  Or do you want us to work on it?

kraj: Will do but it’s ok if you have cycles

PR #1913, opened 2024-04-07
Support standard deviation(STDDEV) in report output

ziming-zh:
Hello @namhyung,

I've tried to incorporate STDDEV calculation in the report output. It can now be displayed from `report -f all` command. In order to minimize the calculation overhead, I employ efficient running STDDEV calculation following the method according [this post](https://stackoverflow.com/a/1175084):

`stdev = sqrt((sum_x2 / n) - (mean * mean))`, where `mean = sum_x / n`.

This method involves sqrt calculation, which is the main calculation overhead, the sqrt result should be rounded to uint_64 (ns) in consistency with the format of other statistics.  I've tried optimization targeted on (int) -> (int) sqrt calculation. However, it turns out that the fastest way is still using the std C math library <math.h>, and cast the result to an int value. Relevant discussion could be found in [this post](https://stackoverflow.com/questions/4930307/fastest-way-to-get-the-integer-part-of-sqrtn)

`result = (int) sqrt(i)`

(*Note: In that case, the compilation tag -lm should be added to the makefile*)

I've run on the tests and see that the pass/fail condition would be the same before / after the modification. I've also run defect detection tool *Infer* to guarantee that no more new defects have been introduced.

Subsequently, I've run the uftrace w/ & w/o STDDEV calculation on the same `mnist.py` code (adopted from `pytorch/example`). Despite the two new columns added to the report, there seems no significant difference of the avg/min/max execution time in other columns, which means that the added STDDEV feature would not evidently interfere the original tracer functionality.

- Without STDDEV calculation: [log_with_std.txt](https://github.com/namhyung/uftrace/files/14898180/log_with_std.txt)
- With STDDEV calculation: [log_without_std.txt](https://github.com/namhyung/uftrace/files/14898181/log_without_std.txt)

I've also run Facebook's Infer tool to check for potential defection in the current code base. I've made three adjustment to the file handling (no close for open) and memory deallocation (no free after xalloc) in `utils/dwarfs.c`. I've attached the complete infer report in [bugs.txt](https://github.com/namhyung/uftrace/files/14898292/bugs.txt) and [report.csv](https://github.com/namhyung/uftrace/files/14898288/report.csv), please see if you need help to check and correct the remaining of them.

Finally, I must say this is such a good projects with very well-done documentation and superb functionality. Thank you for your effort! Please let me know for further revision of this PR.


Makefile
```diff
@@ -61,7 +61,7 @@ COMMON_CFLAGS += -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(AR
 COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
 COMMON_CFLAGS += -Wdeclaration-after-statement -Wstrict-prototypes
 
-COMMON_LDFLAGS := -ldl -pthread -Wl,-z,noexecstack $(LDFLAGS)
+COMMON_LDFLAGS := -ldl -lm -pthread -Wl,-z,noexecstack $(LDFLAGS)
```
namhyung, line 64: I don't think we need the math library for libmcount or other places.  Please add it to `UFTRACE_LDFLAGS` instead.

cmds/tui.c
```diff
@@ -182,11 +182,11 @@ static const char *graph_field_names[NUM_GRAPH_FIELD] = {
...
 };
 
-#define NUM_REPORT_FIELD 10
+#define NUM_REPORT_FIELD 12
 
 static const char *report_field_names[NUM_REPORT_FIELD] = {
-	"TOTAL TIME", "TOTAL AVG", "TOTAL MIN", "TOTAL MAX", "SELF TIME",
-	"SELF AVG",   "SELF MIN",  "SELF MAX",	"CALL",	     "SIZE",
+	"TOTAL TIME", "TOTAL AVG", "TOTAL MIN", "TOTAL MAX", "TOTAL STD", "SELF TIME",
+	"SELF AVG",   "SELF MIN",  "SELF MAX",	"SELF STD",  "CALL",	  "SIZE",
```
namhyung, line 189: Maybe "STD" looks confusing, how about "STDV"?

utils/dwarf.c
```diff
@@ -1764,6 +1768,7 @@ static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symt
 	for (i = 0; i < rets->nr; i++)
 		free_filter_pattern(&ret_patt[i]);
 	free(ret_patt);
+	free(dinfo->locs);
```
namhyung, line 1771: Thanks for catching these, but they are totally unrelated changes.  You should have them in a separate commit, preferably in a separate PR.

utils/report.h
```diff
@@ -17,7 +17,10 @@ enum avg_mode {
 struct report_time_stat {
 	uint64_t sum;
 	uint64_t rec; /* time in recursive call */
+	uint64_t sum_sq;
+	uint64_t rec_sq;
 	uint64_t avg;
+	uint64_t std;
```
namhyung, line 23: It doesn't need to be an integer, why not using `double` type?

namhyung (commented):
Thanks for your work!  In general, they look good.  I've added some comments for some cleanups.

But I'd like to update the the commit message to have a "report:" prefix and remove the unrelated changes.  Also it'd be nice if you would add some example output of the new user-facing changes.  You can use  the fibonacci example to show the stddev in the report.  Ideally we can add it to the test suite to verify the behavior.  Finally please update the uftrace-report and uftrace-tui documentation to add the new fields.

cmds/tui.c
namhyung: Or just "SD" according to the [wikipedia](https://en.wikipedia.org/wiki/Standard_deviation).

cmds/tui.c
```diff
@@ -369,20 +369,22 @@ REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME, total, total.sum, total, "TOTAL TIME");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_AVG, total-avg, total.avg, total_avg, "TOTAL AVG");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MIN, total-min, total.min, total_min, "TOTAL MIN");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MAX, total-max, total.max, total_max, "TOTAL MAX");
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_STD, total-std, total.std, total_std, "TOTAL STD");
```
namhyung, line 372: I'm curious if the standard deviation still has the same unit with the original values.  I guess it's just a pure number, isn't it?

cmds/tui.c
ziming-zh: Thanks for the suggestion! I'll change the name to STDV here

utils/dwarf.c
ziming-zh: Sure. I'll sumit a separate issue for these handler fix!

utils/report.h
ziming-zh: Hi, that is a good point! I'm using the `uint64_t` in consistent with the argument type of the print_time function. 

(line 229 in tui.c) `static void print_time(uint64_t ntime)`

I think we do not need to break this consistency to STDV calculation, and it is sufficiently accurate to reflect the distribution. ^_^

cmds/tui.c
ziming-zh: Yes, exactly. It is just a pure number and has the same unit. So what I am trying to do here is to maintain the consistency.

Makefile
ziming-zh: Thanks for the suggestions. I'm removing it to `UFTRACE_LDFLAGS`!

cmds/tui.c
namhyung: Are you sure it has the same unit?  I thought stddev numbers don't have units.

ziming-zh: Sorry for the confusion here. Indeed, stddev numbers don't have units and is a pure number. The units are generated by the `print_time` function. I'll try to update the commit later today!

ziming-zh:
@namhyung Thank you for the feedback! I've made the requested changes:

- [x] Updated the commit message to include a "report:" prefix.
- [x] Removed unrelated changes from the commit.
- [x] Added example output of the new user-facing changes, including the standard deviation of the `Fibonacci` and `abc` example in the report.
- [x] Updated the test suite to verify the behavior of the new changes.
- [x] Updated the documentation for uftrace-report and uftrace-tui to include information about the new fields.

Please review the changes and let me know if there are any further suggestions or modifications needed.

namhyung (commented):
Please add an example output in the commit message.  I think `uftrace report -f total-avg,total-stdv` would be fine.

cmds/tui.c
namhyung: No, I found stddev would have the same unit of the original data.

https://medium.com/@tvsdhanan009/does-standard-deviation-have-units-40ed74396bba

tests/t007_library.py
namhyung, line 1: It seems you deleted this file.

doc/uftrace-report.md
```diff
@@ -24,25 +24,26 @@ REPORT OPTIONS
...
     show all fields.
-    Default is 'total,self,call'.  See *FIELDS*.
+    Default is 'total,self,call'. See *FIELDS*. `stdv` stands for standard deviation, it would only contain value
+    if the function is invoked multiple times.
 
 -s *KEYS*[,*KEYS*,...], \--sort=*KEYS*[,*KEYS*,...]
 :   Sort functions by given KEYS.  Multiple KEYS can be given, separated by
     comma (,).  Possible keys are `total` (time), `total-avg`, `total-min`,
-    `total-max`, `self` (time), `self-avg`, `self-min`, `self-max`, `size`,
+    `total-max`, `total-stdv`, `self` (time), `self-avg`, `self-min`, `self-max`, `self-stdv`, `size`,
```
namhyung, line 37: Please keep the line length as similar as others.  I think I kept it around 80 characters per line.  It's not the part of the auto formatting but it'd be nice if you follow the convention even if it ends up touching unchanged lines.

doc/uftrace-report.md
```diff
@@ -24,25 +24,26 @@ REPORT OPTIONS
...
     the possible keys can be `avg`, `min` and `max` that apply to total or self
     time respectively.
 
 \--avg-total
-:   Show average, min, max of each function's total time.
+:   Show average, min, max, stdv of each function's total time.
 
 \--avg-self
-:   Show average, min, max of each function's self time.
+:   Show average, min, max, stdv of each function's self time.
```
namhyung, line 46: You can say "standard deviation" here.

doc/uftrace-report.md
```diff
@@ -360,7 +355,8 @@ Each field has following meaning:
 
 The default value is 'total,self,call'.  If given field name starts with "+",
 then it'll be appended to the default fields.  So "-f +total-avg" is as same as
-"-f total,self,call,total-avg".  And it also accepts a special field name of
+"-f total,self,call,total-avg", and "-f +total-avg,self-stdv" is as same as
+"-f total,self,call,total-avg,self-stdv".  And it also accepts a special field name of
```
namhyung, line 359: I think this part explains what "+" means so you can drop this change.

doc/uftrace-tui.md
```diff
@@ -27,12 +27,14 @@ TUI OPTIONS
...
     this option indicates report fields.  Possible values are total, total-avg,
-    total-min, total-max, self, self-avg, self-min, self-max, call and size.
+    total-min, total-max, total-stdv, self, self-avg, self-min, self-max, self-stdv, call and size.
     The default value is 'total,self,call'.
     Multiple fields can be set by using comma.
     If given field name starts with "+", then it'll be appended to the default fields.
     The special field 'none' can be used (solely) to hide all fields.
     See `uftrace-graph`(1) or `uftrace-report`(1) for an explanation of fields.
+    The `stdv` in total-stdv and self-stdv stands for standard deviation, it would only contain value
+    if the function is invoked multiple times.
```
namhyung, line 37: Ditto, keep the line length please.

tests/t044_report_avg_total.py
```diff
@@ -5,15 +5,15 @@
...
-   39.967 us   39.801 us   40.275 us  loop
-    0.701 us    0.701 us    0.701 us  __monstartup
-    0.270 us    0.270 us    0.270 us  __cxa_atexit
+   Total avg   Total min   Total max  Total stdv  Function
+  ==========  ==========  ==========  ==========  ====================
+   10.277 ms   10.277 ms   10.277 ms              main
+   10.224 ms   10.224 ms   10.224 ms              bar
+   10.210 ms   10.210 ms   10.210 ms              usleep
+    7.849 us    7.713 us    7.985 us    0.136 us  foo
+    2.499 us    2.483 us    2.547 us    0.046 us  loop
```
namhyung, line 14: I still think it's worth printing it in percentage instead.

namhyung (commented):
Can you please break the long lines?  I think the convention is 72 characters per line in the commit message.  Also it'd be nice if you indent the examples by 4 spaces so that it can be displayed properly in markdown.

ziming-zh:
Hi @namhyung,

I've changed everything to Coefficient of Variance (CV) now. Please check whether the current version works. Thank you! BTW, our instructor suggests that we should complete our contribution by the end of the semester (April 25th). Could you please kindly do me a favor and merge the PR before the deadline? I can make some in-time revision if you still find there could be some room for improvement. Much thanks for your help!

namhyung:
Change looks good now.  While it's technically correct to use the term "CV" it can be confusing with Curriculum Vitae (resume), so I think it's better to keep "STDV" and explain it in the documentation that it's actually "relative" standard deviation (RSD).

ziming-zh:
@namhyung Thanks for the comment. I've switched back to the `stdv` notation. Please review the revision. Thanks!

cmds/tui.c
```diff
@@ -369,20 +369,22 @@ REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME, total, total.sum, total, "TOTAL TIME");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_AVG, total-avg, total.avg, total_avg, "TOTAL AVG");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MIN, total-min, total.min, total_min, "TOTAL MIN");
 REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_MAX, total-max, total.max, total_max, "TOTAL MAX");
+REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME_STDV, total-stdv, total.stdv, total_stdv, "TOTAL STDV");
```
namhyung, line 372: Should be `REPORT_FIELD_PERCENTAGE`.

cmds/tui.c
```diff
@@ -369,20 +369,22 @@ REPORT_FIELD_TIME(REPORT_F_TOTAL_TIME, total, total.sum, total, "TOTAL TIME");
...
 REPORT_FIELD_TIME(REPORT_F_SELF_TIME, self, self.sum, self, "SELF TIME");
 REPORT_FIELD_TIME(REPORT_F_SELF_TIME_AVG, self-avg, self.avg, self_avg, "SELF AVG");
 REPORT_FIELD_TIME(REPORT_F_SELF_TIME_MIN, self-min, self.min, self_min, "SELF MIN");
 REPORT_FIELD_TIME(REPORT_F_SELF_TIME_MAX, self-max, self.max, self_max, "SELF MAX");
+REPORT_FIELD_TIME(REPORT_F_SELF_TIME_STDV, self-stdv, self.stdv, self_stdv, "SELF STDV");
```
namhyung, line 377: Ditto.

namhyung (commented):
Mostly looks good.  Can you please use the full name in the Signed-off-by line?

namhyung (commented):
A few nitpicks on the commit message:
* Add a blank line between the subject line and the message body
* Break long lines in the message around 72 characters per line
* Merge the example usage and the output into a single block
* Use the actual output (verbatim) in the example
* Indent the example block with 4 spaces

namhyung (approved):
LGTM, thanks!

namhyung:
Hmm.. I found that some report diff tests are failing.

```
$ make -j8 runtest TESTARG=report_diff
  TEST     test_run
Start 10 tests with 8 worker

Compiler                  gcc                                           clang                                       
Runtime test case         pg             finstrument-fu fpatchable-fun  pg             finstrument-fu fpatchable-fun
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os O0 O1 O2 O3 Os  O0 O1 O2 O3 Os O0 O1 O2 O3 Os O0 O1 O2 O3 Os
067 report_diff         : SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG  SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG
158 report_diff_policy1 : SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG  SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG
159 report_diff_policy2 : SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG  SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG
160 report_diff_policy3 : SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG  SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG
177 report_diff_policy4 : SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG  SG SG SG SG SG SG SG SG SG SG SG SG SG SG SG
239 report_diff_field1  : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK  OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK
240 report_diff_field2  : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK  OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK
241 report_diff_field3  : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK  OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK
242 report_diff_field4  : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK  OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK
243 report_diff_field5  : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK  OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK


runtime test stats
====================
total   300  Tests executed (success: 50.00%)
  OK:   150  Test succeeded
  OK:     0  Test succeeded (with some fixup)
  NG:     0  Different test result
  SG:   150  Abnormal exit by signal
  TM:     0  Test ran too long
  BI:     0  Build failed
  SK:     0  Skipped
```

ziming-zh:
That is strange. I think I don't change any of the features related to diff here

ziming-zh:
I've tried to directly running the test on the master branch, it also fails the same test. It seems that the test fail derives from previous commits.

namhyung:
Really?  I don't see the failures in the master branch.

namhyung:
I can reproduce it easily like this:
```
$ uftrace record --force pwd
/home/namhyung/tmp

$ uftrace record --force pwd
/home/namhyung/tmp

$ uftrace report --diff=uftrace.data.old
Segmentation fault.
```

With gdb and build with `DEBUG=1`, I can see this.
```
Program received signal SIGSEGV, Segmentation fault.
0x000055555559a79b in add_field (output_fields=0x5555555e9030 <output_fields>, field=0x0) at /home/namhyung/project/uftrace/utils/field.c:80
80		if (field->used)
```
It seems to access the NULL pointer for `field`.

Here's the backtrace.
```
(gdb) bt
#0  0x000055555559a79b in add_field (output_fields=0x5555555e9030 <output_fields>, field=0x0) at /home/namhyung/project/uftrace/utils/field.c:80
#1  0x00005555555b276f in setup_default_field (fields=0x5555555e9030 <output_fields>, opts=0x7fffffffd9d0, 
    p_field_table=0x5555555ea920 <field_diff_table>) at /home/namhyung/project/uftrace/utils/report.c:959
#2  0x000055555559a918 in setup_field (output_fields=0x5555555e9030 <output_fields>, opts=0x7fffffffd9d0, 
    setup_default_field=0x5555555b2711 <setup_default_field>, field_table=0x5555555ea920 <field_diff_table>, field_table_size=10)
    at /home/namhyung/project/uftrace/utils/field.c:125
#3  0x00005555555b2a00 in setup_report_field (output_fields=0x5555555e9030 <output_fields>, opts=0x7fffffffd9d0, avg_mode=AVG_NONE)
    at /home/namhyung/project/uftrace/utils/report.c:1024
#4  0x0000555555580467 in report_diff (handle=0x7fffffffd7b0, opts=0x7fffffffd9d0) at /home/namhyung/project/uftrace/cmds/report.c:471
#5  0x00005555555808df in command_report (argc=0, argv=0x7fffffffdca0, opts=0x7fffffffd9d0) at /home/namhyung/project/uftrace/cmds/report.c:546
#6  0x00005555555650aa in main (argc=0, argv=0x7fffffffdca0) at /home/namhyung/project/uftrace/uftrace.c:1543
```

namhyung:
Ok, I think it's because you didn't implement stdv for diff mode.  The diff field table doesn't have total_stdv and self_stdv.  But the default field table has 'call' column which is not out of range in the field table due to the new stdv fields.  It's questionable that comparing standard deviation though.  You can either implement the diff mode or move stdv fields after `REPORT_F_SIZE` in utils/field.h.

ziming-zh:
I see. Thanks for the guidance. I think I can choose the later approach. Comparing stdv in diff mode seems not very much necessary?

namhyung:
> Comparing stdv in diff mode seems not very much necessary?

Well.. I don't think it's in the high priority at least. :)

ziming-zh:
Hi @namhyung, I've tested the current version. Should succeed all `report`  test case for now

namhyung (approved):
LGTM!

doc/uftrace-report.md
```diff
@@ -24,25 +24,34 @@ REPORT OPTIONS
...
+    Default is 'total,self,call'. See *FIELDS*.
+    `stdv` stands for the relative standard deviation (RSD),
+    which is the standard deviation expressed as a percentage of the mean.
+    The relative standard deviation is calculated using the formula:
+
+    \[ RSD = \left( \frac{\sigma}{\mu} \right) \times 100\% \]
+
+    where:
+    - \( \sigma \) represents the standard deviation.
+    - \( \mu \) represents the mean of the function's time.
```
honggyukim, line 40: I'm leaving a comment very late but this part isn't shown properly in its markdown format.
https://github.com/namhyung/uftrace/blob/master/doc/uftrace-report.md#report-options

It is shown as follows in its man page.
```
              [ RSD = ( ) % ]

              where: - ( ) represents the standard deviation.  - ( ) represents the mean of the function’s time.
```
Can we fix it?

PR #1914, opened 2024-04-08
build: Add "check-syntax" make target

clementguidi:
Hi,

Do you have any interest in supporting syntax checkers such as Emacs' Flymake?

It would require a `check-syntax` target to be added to the Makefile.

See https://www.gnu.org/software/emacs/manual/html_node/flymake/Example_002d_002d_002dConfiguring-a-tool-called-via-make.html (bottom of the page).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -488,7 +488,10 @@ help:
 	@echo "  make O=dir [targets]        - Set directory as objdir (default: $(srcdir))"
 	@echo ""
 
+check-syntax:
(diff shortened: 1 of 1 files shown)


namhyung:
I'm not a fan of the quick syntax checker.  Eventually we will find the errors when we build the source, right?

clementguidi:
Right. I'll close this PR for now.

PR #1918, opened 2024-04-15
Fix close file descriptor in setup_dwarf_info function to release acquired resource

ziming-zh:
Hello @namhyung,

I've also run Facebook's Infer tool to check for potential defection in the current code base. I've made three adjustment to close file descriptor in setup_dwarf_info function to release acquired resource in `utils/dwarfs.c`. I've attached the complete infer report in [bugs.txt](https://github.com/namhyung/uftrace/files/14898292/bugs.txt) and [report.csv](https://github.com/namhyung/uftrace/files/14898288/report.csv). The other detected defects seems trivial on my side. I'd be happy if you could take a look at them as well.


utils/dwarf.c
```diff
@@ -326,6 +328,8 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
 
 	dinfo->file_type = elf_file_type(dinfo);
 
+	close(fd); // Close the file descriptor to release the acquired resource
```
namhyung, line 331: I'm not sure if it's ok to close the fd here.  libdw/libdwfl will access the file later and they assume the file is open during the processing.

utils/dwarf.c
```diff
@@ -310,6 +311,7 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
 		pr_dbg2("failed to setup debug info: %s\n", dwfl_errmsg(dwfl_errno()));
 		dwfl_end(dinfo->dwfl);
 		dinfo->dwfl = NULL;
+		close(fd); // Close the file descriptor to release the acquired resource
```
namhyung, line 314: I thought `dwfl_end()` would close the fd since it can remember the fd from the c`dwfl_report_offline()`.  Can you please check if it's the case?

namhyung (commented):
Please add "dwarf:" prefix to the subject line.  Also it would be short like "dwarf: Close file descriptors in setup_dwarf_info()".  And you can add the details (like it was detected by xxx tool) in the message body.

utils/dwarf.c
```diff
@@ -301,6 +301,7 @@ static int setup_dwarf_info(const char *filename, struct uftrace_dbg_info *dinfo
 	dinfo->dwfl = dwfl_begin(&dwfl_callbacks);
 	if (dinfo->dwfl == NULL) {
 		pr_dbg2("failed to begin libdwfl: %s\n", dwfl_errmsg(dwfl_errno()));
+		close(fd); // Close the file descriptor to release the acquired resource
```
namhyung, line 304: I think the comment is not necessary.

utils/dwarf.c
ziming-zh: You are right, it will be handled by dwfl_end()

utils/dwarf.c
ziming-zh: Yes, it should not be closed here

ziming-zh:
Hi @namhyung, thanks for your comment! It seems that the static analysis tool `Facebook.infer` would generate most false positive in this case. I just put a minor fix to close the file descriptor before returning -1, not sure if it is necessary. Please close this PR and #1919 if you think it is not required. Thanks!

namhyung:
Oops, I clicked wrong button.  Can you please force push the branch with the latest change and remove my merge commit?

namhyung (approved):
LGTM

Issue #1916, opened 2024-04-11
Give monotonic time offset from real time. 


MatthewKhouzam:
It would be nice to have the option to get the offset between `clock_monotonic` and `clock_realtime`. There is a simple-ish solution.

When uftrace starts, record quickly both `clock_realtime` and `clock_monotonic`. In task.txt, the timestamps can then be offset and the traces can then be correlated with lttng traces as well as system logs.

namhyung:
Sounds good.  I think we can add something in the info file where we have record data and elapsed time.

honggyukim:
Hi @MatthewKhouzam,

You explained this issue but I actually couldn't fully understand what the problem is. Could you please explain the following questions?
1. What's the meaning of the offset between `clock_monotonic` and `clock_realtime`?
2. Why this offset adjustment is needed for lttng?
3. Is is possible to change the way to get timestamps from lttng itself simply with either `CLOCK_MONOTONIC` or `CLOCK_REALTIME`?

Thanks.

honggyukim:
Hi @MatthewKhouzam, you may want to have #1923. Please have a look. This patch simply prints the UTC offset, but we might be able to provide a tool that does post processing for each `<tid>.dat` so that you can avoid time adjustment whenever you open tracecompass.

honggyukim:
Another idea is that you can modify uftrace parser of tracecompass so that it can understand this new info field and use it for automatic time adjustment.

PR #1923, opened 2024-05-10
info: Add utc_offset info for external usage

honggyukim:
In some cases, there is a requirement to get the time offset between UTC time and our timestamp from the given clock source, CLOCK_MONOTONIC by default.

This patch provides an additional info to provide UTC time offset in the output of uftrace info.

This number can be used to make time adjustment easier for external tools such as tracecompass[1].

Closes: #1916

[1] https://eclipse.dev/tracecompass
Requested-by: Matthew Khouzam <matthew.khouzam@gmail.com>

### DIFF
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -851,6 +851,41 @@ static int read_uftrace_version(void *arg)
 	return 0;
 }
 
+static int fill_utc_offset(void *arg)
+{
+	struct fill_handler_arg *fha = arg;
+	time_t current_time;
+	struct timespec ts;
+	long offset;
+
+	time(&current_time);
+	clock_gettime(clock_source, &ts);
+
+	/* calculate time offset between UTC and clock_source. */
+	offset = (long)difftime(current_time, ts.tv_sec);
+
+	dprintf(fha->fd, "utc_offset:%ld\n", offset);
+	return 0;
+}
+
+static int read_utc_offset(void *arg)
+{
+	struct read_handler_arg *rha = arg;
+	struct uftrace_data *handle = rha->handle;
+	struct uftrace_info *info = &handle->info;
+	char *buf = rha->buf;
+
+	if (fgets(buf, sizeof(rha->buf), handle->fp) == NULL)
+		return -1;
+
+	if (strncmp(buf, "utc_offset:", 11))
+		return -1;
+
+	info->utc_offset = copy_info_str(&buf[11]);
+
+	return 0;
+}
+
 struct uftrace_info_handler {
 	enum uftrace_info_bits bit;
 	int (*handler)(void *arg);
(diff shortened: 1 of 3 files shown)


honggyukim:
This patch is to make the UTC time adjustment easier than https://youtu.be/KtVdeHfD45o?si=hIMvE10XI87XqXIW.

honggyukim:
The remaing job is to make this feature doesn't break previously recorded data to support backward compatibility.

cmds/info.c
```diff
@@ -999,6 +1037,12 @@ void process_uftrace_info(struct uftrace_data *handle, struct uftrace_opts *opts
 	else
 		process(data, "# %-20s: %s", "recorded on", ctime(&statbuf.st_mtime));
 
+#if 0
+	/* This is better not to be printed. */
+	if (info_mask & UTC_OFFSET)
+		process(data, fmt, "utc offset", info->utc_offset);
```
honggyukim, line 1043: I think the `info->utc_offset` is for internal information so better not to be printed in `uftrace info` output.

honggyukim: It's now removed after another update.

honggyukim:
I just fixed `uftrace dump` crash with https://github.com/namhyung/uftrace/compare/96da52d61136ba055102603644318eedee75562c..5d83c874643b4668f99092faf7ff8c23f8803fe5.

@namhyung  @MatthewKhouzam I think it's good to be merged now. Please have a look.

cmds/info.c
```diff
@@ -851,6 +851,41 @@ static int read_uftrace_version(void *arg)
...
+	struct timespec ts;
+	double offset;
+
+	time(&current_time);
+	clock_gettime(clock_source, &ts);
+
+	/* calculate time offset between UTC and clock_source. */
+	offset = difftime(current_time, ts.tv_sec);
+
+	dprintf(fha->fd, "utc_offset:%ld\n", (time_t)offset);
```
namhyung, line 867: For `%ld`, you'd better cast it to `unsigned long`.

namhyung (commented):
It's needed to correlate the uftrace trace with external trace data.  Anyway it looks ok, only a nitpick below.

cmds/info.c
honggyukim: As I understand, `time_t` is `long` type so `%ld` is used instead of `%lu`.
```i
# 143 "/usr/include/x86_64-linux-gnu/bits/types.h" 2 3 4
        ...
typedef long int __time_t;
        ...
typedef __time_t time_t;
```
Since this `offset` is used to adjust the other `time_t` value for UTC time, I made it casted to the same type `time_t`. Should I change it to `unsigned long` and print it with `%lu` instead?

namhyung: Oh, sorry.  I meant `long`.  But it could be different on different architecture or distro.  So I think it's better to explicitly cast to `long` instead of `time_t`.

honggyukim: OK. I've just changed it to use `long` for `offset` type itself.

namhyung (approved):
LGTM

PR #1924, opened 2024-05-11
cmds: static variable used stack after return

### LINKED ISSUE #1915: cmds: static `tmp_dirname` used stack after return
In `cmds/live.c`, There was stack use after return code.
`tmp_dirname` static char pointer would be not pointed stack allocation after `command_live()` call ended.
We can use `tmp_dirname` heap or fix sized array.
```c
static char *tmp_dirname; // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
static void cleanup_tempdir(void)
{
	if (!tmp_dirname)
		return;

	remove_directory(tmp_dirname);
	tmp_dirname = NULL;
}
```

```c
int command_live(int argc, char *argv[], struct uftrace_opts *opts)
{
#define LIVE_NAME "uftrace-live-XXXXXX"
	char template[32] = "/tmp/" LIVE_NAME; // <<<<<<<<<<<<<<<<<<<
	int fd;
	struct sigaction sa = {
		.sa_flags = SA_RESETHAND,
	};
	int ret;

	if (!opts->record) {
		tmp_dirname = template; // <<<<<<<<<<<<<<<<<<<<<<<<<<<
```

yskelg:
ASAN detect stack after return on this scenario.
```
$ uftrace record --agent --trace=off ./valkey-server

$ uftrace --pid `pidof valkey-server` --trace=on
$ uftrace --pid `pidof valkey-server` --trace=off
```
This patch fix #1915 `static tmp_dirname` variable not pointed `command_live()`'s char template array stack variable.


cmds/live.c
```diff
@@ -13,14 +13,13 @@
...
-static char *tmp_dirname;
+#define TMP_DIR_NAME_SIZE 32
+
+static char tmp_dirname[TMP_DIR_NAME_SIZE] = { '\0', };
 static void cleanup_tempdir(void)
 {
-	if (!tmp_dirname)
-		return;
-
 	remove_directory(tmp_dirname);
```
yskelg, line 21: I think wrong `dirname` would be covered with `opendir()` cheking.

cmds/live.c
```diff
@@ -13,14 +13,15 @@
 #include "utils/socket.h"
 #include "utils/utils.h"
 
-static char *tmp_dirname;
+#define TMP_DIR_NAME_SIZE 32
+
+static char tmp_dirname[TMP_DIR_NAME_SIZE] = {
+	'\0',
```
namhyung, line 19: I think you can initialize it with `LIVE_NAME` below.

cmds/live.c
namhyung: But let's check if `tmp_dirname` has a valid string first.

cmds/live.c
```diff
@@ -13,14 +13,15 @@
...
+};
 static void cleanup_tempdir(void)
 {
-	if (!tmp_dirname)
-		return;
-
 	remove_directory(tmp_dirname);
-	tmp_dirname = NULL;
+	memset(tmp_dirname, 0, sizeof(char) * TMP_DIR_NAME_SIZE);
```
namhyung, line 24: You can just use `sizeof(tmp_dirname)`;

cmds/live.c
```diff
@@ -416,15 +417,14 @@ static int forward_options(struct uftrace_opts *opts)
 int command_live(int argc, char *argv[], struct uftrace_opts *opts)
 {
 #define LIVE_NAME "uftrace-live-XXXXXX"
-	char template[32] = "/tmp/" LIVE_NAME;
+	char template[TMP_DIR_NAME_SIZE] = "/tmp/" LIVE_NAME;
```
namhyung, line 420: Now we can remove the `template`.

cmds/live.c
```diff
@@ -438,12 +438,13 @@ int command_live(int argc, char *argv[], struct uftrace_opts *opts)
...
 				pr_err("cannot create temp name");
-			tmp_dirname = template;
 		}
 
 		close(fd);
-		unlink(tmp_dirname);
 
+		unlink(template);
+
+		snprintf(tmp_dirname, strlen(template), "%s", template);
```
namhyung, line 447: You should use `sizeof(tmp_dirname)` instead of `strlen(template)`, otherwise it might not set the last NUL byte.

cmds/live.c
paranlee: Thank for idea!

cmds/live.c
paranlee: Done

cmds/live.c
paranlee: Thank you @namhyung. good advice! 👍

cmds/live.c
```diff
@@ -415,30 +423,26 @@ static int forward_options(struct uftrace_opts *opts)
...
 		umask(022);
-		fd = mkstemp(template);
+		fd = mkstemp(tmp_dirname);
 		if (fd < 0) {
-			/* can't reuse first template because it was trashed by mkstemp */
-			strcpy(template, LIVE_NAME);
-
 			if (errno != EPERM && errno != ENOENT)
-				pr_err("cannot access to /tmp");
+				pr_err("cannot access to " TMP_LIVE_NAME);
```
paranlee, line 438: I think non-sense situation that someone makes directory or symbolic link name `/tmp/uftrace-live-XXXXXX`.

cmds/live.c
```diff
@@ -13,14 +13,22 @@
 #include "utils/socket.h"
 #include "utils/utils.h"
 
-static char *tmp_dirname;
+#define LIVE_NAME "uftrace-live-XXXXXX"
+#define TMP_LIVE_NAME "/tmp/" LIVE_NAME
```
paranlee, line 17: String comparison makes macro going to top of code.

cmds/live.c
```diff
@@ -13,14 +13,22 @@
...
+
+#define TMP_DIR_NAME_SIZE 32
+
+static char tmp_dirname[TMP_DIR_NAME_SIZE];
 static void cleanup_tempdir(void)
 {
-	if (!tmp_dirname)
+	if (strncmp(TMP_LIVE_NAME, tmp_dirname, strlen(TMP_LIVE_NAME)) != 0 ||
+	    strncmp(LIVE_NAME, tmp_dirname, strlen(LIVE_NAME)) != 0) {
```
namhyung, line 25: This condition is always true.  You need to check if it has non-zero contents.

cmds/live.c
namhyung: `mkstemp` would replace "XXXXXX" to a random string.  You need to print the actual string in `tmp_dirname`.

cmds/live.c
```diff
@@ -415,30 +423,26 @@ static int forward_options(struct uftrace_opts *opts)
...
 
-			fd = mkstemp(template);
+			/* can't reuse first template because it was trashed by mkstemp */
+			snprintf(tmp_dirname, sizeof(tmp_dirname), "%s", LIVE_NAME);
+			fd = mkstemp(tmp_dirname);
 
 			if (fd < 0)
-				pr_err("cannot create temp name");
-			tmp_dirname = template;
+				pr_err("cannot create " LIVE_NAME);
```
namhyung, line 445: Ditto.

yskelg: Thank you @namhyung, I understand the `mkstemp would replace "XXXXXX"`. Done!

cmds/live.c
yskelg: Done @namhyung 👍

cmds/live.c
namhyung: I think it's ok to check the first byte only.

```c
if (tmp_dirname[0] == '\0')
    return;
```

namhyung (commented):
Also, this change is specific to `live` command.  So the commit subject should start with "live:" instead of generic "cmds:".

cmds/live.c
yskelg: I agree, Done!

namhyung (approved):
LGTM

Issue #1934, opened 2024-06-19
Compile warning caused by Python3.12 header from `-Wdeclaration-after-statement`


yskelg:
In `Makefile`, We use `C11`.
- https://github.com/namhyung/uftrace/commit/4c9117d847543b4986cf1a50b3bd6b56b4573a85
But because C89 based compiler warning flag `-Wdeclaration-after-statement`, latest compiler (gcc version 14.1.1 20240522, clang version 18.1.6) told me that some warning.
- https://github.com/namhyung/uftrace/commit/1e2a3f393aac3303680d0050aab8347fe89b015d
```
COMMON_CFLAGS := -std=gnu11 -D_GNU_SOURCE $(CFLAGS) $(CPPFLAGS)
COMMON_CFLAGS += -iquote $(srcdir) -iquote $(objdir) -iquote $(srcdir)/arch/$(ARCH)
COMMON_CFLAGS += -W -Wall -Wno-unused-parameter -Wno-missing-field-initializers
COMMON_CFLAGS += -Wdeclaration-after-statement -Wstrict-prototypes
```
Here is the build log.
```
In file included from /usr/include/python3.12/Python.h:44,
                 from /home/paran/uftrace/utils/script-python.h:20,
                 from /home/paran/uftrace/utils/script.h:13,
                 from /home/paran/uftrace/utils/script-luajit.c:9:
/usr/include/python3.12/object.h: In function ‘Py_SIZE’:
/usr/include/python3.12/object.h:233:5: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  233 |     PyVarObject *var_ob = _PyVarObject_CAST(ob);
      |     ^~~~~~~~~~~
In file included from /usr/include/python3.12/Python.h:53:
/usr/include/python3.12/cpython/longintrepr.h: In function ‘_PyLong_CompactValue’:
/usr/include/python3.12/cpython/longintrepr.h:121:5: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  121 |     Py_ssize_t sign = 1 - (op->long_value.lv_tag & _PyLong_SIGN_MASK);
      |     ^~~~~~~~~~
  CC       utils/script-python.o
In file included from /usr/include/python3.12/Python.h:44,
                 from /home/paran/uftrace/utils/script-python.h:20,
                 from /home/paran/uftrace/utils/script-python.c:15:
[...]

PR #1935, opened 2024-06-19
build: Remove -Wdeclaration-after-statement flag

yskelg:
Remove ISO C90 forbids mixed declarations and code in C warning flag. #1934

Signed-off-by: Yunseong Kim <yskelg@gmail.com>


honggyukim:
Thanks but we would like to keep avoiding from mixed declarations in uftrace.

Issue #1316, opened 2021-08-23
sub-command specific help


akshit-sharma:
It will be useful to have the help specific to use case.

Existing Behavior:
``` uftrace --help ``` and ``` uftrace replay --help ``` gives the same help prompt.

New Behavior:
``` uftrace --help ``` can give the help for entire application.
``` uftrace replay --help ``` can give the  help for uftrace replay only.

Reason:
1. This will help narrow down the options available for subprograms (uftrace reply) as compared to uftrace.
2. Removes unnecessary options from help section. It will avoid showing the options which are not compatible with the subprogram.
3. User can see all the necessary options. Without this, the only way for user to know which configuration is used by going through [wiki\Tutorial](https://github.com/namhyung/uftrace/wiki/Tutorial) but it may also not have all valid configurations. Example: -F helps to focus on particular function. If user is profiling a specific function, with record and report, can he/she use it with both record and report or only with record or only with report. This will also help user to look at help section as oppose to seeing example command line on internet.

Motivation example:
``` git --help ``` v/s ``` git pull --help ```

**Comment by namhyung** [comment]
Thanks for raising this issue.  I totally agree with you.

**Comment by honggyukim** [comment]
We can redirect subcommand help to man pages just like `git` and `perf` does.

**Comment by kangtegong** [comment]
I think I can solve this one, if @namhyung and @honggyukim's ok :)

**Comment by RndelQndel** [comment]
Is it valid Issue?
I made look like this.
![image](https://github.com/namhyung/uftrace/assets/44370237/127b89e3-9d89-45fd-8559-44bb8fab0433)

**Comment by kangtegong** [comment]
hi @RndelQndel, I guess it'll be better if it prints
available options (and explanation) for specific command.

For example,

```
$ uftrace replay --help
 /* prints options for replay ... */
 /* prints options for replay ... */
 /* prints options for replay ... */ 

$ uftrace recv --help
 /* prints options for recv ... */
 /* prints options for recv ... */
 /* prints options for recv ... */ 
```

I once worked on this issue but I lost the work for some reason.🥲
If you want to do this work, you can solve it. If you don't want to, I'll wrap this up.

honggyukim:
Hi @ParkSeungHyeok,

It's good to have subprogram specific help but I think it will require many changes so I cannot recommend you to work on this especially as a new contributor.

If you want to continue to work on this, then try to understand the option system in `perf` and `git` and investigate how they implemented it before trying to jump on this work.

[...]

PR #1946, opened 2024-08-14
uftrace: Add sub-command specific help

ParkSeungHyeok:
uftrace: Add sub-command specific help

i redirect subcommand help to man pages just like git and perf do.
```c
	case -3:
+		if (opts.mode) {
+			execlp("man", "man", "uftrace", argv[1], (char *)NULL);
+			abort();
+		}
		if (opts.use_pager)
			start_pager(setup_pager());
		pr_out(uftrace_usage);
		pr_out(uftrace_help);
		wait_for_pager();
		ret = 0;
		goto cleanup;
	}
```

Fixed: #1316

Signed-off-by: ParkSeungHyeok <tmdgur1324@naver.com>

### DIFF
--- a/uftrace.c
+++ b/uftrace.c
@@ -1383,6 +1383,19 @@ __used static void apply_default_opts(int *argc, char ***argv, struct uftrace_op
 	}
 }
 
+__used static void show_man_page(char *cmd)
+{
+	char *cmdstr = NULL;
+
+	if (cmd)
+		xasprintf(&cmdstr, "uftrace-%s", cmd);
+	else
+		cmdstr = xstrdup("uftrace");
+	execlp("man", "man", cmdstr, (char *)NULL);
+	/* fall through if man command itself is not found */
+	free(cmdstr);
+}
+
 #ifndef UNIT_TEST
 int main(int argc, char *argv[])
 {
@@ -1428,6 +1441,8 @@ int main(int argc, char *argv[])
 		ret = 0;
 		goto cleanup;
 	case -3:
+		if (opts.mode)
+			show_man_page(argv[1]);
 		if (opts.use_pager)
 			start_pager(setup_pager());
 		pr_out(uftrace_usage);


gichoel:
First, thanks for posting the commit that fixes #1316.

However, for a better commit message, you should follow the **50/72 rule[1]**, and the body of this commit message doesn't seem to do that.

Could you please make this change?

And this is my personal opinion based on existing https://github.com/namhyung/uftrace/pull/1701/commits/cefd9be3d463a3e137cd891e8646bac7322bebf0,  https://github.com/namhyung/uftrace/pull/1923/commits/0ac2cef520a40f14f9343f517e3d0acdca22b3b3 PRs, but I think the `**` before `Fixed` is unnecessary :)


[1] https://www.gitkraken.com/learn/git/best-practices/git-commit-message#git-commit-message-structure

honggyukim:
Thanks for the PR @ParkSeungHyeok and thanks for the review @gichoel.
In addition, please remove the diff in the commit message.

uftrace.c
```diff
@@ -1428,11 +1428,16 @@ int main(int argc, char *argv[])
 		ret = 0;
 		goto cleanup;
 	case -3:
-		if (opts.use_pager)
-			start_pager(setup_pager());
-		pr_out(uftrace_usage);
-		pr_out(uftrace_help);
-		wait_for_pager();
+		if (!(opts.mode)) {
```
yskelg, line 1431: Hello @ParkSeungHyeok, We can follow the convention simply with `!opts.mode`.

ParkSeungHyeok:
thank you for review @gichoel , @honggyukim, @yskelg 
I fixed the commit message to follow the 50/72 rule.
and regarding "the diff" (If i understand correctly), there was in the first commit message, but not now.
lastly i also fixed the code to follow the convention wiht `!opts.mode`

yskelg:
Thank you @ParkSeungHyeok, LGTM

uftrace.c
```diff
@@ -1428,11 +1428,16 @@ int main(int argc, char *argv[])
...
+		if (!opts.mode) {
+			if (opts.use_pager)
+				start_pager(setup_pager());
+			pr_out(uftrace_usage);
+			pr_out(uftrace_help);
+			wait_for_pager();
+		}
+		else {
+			execlp("man", "man", "uftrace", argv[1], (char *)NULL);
+		}
```
honggyukim, line 1440: How about making it as follows?
```c
                if (opts.mode) {
                        execlp("man", "man", "uftrace", argv[1], (char *)NULL);
                        abort();
                }
                if (opts.use_pager)
                        start_pager(setup_pager());
                pr_out(uftrace_usage);
                pr_out(uftrace_help);
                wait_for_pager();
                ret = 0;
                goto cleanup;
```

honggyukim (commented):
Thanks for the update. Your name is currently written as `ParkSeungHyeok`, but it is better to be changed because I don't think you write your name without having spaces when you write it.

Maybe one of the following is better.
1. `Park SeungHyeok`
2. `SeungHyeok Park`
3. `Seunghyeok Park`

You can choose which one to use. Please make sure you update your name in both in the commit author name and `Signed-off-by`.

ParkSeungHyeok:
@honggyukim  i agree your recommendation. it looks better.
and i changed my name in the commit author name and Signed-off-by.

honggyukim:
Thanks. One more request is to change the error message when `man uftrace <command>` isn't available in the system. It currently shows as follows.
```
# ./uftrace record --help
No manual entry for uftrace
No manual entry for record
```
Maybe we better handle it just like `perf` as follows.
```
# perf record --help
No manual entry for perf-record
```

ParkSeungHyeok:
@honggyukim Okay i agree then i think how about below code 
i tried to implement that similar to Git or Perf.

```c
+static void show_man_page(char *sub_command)
+{
+	char * page = xmalloc(sizeof("uftrace-"));
+	strcpy(page,"uftrace-");
+	page = strjoin(page,sub_command,"");
+	execlp("man", "man", page, (char *)NULL);
+}

{
	...
	case -3:
+		if (opts.mode) {
+			show_man_page(argv[1]);
+			abort();
+		}
		if (opts.use_pager)
			start_pager(setup_pager());
		pr_out(uftrace_usage);
		pr_out(uftrace_help);
		wait_for_pager();
		ret = 0;
		goto cleanup;
	}
	...
}
```

i not sure if it is okay to make new function
so here's an alternative approach:
```c
{
	...
+	char * page = xmalloc(sizeof("uftrace-"));
	...
	case -3:
+		if (opts.mode) {
+			strcpy(page,"uftrace-");
+			page = strjoin(page,sub_command,"");
+			execlp("man", "man", page, (char *)NULL);
+			abort();
+		}
		if (opts.use_pager)
			start_pager(setup_pager());
		pr_out(uftrace_usage);
		pr_out(uftrace_help);
		wait_for_pager();
		ret = 0;
		goto cleanup;
	}
	...
}
```
how about them?

uftrace.c
```diff
@@ -1428,6 +1428,10 @@ int main(int argc, char *argv[])
 		ret = 0;
 		goto cleanup;
 	case -3:
+		if (opts.mode) {
+			execlp("man", "man", "uftrace", argv[1], (char *)NULL);
+			abort();
```
namhyung, line 1433: I don't think we need `abort()` here and maybe it's better to call `exit(1)` instead.  But we used to show the full message for the `--help` option, so it'd be natural to fall back to the old behavior if no manual pages are available.

namhyung (commented):
Thanks for the contribution!

uftrace.c
honggyukim: `abort()` here is to show this is unreachable. It doesn't hit the `abort()` even if `No manual entry for ...` error message is shown.

But it may be reachable if `man` command itself is not found in the system, which is very unlikely. Maybe the following is better to cover such case as well.
```c
        case -3:
                if (opts.mode) {
                        execlp("man", "man2", "uftrace", argv[1], (char *)NULL);
                        /* fall through if man command itself is not found */
                }
                if (opts.use_pager)
                        start_pager(setup_pager());
                pr_out(uftrace_usage);
                pr_out(uftrace_help);
                wait_for_pager();
                ret = 0;
                goto cleanup;
```

honggyukim: > But we used to show the full message for the --help option, so it'd be natural to fall back to the old behavior if no manual pages are available.

Once the `man` command is executed by `execlp`, then there is no way to go back to the fallback routine. If that is really needed, then we should check whether the proper man page can be found before calling `execlp`. But I think we don't have to make it too much complicated.

namhyung: Oh.. ok.  But sometimes `man` might not be installed.  I admit it's a rare condition. :)

honggyukim:
Considering all our conversation, we can think about this.
```c
static void show_man_page(char *cmd)
{
        char *page = NULL;

        if (cmd)
                xasprintf(&page, "uftrace-%s", cmd);
        else
                page = xstrdup("uftrace");
        execlp("man", "man", page, (char *)NULL);
        /* fall through if man command itself is not found */
        free(page);
}
    ...
{
            ...
        switch (parse_options(argc, argv, &opts)) {
            ...
        case -3:
                if (opts.mode)
                        show_man_page(argv[1]);
                if (opts.use_pager)
                        start_pager(setup_pager());
                pr_out(uftrace_usage);
                pr_out(uftrace_help);
                wait_for_pager();
                ret = 0;
                goto cleanup;
        }
        ...
}
```

ParkSeungHyeok:
thank you for review @namhyung @honggyukim
i reflected your review

but i wonder how to reach below code
```c
        else
                cmdstr = xstrdup("uftrace");
```
is it for another case later?

honggyukim:
My mistake. `if (cmdstr)` should be `if (cmd)`.

honggyukim:
You should put `show_man_page` inside `#ifndef UNIT_TEST` right above main function to avoid the following unittest error.
```
/home/runner/work/uftrace/uftrace/uftrace.c:1386:13: error: ‘show_man_page’ defined but not used [-Werror=unused-function]
 1386 | static void show_man_page(char *cmd)
      |             ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
```
https://github.com/namhyung/uftrace/actions/runs/10432184123/job/28892638730?pr=1946

uftrace.c
```diff
@@ -1428,6 +1441,9 @@ int main(int argc, char *argv[])
 		ret = 0;
 		goto cleanup;
 	case -3:
+		if (opts.mode) {
+			show_man_page(argv[1]);
+		}
```
honggyukim, line 1446: You don't need curly braces for a single statement.

honggyukim:
Looks good now. The last request is please read other commit message and make it similar.

I'm typing in my phone so can't tell you the details but please read other commit messages more to learn the convention.

ParkSeungHyeok:
thank you for your detailed review @honggyukim 
i missed  `if (cmdstr)` too
i changed that to `if(cmd)`
and i added `__used` front of the function to avoid unittest error like `apply_default_opts()` funciton

but although it is if(cmd), i still wonder how to reach below code
```c
        else
                cmdstr = xstrdup("uftrace");
```
i mean...
`show_man_page` function is only called as `show_man_page(argv[1])`
and when `show_man_page(argv[1])` is called, `argv[1]` always isn't `NULL` because of `if(opts.mode)`

and...
i will change commit message after i read other commit message

honggyukim:
> thank you for your detailed review @honggyukim

Thanks for your work and effort as well!

> but although it is if(cmd), i still wonder how to reach below code
> ```
>         else cmdstr = xstrdup("uftrace");
> ```
> i mean...
> show_man_page function is only called as show_man_page(argv[1])

That is the caller dependent point of view. When you write a function, you better handle all the possible cases regardless of the input argument.

> i will change commit message after i read other commit message

I would write the commit message as follows.
```
uftrace: Add sub-command specific help

This patch adds sub-command specific help by redirecting the help
message to their man pages of each command just like git and perf do.

The simple usage is as follows.

  $ uftrace [COMMAND] --help

Then you will see the man page of given `[COMMAND]` such as record,
replay and so on.

Fixed: #1316

Signed-off-by: Seunghyeok Park <tmdgur1324@naver.com>
```

honggyukim:
It's just a nitpick but please begin a sentence in capital letter unless you have a strong reason for that.

You can think the commit message like sending an email to someone.

ParkSeungHyeok:
thank you @honggyukim 
i reflected your review

honggyukim (approved):
LGTM! Thanks for your persistence.

PR #1947, opened 2024-08-14
misc: Add checkout-pr.sh script for easier checkout

honggyukim:
It takes some time to get the patches for review but it'd be useful to have a script for that.

The script accepts only a parameter to get the pull request ID. It fetches from the source repo and checkouts to the fetched head.

### DIFF
--- a/misc/checkout-pr.sh
+++ b/misc/checkout-pr.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+if [ $# -ne "1" ]; then
+	echo "usage: $0 <PR number>"
+	exit 1
+fi
+
+if ! command -v curl > /dev/null 2>&1 || ! command -v jq > /dev/null 2>&1; then
+	echo "You need both 'curl' and 'jq' to run this script."
+	exit 1
+fi
+
+pr=$1
+pr_json="pr.json"
+
+curl -o $pr_json https://api.github.com/repos/namhyung/uftrace/pulls/$pr
+repo=$(jq -r '.head.repo.html_url' $pr_json)
+refspec=$(jq -r '.head.ref' $pr_json)
+git fetch $repo $refspec && git checkout -B pull/$pr FETCH_HEAD


misc/checkout-pr.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ ! -x "$(command -v curl)" ]; then
+	echo "You need curl to run this script."
+	exit 1
+fi
+
+pr=$1
+pr_json="pr.json"
+
+curl -o $pr_json https://api.github.com/repos/namhyung/uftrace/pulls/$pr
+repo=$(jq -r '.head.repo.html_url' $pr_json)
```
namhyung, line 17: What is `jq`?

namhyung (commented):
Don't we have `misc/wget-pr.sh` already?

misc/checkout-pr.sh
honggyukim: https://jqlang.github.io/jq

honggyukim (commented):
> Don't we have `misc/wget-pr.sh` already?

`wget-pr.sh` just downloads `.patch` file, but this fetches the remote branch then checkout to it. If the PR is a bit outdated then the downloaded `.patch` file can make conflicts.

Rather than this, we can just put the PR number then it will create a branch `pull/<pr>`, then checkout to it and we can easily see if the branch needs rebase or not.

The usage to checkout to #1756 is as follows.
```
$ ./misc/checkout-pr.sh 1756
```

honggyukim:
This might be not needed if we're working on already registered remote repos but I felt it's tedious to fetch and checkout some new contributors' branch. This script makes the workflow much easier.

misc/checkout-pr.sh
honggyukim: You can easily understand by running the following command. It's to get the PR sender's uftrace repo.
```
$ curl -s https://api.github.com/repos/namhyung/uftrace/pulls/1947 | jq -r '.head.repo.html_url'
https://github.com/honggyukim/uftrace
```

honggyukim: We can also get the branch name as follows.
```
$ curl -s https://api.github.com/repos/namhyung/uftrace/pulls/1947 | jq -r '.head.ref'
check/checkout-pr.sh
```

namhyung: That's cool.  Is it ok to assume if it's installed on all modern distros?

namhyung (commented):
Ok, I like you used `git checkout -B` to create or update the branch.

misc/checkout-pr.sh
honggyukim: Sorry for the late answer but I guess yes.

honggyukim:
@namhyung I've just update it so please have a look again. Thanks.

namhyung (approved):
LGTM

Issue #1869, opened 2023-12-28
gcc -Wall -Werror: libelf: [ OFF ] even when libelf-devel is installed


vt-alt:
JFYI:
In ALT we compiling with `CFLAGS='-pipe -frecord-gcc-switches -Wall -g -O2 '` by default (but no `-Werror`). With this libelf tests always negative even when libelf-devel is installed.

With this patch
```patch
diff --git a/check-deps/Makefile b/check-deps/Makefile
index 22e33f34..7887b436 100644
--- a/check-deps/Makefile
+++ b/check-deps/Makefile
@@ -58,7 +58,7 @@ LDFLAGS_have_libtraceevent = $(shell pkg-config --libs  libtraceevent 2> /dev/nu
 check-build: $(CHECK_LIST)

 $(CHECK_LIST): %: __%.c $(CHKDIR)/check-tstamp
-        @$(CC) $(CHECK_CFLAGS) -o $(CHKDIR)/$@  $< $(CHECK_LDFLAGS) > /dev/null 2>&1
+        set -x; $(CC) $(CHECK_CFLAGS) -o $(CHKDIR)/$@  $< $(CHECK_LDFLAGS)

 $(CHKDIR)/check-tstamp: PHONY
         @mkdir -p $(dir $@)
```
we see
```
+ gcc -pipe -frecord-gcc-switches -Wall -g -O2 -Werror -o /usr/src/RPM/BUILD/uftrace-0.14/check-deps/have_libelf __have_libelf.c -lelf
__have_libelf.c: In function 'main':
__have_libelf.c:6:19: error: unused variable 'ehdr' [-Werror=unused-variable]
    6 |         GElf_Ehdr ehdr;
      |                   ^~~~
cc1: all warnings being treated as errors
```
In gcc 13.2.1` -Wunused-variable` is included in `-Wall`:
```
$ gcc --help=warning -Q | grep unused-var
  -Wunused-variable                     [disabled]
$ gcc -Wall --help=warning -Q | grep unused-var
  -Wunused-variable                     [enabled]
```
I easily (except the time to find this) workaround this by adding `-Wno-error=unused-variable` to `CFLAGS`.

namhyung:
Ok, I think we can add "unused" attribute to those variables to prevent compiler warnings.

marc-delgado-ferreres:
[...]

PR #1959, opened 2024-09-02
check-deps:changed gelf function in __have_libelf.c

JangSoJin:
The currently reflected __have_libelf.c may cause errors during compilation if unused variables(GElf_Ehdr ehdr) are declared.

so i would like to modify it as follows :

```C
#include <gelf.h>
#include <libelf.h>

int main(void)
{
      Elf *elf;

      elf_version(EV_CURRENT);

      /* check that the gelf function */
      elf = elf_begin(0, ELF_C_READ, NULL);
      gelf_checksum(elf);
      elf_end(elf);

      return 0;
}
```

Thanks,
Fixed : #1869 (gcc -Wall -Werror: libelf: [ OFF ] even when libelf-devel is installed)


vt-alt:
I only wonder why you break formatting of the original file from a tab to 6 spaces.

JangSoJin:
oh, okay thanks @vt-alt 
i forgot to check format in the procedure, sorry. 
i’ll fix and push again.

JangSoJin:
[Re-upload] 

The currently reflected __have_libelf.c may cause errors during compilation if unused variables(GElf_Ehdr ehdr) are declared.
so i would like to modify it as follows :

```C
#include <gelf.h>
#include <libelf.h>

int main(void)
{
      Elf *elf;

      elf_version(EV_CURRENT);

      /* check that the gelf function */
      elf = elf_begin(0, ELF_C_READ, (Elf *)0);
      gelf_checksum(elf);
      elf_end(elf);

      return 0;
}

```

FYI, when i tried `elf_begin(0, ELF_C_READ, NULL)`, the result was incorrect, so i changed `elf_begin(0, ELF_C_READ, (Elf *)0)`.
(i don't understand what the difference is between,,)
※ if it needs to be modified, please let me know.

Sorry for the inconvenience!
Thanks,
Fixed : #1869 (gcc -Wall -Werror: libelf: [ OFF ] even when libelf-devel is installed)

gichoel:
First, Thank you for your efforts to resolve this issue.

However, There are a few rules to follow to write a good readable commit message, and there are three things you should fix in your commit message:
  1. The first letter of the title, except for the 'type' part, should be capitalized.
  2. The title should be 50 characters or less, but currently exceeds 50 characters, including spaces.
  3. The body should also be 72 characters or less, but there are lines of body that exceed 72 characters.

Could you please make this change?

JangSoJin:
Thanks, @gichoel 
i made the changes based on your feedback, thanks for the support!

namhyung (approved):
LGTM

PR #1961, opened 2024-09-03
doc/ko: Add missing options in uftrace-live.md

EmpChan:
These four missing options are translated and added to 'uftrace-live.md'

1. public options, --trace = `STATE`
2. LIVE options, -p *PID*, \--pid=*PID*
3. RECORD config options, -g, \--agent
4. REPLAY options, --no-args


gichoel:
Hello, thank you for contributing to uftrace.

However, before merging that PR, we need to address two of the three checkpoints below:
  1. DCO failed
  2. `pre-commit-check` test failed

DCO failure seems to be due to the email address not being set in the local git environment. Please follow the command below in the uftrace directory.
```
$ git config --global user.email "hss598876@gmail.com"

$ git commit -s --amend
$ git push --force-with-lease
```

Issue #938, opened 2019-10-13
heap-buffer overflow in read_taskinfo()


gy741:
Hello,

I found heap-buffer overflow bug.

If there is a large amount of data(`,`) in the `taskinfo:tids` of the `info` file, it falls into an infinite 
loop.

https://github.com/namhyung/uftrace/blob/2d6c907dedaecb37fb73b9a42f97020842c5b6d4/cmds/info.c#L560-L569

PoC: 
```
cat uftrace.data/info

taskinfo:tids=,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```

Crash info:
```
$ uftrace info
uftrace: malloc.c:2401: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' fa

$bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f5b6b557801 in __GI_abort () at abort.c:79
#2  0x00007f5b6b5aaa91 in __malloc_assert (file=<optimized out>, function=<optimized out>, line=<optimized out>, assertion=<optimized out>) at malloc.c:298
#3  sysmalloc (nb=nb@entry=112, av=av@entry=0x7f5b6b902c40 <main_arena>) at malloc.c:2398
#4  0x00007f5b6b5abff0 in _int_malloc (av=av@entry=0x7f5b6b902c40 <main_arena>, bytes=bytes@entry=100) at malloc.c:4125
#5  0x00007f5b6b5ae0fc in __GI___libc_malloc (bytes=100) at malloc.c:3057
#6  0x00007f5b6b59f525 in _IO_vasprintf (result_ptr=0x7ffe0645e6d0, format=0x7f5b6b6ce7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", args=args@entry=0x7ffe0645e5c0)
    at vasprintf.c:47
[...]

PR #1963, opened 2024-09-07
info : fix heap-buffer overflow in read_taskinfo()

ParkSeungHyeok:
It fixes a heap-buffer overflow issue caused by data in `taskinfo:tids`. The root cause was insufficient exception handling for `tid` values.

This patch added exception handling to manage any value written to tids_str

- The exception handling ensures that:
  1. `tids_str` has the same number of values as `info->nr_tid`.
  2. Each `tid` value in `tids_str` is properly separated by a comma.

Fixed: #938


namhyung (commented):
Looks good.  The unittest code is in the tests/unittest.[ch] files.  But I'm not sure if it's easy to make unittests for the info code.

namhyung (approved):
LGTM

PR #1806, opened 2023-08-20
doc/ko: Add missing options in man pages

jeonghwanin:
The missing options -Z *SIZE*, \--size-filter=*SIZE*,
-L LOCATION' '--loc-filter=LOCATION' and '--with-syms=DIR'
are translated and added to man pages in Korean.

Signed-off-by: Jeonghwan In <jhpc0128@naver.com>

PR #1966, opened 2024-09-09
doc/ko Translate uftrace-script.md in Korean

CTRLHyeon:
added missed contents in document
in line 33-34, 92, 246


gichoel:
First of all, thank you for contributing to improving the Korean documentation of uftrace.

However, I think there are a few things that need to be fixed in this PR, and they are as follows:

1. Please add `:` after `doc/ko` in the commit title.
We need a delimiter to separate the title and the type.

2. Please be more specific in the commit description.
One of the rules for writing a good commit message is to describe what you changed and why you changed it. This way, other people can know what you changed without having to check the code directly using only the `git log` command.

If you need examples of writing commit messages, see https://github.com/namhyung/uftrace/commit/c454d1e11022cf4b22cb7a535cc9b8552a2fce4c, https://github.com/namhyung/uftrace/pull/1785/commits/db895cc3e90002646d852976d95412ed0ea3cc98, https://github.com/namhyung/uftrace/pull/1806/commits/d69b47102640e5bfa4d775a8192c659fa8de7af2.

Finally, if you run into any problems while making changes or don't know what to do, leave a comment and we'll try to help :)

CTRLHyeon:
Thanks for advice. I edited thing you taught. Please review it again.

gichoel:
First of all, I think you should remove line numbers from commits, because if other contributors contribute to the documentation later, the line numbers will change, making it impossible to compare.

Also, I would like the body to be different from the title, with a little more description of what was done and why, like the commit I attached as an example.

Here's what I thought the commit message would look like.
```
doc/ko: Translate uftrace-script.md in Korean

Add `--no-args` option, which is present in the English
`uftrace-script.md` but not in the Korean `uftrace-script.md`, 
and fix some translations.
```

But I'm not sure about this, so We'll wait for an admin to review it :)

CTRLHyeon:
I appreciate for your kind answer. I never thought about further contributes. I pondered your advises and thought it would be better to edit as you told me. your answers helped and taught me a lot. Thank you!

Issue #831, opened 2019-07-10
logfile enhancement


honggyukim:
I would like to create the logfile inside `uftrace.data` directory by default so that external users can send it us simply a single directory. And the name doesn't have to be different all the time.

The current implementation requires explicitly remove the previous logfile because it's open mode is `append`, but we don't have to remove it anymore if it's create inside `uftrace.data`.
The previous discussion is in #707.

In addition, I think we don't have to provide logfile name explicitly. Why don't we just fix the logfile name as `uftrace.data/logfile.txt`? It looks simpler.

namhyung:
OK, I agree with you that having the log file in the data directory (with a fixed name) is good.  But the problem is that we need to consider backward compatibility of the option because it used to have an argument for the file name.  Maybe it'd be nice to use the given name in the data directory.

honggyukim:
I'm okay with it. We should also think how to handle the case when `-H` option is used.

junhoryu:
I'd like to take it

honggyukim:
@junhoryu Thanks! Please take it and let me know if you have any problems.

honggyukim:
@junhoryu Do you still have a plan to work on this?

MichelleJin12:
### How to use --logfile option
`uftrace record --logfile="filename" -v[vv] "executable file"`
-v, -vv, and -vvv are debug levels.
related: https://github.com/namhyung/uftrace/wiki/Debug-message-guidelines

### How to use --host option
In host, `uftrace recv`
In the machine for tracing, `uftrace record --logfile="filename" -v --host="IP addr" "executable file"`

Currently, the --logfile option stores a logfile in the uftrace execution location. 
The --host option + --logfile option stores the logfile in the uftrace execution location, and the host receives it in uftrace.data.

### TODO
- Set the logfile name `logfile.txt` and save it as default inside the uftrace.data directory. 
- Consider the --host option which sends uftrace.data to the host.
related: https://github.com/namhyung/uftrace/pull/707

hjongkim:
I will take it.

hjongkim:
https://github.com/namhyung/uftrace/commit/c829de3145528d12e04b3db4c12de76fa4e42c85

-H option is deleted. Backwards compatibility for --host option needs to be considered.

dlgus8648:
**Hello** @namhyung @honggyukim 

I am a mentee participating in the uftrace open-source project in OSSCA.

[...]

PR #1971, opened 2024-09-12
Signed-off-by: Rihyeon Kim <dlgus8648@naver.com>

dlgus8648:
uftrace: Improve logfile creation location

Ensure that when using "--logfile=<name>", the logfile is created i
nside the uftrace.data directory.

Additionally, when using both "--host" and "--logfile" options simultaneously 
on the clinet, the logfile in the uftrace.data directory is sent to the host.

Fixed: #831


gichoel:
Thanks for your contribution to fix this issue.

However, we can see that the title part has a signature in it, and this line needs to be removed so that the original intended title can be uploaded.

In addition, we can see that the title is `uftrace: Improve logfile creation location`, but this work is more like moving rather than improving the creation location, so how about changing it to `uftrace: Move logfile path to uftrace.data`?

dlgus8648:
It seems that the title recommended by gicheol is more appropriate.  
I have updated the title.

Thank you!

uftrace.c
```diff
@@ -1474,7 +1474,16 @@ int main(int argc, char *argv[])
 		debug = 1;
 
 	if (opts.logfile) {
-		logfp = fopen(opts.logfile, "a");
+		char *logfile_path = NULL;
+		if (create_directory(opts.dirname) < 0) {
```
namhyung, line 1478: I'm not sure about this.  Creating directory includes renaming the existing directory.  This can be a problem either you create a new one for record or dealing with an existing one for replay.

dlgus8648:
I understood @namhyung 's comment to mean that calling the create_directory() function at that point could potentially cause issues elsewhere. 
Therefore, I moved the logic for creating the logfile in the uftrace.data directory to the point when the uftrace terminates.

Issue #1500, opened 2022-07-30
create a real call graph that can have cycles in --graphviz


honggyukim:
`uftrace graph` actually doesn't show a call graph, but just a call tree output. For example, `t-fibonacci` example's output is as follows.
```
$ gcc -pg -o t-fibonacci tests/s-fibonacci.c

$ uftrace record t-fibonacci

$ uftrace graph
# Function Call Graph for 't-fibonacci' (session: 7701e718f823010e)
========== FUNCTION CALL GRAPH ==========
# TOTAL TIME   FUNCTION
   20.354 us : (1) t-fibonacci
    2.199 us :  +-(1) __monstartup
             :  |
    1.129 us :  +-(1) __cxa_atexit
             :  |
   17.026 us :  +-(1) main
   16.499 us :    (1) fib
   15.859 us :    (2) fib
   14.887 us :    (4) fib
   12.782 us :    (8) fib
    8.362 us :    (14) fib
    2.976 us :    (10) fib
    0.390 us :    (2) fib
```
As the output shows there are multiple nodes for `fib` function. It's because there are multiple backtrace or call paths of each `fib` node.

The `--graphviz` output also uses the same logic when building a graph, precisely a tree. So its output doesn't show the multiple nodes for the same function, but it draws multiple edges for the same calls.

The `--graphviz` output of `t-fibonacci` is as follows.
```dot
$ gcc -pg -o t-fibonacci tests/s-fibonacci.c
$ uftrace record t-fibonacci
$ uftrace dump --graphviz
      ...
digraph "t-fibonacci" {
    "__monstartup" [xlabel = "1"]
    "__cxa_atexit" [xlabel = "1"]
    "main" [xlabel = "1"]
    "main" -> "fib" [xlabel = "1"]
    "fib" -> "fib" [xlabel = "2"]
    "fib" -> "fib" [xlabel = "4"]
    "fib" -> "fib" [xlabel = "8"]
    "fib" -> "fib" [xlabel = "14"]
    "fib" -> "fib" [xlabel = "10"]
    "fib" -> "fib" [xlabel = "2"]
}
```
Its image shows too many edges from `fib` to `fib` as follows.
<img width="580" alt="image" src="https://user-images.githubusercontent.com/19642328/181995570-27e1853d-72bf-4d03-b52f-3ad6bad96889.png">

But it would be much better to make it as follows.
```dot
digraph "t-fibonacci" {
    "__monstartup" [xlabel = "1"]
[...]

PR #1978, opened 2024-10-29
dump: Handle cycles in graphviz

deepseafishy:
These commits include skipping the recursive calls during the dump of graphviz, and correctly adding number of calls of the skipped entries to the initial recursive call.

Fixed: #1500


namhyung (commented):
Hello, thanks for your contribution.

First, you don't need to create a new PR for code changes.  Just keep the PR and force-push changes to your branch.

Second, Please remove commits to add NULL checks and squash them to the original changes.

Third, I think you modified logic in the general code which will affect other graph codes that might not want to merge recursive calls.  If so, you may want to pass an option (or set it globally somewhere) to control the behavior.

PR #1979, opened 2024-11-02
record: Fix crash when the root is not existed

zyxeeker:
This patch to fix crash when the root is not existed, due to some embed devices maybe not have it.
source:
```c
static void unlink_shmem_list(void)
{
        ...
	list_for_each_entry_safe(sl, tmp, &shmem_need_unlink, list) {
		...
		struct dirent **shmem_bufs;
		sscanf(sl->id, "/uftrace-%[^-]-%*d-%*d", shmem_session);
		...
		free(shmem_bufs);
                ...
	}
}
```
My SOC do not have `/dev/shm/` and `shmem_bufs`'s value has not initialized, so `scandir()` not flush `shmem_bufs` and it has bad address, then bad thing happened:
```shell
/userdata # ./uftrace record -vv ./prog_test
...
uftrace: flushing /uftrace-bd6de237f8a0b25b-1383-000
uftrace: unlink for session: bd6de237f8a0b25b
munmap_chunk(): invalid pointer
Aborted (core dumped)
```
So I add `access()` to fix it:
```c
static void unlink_shmem_list(void)
{
        ...
	list_for_each_entry_safe(sl, tmp, &shmem_need_unlink, list) {
		...
		/* check the root is existed (due to some embed devices maybe not have it) */
		if (access(uftrace_shmem_root(), F_OK) == 0) {
			num = scandir(uftrace_shmem_root(), &shmem_bufs, filter_shmem, alphasort);
			for (i = 0; i < num; i++) {
				...
			}

			free(shmem_bufs);
		}
		else {
			pr_dbg2("access %s failed, err: %s\n", uftrace_shmem_root(), strerror(errno));
		}
                ...
	}
}
```

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -930,6 +930,13 @@ static int filter_shmem(const struct dirent *de)
 static void unlink_shmem_list(void)
 {
 	struct shmem_list *sl, *tmp;
+	char *shmem_root = (char *)uftrace_shmem_root();
+
+	/* check the root is existed (due to some embed devices maybe not have it) */
+	if (access(shmem_root, F_OK) != 0) {
+		shmem_root = NULL;
+		pr_warn("access shmem root failed and will ignore it, err: %s\n", strerror(errno));
+	}
 
 	/* unlink shmem list (not used anymore) */
 	list_for_each_entry_safe(sl, tmp, &shmem_need_unlink, list) {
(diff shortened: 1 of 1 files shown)


namhyung (commented):
Thanks for your fix.  It looks good but we have 3 calls to `uftrace_shmem_root()` now.  I think it's better to add a local variable for it.

zyxeeker:
Thanks your reply!

> Thanks for your fix. It looks good but we have 3 calls to `uftrace_shmem_root()` now. I think it's better to add a local variable for it.

The suggestion is good and I will check it again, so I will convert the issue to draft.


> Also I'm curious if your system doesn't have a temp directory at all.

I have it, but many embed devices will use static `/dev` when RAM is limited, so maybe not have `shm`.
I will change `pr_dbg2` to `pr_warn` to remind others.

namhyung (commented):
I think you can saved the result of `uftrace_shmem_root()` instead.

But I'm curious if it works on a system without shmem..

zyxeeker:
>I think you can saved the result of `uftrace_shmem_root()` instead.

Thanks, I will modify it 

>But I'm curious if it works on a system without shmem..

Sorry for my bad explain, I mean mostly embedded devices will use static `/dev` and maybe not have `shm` folder.
Although create `shm` folder will solve this problem, but I think that\`s not a best way.

namhyung (approved):
LGTM

PR #1984, opened 2024-12-12
Release v0.17

namhyung:
I'd like to release a new version (v0.17).  Please check out the current version and run the test on your environment.

This is a small release including the following changes.

* global variable watchpoint support
* argument/return value and srcline support for dlopen libraries
* show man page of current command when -h option is used

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-VERSION := 0.16
+VERSION := 0.17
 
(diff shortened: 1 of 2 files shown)


honggyukim:
I see some test failures but those come from test scripts themselves. Testing this PR looks good to me so let's make a release. Thanks!

namhyung:
Thanks for checking.  Time for a new release! :)

PR #1988, opened 2025-01-08
build: Support MemorySanitizer

yskzalloc:
Uninitialized values occur when stack- or heap-allocated memory is read before it is written. LLVM's MSan detects cases
* Uninitialized value was used in a conditional branch.
* Uninitialized pointer was used for memory accesses.
* Uninitialized value was passed or returned from a function call.
* Uninitialized data was passed into some libc calls.
```sh
    $ cd ~/uftrace/
    $ ./configure CC=clang

    $ make MSAN=1

    $ cd tests/
    $ ./runtest.py
```


Makefile
```diff
@@ -135,18 +135,16 @@ ifeq ($(COVERAGE), 1)
...
 
 ifeq ($(ASAN), 1)
-  ASAN_CFLAGS       := -O0 -g -fsanitize=$(DEFAULT_SANITIZERS)
-  UFTRACE_CFLAGS    += $(ASAN_CFLAGS)
-  DEMANGLER_CFLAGS  += $(ASAN_CFLAGS)
-  SYMBOLS_CFLAGS    += $(ASAN_CFLAGS)
-  DBGINFO_CFLAGS    += $(ASAN_CFLAGS)
-  TRACEEVENT_CFLAGS += $(ASAN_CFLAGS)
-  TEST_CFLAGS       += $(ASAN_CFLAGS)
+  SAN=$(DEFAULT_SANITIZERS)
```
namhyung, line 138: It could be `SAN=all`.

Makefile
```diff
@@ -135,18 +135,16 @@ ifeq ($(COVERAGE), 1)
...
+else ifeq ($(MSAN), 1)
+  SAN=memory
 endif
 
 ifneq ($(SAN),)
   ifeq ($(SAN), all)
     SAN_CFLAGS := -O0 -g -fsanitize=$(DEFAULT_SANITIZERS)
+  else ifeq ($(MSAN), 1)
+    SAN_CFLAGS := -O0 -g -fsanitize=$(SAN) -fsanitize-memory-track-origins
```
namhyung, line 147: It seems it should be used when memory sanitizer is requested (even with others).  What about this?

```
else ifneq ($(findstring memory, $(SAN)), )
  SAN_CFLAGS := -O0 -g -fsanitize=$(SAN) -fsanitize-memory-track-origins
```

Makefile
yskzalloc: Thank you @namhyung I applied your comment.

Makefile
yskzalloc: Thanks, I applied your suggestion.

namhyung (commented):
Thanks for the update.

The changes look good but can you please update the commit message?  It should have "build:" prefix in the subject line.  And I don't think it needs instructions to build clang.  (Btw why did you build new clang?)  Just leave the uftrace part and I prefer having code examples with 4 space indentation in the commit message.

yskzalloc:
Thank you @namhyung for the review. I updated the commit message and removed the unnecessary parts.

namhyung:
I think you accidentally added llvm-cov.md file.  Please remove that.

yskzalloc:
Oops, thanks for the kind words, Namhyung. That was my mistake. I updated it.

namhyung (approved):
LGTM

Issue #1451, opened 2022-07-02
adding more fields in graph command


kangtegong:
Personally, I use `total-avg`, `total-min`, `total-max`, `self-avg`, `self-min`, `self-max` fields in report command is very usefully.

But currently `graph` command only support `total`, `self`, `addr` field. 
I think it would be great if we can add these fields in `graph` commands, too.

honggyukim:
Right, we can also add them to `tui` command as well.

dongwooklee96:
If anyone doesn't care of it, I'll give it a try.

holyshitcode:
Can I give it a shot?

dongwooklee96:
> Can I give it a shot?

you can give it a shot

holyshitcode:
Could you give me some time to analysis this issue? Simultaneously assign it to me. I'm first at this library.

PR #1997, opened 2025-05-26
adding more fields in graph command #1451

holyshitcode:
add total_max,total_min,total_avg,self_avg,self_max,self_min commands to graph.


namhyung (commented):
Thanks for your contribution!  Please take a look at [CONTRIBUTING.md](https://github.com/namhyung/uftrace/blob/master/CONTRIBUTING.md) for basic information.

For patch organization, we don't want to keep the buggy commits in the history.  Please squash commits so that each commit can be compiled and run correctly.  I recommend you to read other commit messages to learn how to write it effectively.

holyshitcode:
I will reopen this after i got your recommendation

namhyung:
> is this the right function on total-avg, self-avg?

I don't understand your question.  Are you asking if the behavior looks correct?  I'm afraid not, the original time should be the same with 'avg' if the call count is 1.

holyshitcode:
> > is this the right function on total-avg, self-avg?
> 
> I don't understand your question. Are you asking if the behavior looks correct? I'm afraid not, the original time should be the same with 'avg' if the call count is 1.

Now I understand it. Should I complete all functions or part of it?

holyshitcode:
I edited total-avg,total-min,total-max,self-avg,self-min,self-max cmds & tui

holyshitcode:
And I override my feature branch to master. So I made for one commit

PR #2000, opened 2025-06-15
record: Add shmem directory to exec permission

honggyukim:
The shmem directory under TMPDIR is used to store named pipe for communication.  Since the named pipe is created under the created 'uftrace' directory, the 'uftrace' directory has to have an exec permission so that named pipe files can be created inside it.

This patch fixes the problem by adding exec permission for 'uftrace' directory creation.

Before:
```
  $ uftrace record --force pwd
  mcount: /data/data/com.termux/files/home/uftrace/libmcount/record.c:81:prepare_shmem_buffer
   ERROR: mmap shmem buffer: Permission denied
```
After:
```
  $ uftrace record --force pwd
  /data/data/com.termux/files/home/uftrace
```

### DIFF
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -833,7 +833,7 @@ static void record_mmap_file(const char *dirname, char *sess_id, int bufsize)
 	struct mcount_shmem_buffer *shmem_buf;
 
 	/* write (append) it to disk */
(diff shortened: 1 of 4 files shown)


honggyukim:
I've just addressed the comments given by copilot review.

PR #2002, opened 2025-06-24
misc: Support Rocky Linux 9 in install-deps parts

junyeong0619:
Rocky Linux is rooted in CentOS, but `yum`
is no longer recommended from version 9.
This change departs it from the CentOS part,
adding a dedicated Rocky Linux section.
It also installs `dnf-plugins-core` and enables
`config-manager` for `capstone-devel` installation.

### DIFF
--- a/misc/install-deps.sh
+++ b/misc/install-deps.sh
@@ -16,6 +16,16 @@ install_packages() {
             dnf install $OPT libtraceevent-devel || true
             exit
             ;;
+        "rocky")
+            rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+            dnf install $OPT dnf-plugins-core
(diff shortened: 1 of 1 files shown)


namhyung:
Looks ok, but can you please briefly explain what plugins-core and config-manager does?

junyeong0619:
> Rocky Linux is rooted in CentOS, but `yum` is no longer recommended from version 9. This change departs it from the CentOS part, adding a dedicated Rocky Linux section. It also installs `dnf-plugins-core` and enables `config-manager` for `capstone-devel` installation.

dnf-plugins-core enables the config-manager command, and config-manager is used to activate the CRB repository where capstone-devel is located, ensuring its successful installation and preventing build issues.

namhyung:
Got it.  Thanks for the explanation!

PR #2006, opened 2025-06-29
tests: Keep testdir only when test fails

honggyukim:
The current runtest.py -k/--keep option makes the test directory in /tmp move into the tests directory for later reference.

But it tends to keep too many test_* directories so it'd be better to keep only the test directories of failed tests.


namhyung:
I think it sometimes needs to keep successful test directories.  To get  the failed tests, you can run it again with options to limit the test cases without this change.

honggyukim:
I agree. But we may need to have a way to opt in and out such cases. I saw many random failures in AArch64 tests.

PR #2008, opened 2025-06-30
filter: add time-based trace filter option

junyeong0619:
Add global time filter to start and stop tracing
based on elapsed time since program start.

To handle running time added mcount_global_timer
and if option is existing every entry points
check the time range.

This filter can be simply used like this

$ uftrace --trace=time~time ./prog

### DIFF
--- a/libmcount/mcount.c
+++ b/libmcount/mcount.c
@@ -47,6 +47,14 @@
  * when no filter is specified in the command line and/or single-thread only.
  */
 
+struct {
+	uint64_t start_ns;
+	uint64_t end_ns;
+	bool enabled;
+} mcount_time_filter = { 0, 0, false };
+
+uint64_t mcount_global_timer;
+
 /* time filter in nsec */
 uint64_t mcount_threshold;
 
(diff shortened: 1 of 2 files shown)


junyeong0619:
Hello,

I have designed an option to implement time-based tracing using the --trace=[start]~[end] argument.

My current approach is as follows:

The time range argument is parsed in mcount_startup and stored in a global mcount_time_filter struct.
Based on the elapsed time measured by a dedicated timer, the global mcount_enabled flag is adjusted accordingly at every function entry point.
However, it seems the actual tracing on/off control is not working as expected. My debug logs confirm that the logic is setting mcount_enabled to false when the current time is outside the specified range, but the uftrace replay output still shows the full trace.

My question is: Is there any internal logic within the uftrace runtime that might be overwriting or taking precedence over the mcount_enabled flag? I would appreciate any guidance on what I might be missing.

Thank you for your time and help!

junyeong0619:
```bash
[root@825badae0e12 c]# uftrace ./s-sort

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

mcount_enabled=true

# DURATION     TID      FUNCTION

 105.875 us [  42579] | __monstartup();

   0.542 us [  42579] | __cxa_atexit();

            [  42579] | main() {

            [  42579] |   foo() {

  48.917 us [  42579] |     loop();

  18.541 us [  42579] |     loop();

  10.750 us [  42579] |     loop();

  92.250 us [  42579] |   } /* foo */

            [  42579] |   foo() {

  10.750 us [  42579] |     loop();

  10.708 us [  42579] |     loop();

  14.750 us [  42579] |     loop();

  46.959 us [  42579] |   } /* foo */

            [  42579] |   bar() {

            [  42579] |     usleep() {

  13.302 ms [  42579] |       /* linux:schedule */

  13.338 ms [  42579] |     } /* usleep */

  13.381 ms [  42579] |   } /* bar */

  13.692 ms [  42579] | } /* main */

[root@825badae0e12 c]# uftrace --trace=10s~ ./s-sort

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

mcount_enabled=false

# DURATION     TID      FUNCTION

 122.917 us [  42584] | __monstartup();

   0.834 us [  42584] | __cxa_atexit();

            [  42584] | main() {

            [  42584] |   foo() {

  51.916 us [  42584] |     loop();

  15.250 us [  42584] |     loop();

  10.792 us [  42584] |     loop();

  89.958 us [  42584] |   } /* foo */

            [  42584] |   foo() {

  10.750 us [  42584] |     loop();

  10.708 us [  42584] |     loop();

  10.708 us [  42584] |     loop();

  36.584 us [  42584] |   } /* foo */

            [  42584] |   bar() {

            [  42584] |     usleep() {

  11.856 ms [  42584] |       /* linux:schedule */

  11.888 ms [  42584] |     } /* usleep */

  11.925 ms [  42584] |   } /* bar */

  12.213 ms [  42584] | } /* main */

[root@825badae0e12 c]#```

junyeong0619:
```c
1609
1610                 /* restore return address of parent */
1611                 if (mcount_auto_recover)
1612                         mcount_auto_restore(mtdp);
1613         }
1614
1615         mcount_entry_filter_record(mtdp, rstack, &tr, regs);
1616         mcount_unguard_recursion(mtdp);
1617         if(mcount_enabled){
1618                 printf("mcount_enabled=true\n");
1619         }else{
1620                 printf("mcount_enabled=false\n");
1621         }
1622         return 0;
1623 }```

above debug message is located in the end of the __mcount_entry();

namhyung:
Sorry for the super late reply.  I think we already have `-r/--time-range` option.  I think it works at replay time because it needs to know the timestamp.  But it can also use elapsed time with `+` prefix.

Please see commits 3a43a133 and b12ffd5a.

junyeong0619:
Thank you for the clarification! I'll look into the existing -r option more carefully and find another area to contribute.

PR #2011, opened 2025-07-12
Add support for dereferencing int* arguments with /ip format

zahra121281:


honggyukim:
Hi @zahra121281, thanks very much for your work!

I can also see that this works fine but it works only for `int*` types. This is a bit difficult to handle all the primitive types, but we need to think about a syntax that covers all the other types.

I think `/ip` is too simple for this. Instead, something like `i*` or `/i32*` might be better. To make it in a more formal syntax, it can be rewritten as follows. (For simplicity, only `ret_spec` is used here, but `arg_spec` can be the same)

```diff
-<ret_spec>    :=  "retval" [ "/" <format> [ <size> ] ]
+<ret_spec>    :=  "retval" [ "/" <format> [ <size> ] [ "*" ] ]
 <format>      :=  "d" | "i" | "u" | "x" | "o" | "s" | "c" | "f" | "S" | "p"
 <size>        :=  "8" | "16" | "32" | "64"
```

@namhyung might have a better idea.

namhyung:
I was a bit hesitant to support pointer dereferences because of potential crashes, but maybe we can leave the risk to the user.  Having said that, I agree with @honggyukim for flexible formats.  Probably we don't want double-pointers and s*.

Also please read [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) for patch submission.

PR #2017, opened 2025-09-03
fixing missing dependencies in Makefile

Meiye-lj:
This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like util/regional.h would not trigger a rebuild of stats.o. The PR fixes this by including them as additional dependencies.


namhyung:
But we don't have util/regional.h and util/stats.c.

PR #2018, opened 2025-09-07
utils: hashmap: Add RISC-V 64-bit support in ARCH64 check

kzall0c:
The hashmap implementation currently defines ARCH64 only for amd64 and aarch64, and now also checks for 64-bit RISC-V by checking __riscv with __riscv_xlen == 64.

This enables uftrace to correctly detect 64-bit RISC-V user programs and apply the appropriate word size.


utils/hashmap.h
```diff
@@ -21,7 +21,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-#if defined(__amd64__) || defined(__aarch64__)
+#if defined(__amd64__) || defined(__aarch64__) || (defined(__riscv) && __riscv_xlen == 64)
```
namhyung: Can we have more general check like below?
```c
#if __SIZEOF_POINTER__ == 8
```

yskzalloc: Thank you Namhyung, for the review! How about this?
`#if __SIZEOF_POINTER__ >= 8`
- https://github.com/llvm/llvm-project/blob/llvmorg-22-init/llvm-libgcc/gcc_s.ver.in#L19

namhyung (approved):
LGTM

PR #1925, opened 2024-05-12
Report: Support max/min timestamp in report output

Cosh-y:
This patch introduces timestamps for special function calls that take the minimum or maximum time.
With this capability, users can identify function calls that exhibit unusual execution times and investigate the underlying reasons.

Example usage:

$ uftrace report -f total-min-ts,total-max-ts
  Total min ts  Total max ts  Function
  ============  ============  ====================
     12.009  h     12.009  h  main
     12.009  h     12.009  h  printf
     12.009  h     12.009  h  __monstartup
     12.009  h     12.009  h  __cxa_atexit

Resolves: #1898

Issue #1898, opened 2024-03-03
Support max/min timestamp in report output


namhyung:
I think it's useful to provide timestamp information in the report so that users can replay at the given time later.  For example, when some function is called a lot, it runs mostly very short but sometimes it takes long.  Then users want to see where it is, and what happened during the call.  So I suggest 4 more output fields - `total-max-ts`, `total-min-ts`, `self-max-ts`, and `self-min-ts`.  They will have a timestamp of function ENTRY for the cases respectively.

`struct uftrace_report_node` has `struct report_time_stat` for total and self time already.  So it's easier to extend it to save timestamp.

honggyukim:
That looks a good idea.

> So I suggest 4 more output fields - total-max-ts, total-min-ts, self-max-ts, and self-min-ts.

But I worry adding many fields as each result of `report -f all` will be too long to fit into a single line. We might have to consider if `-f all` should show the entire field if we want to add many more fields.

namhyung:
Right, but I think `-f all` is too long already and it's not encouraged to use that.  But it shouldn't mean to prevent adding new fields.  I think something like `-f +total-max-ts` is more useful and using `uftrace tui` would be the best as it can change the fields dynamically (with `f` key).

honggyukim:
Yeah, the current `-f all` is already long. How about making `-f` accept pattern matching for simplicity?

namhyung:
Sounds good.

Cosh-y:
Hello @namhyung @honggyukim,

I am an undergraduate student from Beihang University in Beijing, China, currently enrolled in a course focused on open source projects. I believe you both are welcoming to newcomers like me. I appreciate that you've labeled the issue with the "good-first-issue" tag and provided helpful guidance. I'm thankful for that and eager to contribute.

I'd like to start by working on the four additional fields for the report. Given the existing codebase, I believe this task should be manageable.

Additionally, the "too long for display" issue seems interesting and practical. I am also interested in implementing the pattern matching function.

Could I please have the opportunity to contribute to such an excellent project? I would be grateful for the chance to participate.

Thank you.

namhyung:
Hello @Cosh-y!  We always welcome new contributors. :)

Please read the [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) carefully and try to follow the existing convention in the commit history.  I'm looking forward to seeing your PR soon!

Cosh-y:
Hello @namhyung ,

[...]

PR #2032, opened 2026-03-12
Report: Support max/min timestamp in report output

junyeong0619:
This PR is a continuation of the work started by @Cosh-y in #1925 to address the requirements in issue #1898.

Changes & Updates in this PR:
I have addressed the feedback left in the previous PR (#1925) and added a new test case:

* Field Reordering: Moved the new timestamp fields (total-min-ts, total-max-ts, self-min-ts, self-max-ts) after the STDV fields in utils/field.h, utils/report.c, and cmds/tui.c to maintain proper field alignment.
* Naming Convention: Removed the TIME part from the enum names in utils/field.h (e.g., updated to REPORT_F_TOTAL_MIN_TS) as suggested.
* Copilot AI Fixes & Timestamp Logic: Addressed the type mismatch warnings by using PRIu64 for printing timestamps in cmds/tui.c, and resolved the array boundary issues related to NUM_REPORT_FIELD. I also discarded the previous AI suggestion to use "task->timestamp_last" as it caused incorrect results. Instead, I securely kept "task->rstack->time" with a proper NULL check to ensure accurate timestamp calculation.
* New Test Case Added: Added tests/t298_report_field_ts.py and tests/s-ts-report.c to thoroughly verify that the min and max timestamp fields are recorded and evaluated correctly.

Note on Testing & Environment:
I successfully verified the new test case on an x86_64 WSL environment (Rocky Linux 9.7), where all tests passed perfectly and outputted correct timestamps (e.g., 533.xxx).
However, when testing the exact same code on my MacBook Air AArch64 Ubuntu 24.04 LTS VM (via UTM), I noticed anomalous behaviors: the "fpatchable-fun" method occasionally returned NG, and some timestamps underflowed to a massive number (18446744073.7..., near the 64-bit maximum). Since the logic is proven perfectly stable on x86_64, this strongly indicates an underlying architecture-specific issue with uftrace's dynamic patching and timestamp calculation on AArch64, rather than a flaw in this PR.

Please let me know if there are any further adjustments needed.

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -202,10 +202,10 @@ enum tui_mode {
 	TUI_MODE_OTHER,
 };
 
-static char *report_sort_key[] = {
-	OPT_SORT_KEYS, "total_avg", "total_min", "total_max", "self",	    "self_avg",
-	"self_min",    "self_max",  "call",	 "size",      "total_stdv", "self_stdv",
-};
+static char *report_sort_key[] = { OPT_SORT_KEYS,  "total_avg",	   "total_min",	  "total_max",
+				   "self",	   "self_avg",	   "self_min",	  "self_max",
+				   "call",	   "size",	   "total_stdv",  "self_stdv",
+				   "total_min_ts", "total_max_ts", "self_min_ts", "self_max_ts" };
 
 static char *selected_report_sort_key[NUM_REPORT_FIELD];
 
--- a/utils/report.c
+++ b/utils/report.c
@@ -191,6 +191,10 @@ SORT_KEY(self_max, self.max);
 SORT_KEY(self_stdv, self.stdv);
 SORT_KEY(call, call);
 SORT_KEY(size, size);
+SORT_KEY(total_min_ts, total.min_ts);
+SORT_KEY(total_max_ts, total.max_ts);
+SORT_KEY(self_min_ts, self.min_ts);
+SORT_KEY(self_max_ts, self.max_ts);
 
 static int cmp_func(struct uftrace_report_node *a, struct uftrace_report_node *b)
 {
@@ -204,9 +208,11 @@ static struct sort_key sort_func = {
 };
 
 static struct sort_key *all_sort_keys[] = {
-	&sort_total,	&sort_total_avg,  &sort_total_min, &sort_total_max, &sort_self,
-	&sort_self_avg, &sort_self_min,	  &sort_self_max,  &sort_call,	    &sort_func,
-	&sort_size,	&sort_total_stdv, &sort_self_stdv,
+	&sort_total,	   &sort_total_avg,    &sort_total_min,	   &sort_total_max,
+	&sort_self,	   &sort_self_avg,     &sort_self_min,	   &sort_self_max,
+	&sort_call,	   &sort_func,	       &sort_size,	   &sort_total_stdv,
+	&sort_self_stdv,   &sort_total_min_ts, &sort_total_max_ts, &sort_self_min_ts,
+	&sort_self_max_ts,
 };
 
 /* list of used sort keys */


tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,69 @@
...
+                self_max_ts  = float(line[3])
+
+                # all timestamps must be positive
+                if total_min_ts <= 0 or total_max_ts <= 0 or \
+                   self_min_ts <= 0 or self_max_ts <= 0:
+                    result.append('%s : NG' % func)
+                    continue
+
+                # foo() and child() are each called twice with different durations.
+                # So min and max timestamps must differ (each from a different call).
```
namhyung: Just checking timestamp difference is not enough.  Actually I wanted to compare results from `uftrace replay` but running the command would be a headache.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,69 @@
...
+                if func in ('foo', 'child'):
+                    if total_min_ts == total_max_ts or self_min_ts == self_max_ts:
+                        result.append('%s : NG' % func)
+                        continue
+
+                result.append('%s : OK' % func)
```
namhyung: What about checking `total_min_ts` equals to `self_max_ts` and vice versa?

utils/report.c
```diff
@@ -25,10 +28,14 @@ static void update_time_stat(struct report_time_stat *ts, uint64_t time_ns, bool
 		ts->sum_sq += time_ns * time_ns;
 	}
 
-	if (ts->min > time_ns)
+	if (ts->min > time_ns) {
 		ts->min = time_ns;
-	if (ts->max < time_ns)
+		ts->min_ts = timestamp;
+	}
+	if (ts->max < time_ns || ts->max_ts == 0) {
```
namhyung: I think the max_ts check can go away.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'ts-report', """
+foo : OK
+child : OK
```
namhyung: Can you please use an actual output here?  That'd be useful to understand what's going on even if the timestamps are different.  I expect the final output would be like above (after processed by the sort function) and you can put it in a comment.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,81 @@
...
+            if not ln.strip():
+                continue
+            line = ln.split()
+            if not line:
+                continue
+
+            # pass through already-processed lines (e.g. expected string)
+            if len(line) == 3 and line[1] == ':' and line[2] in ('OK', 'NG'):
+                result.append(ln.strip())
+                continue
```
namhyung: Then we can remove this check.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,81 @@
...
+
+            if line[0] in ('Total', 'Self') or line[0].startswith('='):
+                continue
+            # line format: total_min_ts  total_max_ts  self_min_ts  self_max_ts  function
+            if len(line) < 5:
+                continue
```
namhyung: Looks like an error if it's not 5.  Maybe better to fail by return the original output.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,81 @@
...
+
+            func = line[-1]
+            if func.startswith('__') or func == 'main':
+                continue
```
namhyung: You can add `-F main` to the option in the `prepare()` and remove this check.

junyeong0619:
Sorry. I'm testing now through force push and clone it on linux environment. When I corrected all of your suggestions I will mention you. Thank you!!

junyeong0619:
@namhyung, I’ve addressed all of your suggestions and updated the tests to be more consistent with s-ts-report.c.

junyeong0619:
I’ve added one more commit to address an issue I found during testing.

The out-of-bounds error I previously encountered on my AArch64 Ubuntu (UTM) environment actually only happened in TUI mode. I found that this was caused by the missing sort_key registrations for the newly added timestamp fields. Since NUM_REPORT_FIELD increased, iterating through the fields caused an out-of-bounds read. Interestingly, my WSL environment is also the exact same AArch64 architecture, but it didn't cause a visible crash there, which is why I missed it initially.

I remember we agreed to defer the sort and diff features, but I inevitably had to add the missing sort arrays and macros in this patch to prevent the TUI from breaking.

Also, regarding the fpatchable-function-entry NG issue, it fails on AArch64 Ubuntu (UTM) but passes perfectly on my AArch64 WSL. This strongly indicates that it's an OS or VM-specific dynamic patching issue, rather than a logic flaw in this PR.

Please let me know if this looks good to you!

utils/report.c
```diff
@@ -12,9 +12,12 @@
 static void init_time_stat(struct report_time_stat *ts)
 {
 	ts->min = -1ULL;
+	ts->min_ts = 0;
+	ts->max_ts = 0;
```
namhyung: Actually this is not necessary as we expect the struct was zero-initialized.  But it's fine to leave it.

utils/report.c
```diff
@@ -906,7 +927,10 @@ FIELD_UINT(REPORT_F_CALL, call, call, call, "Calls");
 FIELD_UINT(REPORT_F_SIZE, size, size, size, "Size");
 FIELD_PERCENTAGE(REPORT_F_TOTAL_TIME_STDV, total-stdv, total.stdv, total_stdv, "Total stdv");
 FIELD_PERCENTAGE(REPORT_F_SELF_TIME_STDV, self-stdv, self.stdv, self_stdv, "Self stdv");
-
+FIELD_TIMESTAMP(REPORT_F_TOTAL_MIN_TS, total-min-ts, total.min_ts, total_min_ts, "Total min ts");
+FIELD_TIMESTAMP(REPORT_F_TOTAL_MAX_TS, total-max-ts, total.max_ts, total_max_ts, "Total max ts");
+FIELD_TIMESTAMP(REPORT_F_SELF_MIN_TS, self-min-ts, self.min_ts, self_min_ts, "Self min ts");
+FIELD_TIMESTAMP(REPORT_F_SELF_MAX_TS, self-max-ts, self.max_ts, self_max_ts, "Self max ts");
```
namhyung, line 937: Nit: Please keep the blank line between regular fields and diff fields for visual separation.

tests/t298_report_field_ts.py
namhyung: No, I meant the the comment and the result parts should be exchanged.  The result should go to the `sort` function and to be processed in the same way.

tests/t298_report_field_ts.py
```diff
@@ -0,0 +1,87 @@
...
+            line = ln.split()
+            if not line:
+                continue
+
+            if line[0] in ('Total', 'Self') or line[0].startswith('='):
+                continue
+
+            if len(line) < 5:
+                return output
```
namhyung: And you can remove this condition.

cmds/tui.c
```diff
@@ -203,8 +203,10 @@ enum tui_mode {
 };
 
 static char *report_sort_key[] = {
-	OPT_SORT_KEYS, "total_avg", "total_min", "total_max", "self",	    "self_avg",
-	"self_min",    "self_max",  "call",	 "size",      "total_stdv", "self_stdv",
+	OPT_SORT_KEYS,	"total_avg",	"total_min",	     "total_max",
+	"self",		"self_avg",	"self_min",	     "self_max",
+	"call",		"size",		"total_stdv",	     "self_stdv",
+	"total_min_ts", "total_max_ts", "self_total_min_ts", "self_total_max_ts"
```
namhyung: I think the last two should be "self_min_ts" and "self_max_ts".

junyeong0619: My mistake! I'll correct those two to self_min_ts and self_max_ts to keep it consistent. Thanks for the review

namhyung (approved):
LGTM, thanks a lot!

junyeong0619:
> LGTM, thanks a lot!

Thank you so much for the review and the LGTM! I really appreciate your time and the helpful feedback.

Issue #2033, opened 2026-03-13
Build failure in utils/symbol-rawelf.h


junyeong0619:
A compilation error occurs when building uftrace on Rocky Linux 9 (or any environment without libelf-devel installed), where the
  project falls back to its internal ELF parser (utils/symbol-rawelf.h).

Environment
   * OS: Rocky Linux 9 (or Linux systems without libelf-devel)
   * Architecture: x86_64 / AArch64
   * Error Message:
      error: too few arguments to function ‘strcmp’ during compilation of utils/symbol.o.

And also when i  installed elfutils-libelf-devel it is compiled normally.
It is happened to discover while testing in a minimal installation.

Under bash commands is the result when compile without libelf.

```bash
[root@localhost uftrace]# make
  CC       utils/symbol.o
In file included from /tmp/uftrace/utils/symbol.h:23,
                 from /tmp/uftrace/uftrace.h:14,
                 from /tmp/uftrace/utils/symbol.c:22:
/tmp/uftrace/utils/symbol.c: In function ‘load_dynsymtab’:
/tmp/uftrace/utils/symbol-rawelf.h:170:27: error: too few arguments to function ‘strcmp’
  170 |              (iter)->i += strcmp((iter)->comment) + 1, (iter)->comment = (iter)->data + (iter)->i)
      |                           ^~~~~~
/tmp/uftrace/utils/symbol.c:730:25: note: in expansion of macro ‘elf_for_each_comment’
  730 |                         elf_for_each_comment(&elf, &sec_iter) {
      |                         ^~~~~~~~~~~~~~~~~~~~
In file included from /tmp/uftrace/utils/utils.h:20,
                 from /tmp/uftrace/utils/symbol.h:18,
                 from /tmp/uftrace/uftrace.h:14,
                 from /tmp/uftrace/utils/symbol.c:22:
/usr/include/string.h:156:12: note: declared here
  156 | extern int strcmp (const char *__s1, const char *__s2)
      |            ^~~~~~
In file included from /tmp/uftrace/utils/symbol.h:23,
                 from /tmp/uftrace/uftrace.h:14,
                 from /tmp/uftrace/utils/symbol.c:22:
/tmp/uftrace/utils/symbol-rawelf.h:170:54: warning: left-hand operand of comma expression has no effect [-Wunused-value]
  170 |              (iter)->i += strcmp((iter)->comment) + 1, (iter)->comment = (iter)->data + (iter)->i)
      |                                                      ^
[...]

PR #2034, opened 2026-03-13
symbol: Fix build error in elf_for_each_comment()

junyeong0619:
As the issue that I opened #2033
The elf_for_each_comment() macro incorrectly used strcmp() instead of strlen() when calculating the next comment offset. This caused a compilation error when libelf was not present and the internal ELF parser was used.


namhyung (approved):
LGTM

PR #2039, opened 2026-03-25
report: Add cpu field to show per-function CPU execution info

### LINKED ISSUE #1560: Support --percpu option to show replay based on each core
I got a request to show uftrace data per each CPU. uftrace doesn't keep core information for each function entry/exit records, but we can calculate which thread is assigned to which CPU core based on `perf-cpuXX.dat`.

From the dump output, we can watch `linux:sched-in` and `linux:sched-out` events.
```
$ uftrace dump
    ...
reading perf-cpu4.dat
9741636.914809755  23676: [event] linux:sched-in(200001)
9741636.915957646  23676: [event] linux:task-exit(200005)
reading perf-cpu26.dat
9741636.908026775  23661: [event] linux:task-name(200006)
9741636.914712846  23676: [event] linux:task-new(200004)
9741636.914787536  23661: [event] linux:sched-out(200002)
9741636.916127212  23661: [event] linux:sched-in(200001)
9741636.916760918  23661: [event] linux:task-exit(200005)
```
It can be useful to have a usage something like `uftrace --dump --percpu --chrome`.

junyeong0619:
I've implemented the changes based on our discussion in the issue. 
Since the previous field macros weren't suitable for this feature, I've added a temporary workaround. 
I'd appreciate your feedback on this approach.
Thank you!
closed #1560

스크린샷 2026-03-25 오전 10 13 44


utils/report.h
```diff
@@ -33,6 +33,7 @@ struct uftrace_report_node {
 	struct report_time_stat self;
 	struct uftrace_dbg_loc *loc;
 	uint64_t call;
+	uint64_t cpu_mask;
```
namhyung: This means we can only support up to 64 CPUs.

utils/report.c
```diff
@@ -978,6 +980,50 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+			first = false;
+			run_start = -1;
+		}
+	}
+	slen = strlen(buf);
+	lpad = (10 - slen) / 2;
+	rpad = 10 - slen - lpad;
+	pr_out("%*s%s%*s", lpad, "", buf, rpad, "");
+}
+FIELD_STRUCT(REPORT_F_CPU, cpu, cpu, "   CPU    ", 10);
```
namhyung: We cannot fix the field width to 10.

cmds/tui.c
```diff
@@ -501,6 +502,50 @@ REPORT_FIELD_TIMESTAMP(REPORT_F_TOTAL_MIN_TS, total-min-ts, total.min_ts, total_
...
+				len += snprintf(buf + len, sizeof(buf) - len, "%d-%d", run_start,
+						run_end);
+			first = false;
+			run_start = -1;
+		}
+	}
+	slen = strlen(buf);
+	lpad = (10 - slen) / 2;
+	rpad = 10 - slen - lpad;
+	printw("%*s%s%*s", lpad, "", buf, rpad, "");
```
namhyung: Looks like a similar logic.  Maybe you can split the common part to return a string.

namhyung (commented):
Please rebase the change onto the current master.  There are new test cases added.

utils/report.h
junyeong0619: How about change like this?
```c
unsigned long *cpu_mask;
int nr_cpus;
```
This way we can dynamically allocate the mask and support more than 64 CPUs.

utils/report.c
junyeong0619: You're right. Since the previous change allows us to support more CPUs, the CPU string can easily exceed 10 characters. I'll change it to calculate the maximum length dynamically and adjust the field width accordingly.

cmds/tui.c
junyeong0619: I will create a common function to generate and return the CPU string, then reuse it here.

junyeong0619:
Thanks for the feedback! I'll get these changes applied and tested, then mention you for another look.

utils/report.h
namhyung: Yep, dynamic allocation would work.  I think it's better to use `uint64_t` rather than `unsigned long` to make sure the bit size.

utils/report.c
namhyung: But I don't think it's scalable on large systems with 100+ CPUs.

junyeong0619: Regarding the field width and 100+ CPU scaling issue, I've thought about it and believe that using a Fixed-width Truncation combined with Range compression would be the most practical approach for the terminal output.

For example, we can compress consecutive CPUs into ranges, and if the string length exceeds a fixed maximum width (e.g., 15 or 20 chars), we truncate it and append an ellipsis with the number of omitted items:
- e.g., '0-5,10-15...(+4)'

Since `uftrace report` needs to maintain a readable tabular layout, allowing the CPU field to grow dynamically for hundreds of cores could break the UI. By keeping a fixed width, we ensure the table remains clean. If users ever need the full, un-truncated list of CPUs, we could perhaps consider leaving a full dump via a '--verbose' option in the future.

What do you think about setting a fixed width (like 15 or 18) and using this `...` truncation method for the default report view?

namhyung: Looks ok, but note that we are about to add `--format=csv` option to the `report` output (#2041).  Having commas in the CPU field would confuse the CSV format.  As the CSV format is mainly for machine-processing, you may print the whole list of CPUs there but with a different separator.

Also you would need to be careful when truncate the output with big numbers.  Probably we want to add a new unit test.

utils/report.h
junyeong0619: I changed it with uint64_t*.

utils/report.c
junyeong0619: I have updated the truncated format as we discussed previously. 
```bash
[root@localhost uftrace]# ./uftrace record tests/tcpu && ./uftrace report -f cpu
     CPU     Function
  =========  ====================
      3      main
     0-3     pin_to_cpu
     0-3     sched_setaffinity
     0-3     linux:schedule (pre-empted)
      0      task_a
      1      task_b
      2      task_c
      3      task_d
     0-3     work
   ...(+3)   shared_work
      3      sysconf
      3      __monstartup
      3      __cxa_atexit
[root@localhost uftrace]#
```
I also handled the case where the --format=csv option is used, ensuring the CPU data is represented correctly in CSV output.

```bash
[root@localhost uftrace]# ./uftrace record tests/tcpu && ./uftrace report -f cpu --format=csv
#   CPU    ,Function
3,main
0-3,pin_to_cpu
0-3,sched_setaffinity
0-3,linux:schedule (pre-empted)
0,task_a
1,task_b
3,task_d
2,task_c
0-3,work
0/1/3,shared_work
0,sysconf
0,__monstartup
0,__cxa_atexit
[root@localhost uftrace]#
```

cmds/tui.c
junyeong0619: I've implemented make_cpu_str() in utils/report.c and updated both print_cpu() and print_report_cpu() in cmds/tui.c to use it.

junyeong0619:
Hi @namhyung, I've updated the PR with the requested changes (including the new unit test) and rebased it onto the latest master. Let me know if there's anything else I should address before merging. Thank you!

utils/report.c
```diff
@@ -142,6 +143,20 @@ void report_update_node(struct uftrace_report_node *node, struct uftrace_task_re
 	node->loc = loc;
 	if (task->func != NULL)
 		node->size = task->func->size;
+	if (task->current_cpu >= 0) {
```
namhyung: Can you we do it only if user requested 'CPU' field?  I think it's mostly unused then better to reduce the overhead and memory consumption.

utils/report.c
```diff
@@ -142,6 +143,20 @@ void report_update_node(struct uftrace_report_node *node, struct uftrace_task_re
...
+		int cpu = task->current_cpu;
+		int nr_cpus = task->h->info.nr_cpus_possible;
+
+		if (nr_cpus <= cpu)
+			nr_cpus = cpu + 1;
```
namhyung: Looks like a bug.  Did it happen actually?

utils/report.c
```diff
@@ -142,6 +143,20 @@ void report_update_node(struct uftrace_report_node *node, struct uftrace_task_re
...
+		if (node->cpu_mask == NULL) {
+			int nr_words = DIV_ROUND_UP(nr_cpus, 64);
+			node->cpu_mask = xcalloc(nr_words, sizeof(uint64_t));
+			node->nr_cpus = nr_cpus;
```
namhyung, line 153: I think it's better to have the actual number of CPUs this function ran.

utils/report.c
namhyung: I'm still not very convinced about the use case, but it seems better.  But I think we should use '/' consistently to minimized possible confusion.  Otherwise you need to describe the difference in the document.

It'd be ok for small systems which have like a dozen CPUs.  But usually NUMA machines which you care actually come with a lot more CPUs - e.g. 256, 512, ...  In that case I'm not sure how much this limited CPU notation can give the meaningful result (without using the CSV format).

utils/report.c
namhyung: And maybe we need disable this field in the TUI then.

tests/s-per-cpu.c
```diff
@@ -0,0 +1,58 @@
...
+	CPU_ZERO(&cpuset);
+	CPU_SET(cpu, &cpuset);
+	sched_setaffinity(0, sizeof(cpuset), &cpuset);
+}
+
+static void work(void)
+{
+	volatile int i;
+	for (i = 0; i < 10000; i++)
+		;
```
namhyung: I prefer a non-empty statement.  How about replacing it with `continue;`?

tests/t304_report_field_cpu.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python3
```
namhyung, line 1: Please check the latest code how many tests are added recently.

tests/t299_report_field_cpu.py
```diff
@@ -0,0 +1,85 @@
...
+from runtest import TestBase
+
+
+class TestCase(TestBase):
+    def __init__(self):
+        ncpus = os.cpu_count()
+
+        expected_lines = [
+            "CPU      Function",
+            "==========  ====================",
```
namhyung: It's annoying if the lines are not aligned.

tests/t299_report_field_cpu.py
```diff
@@ -0,0 +1,85 @@
...
+        self.option = ''
+        record_cmd = TestBase.runcmd(self)
+        record_cmd = record_cmd.replace('--no-event', '')
+        self.pr_debug('prerun command: ' + record_cmd)
+        sp.call(record_cmd.split())
+        return TestBase.TEST_SUCCESS
+
+    def setup(self):
+        self.subcmd = 'report'
+        self.option = '-f cpu'
```
namhyung: I think we can remove the sort logic or at least make it very simple once you use a filter option.  I think it's better to demonstrate power of uftrace filtering features.  How about this?

```
self.option = '-f cpu -F ^task_@depth=1'
```

namhyung (commented):
Thanks for the update!

utils/report.c
junyeong0619: Thanks for the feedback! 
I've been thinking about the usability concern on large NUMA machines, and I agree that a truncated CPU list in a narrow column is suboptimal for hundreds of CPUs.

To resolve this, I've revised the design as follows:

Normal Report Mode (-f cpu): Now displays the count of distinct CPUs the function ran on.

```bash
Total time   Calls   CPU   Function
==========   =====   ===   ========
1.234 ms      100     1    foo
0.500 ms       50     4    bar
```

This provides a clear metric for CPU affinity/migration at a glance, regardless of the system scale.

CSV Mode (--format=csv -f cpu): Provides the full CPU list using a consistent notation (- for ranges and / for separators).

```bash
# CPU_List,Function
0-3,main
0,foo
0/2/4-6,bar
```
This dual approach keeps the TUI clean and scalar while preserving detailed data for script-based analysis in CSV. It also eliminates the notation inconsistency between formats.

Does this direction look good to you?

namhyung: I think we could have multiple CPU fields like below:
* cpu_count: number of distinct CPUs
* cpu_mask: bitmask of CPUs (32-bit unit): for example "00000001 000000ff" means 0-7/32.  This can be fixed-width according to the max possible CPU number.
* cpu_list: string format of CPUs like "0-7/32".  This can be variable width with a limit.

junyeong0619: For the cpu_mask of 512 cores, I will follow the kernel standard by printing 16 blocks of 32-bit hex units separated by spaces.

Like this:
ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff 
ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff ffffffff

I'll apply a fixed-width limit (around 20 characters) and use a simple ... suffix when it exceeds the limit. 

Like this:
0-63/64/65/66/...

Also we will develop the cpu_count field, users can easily see how many CPUs are missing from the truncated list.

Could you confirm if I've understood your suggestions correctly? I want to make sure we're on the same page before I finalize the implementation.

namhyung: I think we don't need a limit for cpu_mask.  Yes, it can be very long on large systems but it's deterministic.  It's up to users to use this field or not.  But cpu_list is unpredictable so better to have a limit.

junyeong0619: I will continue developing the project along these lines. Thanks for your review!

junyeong0619:
Hi @namhyung,

I have addressed all the review comments and cleaned up the branch into two commits. Here is a summary of the changes:

1. report: Add cpu-count, cpu-mask, and cpu-list fields

Field Separation: As discussed, the original cpu field has been split into three separate fields (cpu-count, cpu-mask, cpu-list). TUI support has been removed.

Optimization: Updates in report_update_node() are now gated behind report_cpu_field_used() to avoid unnecessary overhead when these fields are not requested.

Logic Cleanup: Removed the defensive if (nr_cpus <= cpu) check, as current_cpu is consistently cleared to -1 on sched-out, making the check redundant.

Accurate Tracking: The cpu-count field now reports the actual number of distinct CPUs a function ran on by performing a popcount on the cpu_mask, rather than just reporting the system's total possible CPUs.

2. tests: Add per-cpu test binary and t304

Simplified Logic: Rewrote the test using -f cpu-list -F ^task_@depth=1 to filter only top-level task functions, which significantly simplifies the sorting logic.

Update: The test has been updated to t304 to align with the latest test numbering.

Finally, I have rebased the branch onto the current master.

Please let me know if there are any further points to address. Thanks!

utils/report.c
junyeong0619: done.

utils/report.c
junyeong0619: done.

utils/report.c
junyeong0619: I think that i handled it with cpu-count maybe.

tests/s-per-cpu.c
junyeong0619: done.

tests/t304_report_field_cpu.py
junyeong0619: done.

tests/t299_report_field_cpu.py
junyeong0619: I checked the lines.

tests/t299_report_field_cpu.py
junyeong0619: I used the command for the test.

utils/report.c
```diff
@@ -987,16 +1002,232 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
 FIELD_TIME(REPORT_F_TASK_SELF_TIME, self, self.sum, task_self, "Self time");
 FIELD_TID(REPORT_F_TASK_TID, tid, task_tid, "TID");
 FIELD_UINT(REPORT_F_TASK_NR_FUNC, func, call, task_nr_func, "Num funcs");
+
+static int cpu_field_width = 10;
```
namhyung: Why 10?  I think we can start with the minimal length (8).  Probably you need to define it as a macro constant.

utils/report.c
```diff
@@ -987,16 +1002,232 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+
+int make_cpu_str(uint64_t *mask, int nr_cpus, char *buf, int bufsize)
```
namhyung: Maybe `make_cpulist_str()`?  I think it's better to split this function and the unit test into a separate commit before introducing new report fields.  I feel like they should be in utils.c file.

utils/report.c
```diff
@@ -987,16 +1002,232 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+	}
+
+	slen = snprintf(buf, sizeof(buf), "%d", count);
+	lpad = (9 - slen) / 2;
+	rpad = 9 - slen - lpad;
+	pr_out("%*s%s%*s", lpad, "", buf, rpad, "");
+}
+FIELD_STRUCT(REPORT_F_CPU_COUNT, cpu-count, cpu_count, "CPU Count", 9);
+
+static int make_cpu_hex_str(uint64_t *mask, int nr_cpus, char *buf, int bufsize)
```
namhyung: Similarly, it could be `make_cpumask_str()` and move to utils.c.

tests/s-per-cpu.c
```diff
@@ -0,0 +1,58 @@
...
+
+static void shared_work(void)
+{
+	work();
+}
+
+void task_a(void)
+{
+	pin_to_cpu(0);
+	shared_work();
```
namhyung, line 27: Hmm.. I suggested to rename the test functions in the previous comment.  I also realized that this function can be in any CPU before calling `pin_to_cpu(0)`.  Is that saved in the cpu list?

tests/t304_report_field_cpu.py
```diff
@@ -0,0 +1,64 @@
...
+from runtest import TestBase
+
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'per-cpu', result="""\
+ CPU LIST   Function
+==========  ====================
+         0  task_a
+         1  task_b""")
```
namhyung: Maybe the right test would be to check the first function contains CPU 0 rather than equals to 0.

tests/s-per-cpu.c
```diff
@@ -0,0 +1,58 @@
...
+void task_c(void)
+{
+	pin_to_cpu(2);
+	work();
+}
+void task_d(void)
+{
+	pin_to_cpu(3);
+	shared_work();
+}
```
namhyung: Probably simpler to remove task_c and task_d.

tests/t304_report_field_cpu.py
```diff
@@ -0,0 +1,64 @@
...
+            if 'CPU' in line and 'LIST' in line:
+                header.append('CPU LIST  Function')
+                continue
+            if line[0].startswith('='):
+                header.append('==========  ====================')
+                continue
+            cpu = line[0]
+            func = ' '.join(line[1:])
+            if func in ('task_a', 'task_b'):
+                body.append('%s  %s' % (cpu, func))
```
namhyung: I think you basically accept all lines.  Then you could make it simply oneliner to return `output`.

namhyung (commented):
Thanks for the update.  Please update documentation when you add any user-visible changes.

tests/s-per-cpu.c
junyeong0619: No, it is not saved in the CPU list.

While it's true that task_a starts on an arbitrary CPU (CPU 2 in this case) before the pin_to_cpu(0) call, this initial CPU is not recorded. This is because report_update_node is only invoked at the task's EXIT point.

Since the recording happens only when the task finishes—at which point it has already been pinned to the target CPU—the CPU it occupied prior to migration is ignored. Consequently, the cpu-list only contains the pinned CPU (e.g., "0" for task_a), making the current test assertion task_a == "0" stable and reliable.

junyeong0619:
I've addressed most of the review comments. The branch is now organized into 4 clean commits.

Resolved

cpu_field_width = 10: Defined as CPU_LIST_FIELD_WIDTH macro (set to 8). Both FIELD_STRUCT and the minimum-width clamp in set_cpu_field_width() now use this macro consistently.

Rename / move helpers to utils.c: make_cpu_str() → make_cpulist_str() and make_cpu_hex_str() → make_cpumask_str(). Both have been moved to utils/utils.c along with their unit tests, split into a dedicated commit preceding the report-field changes.

Documentation: Added entries for cpu-count, cpu-mask, and cpu-list to the FIELDS section of uftrace-report.md, including a note that these fields require perf context-switch events.

Remove task_c and task_d: Done. The test binary now only pins task_a to CPU 0 and task_b to CPU 1. The ncpus check in main() has also been removed.

Sort() oneliner: Simplified to return output.rstrip(). To ensure deterministic output across different compilers and optimization levels (where natural total-time order can flip), I added -s func to the report options. This ensures a stable alphabetical order, removing the need for further sorting in Python.

tests/s-per-cpu.c
junyeong0619: Should I keep the equals check, or would you still prefer switching to contains 0 for safety? I'll go with whichever you prefer

namhyung: Please check if it contains 0.

utils/report.c
```diff
@@ -987,16 +1002,150 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+	char buf[16];
+	int slen, lpad, rpad;
+
+	if (format_mode == FORMAT_CSV) {
+		pr_out("%d", count);
+		return;
+	}
+
+	slen = snprintf(buf, sizeof(buf), "%d", count);
+	lpad = (9 - slen) / 2;
```
namhyung: Please define a macro for magic number 9.

utils/report.c
```diff
@@ -987,16 +1002,150 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+	rpad = 9 - slen - lpad;
+	pr_out("%*s%s%*s", lpad, "", buf, rpad, "");
+}
+FIELD_STRUCT(REPORT_F_CPU_COUNT, cpu-count, cpu_count, "CPU Count", 9);
+
+static void print_cpu_mask(struct field_data *fd)
+{
+	struct uftrace_report_node *node = fd->arg;
+	char buf[512];
```
namhyung: Is 512 always enough?

utils/report.c
```diff
@@ -987,16 +1002,150 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+
+void set_cpu_field_width(int width)
+{
+	static char cpu_header[64];
+	int lpad, rpad;
+
+	if (width < CPU_LIST_FIELD_WIDTH)
+		width = CPU_LIST_FIELD_WIDTH;
+	if (width > CPU_STR_MAX_LEN)
+		width = CPU_STR_MAX_LEN;
```
namhyung: They can be renamed to CPU_LIST_MIN/MAX_LEN.

utils/report.c
```diff
@@ -987,16 +1002,150 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+{
+	return cpu_field_width;
+}
+
+int report_calc_cpu_width(struct rb_root *root)
+{
+	struct uftrace_report_node *node;
+	struct rb_node *n = rb_first(root);
+	char buf[256];
+	int max_width = 1;
```
namhyung: It should start with CPU_LIST_MIN_LEN.

utils/report.c
```diff
@@ -987,16 +1002,150 @@ FIELD_TIME(REPORT_F_TASK_TOTAL_TIME, total, total.sum, task_total, "Total time")
...
+
+	while (n) {
+		int w;
+
+		node = rb_entry(n, typeof(*node), name_link);
+		w = make_cpulist_str(node->cpu_mask, node->nr_cpus, buf, sizeof(buf));
+		if (w > max_width)
+			max_width = w;
```
namhyung, line 1146: You can break the loop if max_width is greater than CPU_LIST_MAX_LEN.

tests/s-per-cpu.c
```diff
@@ -0,0 +1,41 @@
...
+		continue;
+}
+
+static void shared_work(void)
+{
+	work();
+}
+
+void task_a(void)
+{
```
namhyung, line 25: In general, we say 'task' for 'thread'.  How about naming it more directly like `func_cpu0`?

utils/utils.c
```diff
@@ -1040,6 +1040,90 @@ int copy_file(const char *path_in, const char *path_out)
...
+				n = snprintf(tmp, sizeof(tmp), "%d", run_start);
+			else
+				n = snprintf(tmp, sizeof(tmp), "%d-%d", run_start, run_end);
+
+			prefix = first ? 0 : 1;
+
+			if (len + prefix + n + 4 >= bufsize) {
+				/* can't fit: append separator (if any) then ellipsis */
+				if (!first && len + 1 + 4 <= bufsize)
+					buf[len++] = '/';
```
namhyung, line 1099: What about if the buffer is almost full and no space for the slash and ellipsis?  I think it should go back and delete the last entry and overwrite it with a slash and ellipsis.

utils/utils.c
```diff
@@ -1159,4 +1243,113 @@ TEST_CASE(utils_parse_timestamp)
...
+	pr_dbg("large CPU number (CPU 63)\n");
+	mask[0] = 1ULL << 63;
+	TEST_EQ(make_cpulist_str(mask, 64, buf, sizeof(buf)), 2);
+	TEST_STREQ(buf, "63");
+
+	pr_dbg("truncation with ellipsis\n");
+	/* CPUs 0,2,4,6,8 - with bufsize=8 */
+	mask[0] = 0x155UL;
+	make_cpulist_str(mask, 9, buf, 8);
+	TEST_STREQ(buf, "0/2/...");
```
namhyung, line 1312: Can you add more complex scenarios like "0-32/34-40/42-45/..."?  I think it's better to confirm if it overwrites the existing entry properly when buffer is full.

namhyung (commented):
Thanks for the update.  Nit: technically it doesn't fix any bugs, so it's better to say it "closed #xxx" rather than "fixed #xxx".

utils/report.c
junyeong0619: Done. Defined it as CPU_COUNT_LEN.

utils/report.c
junyeong0619: Done. Switched to a dynamic allocation sized from node->nr_cpus.

utils/report.c
junyeong0619: Done.

utils/report.c
junyeong0619: Done.

utils/report.c
junyeong0619: Done.

tests/s-per-cpu.c
junyeong0619: Done. Switched it to func_cpux.

utils/utils.c
junyeong0619: Done.  Reworked truncation: when an entry doesn't fit and /... can't be appended cleanly at the current
position, scan back through the buffer for the latest / where /... fits and overwrite from there.  The
walk-back can drop multiple entries when needed; if no separator works, the buffer is reset to ....

Added tests covering one-level overwrite ("0-32/34-40/42-45/..."), multi-level overwrite
("0-32/34-40/..."), and the full-reset fallback ("...").

junyeong0619:
Always Thanks for your review. I Also treated other magic numbers — added CPU_MASK_MIN_LEN and CPU_MASK_GROUP_LEN in report.h and replaced the related 8s and 9s in set_cpu_field_width, set_cpu_mask_field_width, and print_cpu_mask.

junyeong0619:
Updated the sort() in t304 to assert containment instead of exact equality. Also expanded the utils_make_cpumask_str test cases — added scattered single-group bits, bits straddling a group boundary, an arbitrary 64-bit pattern, and a middle-group-only mask

utils/utils.c
junyeong0619: Handled it above.

utils/utils.c
namhyung: Thanks!

utils/utils.c
```diff
@@ -1040,6 +1040,113 @@ int copy_file(const char *path_in, const char *path_out)
...
+		goto empty;
+
+	return len;
+
+empty:
+	snprintf(buf, bufsize, "-");
+	return 1;
+}
+
+int make_cpumask_str(uint64_t *mask, int nr_cpus, char *buf, int bufsize)
```
namhyung: I'm curious what if we can use uint32_t for mask.

utils/utils.c
```diff
@@ -1040,6 +1040,113 @@ int copy_file(const char *path_in, const char *path_out)
...
+	int i;
+
+	if (mask == NULL || nr_cpus == 0) {
+		snprintf(buf, bufsize, "-");
+		return 1;
+	}
+
+	/* print high groups first (kernel cpumask style) */
+	for (i = ngroups - 1; i >= 0; i--) {
+		uint32_t word = (uint32_t)(mask[i / 2] >> ((i % 2) * 32));
```
namhyung: Then this line would be simpler.

utils/utils.c
```diff
@@ -1159,4 +1266,167 @@ TEST_CASE(utils_parse_timestamp)
...
+	TEST_EQ(make_cpulist_str(mask, 5, buf, sizeof(buf)), 5);
+	TEST_STREQ(buf, "0/2/4");
+
+	pr_dbg("large CPU number (CPU 63)\n");
+	mask[0] = 1ULL << 63;
+	TEST_EQ(make_cpulist_str(mask, 64, buf, sizeof(buf)), 2);
+	TEST_STREQ(buf, "63");
+
+	pr_dbg("truncation with ellipsis (multi-level backtrack)\n");
+	/* CPUs 0,2,4,6,8 with bufsize=8: writes "0/2/4/6", then "/8" doesn't
```
namhyung: For multi-line comments, we usually start with an empty line just with "/*".

utils/utils.c
```diff
@@ -1159,4 +1266,167 @@ TEST_CASE(utils_parse_timestamp)
...
+
+	pr_dbg("truncation backtracks the last entry to fit /...\n");
+	/* CPUs 0-32, 34-40, 42-45, 47, 50 with bufsize=21:
+	 * writes "0-32/34-40/42-45/47" then "/50" doesn't fit; backtrack uses
+	 * '/' at index 16 and overwrites the last "/47" entry with "/..."
+	 */
+	mask[0] = ((1ULL << 33) - 1) | (0x7FULL << 34) | (0xFULL << 42) | (1ULL << 47) |
+		  (1ULL << 50);
+	make_cpulist_str(mask, 64, buf, 21);
+	TEST_STREQ(buf, "0-32/34-40/42-45/...");
```
namhyung, line 1328: Let's start with the full string first.  I mean we should check it'd produce "0-32/34-40/42-45/47/50".

utils/utils.c
junyeong0619: I think you're right. Since the output is already grouped in 32-bit hex words, uint32_t fits better and lets us drop the shift/mask in make_cpumask_str(). Will update.

utils/utils.c
junyeong0619: done.

utils/utils.c
junyeong0619: Done. Added a full-string check ("0-32/34-40/42-45/47/50") before the truncation cases, and the same mask is reused for the bufsize=21/20 tests below.

namhyung (approved):
LGTM, thanks!

junyeong0619:
> LGTM, thanks!

Thank you for helping me wrap this up.

PR #2041, opened 2026-04-13
report: Support --format=csv option

kangtegong:
report: Support --format=csv option

Add the --format=csv option to print CSV output.

```
Example:
  $ gcc -pg -o abc tests/s-abc.c
  $ uftrace record ./abc
  $ uftrace report --format=csv
  #Total time,Self time,Calls,Function
  1450,210,1,main
  1240,170,1,a
  1070,230,1,b
  840,450,1,c
  600,600,1,__monstartup
  390,390,1,getpid
  180,180,1,__cxa_atexit
```

### DIFF
--- a/cmds/report.c
+++ b/cmds/report.c
@@ -26,6 +26,44 @@ static void print_field(struct uftrace_report_node *node, int space)
 	print_field_data(&output_fields, &fd, space);
 }
 
+static bool print_csv_field(struct uftrace_report_node *node)
+{
+	struct field_data fd = {
+		.arg = node,
+	};
+	struct display_field *field;
+	bool first = true;
+
+	if (list_empty(&output_fields))
+		return false;
+
+	list_for_each_entry(field, &output_fields, list) {
+		pr_out("%s", first ? "" : ",");
+		field->print(&fd);
+		first = false;
+	}
+
+	return true;
+}
+
+static void print_csv_header(const char *postfix)
+{
+	struct display_field *field;
+	bool first = true;
+
+	if (list_empty(&output_fields)) {
+		pr_out("#%s\n", postfix);
+		return;
+	}
+
+	list_for_each_entry(field, &output_fields, list) {
+		pr_out("%s%s", first ? "#" : ",", field->header);
+		first = false;
+	}
+
+	pr_out(",%s\n", postfix);
+}
+
 static void insert_node(struct rb_root *root, struct uftrace_task_reader *task, char *symname,
 			struct uftrace_dbg_loc *loc)
 {
@@ -233,6 +271,19 @@ static void print_function(struct uftrace_report_node *node, void *unused, int s
 	pr_out("\n");
 }
 
+static void print_function_csv(struct uftrace_report_node *node, void *unused, int space)
+{
+	if (print_csv_field(node))
+		pr_out(",%s", node->name);
+	else
+		pr_out("%s", node->name);
+
+	if (node->loc)
+		pr_out(" [%s:%d]", node->loc->file->name, node->loc->line);
+
+	pr_out("\n");
+}
+
 static void print_line(struct list_head *output_fields, int space)
 {
 	struct display_field *field;
@@ -265,6 +316,12 @@ static void report_functions(struct uftrace_data *handle, struct uftrace_opts *o
 		return;
 
 	setup_report_field(&output_fields, opts, avg_mode);
+
+	if (format_mode == FORMAT_CSV) {
+		print_csv_header("Function");
+		print_and_delete(&sort_root, true, NULL, print_function_csv, field_space);
+		return;
+	}
 
 	print_header_align(&output_fields, "  ", "Function", field_space, ALIGN_RIGHT, false);
 	if (!list_empty(&output_fields)) {
@@ -364,6 +421,21 @@ static void print_task(struct uftrace_report_node *node, void *arg, int space)
 	pr_out("%-*s\n", TASK_COMM_LEN, t->comm);
 }
 
+static void print_task_csv(struct uftrace_report_node *node, void *arg, int space)
+{
+	int pid;
+	struct uftrace_task *t;
+	struct uftrace_data *handle = arg;
+
+	pid = strtol(node->name, NULL, 10);
+	t = find_task(&handle->sessions, pid);
+
+	if (print_csv_field(node))
+		pr_out(",%s\n", t->comm);
+	else
+		pr_out("%s\n", t->comm);
+}
+
 static void report_task(struct uftrace_data *handle, struct uftrace_opts *opts)
 {
 	struct uftrace_record *rstack;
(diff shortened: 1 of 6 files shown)


cmds/report.c
```diff
@@ -233,6 +271,22 @@ static void print_function(struct uftrace_report_node *node, void *unused, int s
 	pr_out("\n");
 }
 
+static void print_function_csv(struct uftrace_report_node *node, void *unused, int space)
+{
+	(void)unused;
+	(void)space;
```
namhyung: I don't think we need this.

cmds/report.c
```diff
@@ -364,6 +424,22 @@ static void print_task(struct uftrace_report_node *node, void *arg, int space)
 	pr_out("%-*s\n", TASK_COMM_LEN, t->comm);
 }
 
+static void print_task_csv(struct uftrace_report_node *node, void *arg, int space)
+{
+	int pid;
+	struct uftrace_task *t;
+	struct uftrace_data *handle = arg;
+
+	(void)space;
```
namhyung: And here as well.

cmds/report.c
```diff
@@ -26,6 +26,44 @@ static void print_field(struct uftrace_report_node *node, int space)
...
+		pr_out("%s\n", postfix);
+		return;
+	}
+
+	list_for_each_entry(field, &output_fields, list) {
+		pr_out("%s%s", first ? "" : ",", field->header);
+		first = false;
+	}
+
+	pr_out(",%s\n", postfix);
```
namhyung, line 64: I'd like to add `#` at the beginning for the header.  How about this?

```c
static void print_csv_header(const char *postfix)
{
    struct display_field *field;
    bool first = true;

    list_for_each_entry(field, &output_fields, list) {
        pr_out("%s%s", first ? "#" : ",", field->header);
        first = true;
    }
    pr_out("%s%s", first ? "#" : ",", postfix);
}
```

namhyung (commented):
I think the CSV format is for machine processing and the output can be a little bit different to be parsed easily.

For example, the first line (header) is for human so it's better to start with a `#` like in comment.  And the each column doesn't need to contain whitespaces to align output.  And the time unit is not needed so it can just print them in nsec.

Something like below would be great.

```
#Total time,Self time,Calls,Function
1120,170,1,main
950,130,1,a
820,180,1,b
640,330,1,c
310,310,1,getpid
290,290,1,__monstartup
100,100,1,__cxa_atexit
```

kangtegong:
Thank you for your valuable feedback.
I’ve just made the revisions :)

To make parsing easier, I included a # in the first line,
and structured the output so that each row contains only numbers, without units like ms, ns, or %.

If there’s anything else to revise, please let me know. Thanks! 😊

doc/uftrace-report.md
```diff
@@ -53,6 +53,15 @@ REPORT OPTIONS
...
 :   Show average, min, max, relative standard deviation of each function's self time.
 
+\--csv
+:   Show report output in comma-separated values (CSV) format.
+    The CSV format is supported in all report modes below:
+    default function report, `--avg-total`, `--avg-self`, `--task`,
+    and `--diff` (including `--diff-policy` and `--sort-column`).
+    It can also be combined with field/sort/filter options such as
+    `-f`, `-s`, `-F`, `-N`, `-C`, `-D`, `-t`, `-r`, `--tid`,
+    `--no-libcall`, and `--srcline`.
```
namhyung: This looks a bit verbose.  I think it should work with all report modes and filter options.  So I think we can just keep the first sentence here.

Also I remember we wanted to reuse the existing --format option.  So please add `--format=csv` instead of `--csv`.  I think we only support `--format=html` for mermaid dump.  Maybe we need to update the existing document, but it could be done in a separate PR.

namhyung (commented):
Looks good, but I'd like to change the option name.  And can you please add a test case?  I think you can check the number of fields (or commas) and only pick the call count and function name for comparison.

kangtegong:
Of course!

I changed `--csv` to `--format=csv` and amended the documentation (and commit message).

I also added test cases. The test results are as follows:

```
$ python3 runtest.py -j1 -c gcc 302

Compiler                  gcc                                         
Runtime test case         pg             finstrument-fu fpatchable-fun
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os O0 O1 O2 O3 Os
302 report_csv          : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK

$ python3 runtest.py -j1 -c clang 302

Compiler                  clang                                       
Runtime test case         pg             finstrument-fu fpatchable-fun
------------------------: O0 O1 O2 O3 Os O0 O1 O2 O3 Os O0 O1 O2 O3 Os
302 report_csv          : OK OK OK OK OK OK OK OK OK OK OK OK OK OK OK
```

tests/t302_report_csv.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'sort', """
+1 main
+1 bar
+1 usleep
+2 foo
+6 loop
```
namhyung: No, please copy the actual uftrace output here.  It should be handled in the same way.

tests/t302_report_csv.py
```diff
@@ -0,0 +1,50 @@
...
+        self.subcmd = 'report'
+        self.option = '--format=csv'
+
+    def sort(self, output):
+        if ',' not in output:
+            result = []
+            for ln in output.split('\n'):
+                if ln.strip():
+                    result.append(ln.strip())
+            return '\n'.join(result)
```
namhyung: Then you can remove this part.

utils/report.c
```diff
@@ -784,11 +784,16 @@ void report_sort_tasks(struct uftrace_data *handle, struct rb_root *name_root,
 						      .list = LIST_HEAD_INIT(                      \
 							      field_##_func.list) };
 
+static bool report_csv_output;
```
namhyung: Now I think you can remove this variable.

utils/report.c
```diff
@@ -784,11 +784,16 @@ void report_sort_tasks(struct uftrace_data *handle, struct rb_root *name_root,
...
+
 #define FIELD_TIME(_id, _name, _field, _func, _header)                                             \
 	static void print_##_func(struct field_data *fd)                                           \
 	{                                                                                          \
 		struct uftrace_report_node *node = fd->arg;                                        \
-		print_time_unit(node->_field);                                                     \
+		if (report_csv_output)                                                             \
```
namhyung: And use `if (format_mode == FORMAT_CSV)` instead.

tests/t302_report_csv.py
```diff
@@ -0,0 +1,50 @@
...
+
+        result = []
+
+        for ln in output.split('\n'):
+            if ln.strip() == '':
```
namhyung: Maybe like this.
```python
if ln.strip() == '' or ln.startswith('#'):
```

tests/t302_report_csv.py
```diff
@@ -0,0 +1,50 @@
...
+                continue
+
+            if ln.count(',') != 3:
+                return 'invalid csv field count'
+
+            line = [item.strip() for item in ln.split(',')]
+            if line[0].lstrip('#') == 'Total time':
+                continue
```
namhyung: Then you can remove this.

tests/t302_report_csv.py
```diff
@@ -0,0 +1,50 @@
...
+
+            if line[-1].startswith('__'):
+                continue
```
namhyung: I'd like to remove this line and add `self.option = '-F main'` in the prepare() function.

kangtegong:
Thank you for your feedback :)

I just (1) amended the test case, and (2) refactored the redundant variable `report_csv_output`!

namhyung (approved):
LGTM, thanks!

PR #2046, opened 2026-05-12
github: Fix exit code of nightly check_commits step

junyeong0619:
After the nightly workflow was changed to skip when no commits were merged to master, the check step returned a non-zero exit code in the "commits exist" case and caused the run to fail instead of proceeding to the tests.  Rewrite it as a plain 'if' so the step always exits cleanly and only sets should_run=false when there really are no recent commits.


junyeong0619:
Hi @namhyung, I just noticed the nightly test failed. Please take a look.

namhyung (approved):
LGTM, thanks for the quick fix.

PR #2047, opened 2026-05-18
replay: Show callsite location with --srcline

### LINKED ISSUE #2027: Show source line numbers for function callsites.
Thanks for the excellent tool — it has been extremely helpful in my workflow. I’m wondering whether `uftrace` can report the source line of each callsite, not just the source line of the callee’s definition.

Consider the following simple workflow:

``` Shell
gcc -o t-abc -g -pg tests/s-abc.c
uftrace record --srcline t-abc
uftrace replay --srcline
```
This produces output such as:
```
# DURATION     TID      FUNCTION [SOURCE]
   0.918 us [   5280] | __monstartup();
   0.570 us [   5280] | __cxa_atexit();
            [   5280] | main() { /* tests/s-abc.c:26 */
            [   5280] |   a() { /* tests/s-abc.c:11 */
            [   5280] |     b() { /* tests/s-abc.c:16 */
            [   5280] |       c() { /* tests/s-abc.c:21 */
   1.225 us [   5280] |         getpid();
   1.947 us [   5280] |       } /* c */
   2.500 us [   5280] |     } /* b */
   2.893 us [   5280] |   } /* a */
   3.316 us [   5280] | } /* main */
```
For example, in the line:
```
            [   5280] |   a() { /* tests/s-abc.c:11 */
```

`uftrace` shows the definition of `a()` at line 11.  
Is it possible for `uftrace` to also report the callsite location, such as `/* tests/s-abc.c:33 */`, indicating where `a()` was invoked within `main()`?

This would be very useful for understanding call relationships in more complex codebases.

Thanks for your time.

junyeong0619:
closes #2027

I've added the `callsite` trigger action proposed in the issue.
Setting `-T <func>@callsite` records the return address of the matched
call, and `uftrace replay --srcline` shows it after the existing source
line, e.g. `a() { /* file.c:11 from file.c:33 */`.

Changes:
* Added `TRIGGER_FL_CALLSITE` / `EVENT_ID_CALLSITE` and the `callsite`
  trigger action parser.
* libmcount emits the event (8-byte return address payload) right
  before the matching ENTRY when the trigger fires.
* Added `task_find_exact_loc_addr()` that uses libdw's
  `dwarf_addrdie()` + `dwarf_getsrc_die()` for PC-precise line lookup
  (the existing `task_find_loc_addr()` only knows the symbol's entry
  line). `load_debug_info()` now keeps libdw open at replay time so
  this lookup works there.
* replay intercepts the CALLSITE event, stashes the address on the
  task, and appends `from file:line` to the source comment on the
  next ENTRY.
* Added t305 covering both the live and `--srcline` paths.
* Documented the new action in `doc/uftrace-record.md` with an example
  and a short note about line precision under optimization.

Testing on Rocky Linux 9.7 (GCC 11.5, x86_64):

```
DURATION     TID      FUNCTION [SOURCE]
   0.510 us [3615820] | __monstartup();
   0.187 us [3615820] | __cxa_atexit();
            [3615820] | main() { /* s-abc.c:26 */
            [3615820] |   a() { /* s-abc.c:11 from s-abc.c:33 */
            [3615820] |     b() { /* s-abc.c:16 from s-abc.c:13 */
            [3615820] |       c() { /* s-abc.c:21 from s-abc.c:18 */
   0.250 us [3615820] |         getpid();
   1.449 us [3615820] |       } /* c */
   1.744 us [3615820] |     } /* b */
   2.317 us [3615820] |   } /* a */
   2.604 us [3615820] | } /* main */
```

One thing I'd like your feedback on:
`save_callsite_event()` writes the 8-byte payload with a direct
`*(uint64_t *)event->data = ...` rather than the existing
`mcount_memcpy4()`. Using the memcpy helper here recorded zero on my
setup. My guess is strict aliasing — `mcount_memcpy4` reinterprets the
source as `unsigned int *`, and the 8-byte return address write through
that path gets optimized away. The direct write goes through the
`uint8_t event->data[]` member, which is well-defined. I'm happy to add
an `mcount_memcpy8` helper (or update `mcount_memcpy4` to use byte-wise
loads) if you prefer either of those.

Also a precision note (also in the doc): the resolved call site line
depends on the dwarf line program. `-O0` / `-Og` builds report the
exact line; with `-O1` or higher the call instruction can be folded
with surrounding code so the reported line may shift by 1–2 lines.
`-finstrument-functions` builds are stable across optimization levels.
t305 forces `-O0` in cflags for this reason.

Please let me know if you'd like changes to the approach or naming.
Thanks!

### DIFF
--- a/cmds/info.c
+++ b/cmds/info.c
@@ -480,11 +480,11 @@ static int fill_taskinfo(void *arg)
 		.root = RB_ROOT,
 		.tasks = RB_ROOT,
 	};
+	char *dir = fha->opts->dirname;
 	int i;
 
-	if (read_task_txt_file(&link, fha->opts->dirname, fha->opts->dirname, false, false, false) <
-		    0 &&
-	    read_task_file(&link, fha->opts->dirname, false, false, false) < 0)
+	if (read_task_txt_file(&link, dir, dir, false, false, false, false) < 0 &&
+	    read_task_file(&link, dir, false, false, false, false) < 0)
 		return -1;
 
 	walk_tasks(&link, build_tid_list, &tlist);
@@ -1140,7 +1140,7 @@ void process_uftrace_info(struct uftrace_data *handle, struct uftrace_opts *opts
 
 		/* ignore errors */
 		read_task_txt_file(&handle->sessions, opts->dirname, opts->dirname, false, false,
-				   false);
+				   false, false);
 
 		process(data, "# %-20s: %d\n", "number of tasks", nr);
 
@@ -1401,7 +1401,7 @@ int command_info(int argc, char *argv[], struct uftrace_opts *opts)
 	if (opts->show_task) {
 		/* ignore errors */
 		read_task_txt_file(&handle.sessions, opts->dirname, opts->dirname, false, false,
-				   false);
+				   false, false);
 
 		if (handle.hdr.feat_mask & PERF_EVENT) {
 			if (setup_perf_data(&handle) == 0)
--- a/cmds/replay.c
+++ b/cmds/replay.c
@@ -731,6 +731,38 @@ next:
 	}
 }
 
+/*
+ * Fallback for the @callsite trigger when DWARF info isn't available:
+ * format the address as "symbol+offset" using the ELF symbol table.
+ * Returns a pointer to @buf on success, NULL when no symbol matches.
+ */
+static const char *format_callsite_symbol(struct uftrace_session_link *sessions,
+					  struct uftrace_task_reader *task, uint64_t time,
+					  uint64_t addr, char *buf, size_t bufsize)
+{
+	struct uftrace_session *sess;
+	struct uftrace_mmap *map;
+	struct uftrace_symbol *sym;
+	uint64_t mod_addr;
+
+	sess = find_task_session(sessions, task->t, time);
+	if (sess == NULL)
+		return NULL;
+
+	map = find_map(&sess->sym_info, addr);
+	if (map == NULL || map->mod == NULL)
+		return NULL;
+
+	mod_addr = addr - map->start;
+	sym = find_sym(&map->mod->symtab, mod_addr);
+	if (sym == NULL)
+		return NULL;
+
+	snprintf(buf, bufsize, "%s+%#lx", sym->name, (unsigned long)(mod_addr - sym->addr));
+	buf[bufsize - 1] = '\0';
+	return buf;
+}
+
 static int print_graph_rstack(struct uftrace_data *handle, struct uftrace_task_reader *task,
 			      struct uftrace_opts *opts)
 {
@@ -781,11 +813,40 @@ static int print_graph_rstack(struct uftrace_data *handle, struct uftrace_task_r
 		}
 	}
 
-	if (opts->srcline) {
-		loc = task_find_loc_addr(sessions, task, rstack->time, rstack->addr);
-		if (opts->comment && loc)
+	if (opts->comment) {
+		struct uftrace_dbg_loc csloc;
+		char cs_buf[128];
+		const char *cs_str = NULL;
+
+		if (opts->srcline)
+			loc = task_find_loc_addr(sessions, task, rstack->time, rstack->addr);
+
+		if (opts->callsite && rstack->type == UFTRACE_ENTRY && task->callsite_ip) {
+			if (task_find_exact_loc_addr(sessions, task, rstack->time,
+						     task->callsite_ip, &csloc)) {
+				snprintf(cs_buf, sizeof(cs_buf), "%s:%d", csloc.file->name,
+					 csloc.line);
+				cs_buf[sizeof(cs_buf) - 1] = '\0';
+				cs_str = cs_buf;
+			}
+			else {
+				/* fallback to function+offset when no dwarf info */
+				cs_str = format_callsite_symbol(sessions, task, rstack->time,
+								task->callsite_ip, cs_buf,
+								sizeof(cs_buf));
+			}
+		}
+
+		if (loc && cs_str)
+			xasprintf(&str_loc, "%s:%d from %s", loc->file->name, loc->line, cs_str);
+		else if (loc)
 			xasprintf(&str_loc, "%s:%d", loc->file->name, loc->line);
+		else if (cs_str)
+			xasprintf(&str_loc, "from %s", cs_str);
 	}
+
+	if (rstack->type == UFTRACE_ENTRY)
+		task->callsite_ip = 0;
 
 	if (rstack->type == UFTRACE_ENTRY) {
 		struct uftrace_task_reader *next = NULL;
@@ -944,6 +1005,13 @@ lost:
 		struct uftrace_task_reader *next = NULL;
 		struct uftrace_record rec = *rstack;
 		uint64_t evt_id = rstack->addr;
+
+		/* stash callsite return address for the next ENTRY to consume */
+		if (evt_id == EVENT_ID_CALLSITE) {
+			if (task->args.data && task->args.len >= sizeof(uint64_t))
+				memcpy(&task->callsite_ip, task->args.data, sizeof(uint64_t));
+			goto out;
+		}
 
 		depth = task->display_depth;
 
--- a/doc/uftrace-record.md
+++ b/doc/uftrace-record.md
@@ -459,7 +459,7 @@ The BNF for trigger specification is as follows:
     <actions>    :=  <action>  | <action> "," <actions>
     <action>     :=  "depth="<num> | "trace" | "trace_on" | "trace_off" |
                      "time="<time_spec> | "size="<num> | "read="<read_spec> |
-                     "finish" | "filter" | "notrace" | "recover"
+                     "finish" | "filter" | "notrace" | "recover" | "callsite" |
                      "filter" | "notrace" | "recover" | "if:"<cond_spec>
     <time_unit>  :=  "ns" | "nsec" | "us" | "usec" | "ms" | "msec" | "s" | "sec" | "m" | "min"
     <read_spec>  :=  "proc/statm" | "page-fault" | "pmu-cycle" | "pmu-cache" | "pmu-branch"
@@ -524,6 +524,29 @@ The results are printed as events (comments) like below.
 
 The 'finish' trigger is to end recording.  The process still can run and this
 can be useful to trace unterminated processes like daemon.
+
+The 'callsite' trigger records the return address of the matching function
+call so that the replay command can show where the function was called from.
+The trigger automatically saves debug info too, so a separate `--srcline`
+is not required at record time and `uftrace replay` shows the call site by
+default like below.
+
+    $ uftrace record -T 'a@callsite' ./abc
+    $ uftrace replay
+    # DURATION     TID     FUNCTION
+                [ 1234] | main() {
+                [ 1234] |   a() { /* from tests/s-abc.c:33 */
+                [ 1234] |     b() {
+                ...
+
+Combine it with `--srcline` to also show the callee's definition line:
+
+    $ uftrace replay --srcline
+                [ 1234] |   a() { /* tests/s-abc.c:11 from tests/s-abc.c:33 */
+
+The call site display can be suppressed with `--no-callsite` at replay
+time.  The reported line itself comes from the compiler's dwarf line
+program and is shown as-is.
 
 The 'filter' and 'notrace' triggers have same effect as `-F`/`--filter` and
 `-N`/`--notrace` options respectively.  And it can have a condition.
--- a/tests/t305_replay_callsite.py
+++ b/tests/t305_replay_callsite.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+import re
+
+from runtest import TestBase
+
+class TestCase(TestBase):
+    def __init__(self):
+        TestBase.__init__(self, 'abc', """
+# DURATION     TID     FUNCTION [SOURCE]
+   0.234 us [ 19939] | __cxa_atexit();
+            [ 19939] | main() {
+            [ 19939] |   a() { /* from s-abc.c:N */
+            [ 19939] |     b() { /* from s-abc.c:N */
+            [ 19939] |       c() { /* from s-abc.c:N */
+   1.120 us [ 19939] |         getpid();
+   1.697 us [ 19939] |       } /* c */
+   2.044 us [ 19939] |     } /* b */
+   2.329 us [ 19939] |   } /* a */
+   2.644 us [ 19939] | } /* main */
+""", cflags='-g')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not 'dwarf' in self.feature:
+            return TestBase.TEST_SKIP
+        return TestBase.build(self, name, cflags, ldflags)
+
+    def setup(self):
+        self.option = "-T '^.$@callsite'"
+
+    def runcmd(self):
+        # callsite events must not be filtered out by the default --no-event
+        return TestBase.runcmd(self).replace('--no-event', '')
+
+    def sort(self, output):
+        """ Replace exact line numbers with N to ignore compiler-dependent
+            shifts, and strip directory prefix from paths so the test is
+            stable across optimization levels and build environments. """
+        result = []
+        before_main = True
+        for ln in output.split('\n'):
+            if ln.find(' | main()') > 0:
+                before_main = False
+            if before_main:
+                continue
+            if ln.strip() == '':
+                break
+
+            rest = ln.split('|', 1)[-1]
+            # strip directory prefix from filenames and normalize line numbers
+            rest = re.sub(r'/[\w./-]+/(\S+\.c):\d+', r'\1:N', rest)
+            result.append(rest.rstrip())
+
+        return '\n'.join(result)
--- a/uftrace.h
+++ b/uftrace.h
@@ -305,6 +305,7 @@ struct uftrace_opts {
 	bool list_event;
 	bool event_skip_out;
 	bool no_event;
+	bool callsite;
 	bool no_sched;
 	bool no_sched_preempt;
 	bool nest_libcall;
@@ -354,9 +355,10 @@ static inline void close_data_file(struct uftrace_opts *opts, struct uftrace_dat
 }
 
 int read_task_file(struct uftrace_session_link *sess, char *dirname, bool needs_symtab,
-		   bool sym_rel_addr, bool needs_srcline);
+		   bool sym_rel_addr, bool needs_srcline, bool needs_callsite);
 int read_task_txt_file(struct uftrace_session_link *sess, char *dirname, char *symdir,
-		       bool needs_symtab, bool sym_rel_addr, bool needs_srcline);
+		       bool needs_symtab, bool sym_rel_addr, bool needs_srcline,
+		       bool needs_callsite);
 
 char *get_libmcount_path(struct uftrace_opts *opts);
 void put_libmcount_path(char *libpath);
@@ -484,7 +486,7 @@ extern struct uftrace_session *first_session;
 
 void create_session(struct uftrace_session_link *sess, struct uftrace_msg_sess *msg, char *dirname,
 		    char *symdir, char *exename, bool sym_rel_addr, bool needs_symtab,
-		    bool needs_srcline);
+		    bool needs_srcline, bool needs_callsite);
 struct uftrace_session *find_task_session(struct uftrace_session_link *sess,
 					  struct uftrace_task *task, uint64_t timestamp);
 void create_task(struct uftrace_session_link *sess, struct uftrace_msg_task *msg, bool fork);
@@ -494,7 +496,7 @@ void delete_session_map(struct uftrace_sym_info *sinfo);
 void update_session_map(const char *filename);
 struct uftrace_session *get_session_from_sid(struct uftrace_session_link *sess, char sid[]);
 void session_add_dlopen(struct uftrace_session *sess, uint64_t timestamp, unsigned long base_addr,
-			const char *libname, bool needs_srcline);
+			const char *libname, bool needs_srcline, bool needs_callsite);
 void session_setup_dlopen_argspec(struct uftrace_session *sess,
 				  struct uftrace_filter_setting *setting, bool is_retval);
 struct uftrace_dlopen_list *session_find_dlopen(struct uftrace_session *sess, uint64_t timestamp,
--- a/uftrace.h
+++ b/uftrace.h
@@ -63,6 +63,7 @@ enum uftrace_feat_bits {
 	DEBUG_INFO_BIT,
 	ESTIMATE_RETURN_BIT,
 	SYM_SIZE_BIT,
+	CALLSITE_BIT,
 
 	FEAT_BIT_MAX,
 
@@ -80,6 +81,7 @@ enum uftrace_feat_bits {
 	DEBUG_INFO = (1U << DEBUG_INFO_BIT),
 	ESTIMATE_RETURN = (1U << ESTIMATE_RETURN_BIT),
 	SYM_SIZE = (1U << SYM_SIZE_BIT),
+	CALLSITE = (1U << CALLSITE_BIT),
 };
 
 enum uftrace_info_bits {
--- a/uftrace.h
+++ b/uftrace.h
@@ -629,6 +629,7 @@ enum uftrace_event_id {
 	EVENT_ID_DIFF_PMU_BRANCH,
 	EVENT_ID_WATCH_CPU,
 	EVENT_ID_WATCH_VAR,
+	EVENT_ID_CALLSITE,
 
 	/* supported perf events */
 	EVENT_ID_PERF = 200000U,
--- a/uftrace.h
+++ b/uftrace.h
@@ -514,6 +514,8 @@ struct uftrace_symbol *task_find_sym_addr(struct uftrace_session_link *sess,
 struct uftrace_dbg_loc *task_find_loc_addr(struct uftrace_session_link *sess,
 					   struct uftrace_task_reader *task, uint64_t time,
 					   uint64_t addr);
+bool task_find_exact_loc_addr(struct uftrace_session_link *sess, struct uftrace_task_reader *task,
+			      uint64_t time, uint64_t addr, struct uftrace_dbg_loc *out);
 
 typedef int (*walk_sessions_cb_t)(struct uftrace_session *session, void *arg);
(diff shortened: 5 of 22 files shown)


namhyung:
Thanks for working on this!  A few comments before looking at the code.

I feel like requiring `--srcline` looks like unnecessary.  It'd be nice if it can always show callsite info once recorded.  Maybe we can add `--no-callsite` option to disable it at replay time.  Then the output becomes compact and can be used with `--srcline` if needed.

For memcpy, there's `mcount_memcpy1()` for byte-wise copy.  I'm not sure if it's a strict alias issue but unaligned access on ARM may cause problems.

Regarding the precision of the location, I think it's out of scope.  We can just show whatever compiler generated for the code and don't care about the value itself (unless it's definitely wrong).  Please don't change the test code or settings just to pass the test, it'd be better to leave it failing so that we can be aware of the issue - or we can check whether it generates any callsite info.

junyeong0619:
Thanks for the review!

1. Decouple callsite from `--srcline` so that running `uftrace record -T 'a@callsite'` followed by a plain `uftrace replay` shows the callsite info on its own, as an independent trigger output.

2. Switch the payload write to `mcount_memcpy1()`.

3. Keep the cflags untouched and adjust the test instead.

Will push the updates along these lines.

junyeong0619:
Pushed the revisions. Now we can use for 4 versions.

Usage:
```bash
$ uftrace record -T 'a@callsite' tests/t-abc 

$ uftrace replay
# DURATION     TID      FUNCTION
   0.714 us [ 189072] | __monstartup();
  12.921 us [ 189072] | __cxa_atexit();
            [ 189072] | main() {
            [ 189072] |   a() { /* from s-abc.c:33 */
            [ 189072] |     b() {
            [ 189072] |       c() {
   0.390 us [ 189072] |         getpid();
   0.989 us [ 189072] |       } /* c */
   1.357 us [ 189072] |     } /* b */
   1.905 us [ 189072] |   } /* a */
   2.186 us [ 189072] | } /* main */

$ uftrace replay --srcline
# DURATION     TID      FUNCTION [SOURCE]
   0.714 us [ 189072] | __monstartup();
  12.921 us [ 189072] | __cxa_atexit();
            [ 189072] | main() { /* s-abc.c:26 */
            [ 189072] |   a() { /* s-abc.c:11 from s-abc.c:33 */
            [ 189072] |     b() { /* s-abc.c:16 */
            [ 189072] |       c() { /* s-abc.c:21 */
   0.390 us [ 189072] |         getpid();
   0.989 us [ 189072] |       } /* c */
   1.357 us [ 189072] |     } /* b */
   1.905 us [ 189072] |   } /* a */
   2.186 us [ 189072] | } /* main */

$ uftrace replay --no-callsite
# DURATION     TID      FUNCTION
   0.714 us [ 189072] | __monstartup();
  12.921 us [ 189072] | __cxa_atexit();
            [ 189072] | main() {
            [ 189072] |   a() {
            [ 189072] |     b() {
            [ 189072] |       c() {
   0.390 us [ 189072] |         getpid();
   0.989 us [ 189072] |       } /* c */
   1.357 us [ 189072] |     } /* b */
   1.905 us [ 189072] |   } /* a */
   2.186 us [ 189072] | } /* main */

$ uftrace replay --srcline --no-callsite
# DURATION     TID      FUNCTION [SOURCE]
   0.714 us [ 189072] | __monstartup();
  12.921 us [ 189072] | __cxa_atexit();
            [ 189072] | main() { /* s-abc.c:26 */
            [ 189072] |   a() { /* s-abc.c:11 */
            [ 189072] |     b() { /* s-abc.c:16 */
            [ 189072] |       c() { /* s-abc.c:21 */
   0.390 us [ 189072] |         getpid();
   0.989 us [ 189072] |       } /* c */
   1.357 us [ 189072] |     } /* b */
   1.905 us [ 189072] |   } /* a */
   2.186 us [ 189072] | } /* main */
```

Also Treated the write with mcount_memcpy1().

junyeong0619:
Rewrite the commit message for the changes.

namhyung (commented):
I have some concerns if it would
* work well with other events and arguments.  (I'm a bit worried about the timestamp management now)
* print "function+offset" when DWARF is not available.

Also please pick a single prefix in the commit subject line.  More general and/or important one would be preferred.  Maybe you want to split the commit otherwise.

junyeong0619:
Here is full dump and replay output.
```bash
root@server:~/uftrace# uftrace record -W var:mydata -T 'bar@callsite' -T 'foo@callsite' t-watch
root@server:~/uftrace# uftrace dump | head -50
uftrace file header: magic         = 4674726163652100
uftrace file header: version       = 4
uftrace file header: header size   = 40
uftrace file header: features      = 0x1763 (PLTHOOK | TASK_SESSION | SYM_REL_ADDR | MAX_STACK | PERF_EVENT | AUTO_ARGS | DEBUG_INFO | SYM_SIZE)
uftrace file header: info          = 0xfbff (EXE_NAME | EXE_BUILD_ID | EXIT_STATUS | CMDLINE | CPUINFO | MEMINFO | OSINFO | TASKINFO | USAGEINFO | LOADINFO | RECORD_DATE | PATTERN_TYPE | VERSION | UTC_OFFSET | SYSCTL)

reading 1355229.dat
646618.687034679  1355229: [entry] __monstartup(62efea147090) depth: 0
646618.687036364  1355229: [exit ] __monstartup(62efea147090) depth: 0
646618.687037419  1355229: [entry] __cxa_atexit(62efea1470a0) depth: 0
646618.687037792  1355229: [exit ] __cxa_atexit(62efea1470a0) depth: 0
646618.687038474  1355229: [entry] main(62efea147269) depth: 0
646618.687038640  1355229: [event] trigger:callsite(186ad) depth: 0
646618.687038640  1355229: [data ] length = 8
  trigger:callsite: callsite=0x62efea14729d
646618.687038641  1355229: [entry] foo(62efea147233) depth: 1
646618.687038895  1355229: [event] trigger:callsite(186ad) depth: 0
646618.687038895  1355229: [data ] length = 8
  trigger:callsite: callsite=0x62efea14724a
646618.687038895  1355229: [event] watch:var(186ac) depth: 0
646618.687038895  1355229: [data ] length = 12
  watch:var: mydata=1
646618.687038896  1355229: [entry] bar(62efea14720b) depth: 2
646618.687040199  1355229: [exit ] bar(62efea14720b) depth: 2
646618.687041376  1355229: [event] watch:var(186ac) depth: 0
646618.687041376  1355229: [data ] length = 12
  watch:var: mydata=2
646618.687041377  1355229: [exit ] foo(62efea147233) depth: 1
646618.687041612  1355229: [exit ] main(62efea147269) depth: 0
reading perf-cpu4.dat
646618.675870741  1355229: [event] linux:task-name(200006)
646618.685026396  1355229: [event] linux:sched-out (pre-empted)(200007)
646618.685077278  1355229: [event] linux:sched-in(200001)
646618.687172426  1355229: [event] linux:task-exit(200005)

root@server:~/uftrace# uftrace replay
# DURATION     TID      FUNCTION
   1.685 us [1355229] | __monstartup();
   0.373 us [1355229] | __cxa_atexit();
            [1355229] | main() {
            [1355229] |   foo() { /* from tests/s-watch-global.c:26 */
            [1355229] |     /* watch:var (mydata=1) */
   1.303 us [1355229] |     bar(); /* from tests/s-watch-global.c:15 */
            [1355229] |     /* watch:var (mydata=2) */
   2.736 us [1355229] |   } /* foo */
   3.138 us [1355229] | } /* main */
```

junyeong0619:
Followed up with a deliberate-collision case using watch:var on a global variable changed inside the traced functions 

(tests/s-watch-global.c):
```bash
  $ uftrace record -W var:mydata -T 'foo@callsite' -T 'bar@callsite' t-watch
  $ uftrace dump | head -40

  646618.687038895  [event] trigger:callsite        # entry-1 for bar
  646618.687038895  [event] watch:var (mydata=1)    # same nanosecond
  646618.687038896  [entry] bar
```

Two events do share a timestamp here.  The ordering is still deterministic: `mtdp->event[]` is push-ordered, and
`mcount_entry_filter_record()` calls `save_callsite_event()` before `save_watchpoint()`, so the callsite is written to the trace first and replay processes them in the same order.

The resulting replay output keeps both visible without any mixing:

```bash
  $ uftrace replay
              [1355229] | main() {
              [1355229] |   foo() { /* from tests/s-watch-global.c:26 /
              [1355229] |     / watch:var (mydata=1) /
     1.303 us [1355229] |     bar(); / from tests/s-watch-global.c:15 /
              [1355229] |     / watch:var (mydata=2) /
     2.736 us [1355229] |   } / foo /
     3.138 us [1355229] | } / main */
```

junyeong0619:
Also treated the case that Dwarf is not available.

```bash
root@server:~/uftrace# uftrace record -T 'a@callsite' tests/t-abc-nodbg (t-abc with no -g)
root@server:~/uftrace# uftrace replay
# DURATION     TID      FUNCTION
   0.632 us [1360511] | __monstartup();
   0.282 us [1360511] | __cxa_atexit();
            [1360511] | main() {
            [1360511] |   a() { /* from main+0x41 */
            [1360511] |     b() {
            [1360511] |       c() {
   0.265 us [1360511] |         getpid();
   0.801 us [1360511] |       } /* c */
   1.066 us [1360511] |     } /* b */
   1.477 us [1360511] |   } /* a */
   1.747 us [1360511] | } /* main */
```

namhyung:
Thanks, I've check it with other options and it has some side effects.

```
$ uftrace -a -F main ~/tmp/progs/abc
# DURATION     TID      FUNCTION
            [  10623] | main(1, 0x7ffd623464f8) {
            [  10623] |   a() {
            [  10623] |     b() {
            [  10623] |       c() {
   0.427 us [  10623] |         getpid() = 10623;
   2.495 us [  10623] |       } = 10623; /* c */
   2.913 us [  10623] |     } = 10624; /* b */
   3.169 us [  10623] |   } = 10623; /* a */
   3.950 us [  10623] | } = 0; /* main */
```

```
$ uftrace -T a@read=pmu-cycle -F main ~/tmp/progs/abc
# DURATION     TID      FUNCTION
            [  10657] | main() {
            [  10657] |   a() {
            [  10657] |     /* read:pmu-cycle (cycles=604 instructions=48) */
            [  10657] |     b() {
            [  10657] |       c() {
   0.582 us [  10657] |         getpid();
   1.334 us [  10657] |       } /* c */
   1.579 us [  10657] |     } /* b */
            [  10657] |     /* diff:pmu-cycle (cycles=+5857 instructions=+4554 IPC=0.78) */
  24.601 us [  10657] |   } /* a */
  30.448 us [  10657] | } /* main */
```

And if they are used together, it doesn't work well.
```
$ uftrace -a -T a@read=pmu-cycle,callsite -F main ~/tmp/progs/abc
# DURATION     TID      FUNCTION
            [  10690] | main(1, 0x7ffd169dab78) {
            [  10690] |   a() { /* from tests/s-abc.c:33 */
            [  10690] |     /* read:pmu-cycle (cycles=591 instructions=48) */
            [  10690] |     b() {
            [  10690] |       c() {
   0.446 us [  10690] |         getpid() = 10690;
   2.304 us [  10690] |       } = 10690; /* c */
   2.678 us [  10690] |     } = 10691; /* b */
  23.815 us [  10690] |   } = 10690; /* a */
  24.946 us [  10690] | } = 0; /* main */
```

So it lost the diff:pmu-cycle event.  Removing the argument (`-a`) gives me the correct result.
```
$ uftrace -T a@read=pmu-cycle,callsite -F main ~/tmp/progs/abc
# DURATION     TID      FUNCTION
            [  10710] | main() {
            [  10710] |   a() { /* from tests/s-abc.c:33 */
            [  10710] |     /* read:pmu-cycle (cycles=582 instructions=48) */
            [  10710] |     b() {
            [  10710] |       c() {
   0.468 us [  10710] |         getpid();
   1.228 us [  10710] |       } /* c */
   1.496 us [  10710] |     } /* b */
            [  10710] |     /* diff:pmu-cycle (cycles=+6834 instructions=+4784 IPC=0.70) */
  25.378 us [  10710] |   } /* a */
  30.669 us [  10710] | } /* main */
```

junyeong0619:
Thanks for review and testing! The lost diff event turned out to be a pre-existing bug from combining -a/-A/-R with @read, not from the callsite trigger. I sent a separate PR with the fix: #2048.

utils/event.c
```diff
@@ -121,6 +121,9 @@ char *event_get_name(struct uftrace_data *handle, unsigned evt_id)
 		case EVENT_ID_WATCH_VAR:
 			xasprintf(&evt_name, "watch:var");
 			break;
+		case EVENT_ID_CALLSITE:
+			xasprintf(&evt_name, "trigger:callsite");
```
namhyung: Let's show it as just "callsite".

utils/data-file.c
```diff
@@ -456,17 +456,25 @@ int open_data_file(struct uftrace_opts *opts, struct uftrace_data *handle)
...
 
 		if (handle->hdr.feat_mask & SYM_REL_ADDR)
 			sym_rel = true;
 
+		/*
+		 * also load debug info if the trace has it (e.g. from a
+		 * @callsite trigger), unless the user opted out
+		 */
+		needs_dbg = opts->srcline || (opts->loc_filter != NULL);
+		if (!opts->no_callsite && (handle->hdr.feat_mask & DEBUG_INFO))
```
namhyung: The location filter check is new.  Also I think checking DEBUG_INFO is not enough and you need to add a new CALLSITE feature bit.  Loading dwarf file should depend on the CALLSITE flag.

Also opts->no_callsite doesn't exist at this point which means it'd break the build and bisect.

utils/dwarf.c
```diff
@@ -1768,8 +1768,82 @@ static void build_dwarf_info(struct uftrace_dbg_info *dinfo, struct uftrace_symt
 	free(ret_patt);
 }
 
+/*
+ * match the trailing path against existing entries so callsite filenames
+ * stay consistent with the --srcline output
+ */
+static struct uftrace_dbg_file *find_or_get_debug_file(struct uftrace_dbg_info *dinfo,
```
namhyung: Maybe rename to `get_debug_file_suffix()`?

utils/session.c
```diff
@@ -782,45 +780,85 @@ struct uftrace_dbg_loc *task_find_loc_addr(struct uftrace_session_link *sessions
...
+	loc = &dinfo->locs[sym_idx];
+	if (loc->file != NULL)
+		return loc;
+
 	return NULL;
 }
 
+/*
+ * PC-precise variant of task_find_loc_addr() — returns the source line
+ * at the instruction, not the enclosing function's entry
```
namhyung: It returns a boolean.  Also you need to mention that it uses DWARF directly.

cmds/replay.c
```diff
@@ -781,12 +781,58 @@ static int print_graph_rstack(struct uftrace_data *handle, struct uftrace_task_r
...
+		const char *cs_str = NULL;
+
+		if (opts->srcline)
+			loc = task_find_loc_addr(sessions, task, rstack->time, rstack->addr);
+
+		if (!opts->no_callsite && rstack->type == UFTRACE_ENTRY && task->callsite_ip) {
+			if (task_find_exact_loc_addr(sessions, task, rstack->time,
+						     task->callsite_ip, &csloc)) {
+				snprintf(cs_buf, sizeof(cs_buf), "%s:%d", csloc.file->name,
+					 csloc.line);
```
namhyung, line 828: You need to make sure it has NUL character at the end.

cmds/replay.c
```diff
@@ -781,12 +781,58 @@ static int print_graph_rstack(struct uftrace_data *handle, struct uftrace_task_r
...
+						uint64_t mod_addr =
+							task->callsite_ip - cs_map->start;
+						struct uftrace_symbol *cs_sym;
+
+						cs_sym = find_sym(&cs_map->mod->symtab, mod_addr);
+						if (cs_sym) {
+							snprintf(cs_buf, sizeof(cs_buf), "%s+%#lx",
+								 cs_sym->name,
+								 (unsigned long)(mod_addr -
+										 cs_sym->addr));
```
namhyung: Doesn't look pretty.  Can we reduce the indention level or factor it out as a function?

uftrace.h
```diff
@@ -303,6 +303,7 @@ struct uftrace_opts {
 	bool list_event;
 	bool event_skip_out;
 	bool no_event;
+	bool no_callsite;
```
namhyung: Personally, I'd like to keep the options as positive even if the option name itself is negative.  For example, I think it's natural to check `if (opts->callsite)` rather than `if (!opts->no_callsite)`.

I don't plan to change the existing ones, but I want new ones  in positive form.

tests/t305_replay_callsite.py
```diff
@@ -0,0 +1,56 @@
...
+   2.644 us [ 19939] | } /* main */
+""", cflags='-g')
+
+    def build(self, name, cflags='', ldflags=''):
+        if not 'dwarf' in self.feature:
+            return TestBase.TEST_SKIP
+        return TestBase.build(self, name, cflags, ldflags)
+
+    def setup(self):
+        self.option = '-T a@callsite -T b@callsite -T c@callsite'
```
namhyung: Couldn't it be `-T '^.$@callsite'`?

tests/t305_replay_callsite.py
```diff
@@ -0,0 +1,56 @@
...
+            if before_main:
+                continue
+            if ln.strip() == '':
+                break
+
+            rest = ln.split('|', 1)[-1]
+            # strip directory prefix from filenames in comments
+            rest = re.sub(r'/[\w./-]+/(\S+\.c)', r'\1', rest)
+            # normalize source line numbers
+            rest = re.sub(r'(\.c):\d+', r'\1:N', rest)
```
namhyung: Maybe you can just combine two lines as `re.sub(r'/[\w./-]+/(\S+\.c):\d+', r'\1:N', rest)`.

namhyung (commented):
Thanks for the update!

utils/event.c
junyeong0619: Changed it to callsite.

utils/data-file.c
junyeong0619: Added a CALLSITE feature bit (dwarf auto-load now depends on it), reverted the accidental loc_filter check in read_task_file(), and verified the entire series bisects cleanly.

utils/dwarf.c
junyeong0619: Renamed it to get_debug_file_suffix.

utils/session.c
junyeong0619: Updated the comment to mention DWARF usage and the bool return.

cmds/replay.c
junyeong0619: Added explicit cs_buf[sizeof(cs_buf) - 1] = '\0' after the snprintf (and same in the function+offset helper).

cmds/replay.c
junyeong0619: Extracted into a format_callsite_symbol() helper.

uftrace.h
junyeong0619: Treated it as positive.

tests/t305_replay_callsite.py
junyeong0619: Done. Tests still pass.

tests/t305_replay_callsite.py
junyeong0619: Done.

doc/uftrace-record.md
```diff
@@ -525,6 +525,29 @@ The results are printed as events (comments) like below.
...
+
+Combine it with `--srcline` to also show the callee's definition line:
+
+    $ uftrace replay --srcline
+                [ 1234] |   a() { /* tests/s-abc.c:11 from tests/s-abc.c:33 */
+
+The call site display can be suppressed with `--no-callsite` at replay
+time.  The reported line itself comes from the compiler's dwarf line
+program and is shown as-is.
+
```
namhyung, line 550: Please add this to uftrace live man page too.

utils/data-file.c
```diff
@@ -456,17 +456,26 @@ int open_data_file(struct uftrace_opts *opts, struct uftrace_data *handle)
...
+		callsite_dbg = opts->callsite && (handle->hdr.feat_mask & CALLSITE);
+
 		/* read old task file first and then try task.txt file */
-		if (read_task_file(sessions, opts->dirname, true, sym_rel, opts->srcline) < 0 &&
+		if (read_task_file(sessions, opts->dirname, true, sym_rel,
+				   opts->srcline | callsite_dbg) < 0 &&
 		    read_task_txt_file(sessions, opts->dirname, opts->with_syms ?: opts->dirname,
 				       true, sym_rel,
-				       opts->srcline | (opts->loc_filter != NULL)) < 0) {
+				       opts->srcline | (opts->loc_filter != NULL) | callsite_dbg) <
```
namhyung: I think it's better to split callsite and srcline/loc-filter as the latter only requires the dbg file.  You may introduce a local variable for srcline/loc-filter like `needs_srcline`.  That means you need to pass it to `load_debug_info()` through `create_session()`.

utils/dwarf.c
```diff
@@ -2253,6 +2328,14 @@ void load_debug_info(struct uftrace_sym_info *sinfo, bool needs_srcline)
...
 					needs_srcline);
 		}
+
+		/*
+		 * Keep libdw open per-module so that PC-level line lookups
+		 * (used by the @callsite trigger) can query the dwarf line
+		 * program at replay time.  finish_debug_info() releases it.
+		 */
+		if (needs_srcline && dinfo->dw == NULL)
+			setup_dwarf_info(map->libname, dinfo, map->start, true);
```
namhyung, line 2347: I think this part should belong to the next commit.  Also I don't think it's using callsite info precisely.

namhyung (commented):
Thanks for the update, I think it's almost there.  Please make sure to rebase your changes on to the latest commit whenever you push them.

utils/dwarf.c
junyeong0619: To make this precise, would you prefer introducing a separate flag like needs_callsite_dbg?

For example:

```c
if (needs_callsite_dbg && dinfo->dw == NULL)
     setup_dwarf_info(map->libname, dinfo, map->start, true);
```

namhyung: Yes, something like `needs_callsite`.

utils/data-file.c
junyeong0619: For this to reach load_debug_info(), create_session() would need to take both flags:

```c
create_session(sess, &smsg, dirname, symdir, exename, sym_rel_addr,
                needs_symtab, needs_srcline, needs_callsite_dbg);
```

Does that look right?

utils/dwarf.c
junyeong0619: Thanks for the quick review. I will change the check line in that way.

doc/uftrace-record.md
junyeong0619: Done.

junyeong0619:
Rebased onto master. If there is something wrong let me know. Thanks.

cmds/info.c
```diff
@@ -482,9 +482,9 @@ static int fill_taskinfo(void *arg)
 	};
 	int i;
 
-	if (read_task_txt_file(&link, fha->opts->dirname, fha->opts->dirname, false, false, false) <
-		    0 &&
-	    read_task_file(&link, fha->opts->dirname, false, false, false) < 0)
+	if (read_task_txt_file(&link, fha->opts->dirname, fha->opts->dirname, false, false, false,
+			       false) < 0 &&
```
namhyung: Looks like a whitespace error.

utils/data-file.c
```diff
@@ -456,17 +457,25 @@ int open_data_file(struct uftrace_opts *opts, struct uftrace_data *handle)
...
+		needs_callsite = opts->callsite && (handle->hdr.feat_mask & CALLSITE);
+
 		/* read old task file first and then try task.txt file */
-		if (read_task_file(sessions, opts->dirname, true, sym_rel, opts->srcline) < 0 &&
+		if (read_task_file(sessions, opts->dirname, true, sym_rel, opts->srcline,
+				   needs_callsite) < 0 &&
 		    read_task_txt_file(sessions, opts->dirname, opts->with_syms ?: opts->dirname,
-				       true, sym_rel,
-				       opts->srcline | (opts->loc_filter != NULL)) < 0) {
+				       true, sym_rel, opts->srcline | (opts->loc_filter != NULL),
```
namhyung: Similarly, let's introduce `needs_srcline` variable.

utils/dwarf.c
```diff
@@ -2326,8 +2335,16 @@ void load_debug_info(struct uftrace_sym_info *sinfo, bool needs_srcline)
 
 		if (!debug_info_has_location(dinfo) && !debug_info_has_argspec(dinfo)) {
 			load_debug_file(dinfo, stab, sinfo->symdir, map->libname, map->build_id,
-					needs_srcline);
+					needs_srcline || needs_callsite);
```
namhyung: Ditto.

utils/dwarf.c
```diff
@@ -2297,19 +2297,28 @@ static int load_debug_file(struct uftrace_dbg_info *dinfo, struct uftrace_symtab
...
+			    bool needs_callsite)
 {
 	struct uftrace_dbg_info *dinfo;
 
 	dinfo = &mod->dinfo;
 
 	if (!debug_info_has_location(dinfo) && !debug_info_has_argspec(dinfo)) {
 		load_debug_file(dinfo, &mod->symtab, dirname, mod->name, mod->build_id,
-				needs_srcline);
+				needs_srcline || needs_callsite);
```
namhyung: I don't think this is necessary as callsite doesn't use the debug (.dbg) file.

cmds/info.c
junyeong0619: Whitespace is up to clang-format, so I shortened the line by introducing
char *dir = fha->opts->dirname;. Under 100 columns, clang-format leaves it alone.

utils/data-file.c
junyeong0619: Done — introduced needs_srcline = opts->srcline | (opts->loc_filter != NULL).

utils/dwarf.c
junyeong0619: Right, callsite doesn't read the .dbg file. Dropped needs_callsite from load_debug_file() in both load_debug_info() and load_module_debug_info().

namhyung (approved):
LGTM!  I'm curious when you sleep.. take care. :)

junyeong0619:
Haha, just happened to wake up and couldn't help picking it up :) Thanks for the kind words — and for the thorough reviews!

PR #2053, opened 2026-06-03
python: Use HDRSZ macro in symtab header size warning

junyeong0619:
The warning hardcoded "64" but UFTRACE_PYTHON_SYMTAB_HDRSZ is (48). Use the macro to keep them in sync, matching write_dbginfo().

I found it while i'm look around the python tracing codes.


namhyung (approved):
LGTM

PR #2057, opened 2026-06-17
compiler: Implement memory barriers for RISC-V

### LINKED ISSUE #2056: RISC-V: memory barriers in utils/compiler.h are defined as no-ops
On RISC-V, the memory barrier macros in `utils/compiler.h` are currently defined as plain `nop`, so they provide no ordering guarantee:

```c
#if defined(__riscv)
#define cpu_relax()            asm volatile("nop" ::: "memory")
#define full_memory_barrier()  asm volatile("nop" ::: "memory")
#define read_memory_barrier()  asm volatile("nop" ::: "memory")
#define write_memory_barrier() asm volatile("nop" ::: "memory")
#endif
```

These macros are used for lock-free shared-memory synchronization, e.g.:

- `libmcount/wrap.c` — `write_memory_barrier()` (writer side)
- `utils/perf.c` / `utils/symbol.h` — `read_memory_barrier()` / `full_memory_barrier()` (reader side)

RISC-V uses a weak memory model (RVWMO), so a `nop` does not order the accesses. On a multi-core RISC-V target this can let the reader observe reordered/stale data from the shared ring buffers (a latent data race), whereas on x86 (TSO) the same `nop` happens to be harmless.

The `fence` instruction (I extension) provides the needed ordering. Matching the Linux kernel's SMP barriers:

- `full_memory_barrier()` → `fence rw, rw`
- `read_memory_barrier()`  → `fence r, r`
- `write_memory_barrier()` → `fence w, w`

(Note: this is about *data* ordering only. The `fence.i` instruction for instruction-cache sync during runtime code patching is a separate concern, relevant to dynamic tracing in #1862.)

I found this while working on RISC-V dynamic tracing (#1862). I have a fix ready and can send a PR if this looks correct.

dlgus8648:
The RISC-V memory barrier macros in `utils/compiler.h` were defined as plain `nop`, which provides no ordering on RISC-V's weak memory model (RVWMO). They are used for lock-free shared-memory synchronization (the perf ring buffer in `utils/perf.c` and `utils/symbol.h`, and the libmcount writer in `libmcount/wrap.c`), so on a multi-core RISC-V target the reader could observe reordered or stale data.

This uses the `fence` instruction to provide the proper ordering, matching the Linux kernel SMP barriers:

- `full_memory_barrier()` → `fence rw, rw`
- `read_memory_barrier()`  → `fence r, r`
- `write_memory_barrier()` → `fence w, w`

`cpu_relax()` becomes a compiler barrier. The `fence.i` instruction for instruction-cache sync during runtime code patching is a separate concern and is not handled here.

Resolves #2056

### DIFF
--- a/utils/compiler.h
+++ b/utils/compiler.h
@@ -37,28 +37,23 @@
 #endif
 #endif
 
-/* TODO: not implemented yet (Start)
- *
- *	From RISC-V's "The RISC-V Instruction Set Manual, Volume I: User-Level
- *	ISA, Document Version 20191213", the G Extension consists of "I, M, A,
- *	F, D, Zicsr, Zifencei".
- *
- *	In RISC-V, two instructions exist for memory barriers: the fence
- *	instruction, defined in the I Extension, and the fence.i instruction,
- *	defined in the Zifencei Extension.
- *
- *	So the memory barrier commands we can use are fence, fence.i, and
- *	we'll have to figure out which one we need later when we implement
- *	the functions that call those macro functions.
- *
- */
 #if defined(__riscv)
-#define cpu_relax() asm volatile("nop" ::: "memory")
-#define full_memory_barrier() asm volatile("nop" ::: "memory")
-#define read_memory_barrier() asm volatile("nop" ::: "memory")
-#define write_memory_barrier() asm volatile("nop" ::: "memory")
+/*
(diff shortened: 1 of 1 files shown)


namhyung (approved):
LGTM

PR #2061, opened 2026-07-08
misc: Print enum definitions in dbginfo

kangtegong:
The enum definitions parsed from DWARF were collected but never printed, so the enum types behind argspecs like @arg1/e:color could not be seen.

So this patch prints them ahead of the per-symbol info.

### DIFF
--- a/misc/dbginfo.c
+++ b/misc/dbginfo.c
@@ -14,7 +14,8 @@ void print_debug_info(struct uftrace_dbg_info *dinfo, bool auto_args)
 	char *argspec = NULL;
 	char *retspec = NULL;
 
-	/* TODO: print enum definitions */
+	/* print enum definitions */
+	save_enum_def(&dinfo->enums, stdout);
 
 	for (i = 0; i < dinfo->nr_locs; i++) {
 		struct uftrace_dbg_loc *loc = &dinfo->locs[i];


namhyung:
It's not clear what changed.  Can you please provide an example output before and after?

kangtegong:
Sure! :) dbginfo is a standalone debugging tool that's run directly rather than called from other code, so this is a trivial change that resolves the [TODO comment](https://github.com/namhyung/uftrace/blob/34daaabf4c966bdd473e9117f6cd2d96fa8b9d3d/misc/dbginfo.c#L17).

The enum definitions are already parsed into `dinfo->enums` but were never printed, so the enum behind an argspec like `@arg1/e:color` couldn't be seen. This patch just prints them ahead of the per-symbol info using the existing `save_enum_def()`.

Example (enumtest.c):

```c
enum color { RED, GREEN = 5, BLUE };

int set_color(enum color c)
{
      return (int)c;
}

int main(void)
{
      return set_color(GREEN);
}
```

```
gcc -g -O0 -o enumtest enumtest.c
./misc/dbginfo -a ./enumtest
```

Before:

```
set_color [addr: 1129]
  srcline: enumtest.c:3
  argspec: @arg1/e:color
  retspec: @retval
main [addr: 1139]
  srcline: enumtest.c:8
  retspec: @retval
```

After:

```
E: enum color {RED,GREEN=5,BLUE}   // Added 
set_color [addr: 1129]
  srcline: enumtest.c:3
  argspec: @arg1/e:color
  retspec: @retval
main [addr: 1139]
  srcline: enumtest.c:8
  retspec: @retval
```

namhyung (approved):
LGTM

honggyukim:
Hi @kangtegong, it'd be much easier to understand if those before and after example is shown in the commit message with the complete example c file together.

kangtegong:
Thanks @honggyukim! I've updated the commit message to include the full enumtest.c example along with the before/after output :)

honggyukim (approved):
LGTM. Thanks @kangtegong!

Issue #1060, opened 2019-12-05
use SQLite3 to save Uftrace datas to use ability of SQL and Database.


ParkHanbum:
for currently, uftrace use file to save uftrace datas which contains symbols and tracing information. this is good for portable and independetable. but bad for useable. 

user cannot use tracing datas as way not supported by uftrace. uftrace support many mining way for tracing data. but it's usability is always limited obviusly.

so, it is better to change a way of saving uftrace datas. I think SQLite is good for this. it is Lightweight Database based on File. and support C also. 

so, it can keep the portability like uftrace. and not need overhead for Database such as database deamon. and it can provide common SQL to users. so, even user who did not know uftrace data format, easily mining information from uftrace datas.

how do you think so?

DanielTimLee:
I may not following,

But I'm not sure of improvement of 'usability' by using database.

I guess the reason you comment like this is because uftrace is using its own data tracing format which is hard to use for other use-cases.
(Ex. Importing other tracing data to uftrace.
Or Exporting uftraced traced data to other)

> user cannot use tracing datas as way not supported by uftrace.


IMHO,

namhyung:
@ParkHanbum do you have any specific usecase?  I don't know how many people is good at SQL.

lu-zero:
Saving the tracing data has the first requirement of taking the least amount of it, I doubt sqlite3 is quicker than the status quo.

Converting the data to a sqlite3 db could happen at a later time

namhyung:
Sure, we can have a script to convert/export data to sqlite3.

elfring:
How will the clarification evolve for [the corresponding database schema](https://en.wikipedia.org/wiki/Database_schema "Description for the usage of database schemas")? :thinking:

namhyung:
@ParkHanbum I think we can keep this as a feature request as some people seem to be interested in this.

PR #2062, opened 2026-07-09
script: Support the Firefox Profiler

### LINKED ISSUE #1456: support firefox profiler with uftrace dump --firefox
uftrace currently support chrome trace viewer format with `uftrace dump --chrome` and that is really useful to visualize the workload.

However, there are some requests to support firefox profiler format as well. We can add another firefox option and it can be used as `uftrace dump --firefox`.

The firefox profiler file can be opened at https://profiler.firefox.com/ and the most interesting thing is that it supports workload comparison at https://profiler.firefox.com/compare/.

kangtegong:
Add scripts/dump-firefox.py to convert uftrace data into the Firefox Profiler's Gecko format, which can be opened at
https://profiler.firefox.com/.

Based on the kernel's perf gecko.py.  perf has samples, uftrace has enter/exit, so this makes samples from the enter/exit.

Example usage:
  $ uftrace record <program>
  $ uftrace script -S scripts/dump-firefox.py
  \# open uftrace.data.firefox-trace at https://profiler.firefox.com/

Closes: #1456


scripts/dump-firefox.py
```diff
@@ -0,0 +1,310 @@
...
+		frame_id = self.frameMap.get(frame_str)
+		if frame_id is not None:
+			return frame_id
+		frame_id = len(self.frameTable)
+		self.frameMap[frame_str] = frame_id
+		string_id = self._intern_string(frame_str)
+
+		symbol_name_to_category = KERNEL_CATEGORY_INDEX if frame_str.find('kallsyms') != -1 \
+		or frame_str.find('/vmlinux') != -1 \
+		or frame_str.endswith('.ko)') \
```
namhyung: Have you tested kernel symbols?  I'm not sure if the frame_str would have module names.. anyway, I suspect uftrace frames would have "/vmlinux" or ".ko".

kangtegong: Thank you for your feedback :) Ah, that was my mistake. I had not tested it with kernel symbols, and after doing so, I confirmed that you were right. It turns out that the function name alone does not clearly indicate whether it is a kernel or user function: unlike perf, uftrace does not attach a `/vmlinux` or `.ko` marker and simply passes the bare name, such as `x64_sys_call`, so the name matching classified every kernel frame as a user frame.

This can be solved by exposing the [`is_kernel_record()`](https://github.com/namhyung/uftrace/blob/a0e540779d8ea497bb38bcc391e0644a1878b938/utils/fstack.h#L129)result as a `kernel` field in the script context. For reuse by future scripts (e.g. [SQLite](https://github.com/namhyung/uftrace/issues/1060)), I put this change in a separate commit and updated `dump-firefox.py` to use `ctx["kernel"]` in Python and `ctx.kernel` in Lua.

namhyung (approved):
Sorry for the delay.  I'm not sure about the prefix_id handling but it shouldn't matter.  Thanks for working on this.

LGTM

PR #1974, opened 2024-09-29
tui: Fix TUI display issues in non-UTF-8 locales

GabrielKimm:
This commit addresses the issue #1870 where the TUI displays incorrect characters in non-UTF-8 locales. The fix includes:

- Checking the current locale
- Using ASCII characters for graph display in non-UTF-8 locales
- Adjusting character widths accordingly

Fixes: #1870

Signed-off-by: Gabriel Kim <0xGabriel.kim@gmail.com>

PR #2066, opened 2026-08-20
tui: add ASCII fallback for non-UTF-8 locales

### LINKED ISSUE #1870: uftrace tui in docker shows  ~T~\ ~T~@(1) instead of pseudographics
`uftrace tui` in docker displays like this:
```
  TOTAL TIME : FUNCTION
   17.001  s : (1) apt-get
   19.261 us :   ~T~\ ~T~@(1) mkostemp
             :   ~T~B
    0.409 us :   ~T~\ ~T~@(1) getpid
             :   ~T~B
    0.316 us :   ~T~\ ~T~@(1) __snprintf_chk
             :   ~T~B
   41.176 us :   ~T~\ ~T~@(2) setenv
             :   ~T~B
    4.800 us :   ~T~\ ~T~@(12) signal
             :   ~T~B
   86.704 us :   ~T~\ ~T~@(1) fork
             :   ~T~B
   17.000  s :   ~T~\ ~T~@(1) waitpid
             :   ~T~B
             :   ~T~\ ~T~@(1) execvp
             :   ~T~B
    2.913 us :   ~T~\ ~T~@(1) fstat
             :   ~T~B
   20.147 us :   ~T~T ~T~@(1) unlink
```
`TERM` is `xterm`, `terminfo` is installed. There is no other `TERM*` env variables. It looked OK without docker. Maybe I'm missing something but I don't understand what.

Maybe there could be a option to force ascii or pseudographic lines?

mmustafasenoglu:
When `LANG` is not set to a UTF-8 locale (common in Docker containers), the TUI graph view displays garbled characters like `~T~\ ~T~@(1)` instead of proper box-drawing characters.

This adds locale detection via `LANG`, `LC_CTYPE`, and `LC_ALL`, then falls back to ASCII characters when UTF-8 is not available:

- `│` → `|` (vertical line)
- `└` → `` ` `` (corner)
- `├` → `+` (junction)
- `▶` → `>` (fold indicator)
- `─` → space (unfold connector)

The width correction (`width -= 2` for `snprintf` byte vs column mismatch) now only applies to multi-byte UTF-8 characters, since ASCII chars have 1 byte = 1 display column. This fixes the cursor overflow and alignment issues reported in #1870.

Tested with `LANG=POSIX` and `LANG=en_US.UTF-8`.

Fixes: #1870

### DIFF
--- a/cmds/tui.c
+++ b/cmds/tui.c
@@ -2,12 +2,14 @@
 
 #include <errno.h>
 #include <inttypes.h>
+#include <langinfo.h>
 #include <locale.h>
 #include <ncurses.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
@@ -127,6 +129,7 @@ static struct tui_graph partial_graph;
 static struct tui_list tui_info;
 static struct tui_list tui_session;
 static char *tui_search;
+static bool is_utf8_locale;
 
 static const struct tui_window_ops graph_ops;
 static const struct tui_window_ops report_ops;
@@ -210,6 +213,23 @@ static char *report_sort_key[] = { OPT_SORT_KEYS,  "total_avg",	   "total_min",	
 static char *selected_report_sort_key[NUM_REPORT_FIELD];
 
 static int curr_sort_key = 0;
+
+static void check_utf8_locale(void)
+{
+	/* Use setlocale() to resolve the effective LC_CTYPE from
+	 * LC_ALL, LC_CTYPE, and LANG in priority order. It treats
+	 * empty values as unset and returns NULL when the locale
+	 * cannot be activated (e.g., not installed).
+	 */
+	if (setlocale(LC_CTYPE, "") != NULL) {
+		const char *codeset = nl_langinfo(CODESET);
+		is_utf8_locale =
+			(codeset && (strcasestr(codeset, "utf-8") || strcasestr(codeset, "utf8")));
+	}
+	else {
+		is_utf8_locale = false;
+	}
+}
 
 static void init_colors(void)
 {
@@ -1379,6 +1399,18 @@ static void print_graph_indent(struct tui_graph *graph, struct tui_graph_node *n
 {
 	int i;
 	struct tui_graph_node *parent = (void *)node->n.parent;
+	const char *vertical_line, *corner, *branch;
+
+	if (is_utf8_locale) {
+		vertical_line = "│";
+		corner = "└";
+		branch = "├";
+	}
+	else {
+		vertical_line = "| ";
+		corner = "`-";
+		branch = "+-";
+	}
 
 	for (i = 0; i < depth; i++) {
 		if (width < 3) {
@@ -1393,11 +1425,11 @@ static void print_graph_indent(struct tui_graph *graph, struct tui_graph_node *n
 		}
 
 		if (i < depth - 1 || single_child)
-			printw("  │");
+			printw("  %s", vertical_line);
 		else if (is_last_child(parent, node))
-			printw("  └");
+			printw("  %s", corner);
 		else
-			printw("  ├");
+			printw("  %s", branch);
 	}
 }
 
@@ -1418,7 +1450,10 @@ static void win_display_graph(struct tui_window *win, void *node)
 		return;
 	}
 
-	fold_sign = curr->folded ? "▶" : "─";
+	if (is_utf8_locale)
+		fold_sign = curr->folded ? "▶" : "─";
+	else
+		fold_sign = curr->folded ? ">" : "";
 
 	parent = win_parent_graph(win, node);
 	if (parent == NULL)
@@ -1446,10 +1481,16 @@ static void win_display_graph(struct tui_window *win, void *node)
 			w = COLS - width;
 			width += snprintf(buf, sizeof(buf), "%s(%d) ", fold_sign, curr->n.nr_calls);
 
-			/* handle UTF-8 character length */
+			/* handle character length */
 			if (strcmp(fold_sign, " ")) {
-				width -= 2;
-				w += 2;
+				if (is_utf8_locale) {
+					width -= 2;
+					w += 2;
+				}
+				else {
+					width -= 1;
+					w += 1;
+				}
 			}
 			printw("%.*s", w, buf);
 		}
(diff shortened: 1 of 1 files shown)


GabrielKimm:
Hi @mmustafasenoglu mmustafasenoglu, thanks for picking this up.

I’m the author of #1974, and I didn’t complete the follow-up after the
2024 review. The one-column ASCII glyphs and applying the width
correction only to UTF-8 appear to address the cursor-overflow and
child-alignment issues reported there.

One remaining concern is the locale check. Looking for "UTF-8" in any
of LANG, LC_CTYPE, or LC_ALL can disagree with the effective locale:
LC_ALL overrides the others, and setlocale() can fail when the named
UTF-8 locale is not installed—the latter was part of the original
Docker reproduction.

For example:

- LC_ALL=POSIX LANG=en_US.UTF-8 should not select Unicode.
- LANG=en_US.UTF-8 may still be unusable in a minimal container when
  that locale is not installed.

I’m happy to test a Docker/terminal matrix or help prepare a small
follow-up patch. I won’t race this with a competing update to #1974;
once this PR is ready and maintainers prefer it, I’m happy to close
#1974 in favor of it.

mmustafasenoglu:
Thanks for the thorough review! You're right on both points.

Updated the locale check to:
1. **Honor `LC_ALL` override** — if `LC_ALL` is set and not UTF-8, bail out immediately without checking `LANG`/`LC_CTYPE`
2. **Verify via `setlocale()` + `nl_langinfo(CODESET)`** — this catches the case where `LANG=en_US.UTF-8` is set but the locale isn't installed (minimal Docker images). We now check what the C library actually resolved, not just what the env vars say.

This should handle the `LC_ALL=POSIX LANG=en_US.UTF-8` case and the missing `glibc-locales` Docker scenario from the original issue.

GabrielKimm:
> Thanks for the thorough review! You're right on both points.
> 
> Updated the locale check to:
> 
> 1. **Honor `LC_ALL` override** — if `LC_ALL` is set and not UTF-8, bail out immediately without checking `LANG`/`LC_CTYPE`
> 2. **Verify via `setlocale()` + `nl_langinfo(CODESET)`** — this catches the case where `LANG=en_US.UTF-8` is set but the locale isn't installed (minimal Docker images). We now check what the C library actually resolved, not just what the env vars say.
> 
> This should handle the `LC_ALL=POSIX LANG=en_US.UTF-8` case and the missing `glibc-locales` Docker scenario from the original issue.

Thanks for the quick update. I tested the latest `check_utf8_locale()`
change and ran the TUI in `debian:12-slim`.

The two cases I mentioned earlier now work as expected:

- `LC_ALL=POSIX LANG=en_US.UTF-8` uses the ASCII fallback.
- `LANG=en_US.UTF-8` also uses ASCII when that locale is not installed.

I still see two UTF-8 cases that the environment pre-check misses:

- With `LC_ALL="" LANG=C.UTF-8`, the code uses ASCII. An empty
  `LC_ALL` does not override `LANG`, though, and `locale charmap`
  reports `UTF-8`.
- With `LANG=C.utf8`, the code again uses ASCII. `C.utf8` is available
  in the `debian:12-slim` image, and `locale charmap` reports `UTF-8`.

Could we drop the environment substring check and use the result of
`setlocale(LC_CTYPE, "")` followed by `nl_langinfo(CODESET)`?
`setlocale()` already resolves the effective `LC_CTYPE` from
`LC_ALL`, `LC_CTYPE`, and `LANG` in priority order, treats empty values
as unset, and returns `NULL` when it cannot activate the requested
locale.

I also tried the expanded and collapsed TUI at 40, 60, and 80 columns
in both ASCII and UTF-8 modes. The width and alignment looked correct
in all of those runs.

cmds/tui.c
```diff
@@ -1418,7 +1437,7 @@ static void win_display_graph(struct tui_window *win, void *node)
 		return;
 	}
 
-	fold_sign = curr->folded ? "▶" : "─";
+	fold_sign = curr->folded ? (is_utf8_locale ? "▶" : ">") : (is_utf8_locale ? "─" : " ");
```
honggyukim: Looks too verbose. Using `if` is more readable here.

In addition, `is_utf8_locale` isn't changed for each call so using some character here can be defined once outside.

cmds/tui.c
```diff
@@ -216,34 +216,14 @@ static int curr_sort_key = 0;
...
+	 * cannot be activated (e.g., not installed).
 	 */
-	lang = getenv("LANG");
-	ctype = getenv("LC_CTYPE");
-
-	if (!(all && strstr(all, "UTF-8")) && !(ctype && strstr(ctype, "UTF-8")) &&
-	    !(lang && strstr(lang, "UTF-8"))) {
-		is_utf8_locale = false;
-		return;
-	}
```
honggyukim: Introducing this logic in your first patch, then remove in the next patch isn't a good practice. It can just be squashed into a single patch unless there is some other reason to split.

honggyukim (commented):
Thanks for the PR. I haven't tested this but just leave some comments by reading the change.

cmds/tui.c
```diff
@@ -1393,11 +1412,11 @@ static void print_graph_indent(struct tui_graph *graph, struct tui_graph_node *n
...
 
 		if (i < depth - 1 || single_child)
-			printw("  │");
+			printw("  %s", is_utf8_locale ? "│" : "|");
 		else if (is_last_child(parent, node))
-			printw("  └");
+			printw("  %s", is_utf8_locale ? "└" : "`");
 		else
-			printw("  ├");
+			printw("  %s", is_utf8_locale ? "├" : "+");
```
honggyukim: This part is same, we don't have to make conditional statement here. Whether to use utf8 character or not can be defined once outside.

`check_utf8_locale()` can be a good place for that. If so, then the function name `check_utf8_locale()` is better to be renamed something else.

namhyung:
Thanks for working on this.  I think @GabrielKimm is right and it's better to use `nl_langinfo(CODESET)`.  Please follow the convention in #1974 so that it can minimize code changes.  Also as @honggyukim requested, squash changes into a single commit.

namhyung:
@GabrielKimm are you ok with this change?

GabrielKimm:
> @GabrielKimm are you ok with this change?

Thanks for checking. I tested `d5a425a` under tmux on Debian 12 (glibc) and Alpine 3.22 (musl), 
using POSIX/ASCII and UTF-8 locales at 40, 60, and 80 columns. I also checked collapsing and re-expanding the graph.

The locale detection and UTF-8 output look good. One alignment issue remains in ASCII mode: rows in a single-child chain start one column farther to the right than branch rows. 
The cause seems to be the trailing space in `vertical_line = "| "`, combined with the blank fold marker for expanded nodes.

Changing it to `vertical_line = "|"` restores the alignment, and that version passed the same test matrix. 
Could we drop the trailing space?

image

PR #2067, opened 2026-08-23
uftrace: Make remote debuginfod lookups opt-in

honggyukim:
libdwfl honors DEBUGINFOD_URLS automatically while loading separate debug files.  This can make record and replay wait for a remote server even though the user did not request network access.

If there is some problems connecting the debuginfod remotely, then uftrace record or replay both silently wait forever as follows.
```
  $ uftrace record
  (... wait forever ...)

  $ uftrace replay
  (... wait forever ...)
```
In order to avoid this subtle and critical problem, this patch changes the default behavior not to use remote debuginfod access.

It clears DEBUGINFOD_URLS by default and add --debuginfod to enable remote lookups explicitly.  When enabled, limit commencement to 10 seconds and total download to 60 seconds, and disable retries to bound each remote request.

dwfl_module_getdwarf() returns NULL for network errors, timeouts and missing DWARF alike.  Warn that the request failed or timed out without claiming to distinguish these cases.  Keep unit tests local-only and document the new option.


honggyukim:
This problem happens from sometime ago. It's due to the following environment variable.
```
$ env | grep DEBUG
DEBUGINFOD_URLS=https://debuginfod.ubuntu.com
```
I didn't set this myself explicitly.

namhyung:
I'm ok with the change but not sure what makes the problem in the first place.  The current code already disables it by unsetting the environment variable.  Why did you see the problem?

honggyukim:
Without this unset, it wait forever. Maybe the server has some problems

namhyung:
I mean we have the code already, but maybe it's only for record.

honggyukim:
This problem is confirmed in another Ubuntu 24.04 server. It doesn't wait forever, but takes a long time for waiting network connection in `dwfl_module_getdwarf()`.

Without this patch:
```
$ time uftrace record t-abc

real    0m22.726s
user    0m1.264s
sys     0m0.179s

$ time uftrace replay --no-pager
# DURATION     TID      FUNCTION
   1.357 us [3644849] | __monstartup();
   0.470 us [3644849] | __cxa_atexit();
            [3644849] | main() {
            [3644849] |   a() {
            [3644849] |     b() {
            [3644849] |       c() {
   0.683 us [3644849] |         getpid();
   2.088 us [3644849] |       } /* c */
   2.775 us [3644849] |     } /* b */
   3.233 us [3644849] |   } /* a */
   3.803 us [3644849] | } /* main */

real    0m3.773s
user    0m0.209s
sys     0m0.028s
```

With this patch:
```
$ time uftrace record t-abc

real    0m0.113s
user    0m0.107s
sys     0m0.048s

$ time uftrace replay
# DURATION     TID      FUNCTION
   0.303 us [3648165] | __monstartup();
   0.090 us [3648165] | __cxa_atexit();
            [3648165] | main() {
            [3648165] |   a() {
            [3648165] |     b() {
            [3648165] |       c() {
   0.148 us [3648165] |         getpid();
   0.438 us [3648165] |       } /* c */
   0.590 us [3648165] |     } /* b */
   0.695 us [3648165] |   } /* a */
   0.825 us [3648165] | } /* main */

real    0m0.043s
user    0m0.025s
sys     0m0.018s
```

namhyung:
Oh.. maybe it's from the minidebug support.  It used to disable the debuginfod in the child process, but it seems the the retry logic can be called from all other places.

At this moment, I'd like to have minimal change to disable debuginfod unconditionally in uftrace.c.  We can add the option after v0.20 is released.

honggyukim:
> I mean we have the code already, but maybe it's only for record.

I added some debug code as follows.
```diff
diff --git a/cmds/record.c b/cmds/record.c
index f5fc9fb9de90..a73374e1aa63 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -379,6 +379,7 @@ static void setup_child_environ(struct uftrace_opts *opts, int argc, char *argv[
        setenv("GLIBC_TUNABLES", "glibc.cpu.hwcaps=-IBT,-SHSTK", 1);

        /* disable debuginfo daemon */
+       fprintf(stderr, "[%s:%d: tid=%d] unsetenv(DEBUGINFOD_URLS)\n", __func__, __LINE__, gettid());
        unsetenv("DEBUGINFOD_URLS");
 }

diff --git a/utils/symbol-libelf.c b/utils/symbol-libelf.c
index f27d45e537ef..056ecf53ffe9 100644
--- a/utils/symbol-libelf.c
+++ b/utils/symbol-libelf.c
@@ -100,6 +100,7 @@ int elf_retry(const char *filename, struct uftrace_elf_data *elf)
                goto out;
        }

+       fprintf(stderr, "[%s:%d: tid=%d] calling dwfl_module_getdwarf ...\n", __func__, __LINE__, gettid());
        dw = dwfl_module_getdwarf(mod, &bias);
        if (dw == NULL) {
                pr_dbg2("cannot find debug file: %s\n", dwfl_errmsg(dwfl_errno()));
```
It shows the `unsetenv` is called in a different thread.
```
$ uftrace record t-abc
[setup_child_environ:382: tid=3671515] unsetenv(DEBUGINFOD_URLS)
[elf_retry:104: tid=3671514] calling dwfl_module_getdwarf ...
```

honggyukim:
> At this moment, I'd like to have minimal change to disable debuginfod unconditionally in uftrace.c. We can add the option after v0.20 is released.

If so, are you okay with the following change? I've confirmed that it also fixes the problem.
```diff
diff --git a/cmds/record.c b/cmds/record.c
index f5fc9fb9de90..4158f877e085 100644
--- a/cmds/record.c
+++ b/cmds/record.c
@@ -377,9 +377,6 @@ static void setup_child_environ(struct uftrace_opts *opts, int argc, char *argv[
        put_libmcount_path(libpath);
        setenv("XRAY_OPTIONS", "patch_premain=false", 1);
        setenv("GLIBC_TUNABLES", "glibc.cpu.hwcaps=-IBT,-SHSTK", 1);
-
-       /* disable debuginfo daemon */
-       unsetenv("DEBUGINFOD_URLS");
 }

 static uint64_t calc_feat_mask(struct uftrace_opts *opts)
diff --git a/uftrace.c b/uftrace.c
index cc21488dc80f..4b044c1b0da8 100644
--- a/uftrace.c
+++ b/uftrace.c
@@ -1842,6 +1842,9 @@ int main(int argc, char *argv[])
        if (!opts.libcall && opts.nest_libcall)
                pr_err_ns("cannot use --no-libcall and --nest-libcall options together\n");

+       /* disable debuginfo daemon to avoid unintended network delay */
+       unsetenv("DEBUGINFOD_URLS");
+
        switch (opts.mode) {
        case UFTRACE_MODE_RECORD:
                ret = command_record(argc, argv, &opts);
```

namhyung:
Yep, let's have that for v0.20.  You may want to open a new PR for that.

honggyukim:
The simpler solution is uploaded at ##2068. Please have a look.

PR #2068, opened 2026-08-24
uftrace: Fix debuginfod remote waiting

honggyukim:
The current location of disabling debuginfo daemon is applied only for the record thread, but dwfl_module_getdwarf(), which requires network connection, is called main thread and others as well.

libdwfl honors DEBUGINFOD_URLS automatically while loading separate debug files.  This can make record and replay wait for a remote server even though the user did not request network access.

If there is some problems connecting the debuginfod remotely, then uftrace record or replay both silently wait very long time as follows.
```
  $ time uftrace record t-abc
      ...
  real    0m22.726s
  user    0m1.264s
  sys     0m0.179s

  $ time uftrace replay --no-pager
      ...
  real    0m3.773s
  user    0m0.209s
  sys     0m0.028s
```
In order to avoid this subtle and critical problem, this patch relocate the `unsetenv("DEBUGINFOD_URLS")` logic in the initial stage even before the command check branch so that it can be applied to all the threads being created.


namhyung (approved):
LGTM

PR #2069, opened 2026-08-31
symbol: Fix memory leak when dynamic symbol count is zero

MiliAxe:
When `load_elf_dynsymtab()` finds no valid dynamic symbols (`nr_sym == 0`) , it jumps to the out label. This leaves the pre-allocated symbol array unfreed in `dsymtab->sym`.

This change frees the pre-allocated memory and resets `dsymtab->sym` and `nr_alloc` to prevent memory leaks.


namhyung (approved):
LGTM

PR #2072, opened 2026-09-17
symbol: Fix use of uninitialized uftrace_elf_data and uftrace_elf_iter

yskzalloc:
Some instances of struct uftrace_elf_data and uftrace_elf_iter are used without being initialized. This can result in reads of indeterminate values and trigger the following MemorySanitizer warning:

  SUMMARY: MemorySanitizer: use-of-uninitialized-value utils/symbol.c

  ==195598==WARNING: MemorySanitizer: use-of-uninitialized-value
      #0 0x5597aab57a39 in check_static_binary utils/symbol.c
      ...

This works remove all initialization of the structures before use to prevent uninitialized memory from being accessed in utils/symbol.c.


namhyung (commented):
Can you show me where exactly it detected the uninitialized use?

For uftrace_elf_data, I think it should be fine once `elf_init()` returns 0.  For uftrace_elf_iter, I believe any access macro like `elf_for_each_shdr()` should set it properly.

